PR #7, opened 2026-04-06
fix: reject '.' and '..' in workspace and source name validation

sauldhernandez:
## Summary
- `validate_name()` rejected `/` and `\` but allowed `.` and `..` as workspace/source names
- Since names are joined into filesystem paths via `PathBuf::join()`, this allowed escaping the intended directory (e.g. a source named `..` resolves to the workspace root)
- Added validation to reject `.` and `..` after trimming, with corresponding tests

## Test plan
- [x] New `rejects_path_traversal` test covers `..`, `.`, and whitespace-padded ` .. `
- [x] Existing `rejects_forward_and_backward_slashes` test still passes

### DIFF
--- a/crates/coral-app/src/sources/manager.rs
+++ b/crates/coral-app/src/sources/manager.rs
@@ -91,7 +91,7 @@ impl SourceManager {
             .require_app(request.workspace.as_ref())?;
         let bundled_name = self
             .workspace_manager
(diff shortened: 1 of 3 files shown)


jsummerfield:
@cubic

jsummerfield (approved):

PR #16, opened 2026-04-07
feat(cli): add guided onboarding flow

jsummerfield:
## Summary
Coral's setup story was still a sequence of low-level source commands after install. This change adds a guided onboarding path so first-time setup is easier to discover and follow, while keeping the existing source commands as the underlying primitives.

## What changed
- add `coral onboard` as an interactive source setup and validation flow
- extract reusable source operation helpers and add a non-interactive onboarding integration test
- add a narrow install script and update install, quickstart, CLI, README, and agent-usage docs to align on the new setup path
- add a Coral skill entrypoint and tighten onboarding presentation, navigation, and next-step guidance

## Validation
- `make validate`

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,0 +1,273 @@
+# Coral Setup Plan
+
+## Goal
+
+Improve Coral's setup story using the product model that already exists in this repo:
+
+1. install the CLI
+2. configure at least one source locally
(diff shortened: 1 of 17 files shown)


AlbertQM (approved):
Thank you! Will take it from here 🫡

PR #20, opened 2026-04-07
test(engine): add integration test suite for coral-engine query API

sauldhernandez:
## Summary

Implements the **Engine Integration Test** tier from the Coral Testing Strategy. These 32 tests drive `CoralQuery::execute_sql()` and `list_tables()` directly — no server, no gRPC, no app state — validating backend behavior at low cost.

## Intent

The engine had no integration tests exercising the full manifest → compile → register → SQL → results flow. Existing unit tests cover individual backend internals, but nothing validated the end-to-end query path or cross-source behavior.

## What it enables

- **HTTP backend** (13 tests): SELECT, projection, ordering, limit, filter pushdown, COUNT, three pagination modes (page/offset/link_header), auth header verification, and error handling (500 retries, 401, malformed JSON) via wiremock
- **JSONL backend** (7 tests): full query operations plus glob across subdirectories and missing-file errors
- **Parquet backend** (6 tests): full query operations plus missing-file error with a new empty-schema guard in `infer_schema_expand_dicts`
- **System catalog** (6 tests): `coral.tables`/`coral.columns` metadata, `list_tables` parity, cross-source JOIN, and nonexistent-schema errors
- **CI coverage workflow**: GitHub Actions workflow that runs `cargo-llvm-cov` on every PR and posts a sticky coverage summary comment
- **Makefile `test` target**: runs coverage locally with summary output

## Usage examples

```bash
# Run all engine integration tests
cargo test -p coral-engine --test engine

# Run a specific test
cargo test -p coral-engine --test engine -- pagination_page_mode

# Run coverage locally
make test
```


crates/coral-engine/tests/engine/http_tests.rs
```diff
@@ -0,0 +1,468 @@
+use coral_engine::{CoralQuery, CoreError, StatusCode};
+use serde_json::json;
+use wiremock::matchers::{header, method, path, query_param, query_param_is_missing};
+use wiremock::{Mock, MockServer, ResponseTemplate};
+
+use crate::harness::{
+    TestRuntime, build_source, build_source_with_secrets, execution_to_rows, users_rows,
+};
+
+fn base_http_manifest(
```
antonmry: The caller has to nail YAML indentation exactly. There's already a `parse_source_manifest_value(value: Value)` function in coral-spec that accepts a `serde_json::Value` instead of a YAML string. Tests could
  build manifests as JSON values and merge fields programmatically:

```rust
  fn base_http_manifest(name: &str, base_url: &str) -> serde_json::Value {
      json!({
          "name": name,
          "version": "0.1.0",
          "dsl_version": 3,
          "backend": "http",
          "base_url": base_url,
          "tables": [{
              "name": "users",
              "description": "HTTP users",
              "request": { "method": "GET", "path": "/api/users" },
              "response": { "rows_path": ["data"] },
              "columns": [
                  { "name": "id", "type": "Int64" },
                  { "name": "name", "type": "Utf8" },
                  { "name": "email", "type": "Utf8" },
              ]
          }]
      })
  }
```

Then callers just modify the value before passing it:

```rust
  let mut manifest = base_http_manifest("http_page", &server.uri());
  manifest["tables"][0]["pagination"] = json!({
      "mode": "page",
      "page_param": "page",
      "page_start": 1
  });
```

antonmry (approved):

jsummerfield:
@cubic

jsummerfield (commented):
Description mentions `test_source` but I don't see it.

crates/coral-engine/tests/engine/parquet_tests.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    let error = CoralQuery::execute_sql(
+        &[source],
+        &TestRuntime,
+        "SELECT * FROM parquet_missing.users",
+    )
+    .await
+    .expect_err("missing parquet source should fail");
+
+    match error {
```
jsummerfield: I think we should be stricter on these negative tests, since we're encoding the stable contract. Something more like:

1. error.status_code()
2. the exact CoreError variant
3. one small stable detail fragment

```rust
assert_eq!(error.status_code(), StatusCode::Internal);
match error {
    CoreError::Internal(detail) => {
        assert!(detail.contains("no parquet files found at"));
        assert!(detail.contains("missing"));
    }
    other => panic!("expected internal missing-parquet error, got {other:?}"),
}
```

crates/coral-engine/tests/engine/catalog_tests.rs
```diff
@@ -0,0 +1,238 @@
...
+
+#[tokio::test]
+async fn query_nonexistent_schema_returns_error() {
+    let (_temp, sources) = build_catalog_sources();
+
+    let error = CoralQuery::execute_sql(&sources, &TestRuntime, "SELECT * FROM missing.users")
+        .await
+        .expect_err("missing schema should fail");
+
+    match error {
```
jsummerfield: This assertion is too permissive: it accepts `InvalidInput`, `NotFound`, or `Internal`, so unrelated planner/runtime regressions could still pass. I would assert the intended stable contract here (`status_code()` plus the exact `CoreError` variant) and then only a small stable detail fragment such as `missing`.

crates/coral-engine/tests/engine/jsonl_tests.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    let error =
+        CoralQuery::execute_sql(&[source], &TestRuntime, "SELECT * FROM jsonl_missing.users")
+            .await
+            .expect_err("missing jsonl source should fail");
+
+    assert_nonexistent_schema_error(error, "jsonl_missing");
+}
+
+fn assert_nonexistent_schema_error(error: CoreError, schema_name: &str) {
```
jsummerfield: Same issue here: this helper turns any of three error classes into a pass. For a transport-level integration suite, I would assert the exact contract (`status_code()` and variant) instead of broad matching. Since JSONL intentionally skips missing directories during registration, it may be clearer to assert the resulting missing-relation contract directly or cover this path through `test_source`/`list_tables`.

crates/coral-engine/tests/engine.rs
```diff
@@ -0,0 +1,17 @@
+//! Integration coverage for the transport-neutral `coral-engine` query API.
```
jsummerfield, line 1: The suite description says this covers the transport-neutral query API, but it still does not exercise `CoralQuery::test_source()`. Since that is a public entrypoint with its own single-source initialization path, I would add at least one integration test for it so this suite pins all three API surfaces.

crates/coral-engine/tests/engine/harness.rs
jsummerfield, line 1: Consider adding some test helpers

```rust
use coral_engine::{CoreError, StatusCode};

pub(crate) fn assert_not_found(error: CoreError, needle: &str) {
    assert_eq!(error.status_code(), StatusCode::NotFound);
    match error {
        CoreError::NotFound(detail) => {
            assert!(
                detail.contains(needle),
                "expected not-found detail to contain {needle:?}, got: {detail}"
            );
        }
        other => panic!("expected CoreError::NotFound, got {other:?}"),
    }
}

pub(crate) fn assert_invalid_input(error: CoreError, needle: &str) {
    assert_eq!(error.status_code(), StatusCode::InvalidArgument);
    match error {
        CoreError::InvalidInput(detail) => {
            assert!(
                detail.contains(needle),
                "expected invalid-input detail to contain {needle:?}, got: {detail}"
            );
        }
        other => panic!("expected CoreError::InvalidInput, got {other:?}"),
    }
}

pub(crate) fn assert_internal(error: CoreError, needle: &str) {
    assert_eq!(error.status_code(), StatusCode::Internal);
    match error {
        CoreError::Internal(detail) => {
            assert!(
                detail.contains(needle),
                "expected internal detail to contain {needle:?}, got: {detail}"
            );
        }
        other => panic!("expected CoreError::Internal, got {other:?}"),
    }
}
```

jsummerfield (approved):

jsummerfield (approved):

PR #21, opened 2026-04-07
refactor(cli): redesign onboard landing page with branding and interactive source list

AlbertQM:
## Summary

The goal of `onboard` is to go from "I just installed `coral`" to "this is cool" in an interactive way, with the least amount of friction possible.
We're not going to teach everything here; my goal is to make the 0-to-hero path as short and sweet as possible.

This PR changes what you see when you run `onboard`.
This should make it both easy and clear that you should add some sources if you want `coral` to be useful.

This is not the end game, I plan on improve "next steps" too: they are currently briefly shown after you exit `onboard`; instead, I think we can make it a subsequent step of the interactive wizard.

## Test plan

Manually tested the various paths. Here's one:

image
image
image

### DIFF
--- a/crates/coral-cli/src/onboard.rs
+++ b/crates/coral-cli/src/onboard.rs
@@ -73,7 +73,7 @@ fn select_top_level(
     let first_uninstalled = bundled_sources
         .iter()
         .position(|s| !s.installed)
-        .unwrap_or(0);
+        .unwrap_or(bundled_sources.len());
 
     let selection = Select::with_theme(theme)
         .with_prompt("Choose a source")
--- a/crates/coral-cli/src/onboard.rs
+++ b/crates/coral-cli/src/onboard.rs
@@ -9,42 +9,49 @@ use crate::source_ops;
 
 const SOURCE_DESCRIPTION_PREVIEW_LIMIT: usize = 88;
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-enum OnboardAction {
-    AddBundledSource,
-    ImportSourceManifest,
-    ValidateInstalledSource,
+enum TopLevelChoice {
+    BundledSource(usize),
+    ImportManifest,
     Finish,
+}
+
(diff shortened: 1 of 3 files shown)


crates/coral-cli/src/onboard.rs
```diff
@@ -9,123 +9,160 @@ use crate::source_ops;
...
+            println!("Reconfigured source {}", result.name);
+            maybe_validate_after_install(app, theme, &result.name).await?;
+        }
+        Some(InstalledSourceAction::Back) | None => {}
+    }
+
+    Ok(())
+}
+
+async fn run_add_bundled_source(
```
jsummerfield, line 145: We have an issue here when a workspace contains an imported source whose name matches a bundled source (for example, a customised github manifest). It'll get overwritten by `add_bundled_source` below.

AlbertQM: Thanks! Having slept on this, I think the right way forward is to remove the option to import altogether. This is the onboarding, which we recommend running just after install; don't think users will have .yaml file ready, so IMO this is adding unnecessary mental load

I'm working on the "post onboarding" (ie you've connected some sources, now what?); this _might_ fit in there as a heads-up (more than an interactive process)

jsummerfield:
@cubic

crates/coral-cli/src/onboard.rs
```diff
@@ -1,176 +1,161 @@
...
-    }
-    items.push(("Finish onboarding", OnboardAction::Finish));
-    let labels = items.iter().map(|(label, _)| *label).collect::<Vec<_>>();
+    bundled_sources: &[AvailableSource],
+) -> Result<TopLevelChoice, anyhow::Error> {
+    let name_width = bundled_sources
+        .iter()
+        .map(|s| measure_text_width(&s.name))
+        .max()
+        .unwrap_or(0);
```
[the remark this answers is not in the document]
AlbertQM: Fair point, will address

jsummerfield (approved):
Nice!

PR #22, opened 2026-04-07
feat(cli): auto-generate CLI reference docs from clap metadata

sauldhernandez:
## Intent

The CLI reference (`docs/reference/cli-reference.mdx`) was manually maintained and could drift from the actual clap definitions. This PR makes the docs fully generated from clap metadata so they stay in sync automatically.

## What it enables

- Adding or modifying CLI commands/arguments automatically updates the docs — no manual MDX editing required.
- `make validate` catches drift via `cargo run -p xtask -- --check`.
- `make gen-cli-docs` regenerates the file in one step.

## Usage examples

Regenerate the CLI reference after changing command definitions:
```sh
make gen-cli-docs
# or: cargo run -p xtask
```

Check that the docs are up to date (runs automatically in `make validate`):
```sh
cargo run -p xtask -- --check
```

All command descriptions, argument help text, examples, and notes are defined via clap attributes (`about`, `after_long_help`) in `crates/coral-cli/src/lib.rs`. The xtask walks the command tree and emits the MDX file.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4933,6 +4933,14 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1ffae5123b2d3fc086436f8834ae3ab053a283cfac8fe0a0b8eaae044768a4c4"
 
 [[package]]
+name = "xtask"
+version = "0.1.0"
+dependencies = [
+ "clap",
+ "coral-cli",
+]
+
+[[package]]
 name = "yoke"
 version = "0.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,11 @@
+gen-cli-docs:
+	cargo run -p xtask
+
 validate:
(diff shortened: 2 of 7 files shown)


jsummerfield:
I like the intent, but I think the output is a step backwards in readability and it adds another source of drift since it is manually parsing and rebuilding clap's output. It'll drift again as soon as we add e.g. repeated args, aliases, optional values etc.

Can you see if there's a way to lean more heavily on clap's own renderers for the hard parts (render_usage, render_help). Otherwise let's put a pin in this for now and revisit once the docs have stabilised.

CleanShot 2026-04-07 at 21 54 17@2x

jsummerfield (changes requested):
See comment

sauldhernandez:
@jsummerfield I'm not too fond of the look & feel by using clap's renderers. See https://withcoral-sauldhernandez-auto-cli-docs.mintlify.app/reference/cli-reference

So I'm parking this for now

PR #24, opened 2026-04-07
ci: add release workflow

simonwhitaker:
## Summary

This adds a GitHub Actions release workflow for `coral`.

On pushes to `main`, the workflow watches for `Cargo.toml` version changes and only creates a release when the workspace version actually changes. It also supports manual runs from `main`, with an optional `target_ref` so we can build and publish a release from a specific commit in `main` history.

The workflow now:
- validates release metadata and version/tag invariants up front
- runs the CLI smoke test before publishing
- builds release binaries for:
  - `x86_64-unknown-linux-gnu`
  - `aarch64-unknown-linux-gnu`
  - `x86_64-apple-darwin`
  - `aarch64-apple-darwin`
- packages tarballs and generates SHA-256 checksums
- generates GitHub release notes automatically
- creates the GitHub release against the exact build commit using the release app token

## Why

This gives us a GitHub-native release path that is tied directly to version bumps, while still leaving room to manually release an older commit from `main` when needed. It also makes the produced tag, artifacts, and release notes all line up with the same resolved commit.

## Testing

- `make validate`


jsummerfield:
@cubic

.github/workflows/release.yml
```diff
@@ -0,0 +1,257 @@
...
+          fi
+
+          {
+            echo "build_ref=${resolved_sha}"
+            echo "tag_name=$version"
+            echo "release_name=$version"
+            echo "artifact_prefix=coral"
+          } >> "$GITHUB_OUTPUT"
+
+  smoke-test:
```
jsummerfield: These don't exist atm.

.github/workflows/release.yml
```diff
@@ -0,0 +1,257 @@
...
+      - name: Install cross
+        if: matrix.cross
+        run: |
+          set -euo pipefail
+          cargo install cargo-binstall --locked
+          cargo binstall --no-confirm --force --disable-strategies compile cross@0.2.5
+      - name: Build
+        run: |
+          if [ "${{ matrix.cross }}" = "true" ]; then
+            cross build --locked --release --features otel-receiver --target ${{ matrix.target }}
```
jsummerfield: Nor do these

.github/workflows/release.yml
```diff
@@ -0,0 +1,257 @@
...
+        with:
+          fetch-depth: 0
+      - name: Require main branch
+        if: github.event_name == 'workflow_dispatch'
+        run: |
+          if [ "${GITHUB_REF}" != "refs/heads/main" ]; then
+            echo "::error::Run this workflow from main."
+            exit 1
+          fi
+      - name: Prepare release metadata
```
jsummerfield, line 35: Stepping back, I wonder if there's a simpler tag-driven release process where we release on v* tag pushes and make the tag the explicit release intent. That removes almost all of the custom version-detection and “manual historical release” logic, because releasing an older main commit becomes “tag that commit.”

1. Bump Cargo.toml on main.
2. Create and push a tag like v0.2.0 on that commit.
3. Let the workflow trigger only on push.tags: ['v*'].
4. The release workflow checks that github.ref_name == "v${cargo_version}" and fails if they differ.

simonwhitaker: That's what we used to have. The main downside is if your release build fails and you need to push another commit to fix it. You never want to move a tag (canonical discussion on this is [here](https://git-scm.com/docs/git-tag#_on_re_tagging)), so you end up having to declare a new release. And if that commit also doesn't compile for whatever reason, another release, and so on.

.github/workflows/release.yml
simonwhitaker: You mean the otel-receiver feature?

.github/workflows/release.yml
simonwhitaker: Are they imminent? Should I stub this out or completely remove it?

jsummerfield: Remove

.github/workflows/release.yml
jsummerfield: Ah ok. Thanks.

.github/workflows/release.yml
```diff
@@ -0,0 +1,257 @@
+name: Release
+on:
```
jsummerfield, line 2: Add `concurrency` to the release flow?

.github/workflows/release.yml
```diff
@@ -0,0 +1,257 @@
...
+      - name: Require main branch
+        if: github.event_name == 'workflow_dispatch'
+        run: |
+          if [ "${GITHUB_REF}" != "refs/heads/main" ]; then
+            echo "::error::Run this workflow from main."
+            exit 1
+          fi
+      - name: Prepare release metadata
+        id: meta
+        run: |
```
jsummerfield: Consider moving this into a script (shell or python), it'd make it more readable and ultimately testable.

jsummerfield (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,241 @@
...
+          cd "target/${{ matrix.target }}/release"
+          tar czf "../../../${archive}" coral
+          cd ../../../
+      - name: Upload artifact
+        uses: actions/upload-artifact@v7
+        with:
+          name: ${{ needs.prepare.outputs.artifact_prefix }}-${{ matrix.target }}
+          path: ${{ needs.prepare.outputs.artifact_prefix }}-${{ matrix.target }}.tar.gz
+  checksums:
+    needs: [prepare, build]
```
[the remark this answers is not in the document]
simonwhitaker: See previous comments. The smoke test doesn't exist.

Makefile
```diff
@@ -1,7 +1,7 @@
 validate:
+	python3 -m unittest discover -s .github/scripts -p 'test_prepare_release.py'
```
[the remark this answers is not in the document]
simonwhitaker: This is intentional. The integration tests run in CI.

PR #26, opened 2026-04-08
feat(onboarding): add manifest-driven input hints for source setup

iaptsiauri:
## Summary

Add `onboarding.input_help` to the manifest schema so source authors can provide human-readable hints that are displayed during interactive `source add` and `source import` prompts.

## Testing Strategy

Tested manually 
`make validate`

### DIFF
--- a/crates/coral-spec/src/backends/file.rs
+++ b/crates/coral-spec/src/backends/file.rs
@@ -14,7 +14,7 @@ use serde_json::Value;
 use std::collections::HashSet;
 use url::Url;
 
-use crate::common::{build_source_manifest_common, parse_manifest_data_type};
+use crate::common::{Onboarding, build_source_manifest_common, parse_manifest_data_type};
 use crate::{
     ColumnSpec, FilterSpec, ManifestDataType, ManifestError, ParsedTemplate, Result, SourceBackend,
     SourceManifestCommon, TableCommon, validate_columns, validate_filters_and_column_exprs,
@@ -45,6 +45,7 @@ struct RawFileSourceManifest {
     description: Option<String>,
     backend: SourceBackend,
     tables: Vec<RawFileTableSpec>,
+    onboarding: Option<Onboarding>,
 }
 
 #[derive(Debug, Clone, Deserialize)]
@@ -280,8 +281,9 @@ impl ParquetSourceManifest {
             description: _description,
             backend: _backend,
             tables,
+            onboarding,
         } = raw;
-        let common = build_source_manifest_common(dsl_version, name, version);
+        let common = build_source_manifest_common(dsl_version, name, version, onboarding);
         let tables = tables
             .into_iter()
             .map(|table| table.into_validated_parquet(&common.name))
@@ -310,8 +312,9 @@ impl JsonlSourceManifest {
             description: _description,
             backend: _backend,
             tables,
+            onboarding,
         } = raw;
-        let common = build_source_manifest_common(dsl_version, name, version);
+        let common = build_source_manifest_common(dsl_version, name, version, onboarding);
         let tables = tables
             .into_iter()
             .map(|table| table.into_validated_jsonl(&common.name))
--- a/crates/coral-spec/src/backends/file.rs
+++ b/crates/coral-spec/src/backends/file.rs
@@ -278,12 +278,18 @@ impl ParquetSourceManifest {
             dsl_version,
             name,
             version,
-            description: _description,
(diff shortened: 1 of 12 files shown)


crates/coral-spec/src/inputs.rs
```diff
@@ -34,12 +34,12 @@ impl InputKind {
...
 }
 
-/// One interactive input extracted from a validated source spec.
+/// One install-time input extracted from a validated source spec.
 ///
 /// The app and CLI can map this into prompts, persisted variables, or secret
 /// collection flows without depending on protobuf-specific types.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct ManifestInputSpec {
+pub struct InputSpec {
```
jsummerfield, line 24: Could you rename the other Manifest* structs too

jsummerfield:
Am I reading this correctly? Right now the flow is not:

1. parse manifest into validated Onboarding
2. ask Onboarding for help per input
3. populate InputSpec.help / SourceInputSpec.help

Instead, it does this:

1. parse manifest into Onboarding in common.rs
2. separately scrape onboarding.input_help from raw YAML in collect_input_help() 
3. attach that scraped data to InputSpec.help in collect_source_inputs_value()

iaptsiauri:
> Am I reading this correctly? Right now the flow is not:
> 
> 1. parse manifest into validated Onboarding
> 2. ask Onboarding for help per input
> 3. populate InputSpec.help / SourceInputSpec.help
> 
> Instead, it does this:
> 
> 1. parse manifest into Onboarding in common.rs
> 2. separately scrape onboarding.input_help from raw YAML in collect_input_help()
> 3. attach that scraped data to InputSpec.help in collect_source_inputs_value()

Yeap, that's true. I did rearrange things now. Now we are 
Parsing into ValidatedSourceManifest
Collecting inputs 
Attaching help from the parsed metadata

Put it into a centralized Collector struct to make it more obvious as well

crates/coral-spec/src/backends/file.rs
```diff
@@ -14,7 +14,7 @@ use serde_json::Value;
 use std::collections::HashSet;
 use url::Url;
 
-use crate::common::{build_source_manifest_common, parse_manifest_data_type};
+use crate::common::{Onboarding, build_source_manifest_common, parse_manifest_data_type};
```
antonmry, line 17: Not a blocker, but something to evaluate:

```
 Onboarding is not a manifest DSL type — it's a UX/CLI concern about how to present prompts during setup. It's fundamentally
  different from ColumnSpec or PaginationSpec which describe runtime query behavior.

  What Onboarding actually does in this PR:

  - Gets deserialized from the manifest (common.rs)
  - Gets stored on SourceManifestCommon (every backend carries it)
  - Gets read by InputCollector in inputs.rs to attach help text
  - Gets surfaced to the CLI for display

  Two issues with putting it in common.rs:

  1. Mixed concerns — common.rs defines the query/runtime schema model. Onboarding is purely about install-time UX. It's like putting CLI prompt colors in
  the same file as your SQL column types.
  2. Unnecessary coupling — Every backend (http.rs, file.rs) now imports and threads Onboarding through build_source_manifest_common. The backends shouldn't
   need to know about onboarding — they compile and register tables, they don't prompt users.

  A cleaner structure would be to keep Onboarding in inputs.rs (which already handles the install-time input collection) or in its own small module.
  collect_inputs could pull onboarding data directly from the parsed YAML/value rather than threading it through SourceManifestCommon → every backend → back
   to inputs.rs. The collect_source_onboarding_yaml and collect_source_onboarding_value functions already exist in this PR but are exported alongside core
  manifest types, which feels off.

  Worth raising as a structural suggestion — not a blocker, but it'd keep the separation between "what the engine needs" and "what the CLI needs" cleaner.

```

sources/github/manifest.yaml
```diff
@@ -140,7 +145,7 @@ tables:
         type: Boolean
         nullable: true
         virtual: false
-        description: Whether feedback pull request will be created on assignment 
+        description: Whether feedback pull request will be created on assignment
```
antonmry, line 148: Minor thing but it would be good to add yamlint or yamlfmt to "make validate" to avoid this in the future. Another might be ryl https://github.com/owenlamont/ryl

antonmry (approved):

iaptsiauri:
Closed in favor of https://github.com/withcoral/coral/pull/97

PR #30, opened 2026-04-08
feat(sources): adds source validate command. uses manifest schema file

iaptsiauri:
## Summary
  - All callers get structural checks before semantic validation
  - Remove duplicate Rust validation checks 
  - Add `coral source validate <path>` CLI command
 
  ## Test plan
  Modified accordingly existing tests
`make validate`

### DIFF
--- a/crates/coral-cli/src/main.rs
+++ b/crates/coral-cli/src/main.rs
@@ -83,6 +83,11 @@ enum SourceCommand {
         /// Name of the source to remove
         name: String,
     },
+    /// Validate a source manifest file
+    Validate {
+        /// Path to the source manifest file
+        path: PathBuf,
+    },
 }
 
 #[derive(Debug, Clone, Copy, ValueEnum)]
@@ -170,6 +175,9 @@ async fn main() -> Result<(), anyhow::Error> {
             SourceCommand::Remove { name } => {
                 source_ops::delete_source(&app, &name).await?;
                 println!("Removed source {name}");
+            }
+            SourceCommand::Validate { path } => {
+                source_ops::validate_manifest(&path)?;
             }
         },
         Command::Onboard => {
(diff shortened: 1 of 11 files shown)


jsummerfield (changes requested):
See comments

crates/coral-spec/Cargo.toml
```diff
@@ -16,6 +16,8 @@ serde_yaml.workspace = true
 thiserror.workspace = true
 url.workspace = true
 
+[dependencies.jsonschema]
+workspace = true
```
jsummerfield, line 20: Prefer inline

crates/coral-spec/src/parser.rs
```diff
@@ -147,3 +155,27 @@ fn parse_source_backend(value: &Value) -> Result<SourceBackend> {
         serde_json::from_value(backend).map_err(ManifestError::deserialize)?;
     Ok(backend)
 }
+
+fn validate_manifest_schema(value: &Value) -> Result<()> {
```
jsummerfield, line 159: `coral source validate` and `coral source import` now have different inspection paths. One can pass validation and fail import, e.g. a manifest using the same variable key with conflicting defaults.

Better would be to share a single manifest inspection function and share with validate and import.

crates/coral-cli/src/main.rs
```diff
@@ -83,6 +83,11 @@ enum SourceCommand {
         /// Name of the source to remove
         name: String,
     },
+    /// Validate a source manifest file
+    Validate {
```
jsummerfield, line 87: We have `coral source test` already. I think `validate` too close in meaning.

We also have `add` and `import`, which you could argue makes sense, but perhaps we could simplify here too, wdyt:

coral source install <bundled-name>
coral source install --file <path>
coral source lint <path>
coral source test <name>

iaptsiauri:
Closing in favor of https://github.com/withcoral/coral/pull/49

PR #33, opened 2026-04-08
docs: require docs updates for CLI and MCP changes

jsummerfield:
## Why
CLI and MCP behavior is a user-facing contract, so instruction drift between code changes and `docs/` creates avoidable confusion. This change makes the documentation requirement explicit anywhere those surfaces are governed.

## What changed
- add a repo-wide rule that CLI and MCP surface changes must include matching `docs/` updates
- add crate-scoped guidance in `coral-cli` and `coral-mcp` for the specific documented surfaces each adapter owns
- add the same expectation to `docs/AGENTS.md` so the docs area reinforces the rule from its side

## Validation
- run `make validate`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -26,6 +26,8 @@
   its bootstrap seam today.
 - If a caller needs explicit local server control, prefer `coral-client::local`
   over widening the default client surface.
+- Changes to CLI or MCP surfaces must include corresponding documentation
+  updates under `docs/` in the same change.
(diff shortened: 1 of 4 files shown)


AlbertQM (approved):
Wonder if this will be enough, we might need some code to actually detect changes and raise some flags, so agents won't forget. Still, this is good, thanks for adding 🙏

Bradley-Butcher (approved):

PR #37, opened 2026-04-08
feat(app): add OpenTelemetry tracing and metrics

antonmry:
Adds opt-in OpenTelemetry instrumentation to Coral so that traces, metrics,
and logs can be exported to any OTLP-compatible backend (tested with
OpenObserve). Telemetry is **off by default** -- it activates only when a
`[telemetry]` section is present in `config.toml`.

### New: `crates/coral-app/src/telemetry/`

- **`mod.rs`** -- `init_tracing()` / `shutdown_tracing()` configure a
  `tracing-subscriber` registry with up to three layers:
  - **stderr** (compact fmt) -- active only when `log_filter` is set in
    config.
  - **OTLP traces** -- batch span exporter to `{otel_endpoint}/v1/traces`.
  - **OTLP logs** -- log bridge via `opentelemetry-appender-tracing`.
  - **OTLP metrics** -- periodic reader to `{otel_endpoint}/v1/metrics`.
  - Supports `CORAL_TRACE_PARENT` env var for distributed tracing
    (`FixedTraceIdGenerator` pins all spans to the parent trace ID).
- **`config.rs`** -- `TelemetryConfig` loaded from `[telemetry]` in
  `config.toml`. No environment variable overrides -- all telemetry
  settings live in the config file.
- **`metrics.rs`** -- Four query instruments (`coral.query.count`,
  `coral.query.duration`, `coral.query.errors`, `coral.query.rows`) with
  thread-local test support for isolated metric assertions.

## Configuration

```toml
# ~/.config/coral/config.toml
[otel]
endpoint = "http://localhost:4318"          # OTLP/HTTP base URL
headers = "key1=value1, ..."         # comma-separated key=value
log_filter = "info"                              # stderr tracing (off by default)
trace_filter = "coral_app=trace,coral_engine=trace"  # OTLP span filter
service_name = "coral"                      # service.name resource attribute
```

`CORAL_TRACE_PARENT` is the only env var supported -- it propagates a W3C
traceparent for distributed tracing (e.g. when invoked by an AI agent).

## TODO (following PRs)

- Add Datafusion Tracing (requires Datafusion upgrade)
- Update README.md and CONTRIBUTING.md once we have real feedback from internal users
- Add env variables expansion
- ~~Review the -j1 flag in the makefile~~

## Validation

| # | Test | Result |
|---|------|--------|
| 1 | `cargo build -p coral-cli` compiles | PASS |
| 2 | Normal mode produces no stderr tracing | PASS |
| 3 | `log_filter` in config enables stderr tracing | PASS |
| 4 | `coral.query` trace spans arrive in OpenObserve | PASS |
| 5 | OTLP logs arrive in OpenObserve | PASS  |
| 6 | Metrics (`coral_query_count/duration/rows`) arrive | PASS |
| 7 | `CORAL_TRACE_PARENT` links spans to fixed trace ID | PASS |
| 8 | No `[telemetry]` config = no export, no stderr | PASS |

### DIFF
--- a/crates/coral-app/src/bootstrap/server.rs
+++ b/crates/coral-app/src/bootstrap/server.rs
@@ -22,6 +22,7 @@ use crate::query::service::QueryService;
 use crate::sources::manager::SourceManager;
 use crate::sources::service::SourceService;
 use crate::state::{AppStateLayout, ConfigStore, SecretStore};
+use crate::telemetry::TelemetryConfig;
 
 /// Server-side bootstrap configuration for the Coral server.
 #[derive(Clone)]
@@ -119,6 +120,8 @@ impl ServerBuilder {
                 .or_else(|| env.coral_config_dir_override()),
         )?;
         layout.ensure()?;
+        let telemetry_config = TelemetryConfig::load(&layout)?;
+        crate::telemetry::init_tracing(&telemetry_config);
         let config_store = ConfigStore::new(layout.clone());
         let secret_store = SecretStore::new(layout.clone());
         let source_manager =
@@ -179,6 +182,9 @@ impl RunningServer {
         if let Some(task) = task {
             task.await??;
         }
+        tokio::task::spawn_blocking(crate::telemetry::shutdown_tracing)
+            .await
+            .ok();
         Ok(())
     }
 }
@@ -229,6 +235,7 @@ async fn start_server(
 
 #[cfg(test)]
 mod tests {
+    use std::net::{Ipv4Addr, TcpListener};
     use std::sync::Arc;
 
     use coral_api::v1::query_service_client::QueryServiceClient;
@@ -259,8 +266,16 @@ mod tests {
             .add_engine_extensions_provider(Arc::new(NoopEngineExtensionsProvider));
     }
 
+    fn loopback_sockets_available() -> bool {
+        TcpListener::bind((Ipv4Addr::LOCALHOST, 0)).is_ok()
+    }
+
     #[tokio::test]
     async fn file_tilde_sources_resolve_from_app_owned_runtime_context() {
+        if !loopback_sockets_available() {
+            return;
+        }
+
         let temp = TempDir::new().expect("temp dir");
         let fake_home = temp.path().join("fake-home");
         let config_dir = temp.path().join("coral-config");
--- a/crates/coral-app/src/bootstrap/server.rs
+++ b/crates/coral-app/src/bootstrap/server.rs
@@ -121,7 +121,7 @@ impl ServerBuilder {
         )?;
         layout.ensure()?;
         let telemetry_config = TelemetryConfig::load(&layout)?;
-        crate::telemetry::init_tracing(&telemetry_config);
+        crate::telemetry::init_tracing(&telemetry_config)?;
         let config_store = ConfigStore::new(layout.clone());
         let secret_store = SecretStore::new(layout.clone());
         let source_manager =
--- a/crates/coral-app/src/bootstrap/server.rs
+++ b/crates/coral-app/src/bootstrap/server.rs
@@ -29,6 +29,7 @@ use crate::telemetry::TelemetryConfig;
 pub(crate) struct ServerConfig {
     config_dir: Option<PathBuf>,
     engine_extensions_providers: Vec<Arc<dyn EngineExtensionsProvider>>,
+    enable_stderr_logs: bool,
 }
 
 impl Default for ServerConfig {
@@ -44,6 +45,7 @@ impl ServerConfig {
         Self {
             config_dir: None,
             engine_extensions_providers: Vec::new(),
+            enable_stderr_logs: false,
         }
     }
 
@@ -61,6 +63,12 @@ impl ServerConfig {
     ) -> Self {
         self.engine_extensions_providers
             .push(engine_extensions_provider);
+        self
+    }
+
+    #[must_use]
+    pub(crate) fn with_stderr_logs(mut self, enable_stderr_logs: bool) -> Self {
+        self.enable_stderr_logs = enable_stderr_logs;
         self
     }
 }
@@ -102,6 +110,17 @@ impl ServerBuilder {
         self
     }
 
+    #[must_use]
+    /// Enables or disables local stderr log rendering for this server.
+    ///
+    /// `MCP` stdio adapters can enable this for diagnostics while keeping
+    /// stdout reserved for protocol messages. Other command surfaces should
+    /// leave it disabled and rely on OTEL export for logs.
+    pub fn with_stderr_logs(mut self, enable_stderr_logs: bool) -> Self {
+        self.config = self.config.with_stderr_logs(enable_stderr_logs);
+        self
+    }
+
     /// Starts the Coral gRPC server on loopback TCP.
     ///
     /// Coral keeps a real local gRPC boundary here so the public client talks
@@ -121,7 +140,7 @@ impl ServerBuilder {
         )?;
         layout.ensure()?;
         let telemetry_config = TelemetryConfig::load(&layout)?;
-        crate::telemetry::init_tracing(&telemetry_config)?;
+        crate::telemetry::init_tracing(&telemetry_config, self.config.enable_stderr_logs)?;
         let config_store = ConfigStore::new(layout.clone());
         let secret_store = SecretStore::new(layout.clone());
         let source_manager =
--- a/crates/coral-app/src/bootstrap/server.rs
+++ b/crates/coral-app/src/bootstrap/server.rs
@@ -201,9 +201,6 @@ impl RunningServer {
         if let Some(task) = task {
             task.await??;
         }
-        tokio::task::spawn_blocking(crate::telemetry::shutdown_tracing)
-            .await
-            .ok();
         Ok(())
     }
 }
--- a/crates/coral-app/src/query/service.rs
+++ b/crates/coral-app/src/query/service.rs
@@ -5,15 +5,12 @@ use arrow::ipc::writer::StreamWriter;
 use arrow::record_batch::RecordBatch;
 use coral_api::v1::query_service_server::QueryService as QueryServiceApi;
 use coral_api::v1::{ExecuteSqlRequest, ExecuteSqlResponse, ListTablesRequest, ListTablesResponse};
-use opentelemetry::propagation::{Extractor, TextMapPropagator as _};
-use opentelemetry_sdk::propagation::TraceContextPropagator;
 use tonic::{Request, Response, Status};
 use tracing::Instrument as _;
-use tracing_opentelemetry::OpenTelemetrySpanExt as _;
 
 use crate::bootstrap::core_status;
 use crate::query::manager::QueryManager;
-use crate::transport::{query_status, table_to_proto, workspace_name_from_proto};
+use crate::transport::{grpc_span, query_status, table_to_proto, workspace_name_from_proto};
 
 #[derive(Clone)]
 pub(crate) struct QueryService {
@@ -28,58 +25,36 @@ impl QueryService {
     }
 }
 
-struct MetadataExtractor<'a>(&'a tonic::metadata::MetadataMap);
-
-impl Extractor for MetadataExtractor<'_> {
-    fn get(&self, key: &str) -> Option<&str> {
-        self.0.get(key).and_then(|v| v.to_str().ok())
-    }
-
-    fn keys(&self) -> Vec<&str> {
-        use tonic::metadata::KeyRef;
-        self.0
-            .keys()
-            .filter_map(|k| match k {
-                KeyRef::Ascii(key) => Some(key.as_str()),
-                KeyRef::Binary(_) => None,
-            })
-            .collect()
-    }
-}
-
-fn extract_trace_context(metadata: &tonic::metadata::MetadataMap) -> opentelemetry::Context {
-    TraceContextPropagator::new().extract(&MetadataExtractor(metadata))
-}
-
 #[tonic::async_trait]
 impl QueryServiceApi for QueryService {
     async fn list_tables(
         &self,
         request: Request<ListTablesRequest>,
     ) -> Result<Response<ListTablesResponse>, Status> {
-        let request = request.into_inner();
-        let workspace_name = workspace_name_from_proto(request.workspace.as_ref())?;
-        let tables = self
-            .queries
-            .list_tables(&workspace_name)
-            .await
-            .map_err(query_status)?
-            .into_iter()
-            .map(|table| table_to_proto(&workspace_name, table))
-            .collect();
-        Ok(Response::new(ListTablesResponse { tables }))
+        let span = grpc_span(request.metadata(), "list_tables");
+        let queries = self.queries.clone();
+        async move {
+            let request = request.into_inner();
+            let workspace_name = workspace_name_from_proto(request.workspace.as_ref())?;
+            let tables = queries
+                .list_tables(&workspace_name)
+                .await
+                .map_err(query_status)?
+                .into_iter()
+                .map(|table| table_to_proto(&workspace_name, table))
+                .collect();
+            Ok(Response::new(ListTablesResponse { tables }))
+        }
+        .instrument(span)
+        .await
     }
 
     async fn execute_sql(
         &self,
         request: Request<ExecuteSqlRequest>,
     ) -> Result<Response<ExecuteSqlResponse>, Status> {
-        let parent_cx = extract_trace_context(request.metadata());
-        let span = tracing::info_span!("grpc.execute_sql");
-        let _ = span.set_parent(parent_cx);
-
+        let span = grpc_span(request.metadata(), "execute_sql");
         let queries = self.queries.clone();
-
         async move {
             let inner = request.into_inner();
             let workspace_name = workspace_name_from_proto(inner.workspace.as_ref())?;
--- a/crates/coral-app/src/query/service.rs
+++ b/crates/coral-app/src/query/service.rs
@@ -5,7 +5,11 @@ use arrow::ipc::writer::StreamWriter;
 use arrow::record_batch::RecordBatch;
 use coral_api::v1::query_service_server::QueryService as QueryServiceApi;
 use coral_api::v1::{ExecuteSqlRequest, ExecuteSqlResponse, ListTablesRequest, ListTablesResponse};
+use opentelemetry::propagation::{Extractor, TextMapPropagator as _};
+use opentelemetry_sdk::propagation::TraceContextPropagator;
 use tonic::{Request, Response, Status};
+use tracing::Instrument as _;
+use tracing_opentelemetry::OpenTelemetrySpanExt as _;
 
 use crate::bootstrap::core_status;
 use crate::query::manager::QueryManager;
@@ -22,6 +26,29 @@ impl QueryService {
             queries: query_manager,
         }
     }
+}
+
+struct MetadataExtractor<'a>(&'a tonic::metadata::MetadataMap);
+
+impl Extractor for MetadataExtractor<'_> {
+    fn get(&self, key: &str) -> Option<&str> {
+        self.0.get(key).and_then(|v| v.to_str().ok())
+    }
+
+    fn keys(&self) -> Vec<&str> {
+        use tonic::metadata::KeyRef;
+        self.0
+            .keys()
+            .filter_map(|k| match k {
+                KeyRef::Ascii(key) => Some(key.as_str()),
+                KeyRef::Binary(_) => None,
+            })
+            .collect()
+    }
+}
+
+fn extract_trace_context(metadata: &tonic::metadata::MetadataMap) -> opentelemetry::Context {
+    TraceContextPropagator::new().extract(&MetadataExtractor(metadata))
 }
 
 #[tonic::async_trait]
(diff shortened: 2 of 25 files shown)


crates/coral-app/src/bootstrap/server.rs
```diff
@@ -82,6 +83,12 @@ impl ServerBuilder {
                 .or_else(|| env.coral_config_dir_override()),
         )?;
         layout.ensure()?;
+        let telemetry_config = match TelemetryConfig::load(&layout) {
+            Ok(config) => config,
+            Err(AppError::TomlDecode(_)) => TelemetryConfig::default(),
```
jsummerfield: Falling back to `TelemetryConfig::default()` on `TomlDecode` seems risky because the same `config.toml` is still parsed normally by `ConfigStore` later. That means a malformed config can let the server start and then fail only on the first real operation, which feels worse than failing fast at startup. Is there a reason telemetry decode errors should be treated differently from the rest of config parsing?

crates/coral-cli/src/main.rs
```diff
@@ -111,6 +117,7 @@ async fn main() -> Result<(), anyhow::Error> {
         }
         Command::Source(args) => match args.command {
             SourceCommand::Discover => {
+                let app = build_app().await?;
```
jsummerfield: This looks like unnecessary churn to me. In the current CLI-only model, most of these branches still always need an `AppClient`, so moving `build_app().await?` into each arm mostly makes the control flow noisier without buying much. Is there a concrete startup or ownership problem this refactor is solving?

jsummerfield:
This raises a real design question about telemetry ownership.

Right now telemetry is initialized from server bootstrap in [`crates/coral-app/src/bootstrap/server.rs`](https://github.com/withcoral/coral/pull/37/files#diff-94886dfe31f9ee50ca6948bbeeb0a9d06d0057c7ca7e37f457b5d56d0ee27545R86) and the remote parent is stored globally in [`crates/coral-app/src/telemetry/mod.rs`](https://github.com/withcoral/coral/pull/37/files#diff-ac4663aa8dedb9a912b74670a31b4ba4a2fd5c20784d6d80f31329758af817d6R30), but flushing happens from the CLI entrypoint in [`crates/coral-cli/src/main.rs`](https://github.com/withcoral/coral/pull/37/files#diff-498fe0bcbf1ee26390c481c49d0a87d787b29f6b8d671f2fbbdf04a8b30ca1b8R95). That works for the current short-lived CLI flow, but it muddles who actually owns telemetry: is it the server process, or is it the CLI wrapper that happened to start the server?

That matters because this isn’t how tracing should work once we want a real client -> gRPC -> server trace. In the long-term model, I’d expect:
- the CLI to optionally use `CORAL_TRACE_PARENT` to seed its own top-level client span
- the client transport to inject `traceparent` / `baggage` into outgoing gRPC metadata
- the server to extract that context per request at the gRPC boundary
- request/domain spans like `coral.query` to inherit from that extracted request context naturally

In other words, parentage should become request-scoped at the transport boundary, not process-scoped via a global `CORAL_TRACE_PARENT` value attached inside query code in [`crates/coral-app/src/query/manager.rs`](https://github.com/withcoral/coral/pull/37/files#diff-566b9560dad5fb5bef11d3cf5f769bce32f8ddf28745317dce45a865ad6727f3R187).

I think the instrumentation added here is useful, but before we cement this shape I’d like us to be explicit about the intended ownership model:
- if telemetry is owned by the server, init/shutdown should live with the server process entrypoint
- if `CORAL_TRACE_PARENT` is for distributed tracing, it should be a CLI-side seed that gets propagated over gRPC metadata, not server-global state

That’s the part I’m unsure about in the current design.

crates/coral-cli/src/main.rs
```diff
@@ -12,8 +12,8 @@ use bootstrap::bootstrap;
 
 #[tokio::main]
 async fn main() -> Result<(), anyhow::Error> {
-    let bootstrap::Bootstrap { app, _server } = bootstrap().await?;
-    match coral_cli::run(app).await {
+    let bootstrap = bootstrap().await?;
+    let result = match coral_cli::run(bootstrap.app.clone()).await {
```
jsummerfield: `CORAL_TRACE_PARENT` is documented by the new telemetry helper, but the CLI never enters or holds the root span before issuing RPCs. Since `coral-client` injects `tracing::Span::current()` into gRPC metadata, the env seed needs to be read by the CLI and kept alive around `coral_cli::run(...)`; otherwise the advertised distributed trace path is a no-op.

crates/coral-cli/src/main.rs
```diff
@@ -22,5 +22,7 @@ async fn main() -> Result<(), anyhow::Error> {
             }
             Err(error)
         }
-    }
+    };
+    bootstrap.shutdown().await;
```
jsummerfield: This shutdown only runs on the normal `result` path. The `std::process::exit(1)` inside the `CliExitError` branch bypasses it, which means local server teardown and telemetry flushing are skipped on common command failures. Please return an exit status after shutdown, or otherwise ensure this path runs before process exit.

crates/coral-app/src/telemetry/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+        let _ = span.set_parent(parent_cx);
+    }
+    span.entered()
+}
+
+#[allow(
+    clippy::too_many_lines,
+    reason = "Initialization configures three OTLP pipelines in one place"
+)]
+pub(crate) fn init_tracing(config: &TelemetryConfig) {
```
jsummerfield: The lifecycle here is not restart-safe: `init_tracing` is guarded by a process-wide `Once`, but `shutdown_tracing` drains and shuts down the providers. A later `ServerBuilder::start()` in the same process will hit the `Once` no-op and keep subscriber/metric state backed by shut-down providers. The cleaner shape is to treat tracing as process-global and flush at true process exit, or make the server shutdown path flush without permanently draining unrestartable global state.

crates/coral-app/src/telemetry/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+                .build();
+
+            let headers = parse_headers(config.otel_headers.as_deref().unwrap_or_default());
+
+            let trace_exporter = SpanExporter::builder()
+                .with_http()
+                .with_endpoint(normalize_otlp_endpoint(endpoint, "traces"))
+                .with_headers(headers.clone())
+                .build()
+                .expect("failed to build OTLP span exporter");
```
jsummerfield: This can panic from user-provided telemetry config. Since `TelemetryConfig::load` already returns `AppError` and startup is fallible, `init_tracing` should also be fallible and propagate exporter/header/endpoint construction failures instead of using `expect(...)`.

crates/coral-client/src/client.rs
```diff
@@ -54,8 +49,9 @@ impl AppClient {
         let endpoint = Endpoint::from_shared(endpoint_uri.to_string())?
             .http2_max_header_list_size(HTTP2_MAX_HEADER_LIST_SIZE);
         let channel = endpoint.connect().await?;
-        let source_client = SourceServiceClient::new(channel.clone());
-        let query_client = QueryServiceClient::new(channel)
+        let source_client =
+            SourceServiceClient::with_interceptor(channel.clone(), TraceContextInterceptor);
```
jsummerfield, line 54: Now that both generated clients inject W3C trace metadata, the app side should extract it consistently at the gRPC boundary. Today only `execute_sql` extracts the parent context; source lifecycle calls and `list_tables` discard the metadata, so traces split for those RPCs.

jsummerfield (commented):
I think the overall telemetry direction is good, but the current shape mixes three ownership concerns: server-owned telemetry lifecycle, client-owned gRPC propagation, and CLI-owned environment/process behavior. The cleanest fix is a targeted refactor rather than a larger abstraction: keep OTLP provider/config ownership in `coral-app`, keep trace metadata injection in `coral-client`, move `CORAL_TRACE_PARENT` env handling into `coral-cli`, and make startup/shutdown semantics explicit and fallible.

crates/coral-app/src/telemetry/config.rs
```diff
@@ -0,0 +1,110 @@
...
+#[allow(
+    clippy::struct_field_names,
+    reason = "field names mirror the TOML config keys for direct serde deserialization"
+)]
+pub struct TelemetryConfig {
+    pub(crate) otel_endpoint: Option<String>,
+    pub(crate) otel_headers: Option<String>,
+    pub(crate) otel_log_filter: Option<String>,
+    pub(crate) otel_trace_filter: String,
+    pub(crate) otel_service_name: String,
```
jsummerfield: Do we need these for v1? I'd be tempted to start with just `otel_endpoint`.

crates/coral-app/src/telemetry/config.rs
```diff
@@ -0,0 +1,110 @@
...
+/// Telemetry settings loaded from `config.toml`.
+#[derive(Debug, Clone, Deserialize, PartialEq, Eq)]
+#[serde(default)]
+#[allow(
+    clippy::struct_field_names,
+    reason = "field names mirror the TOML config keys for direct serde deserialization"
+)]
+pub struct TelemetryConfig {
+    pub(crate) otel_endpoint: Option<String>,
+    pub(crate) otel_headers: Option<String>,
```
jsummerfield: I don't think we should be encouraging secrets (bearer tokens) in config.toml. We've used env/secret-store support for secrets so far.

crates/coral-app/src/telemetry/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+/// Enters a root CLI span, optionally parented to the W3C `traceparent` value
+/// in the `CORAL_TRACE_PARENT` environment variable.
+///
+/// The returned [`tracing::span::EnteredSpan`] must be kept alive for the
+/// duration of the CLI invocation — dropping it exits the span.
+#[allow(
+    clippy::disallowed_methods,
+    reason = "CORAL_TRACE_PARENT is intentionally read from the environment as a per-invocation CLI seed"
+)]
+pub fn seed_root_span() -> tracing::span::EnteredSpan {
```
jsummerfield: Even once the CLI wires this in, returning an `EnteredSpan` is not the right async shape. Holding an entered span guard across `.await` points can attach work on the task to the wrong current span; the usual `tracing` pattern is to return/create a `Span` and run the CLI future with `.instrument(span).await` so the context is polled correctly.

crates/coral-client/src/propagation.rs
```diff
@@ -0,0 +1,34 @@
...
+#[derive(Clone)]
+pub struct TraceContextInterceptor;
+
+impl tonic::service::Interceptor for TraceContextInterceptor {
+    fn call(
+        &mut self,
+        mut request: tonic::Request<()>,
+    ) -> Result<tonic::Request<()>, tonic::Status> {
+        let cx = tracing::Span::current().context();
+        TraceContextPropagator::new()
```
jsummerfield: This works, but it is not the usual OTel setup shape. Rather than constructing a fresh `TraceContextPropagator` at every injection/extraction site, configure the process propagator once during telemetry initialization and use the global propagator for inject/extract. That keeps client and server propagation policy in one place and makes it straightforward to add baggage or a composite propagator later.

crates/coral-app/src/telemetry/metrics.rs
```diff
@@ -0,0 +1,204 @@
...
+
+fn build_metrics(meter: &Meter) -> Metrics {
+    Metrics {
+        count: meter
+            .u64_counter("coral.query.count")
+            .with_unit("{queries}")
+            .with_description("Total queries executed")
+            .build(),
+        duration: meter
+            .f64_histogram("coral.query.duration")
```
jsummerfield, line 29: For OTel metrics, this would be more idiomatic if query duration used seconds (`s`) rather than milliseconds, and if common dimensions such as `status` were recorded as metric attributes instead of only as a separate error counter. That gives backends a single query-count/duration stream that can be grouped by outcome without needing Coral-specific instrument conventions.

crates/coral-app/src/query/service.rs
```diff
@@ -24,6 +28,29 @@ impl QueryService {
...
+            .keys()
+            .filter_map(|k| match k {
+                KeyRef::Ascii(key) => Some(key.as_str()),
+                KeyRef::Binary(_) => None,
+            })
+            .collect()
+    }
+}
+
+fn extract_trace_context(metadata: &tonic::metadata::MetadataMap) -> opentelemetry::Context {
```
jsummerfield: This extraction helper is the right kind of operation, but it should become a consistent server transport-boundary pattern rather than a one-off used only by `execute_sql`. Since `coral-client` now injects context for both source and query clients, `coral-app` should extract/instrument incoming RPCs through shared service/interceptor-style plumbing or a small helper applied to every gRPC method.

crates/coral-client/src/lib.rs
```diff
@@ -29,6 +30,7 @@ use coral_api::v1::ExecuteSqlResponse;
 use serde_json::Value;
 
 pub use client::{AppClient, DEFAULT_WORKSPACE_ID, QueryClient, SourceClient, default_workspace};
+pub use coral_app::seed_root_span;
```
jsummerfield: This re-export widens `coral-client` with a helper that reads a CLI-specific env var and manages span lifetime. The cleaner seam is for `coral-client` to inject whatever tracing context is already current on outgoing RPCs, while `coral-cli` owns seeding that context from `CORAL_TRACE_PARENT`. That keeps the client surface transport-focused and avoids pulling app/CLI lifecycle behavior through this crate.

crates/coral-cli/src/main.rs
antonmry: Fixed here https://github.com/withcoral/coral/pull/37/changes/a0fac9834b13bbeba4b30d2fc6d50c25f729fafd

crates/coral-app/src/telemetry/metrics.rs
antonmry: fixed on 2fdbaf9edbd31bf18693158a9974a4cbb8ef94da and 9bfc93be20ad9e84de1b0fc59fc5b386be4f8ac3

crates/coral-app/src/query/service.rs
antonmry: Implemented here d8dc92b45f659a24328196dc1a3b6031557d167f

crates/coral-client/src/client.rs
antonmry: implemented here d8dc92b45f659a24328196dc1a3b6031557d167f

crates/coral-client/src/propagation.rs
antonmry: Fixed here fdb21e5377a44d1daa039613c9e6ccb206f2c02f

crates/coral-app/src/telemetry/mod.rs
antonmry: Fixed with 5dd29bee889830a827e11755bad1b37d29dacab1

crates/coral-cli/src/main.rs
antonmry: Fixed with 5dd29bee889830a827e11755bad1b37d29dacab1 

It's a bigger change that I would like, but I think it's needed.

crates/coral-client/src/lib.rs
antonmry: Fixed with 5dd29bee889830a827e11755bad1b37d29dacab1

crates/coral-app/src/telemetry/config.rs
antonmry: SecretStore is currently tightly scoped to sources, it seems to require a big refactor to make it work here. In addition:

- auth is different between otel vendors. Even if we store the credential in SecretStore, we'll need a way to refer to it in `otel_headers`. For example, New Relic expects a header `api-key` but Datadog expects a header `DD-API-KEY`
- `otel_headers` can be used for things that aren't credentials like custom headers, user agents, etc. 
- Most of the tokens used to report telemetry aren't allowed to do anything more (that's the case of NewRelic or Datadog)

Proposal:
- Support otel_headers as they are given the choice to the user.
- We can open a different PR or add to the backlog the ability to use SecretStore to store credentials used in the config file.

crates/coral-app/src/telemetry/config.rs
antonmry: I think they are needed by v1. We can work without them, but the user experience is going to pretty bad if they can't customize log/trace level (in particular for dependencies).

`otel_service_name` isn't a big deal with the default `coral`, but again, it can be annoying not having the option if the user has some nomenclature or app versioning in place.

crates/coral-app/src/telemetry/config.rs
antonmry: Added to the backlog: SOURCE-491

crates/coral-app/src/telemetry/mod.rs
antonmry: Fixed here 69b954fe78bb1d8cf78dcc7bb0422eaca24b3085

crates/coral-app/src/telemetry/mod.rs
antonmry: Fixed here 69b954fe78bb1d8cf78dcc7bb0422eaca24b3085

antonmry:
@cubic-dev-ai review

crates/coral-app/src/telemetry/config.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/// Telemetry settings loaded from `config.toml`.
+#[derive(Debug, Clone, Deserialize, PartialEq, Eq)]
+#[serde(default)]
+#[allow(
+    clippy::struct_field_names,
+    reason = "field names mirror the TOML config keys for direct serde deserialization"
+)]
+pub struct TelemetryConfig {
```
jsummerfield, line 21: Consider renaming this to `[otel]` and then dropping the `otel_` from each key. This is what Codex does and keeps things tidy. https://developers.openai.com/codex/config-advanced#observability-and-telemetry

jsummerfield:
On the telemetry lifecycle issue: the concern is not the normal one-shot CLI process starting again after exit. It is that `ServerBuilder::start()` / `RunningServer::shutdown()` exposes a same-process local-server lifecycle, and the implementation currently mixes two different lifetimes.

`INIT: OnceLock<_>` makes tracing initialization process-global and one-time, but `shutdown_tracing()` treats the OTLP providers as server-lifetime resources and drains/drops them during every `RunningServer::shutdown()`. In an integration test, embedding app, or long-lived process that starts a local server, shuts it down, and later starts another one, the second `start()` sees `INIT` as already initialized and skips rebuilding providers. That leaves the subscriber/global state pointing at providers that were already shut down or removed.

I think the cleaner contract is to make telemetry process-scoped: server shutdown should stop the gRPC server, while the binary/process owner should flush telemetry at actual process exit. If the intended contract is instead “only one local server lifecycle per process,” then `ServerBuilder::start()` should encode that explicitly and fail clearly on subsequent starts, but that would be surprising for this API and for test usage.

jsummerfield:
`Registry::init()` is still a startup panic path here. Exporter construction now returns `AppError`, but subscriber installation still assumes Coral owns the process-global tracing subscriber. That is not always true for this API surface: tests, embedding apps, or another binary in the same process may already have installed a subscriber before calling `ServerBuilder::start()`.

Since this runs under startup rather than process `main`, I think this should use `try_init()` and convert the failure into a clear `AppError`, or intentionally define an existing-subscriber policy and make that a no-op. Either way, `ServerBuilder::start()` should not be able to panic because global tracing was already configured by the host process.

jsummerfield:
One more propagation edge case: `TraceContextInterceptor` now injects through the global text-map propagator, but that propagator is only installed by local `coral-app` telemetry initialization. A pure `AppClient::connect()` path, including external endpoint / test-server usage, can therefore keep the default no-op propagator and silently omit `traceparent` even when the caller has an active tracing span.

I think propagation setup needs to be independent from OTLP exporter/server startup, or the client needs an explicit propagation policy instead of depending on `coral-app` having initialized globals first. Otherwise distributed tracing works for the local bootstrap path but not for the thinner client-only path.

jsummerfield (approved):
Looking great now, thanks for all the hard work on this. Left a few final nits.

crates/coral-app/src/telemetry/config.rs
antonmry: Fixed here 6399d566c2a6419438d14e528715c1bf0d57103a

antonmry:
Thanks for the detailed review.  Summary of how each one was resolved below.

#### 1. Telemetry lifecycle mixed with server lifecycle

Make telemetry process-scoped: server shutdown only stops the
gRPC server, the binary owner flushes telemetry at process exit. Did not pick
the alternative ("one-server-per-process, fail on second start") — that
contract would have been surprising for tests and embedders.

**Commit:** b0c1aae — `test(server): add lifecycle repeatability test and fix tracing shutdown scope`

- `RunningServer::shutdown_inner` no longer calls `shutdown_tracing`.
- `coral-cli/src/main.rs` now calls `coral_app::shutdown_tracing` once after `bootstrap.shutdown().await`, before either return path (covers the `process::exit(1)` branch too).
- Pinned by `server_lifecycle_can_repeat_within_process` in `crates/coral-app/tests/grpc/server_lifecycle_tests.rs` — does two start/shutdown cycles in the same process and asserts both serve a `list_sources` request.

#### 2. `Registry::init()` panic on host-owned subscriber

Went with the no-op-success policy you mentioned as the second
option. Considered three alternatives:

- **Typed `AppError::FailedPrecondition`.** First implementation; works but
  forces every caller (CLI, embedders) to handle a non-recoverable error
  branch. Also caused a subtle variant-flip in the `OnceLock<Result<(), String>>`
  cache (subsequent calls reconstructed the error as `InvalidInput`).
- **Typed `TelemetryStatus { Active, HostOwned }` return.** Cleaner than the
  error variant, but pushes the policy decision out to every caller for an
  edge case where there's only one sensible policy ("warn and continue").
- **No-op success with `tracing::warn!`** (chosen). `try_init` failure now
  emits a single warning through the host's already-installed subscriber and
  returns `Ok(())`. Trade-off: the OTLP providers built before `try_init`
  failed are orphaned in the static slots — they're drained as no-ops by
  `shutdown_tracing` at process exit.

The variant-flip in the `OnceLock` cache went away as a side effect: the only
error path remaining is exporter construction (already `InvalidInput`).

**Commit:** 684a83c — `fix(telemetry): gracefully skip subscriber init when host already has one`

- Both `try_init` call sites in `try_init_tracing` now `tracing::warn!` and `return Ok(())` on conflict.
- Pinned by `host_subscriber_does_not_block_server_startup` in `crates/coral-app/tests/telemetry_host_subscriber.rs`. Lives in its own integration test binary because the `INIT: OnceLock` is process-global — co-locating with the lifecycle test would let the cached success short-circuit `try_init` and hide the conflict path.

#### 3. Propagator install coupled to coral-app telemetry init

Each crate owns its boundary: `coral-app` keeps installing the
propagator on server start (server-side extraction), `coral-client` now
installs it on `AppClient::connect` (client-side injection). Both call
`set_text_map_propagator(TraceContextPropagator::new())`, both gated by
`OnceLock`, so when both crates run in the same process they end up at the
same end state. Considered injecting the propagator into the interceptor
explicitly (no global state) but it ripples through `SourceClient` /
`QueryClient` type aliases for no real win unless we want multiple
propagators per process.

**Commit:** ab53c74 — `feat(client): ensure W3C trace propagator is installed before connecting`

- New `ensure_global_propagator()` in `crates/coral-client/src/propagation.rs`, `OnceLock`-guarded.
- `AppClient::connect` calls it before establishing the channel.
- New direct dep on `opentelemetry_sdk` in `coral-client` (the SDK crate owns `TraceContextPropagator`).

PR #38, opened 2026-04-08
ci(release): trigger homebrew tap bump

simonwhitaker:
This updates the release workflow to dispatch `coral-release` to `withcoral/homebrew-tap` after publishing a GitHub release so the Homebrew cask can be bumped automatically.
The prepare step now exposes the raw workspace version as a workflow output, which lets the dispatch payload use the existing semver string directly instead of reconstructing it from the `v`-prefixed tag.
The release job also scopes the GitHub App token to `homebrew-tap` so the dispatch call can access the target repository.
Validation: `make validate`.

### DIFF
--- a/.github/scripts/prepare_release.py
+++ b/.github/scripts/prepare_release.py
@@ -188,6 +188,7 @@ def prepare_outputs() -> dict[str, str]:
 
     return {
         "should_release": "true",
+        "version": current_version,
         "build_ref": resolved_sha,
         "tag_name": tag_name,
         "release_name": tag_name,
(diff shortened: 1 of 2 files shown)


simonwhitaker:
@jsummerfield: Worth pointing out, the workflow that this triggers [over in withcoral/homebrew-tap](https://github.com/withcoral/homebrew-tap/blob/main/.github/workflows/bump-coral.yml) currently fails, because it tries to fetch release details from this repo's releases (e.g. checksum SHAs) in order to update the Homebrew cask, and this repo isn't yet public, so that attempt fails. The workaround to have it work with a private repo is involved, and not worth the effort; once this repo is public, that workflow will work as intended.

jsummerfield (approved):

PR #45, opened 2026-04-09
ci(validate): split validation checks and add status gate

simonwhitaker:
## Summary
- split `.github/workflows/validate.yml` into change detection, Python helper tests, Rust validation, GitHub Actions security audit, and a final status job
- add path filters so the Python helper and Rust validation jobs only run when their relevant files change
- add GHA security audit using [zizmor](https://docs.zizmor.sh/)
- rename the local Rust validation target to `make rust-checks`, to make its responsibilities clearer, and update the contributor docs to match

## Testing
- `zizmor .github/workflows/*.yml`
- `make rust-checks`


.github/workflows/validate.yml
```diff
@@ -17,44 +16,78 @@ env:
...
 
       - name: Ruff lint
-        if: steps.release-script-changes.outputs.release_python == 'true'
         run: uvx ruff check .github/scripts
 
       - name: Run release script integration tests
-        if: steps.release-script-changes.outputs.release_python == 'true'
         run: python3 -m unittest discover -s .github/scripts -p 'test_prepare_release_integration.py'
 
+  validate-rust:
```
jsummerfield: nit, pick one: call it rust-checks, or rename the Make command

.github/workflows/validate.yml
```diff
@@ -17,44 +16,78 @@ env:
...
+      release_python: ${{ steps.filter.outputs.release_python }}
+      github_actions: ${{ steps.filter.outputs.github_actions }}
+      validate_rust: ${{ steps.filter.outputs.validate_rust }}
 
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           persist-credentials: true
 
       - name: Detect release script changes
```
jsummerfield: This name seems stale now?

jsummerfield (approved):

PR #49, opened 2026-04-09
feat(cli): add lint command

iaptsiauri:
## Summary

Adds `coral source lint` and consolidates manifest validation into a single acceptance path in `coral-spec`.

**Spec validation API (coral-spec)**
- `parse_manifest_and_inputs(raw)` returns `(ValidatedSourceManifest, Vec<ManifestInputSpec>)`. Becomes single acceptance path used by `lint, source add --file, server-side import, bundled install, and discover`. 
- Manifest is now validated before prompting for variables/secrets.

**Performance: single YAML parse**
- Enabled `serde_json/preserve_order`, `inputs.rs` now uses `serde_json::Value`

**Cleanups**
- `catalog::describe_manifest `collapsed from separate parse + input collection calls into a single `parse_manifest_and_inputs` call.
- Removed duplicate private InputKind enum (use ManifestInputKind directly).

## Test plan

- [x] cargo test --workspace passes (162 tests)
- [x] coral source lint <valid.yaml> exits 0 and prints "Manifest is valid"
- [x] coral source lint <missing-backend.yaml> prints schema validation error
- [x] coral source lint <duplicate-columns.yaml> prints semantic validation error
- [x] coral source lint <inconsistent-inputs.yaml> prints "declared inconsistently"
- [x] coral source add --file <broken.yaml> fails before prompting


antonmry (approved):

crates/coral-cli/tests/lint.rs
```diff
@@ -0,0 +1,111 @@
...
+)]
+
+use std::io::Write;
+use std::process::{Command, Stdio};
+
+use tempfile::NamedTempFile;
+
+fn coral_lint(file: &std::path::Path) -> std::process::Output {
+    Command::new(env!("CARGO_BIN_EXE_coral"))
+        .args(["source", "lint", file.to_str().unwrap()])
```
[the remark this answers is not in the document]
antonmry: This one seems valid

crates/coral-cli/tests/lint.rs
```diff
@@ -0,0 +1,128 @@
...
+fn temp_path(suffix: &str) -> PathBuf {
+    let nanos = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .expect("system time before unix epoch")
+        .as_nanos();
+    std::env::temp_dir().join(format!(
+        "coral-lint-test-{}-{nanos}{suffix}",
+        std::process::id()
+    ))
+}
```
jsummerfield: Let's use tempfile::tempdir() for this.

jsummerfield:
One behavioral gap in the new `coral source lint` flow: it only runs `parse_source_manifest_yaml`, while `source add --file` and import/discovery also run input collection via `collect_source_inputs_yaml`.

That means lint is currently weaker than install. A manifest can print `Manifest is valid` and still be rejected later during add/import if its variable/secret declarations are inconsistent.

Concrete example: declaring `API_BASE` twice with different defaults passes lint, but `collect_source_inputs_yaml` rejects it with `manifest input 'API_BASE' is declared inconsistently`.

I think `lint` should exercise the same input-collection validation path as add/import, otherwise it can produce false positives.

jsummerfield:
Architecture-wise, I think the underlying issue is that “is this manifest valid?” does not have a single owner yet.

Right now that answer is assembled from two separate `coral-spec` paths:
- `parse_source_manifest_yaml(...)` for syntax/schema/manifest semantics
- `collect_source_inputs_yaml(...)` for install-time input consistency

That split is fine internally, but I would not leave it for the CLI to compose. Per the repo rules, the CLI should stay thin and avoid owning business logic.

I think the cleaner shape is to make `coral-spec` expose one authoritative manifest-acceptance path that all callers use (`lint`, `source add --file`, app import/discovery, catalog, etc.). Either:
- fold input-consistency validation into the main parse/validate path, or
- add a higher-level `coral-spec` API that returns both the validated manifest and normalized inputs

The important part is that the validity contract lives in `coral-spec`, not in `coral-cli`, so we do not get drift between lint/install/discovery again.

crates/coral-cli/tests/lint.rs
iaptsiauri: I used it, and it was flagged with clappy. As it was not used in the main file, adding it to main file just for tests felt little bit hacky. Will bring it back so you can take a look and let's decide

iaptsiauri:
> Architecture-wise, I think the underlying issue is that “is this manifest valid?” does not have a single owner yet.
> 
> Right now that answer is assembled from two separate `coral-spec` paths:
> 
> * `parse_source_manifest_yaml(...)` for syntax/schema/manifest semantics
> * `collect_source_inputs_yaml(...)` for install-time input consistency
> 
> That split is fine internally, but I would not leave it for the CLI to compose. Per the repo rules, the CLI should stay thin and avoid owning business logic.
> 
> I think the cleaner shape is to make `coral-spec` expose one authoritative manifest-acceptance path that all callers use (`lint`, `source add --file`, app import/discovery, catalog, etc.). Either:
> 
> * fold input-consistency validation into the main parse/validate path, or
> * add a higher-level `coral-spec` API that returns both the validated manifest and normalized inputs
> 
> The important part is that the validity contract lives in `coral-spec`, not in `coral-cli`, so we do not get drift between lint/install/discovery again.

Agreed, `coral-spec` should be the central owner of the manifest and manifest validation rules. Will do the rearrangement to address that

jsummerfield (approved):

simonwhitaker (approved):

PR #60, opened 2026-04-10
feat(cli): auto-validate after adding a source

AlbertQM:
## Summary

- `source add`  now automatically validate the source after installation and pretty-print discovered tables (with truncation). Validation failures are surfaced as warnings — the command still exits successfully since the source is already persisted.
- `source test` now uses the same rich tree format (without truncation, showing all tables).
- Moves `print_validation_pretty` from `onboard.rs` into `source_ops.rs` so onboarding, `source add`, `source import`, and `source test` all share one code path.

## Test plan

image
image
image

### DIFF
--- a/crates/coral-cli/src/source_ops.rs
+++ b/crates/coral-cli/src/source_ops.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeMap;
 use std::io::{IsTerminal, stdin, stdout};
 use std::path::Path;
 
@@ -10,8 +11,11 @@ use coral_client::{AppClient, default_workspace};
 use coral_spec::{
     ManifestInputKind, ManifestInputSpec, ValidatedSourceManifest, parse_manifest_and_inputs,
 };
+use dialoguer::console::style;
 use dialoguer::{Input, Password, theme::ColorfulTheme};
 use tonic::Request;
+
+pub(crate) const MAX_TABLES_PER_SCHEMA: usize = 9;
 
 pub(crate) async fn discover_sources(
     app: &AppClient,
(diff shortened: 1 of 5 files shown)


iaptsiauri:
Check this PR https://github.com/withcoral/coral/pull/49 it adds `lint` command. Additionally it forces sources to be validated before installation. But it for sure is lacking the pretty printing. You can take this branch as your base and add pretty printing to it

simonwhitaker (approved):

crates/coral-cli/src/source_ops.rs
```diff
@@ -184,17 +188,74 @@ pub(crate) fn source_origin_label(origin: i32) -> &'static str {
     }
 }
 
-pub(crate) fn print_validation_success(
+pub(crate) async fn validate_and_print(
+    app: &AppClient,
+    source_name: &str,
+    max_tables: Option<usize>,
```
jsummerfield: `None` meaning “show all tables” is not very self-describing. I’d prefer something like:

```rust
enum TableDisplayLimit {
    All,
    Max(usize),
}
```

or

```rust
struct ValidationPrintOptions {
    max_tables_per_schema: Option<usize>,
}
```

jsummerfield:
Worth calling out this changes the output from something semi-machine parseable to something human-focussed (borrowing the interactive onboarding formatting).

That's fine for now, but we should aim to support `--format json` for all commands (just `coral sql` today) to better support scripting.

jsummerfield (approved):

crates/coral-cli/src/source_ops.rs
AlbertQM: Much better, will update. Thanks!

PR #61, opened 2026-04-10
feat(engine): add structured QueryError contract and wire format

mystic123:
## Summary

Adds the `QueryError` type, its JSON wire format, and a client-side decoder. This is a pure foundation change — no production code path calls it yet. Later changes in the structured-error work will have the engine produce these at the failure origin and the CLI/MCP render them, but the type and wire format land here so those changes stay focused on their own concerns.

### Motivation

`CoreError` today has 6 generic buckets, each wrapping a bare `String`. When a query fails on a missing filter, users see `Error: failed precondition: github.issues requires WHERE repo = <constant>` — the facts are correct but there's no actionable hint, no structured fields, and MCP agents can't pattern-match on the failure. A structured error type lets the engine attach a hint and populated fields at the failure origin that the CLI, MCP, and any scripted consumer can surface consistently.

### What this adds

**`crates/coral-engine/src/contracts/query_error.rs`** (new)
- `QueryError { schema_version, code, summary, detail, hint, retryable, fields }`.
- `QueryErrorCode` — 8 variants for the local-only release: `Unknown`, `EmptyQuery`, `SqlError`, `MissingRequiredFilter`, `UnknownField`, `TableNotFound`, `InvalidQueryShape`, `ProviderRequestFailed`.
- `QueryErrorFields { schema, table, field, source, http_status, http_method, url }` — optional structured fields for programmatic consumers.
- Constructors populate hints at the failure origin:
  - `table_not_found` — schema-aware dispatch (`coral` / `public` / `<source>`). Source-specific arm suggests `coral source add <name>`.
  - `provider_request` — HTTP-status dispatch (400 → invalid shape, 401/403 → auth, 404 → not found, 429/5xx → retryable server error). 401 hint suggests `coral source add <name>` to re-auth.
  - `unknown_field(field, hint)` — caller-supplied "Did you mean" hint.
- `to_plain_message()` — degraded fallback renderer (summary / detail / `Hint: …`) for the `Status::message()` path when `Status::details()` is stripped or undecodable.

**Wire format**
- `QueryError::to_json_bytes()` / `QueryError::from_json_bytes(&[u8])` — serde_json via a required `schema_version = "coral.query_error.v1"` sentinel. Payloads missing or mismatching the sentinel are rejected so unrelated bytes in `tonic::Status::details()` cannot be misclassified as a structured query error.

**`crates/coral-client/src/error.rs`**
- `query_error_from_status(&tonic::Status) -> Option<QueryError>` — reads `Status::details()` and returns `None` for empty, unrelated, or mismatched-version payloads so callers can fall back to `Status::message()`.
- Re-exported from `coral-client`'s crate root.

**SQL and shell escaping in hints.** Source and schema names can currently contain spaces, quotes, and shell metacharacters (workspace validation in `coral-app/src/workspaces/manager.rs:48` only rejects path separators). Names interpolated into copy-paste-able SQL literals and shell commands are routed through private `sql_literal` / `shell_arg` helpers so pathological names don't break the rendered hint:
- `sql_literal` — ANSI doubling of embedded single quotes (`o'brien` → `'o''brien'`).
- `shell_arg` — verbatim when `[a-zA-Z0-9_-]+`, otherwise POSIX single-quoted with `'\''` escaping (`foo bar` → `'foo bar'`, `foo'bar` → `'foo'\''bar'`).

### What this does not touch

- `coral-engine/src/runtime/query.rs` — the engine still emits the old string-flattened `CoreError` variants.
- `coral-app/src/bootstrap/error.rs` — no `tonic::Status` encoder lives here yet.
- The CLI and MCP rendering paths.

## Test plan

31 new unit tests across 2 crates. `make validate` green end-to-end.

- [x] `cargo check --workspace --all-targets --all-features --locked`
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo test --workspace --all-targets --all-features --locked` — **182 passing, 1 pre-existing ignored, 0 failures**
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --all-features --no-deps`
- [x] Regression test for missing `schema_version` sentinel
- [x] Regression test for mismatched `schema_version`
- [x] Regression test for unrelated JSON payload
- [x] Round-trip test for every constructor output through `to_json_bytes` / `from_json_bytes`
- [x] `table_not_found_escapes_unsafe_schema_in_hint` — `foo'bar` renders as `'foo''bar'` in SQL and `'foo'\''bar'` in shell
- [x] `provider_request_401_quotes_unsafe_source_in_hint` — `foo bar` renders as `'foo bar'` in shell
- [x] End-to-end decoder test building `Status::with_details` manually and extracting via `query_error_from_status`


crates/coral-client/src/error.rs
```diff
@@ -27,3 +29,61 @@ pub enum QueryResultError {
     #[error("invalid query result: {0}")]
     InvalidResponse(String),
 }
+
+/// Extracts a structured [`QueryError`] from the `details` payload of a
+/// `tonic::Status`, if one is present.
+///
+/// The engine and app attach a JSON-encoded `QueryError` to query failures via
+/// `coral_app::bootstrap::status_with_query_error`. CLI and MCP consumers call
```
antonmry: We'll have other consumers in the future, there's a risk of this comment to be stale soon

mystic123: Thanks — fair point. Fixed in 085104a. I extended the scope beyond this one line: the same "CLI and MCP consumers" / "MCP agents" / "CLI renderers" pattern showed up in five more places across `query_error.rs` and this file, so I swept all six at once and replaced them with generic terms (`consumers`, `renderers`, `automated callers`) as appropriate.

One reference at `query_error.rs:10` (the "`coral-app` status encoder and `coral-client` status decoder" line in the module doc) is kept as-is — it describes where encode/decode code lives in the crate topology rather than enumerating callers, so it doesn't carry the same staleness risk.

crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -0,0 +1,793 @@
...
+//! [`SCHEMA_VERSION`] sentinel keeps decoders from mistaking unrelated
+//! `tonic::Status` details payloads for ours.
+
+use serde::{Deserialize, Serialize};
+
+/// Wire-format sentinel carried as the `schema_version` field on every encoded
+/// [`QueryError`]. Decoders reject payloads whose `schema_version` does not
+/// match, so unrelated bytes in `tonic::Status::details()` don't get mistaken
+/// for a structured query error.
+pub const SCHEMA_VERSION: &str = "coral.query_error.v1";
```
antonmry, line 20: I wonder if it would make sense to is to move the contract into coral-api and version it
  structurally, instead of relying on a JSON sentinel string.

This is what Codex proposed:

Move structured `QueryError` into `coral-api` and version it structurally instead of treating it as an engine-owned Rust type serialized as ad-hoc JSON with a `SCHEMA_VERSION` sentinel.

Why this is a better fit

- `QueryError` crosses the client/server boundary, so it is transport contract, not engine-internal contract.
- The repo already versions transport structurally through `coral.v1` protobuf packages in `coral-api`.
- This avoids pressure for `coral-client` to depend on `coral-engine`.
- It replaces the JSON sentinel string with an explicit typed envelope.

Proposed wire shape

Add a new protobuf message in `crates/coral-api/proto/coral/v1/query.proto` or a new `errors.proto`:

```proto
message StatusDetail {
  oneof kind {
    QueryErrorDetail query_error_v1 = 1;
  }
}

message QueryErrorDetail {
  QueryErrorCode code = 1;
  string summary = 2;
  string detail = 3;
  optional string hint = 4;
  bool retryable = 5;
  QueryErrorFields fields = 6;
}

enum QueryErrorCode {
  QUERY_ERROR_CODE_UNSPECIFIED = 0;
  QUERY_ERROR_CODE_EMPTY_QUERY = 1;
  QUERY_ERROR_CODE_SQL_ERROR = 2;
  QUERY_ERROR_CODE_MISSING_REQUIRED_FILTER = 3;
  QUERY_ERROR_CODE_UNKNOWN_FIELD = 4;
  QUERY_ERROR_CODE_TABLE_NOT_FOUND = 5;
  QUERY_ERROR_CODE_INVALID_QUERY_SHAPE = 6;
  QUERY_ERROR_CODE_PROVIDER_REQUEST_FAILED = 7;
}

message QueryErrorFields {
  optional string schema = 1;
  optional string table = 2;
  optional string field = 3;
  optional string source = 4;
  optional uint32 http_status = 5;
  optional string http_method = 6;
  optional string url = 7;
}
```

Why the `StatusDetail` wrapper matters

- `tonic::Status::details()` is just opaque bytes.
- A raw protobuf payload is not a strong discriminator on its own.
- The outer `oneof` gives a real type tag: decode failed means not ours, empty kind means not ours, `query_error_v1` means ours.
- This replaces the role currently played by `pub const SCHEMA_VERSION: &str = "coral.query_error.v1";`.

Ownership split

- `coral-api` owns the wire types.
- `coral-engine` owns query classification and hint generation.
- `coral-app` owns mapping query failures into `tonic::Status`.
- `coral-client` owns decoding plus fallback behavior.

That means the query-specific constructors such as `table_not_found(...)` and `provider_request(...)` do not need to move into `coral-api`. Only the transport-visible message shape moves there.

Compatibility story

- Additive changes stay in `coral.v1` by adding new fields with new field numbers.
- Old clients ignore unknown fields.
- New clients tolerate missing fields via defaults and `optional`.
- Breaking changes can use a new `oneof` arm such as `query_error_v2 = 2`, or a future `coral.v2` package.

Recommended API shape in `coral-client`

Keep the low-level decode semantics optional, because "no structured payload present" is a normal case. Do not expose only `Option<QueryError>` as the main public API, because that leaks fallback policy to callers.

Prefer a wrapper like:

```rust
pub enum DecodedStatusError {
    Structured(QueryErrorDetail),
    Plain(String),
}
```

Then `coral-client` can centralize:

- decode `StatusDetail` from `status.details()`
- return `Structured(...)` when `query_error_v1` is present
- fall back to `status.message()` otherwise

Migration plan

1. Define `StatusDetail`, `QueryErrorDetail`, `QueryErrorCode`, and `QueryErrorFields` in `coral-api`.
2. Regenerate `coral-api` bindings.
3. Convert engine-side query classification code to build the API message shape or convert into it at the app boundary.
4. Update `coral-app` status encoding to attach protobuf-encoded `StatusDetail` bytes to `tonic::Status::details()`.
5. Update `coral-client` to decode `StatusDetail` and expose a higher-level wrapper that handles the plain fallback centrally.
6. Keep `status.message()` populated with the degraded plain-text rendering so stripped details and version skew still produce useful output.
7. Remove the JSON `schema_version` sentinel approach once all call sites use the protobuf detail path.

Tradeoff

This is a better ownership and compatibility model, but it is a larger change than the current PR. It touches proto definitions, generated code, encoding/decoding helpers, and the status boundary in `coral-app`.

antonmry (changes requested):
I've added some changes to be evaluated, I'm particularly worried about  SCHEMA_VERSION being part of contracts:

```rust
pub use contracts::{
    ColumnInfo, CoreError, QueryError, QueryErrorCode, QueryErrorFields, QueryExecution,
    QueryRuntimeContext, QueryRuntimeProvider, QuerySource, SCHEMA_VERSION, StatusCode, TableInfo,
};
```

crates/coral-engine/src/contracts/query_error.rs
jsummerfield: I had the same reaction. Right now the QueryError mixes three concerns:
- engine semantics: unknown_field, missing_required_filter, provider_request
- transport encoding: to_json_bytes, from_json_bytes, SCHEMA_VERSION
- UX/rendering policy: plain-text summaries, hints, retryable flag

The dependency invariant is there to prevent mixing of concerns like this, reexporting an engine-owned model to the client is not the right fix.

I think there are two layers here:
- Transport layer: a stable client-facing error details model, owned by `coral-api`, used by app/client/CLI/MCP
- Engine layer: an internal query failure model, owned by coral-engine, used to preserve structure from DataFusion/providers

These two layers can start similar, and the app layer can map field-to-field, but the ownership split matters longer term and helps prevent the engine crate become the de facto owner of a public client contract.

mystic123:
## Superseded — closing as part of the V2 plan

Thanks @antonmry and @jsummerfield for the architectural review in the `query_error.rs:20` thread. You're both right:

- The `QueryError` contract crosses the client/server boundary, so it belongs in `coral-api` (as a protobuf type alongside the other `coral.v1` messages), not in `coral-engine`.
- The JSON `schema_version` string sentinel should be replaced by a structural `StatusDetail { oneof kind { QueryErrorDetail query_error_v1 = 1; } }` discriminator — a real type tag consistent with the rest of the transport contract.
- My `coral-app` re-export in `59405bf` preserved the letter of the dependency graph while violating its spirit. The fix isn't to re-export the type from a different crate; the fix is that the type shouldn't be owned by `coral-engine` in the first place. @jsummerfield called this out directly and I agree — the band-aid should come off.

**Closing this PR as superseded.** The V2 plan landed in `RELEASE_REVIEW_P1_PLAN.md` on `main` (or will once the branch containing it is pushed). Summary of the new five-PR sequence:

1. **PR1′** — Add `StatusDetail` / `QueryErrorDetail` / `QueryErrorCode` / `QueryErrorFields` protobuf types to `crates/coral-api/proto/coral/v1/errors.proto` + a `DecodedStatusError { Structured, Plain }` wrapper in `coral-client` that centralises fallback policy.
2. **PR2′** — Engine internal `QueryError` stays in `coral-engine` but becomes a **private** classification type with constructors and hint generation. `coral-app::core_status` maps it to the protobuf transport type at the `tonic::Status` boundary and attaches prost-encoded `StatusDetail` bytes via `Status::with_details`. Rewrites `provider_error_to_core` to produce `CoreError::Structured(Box<engine::QueryError>)`. Includes the secret-redacting URL sanitisation and widened reauth hints that landed in PR #64.
3. **PR3′** — `coral-cli` renders `DecodedStatusError` on `coral sql` failures, using the structured content when present and the plain `Status::message()` verbatim otherwise. No JSON failure envelope on stdout — preserves `main`'s scripting contract.
4. **PR4′** — `coral-mcp` replaces `tool_error_from_status`'s static gRPC-code dispatch with `DecodedStatusError` consumption, so agents see the same hint text the CLI does.
5. **PR5′** — `coral-engine::datafusion_to_core` enrichment (did-you-mean for unknown columns, schema-aware table-not-found).

**Code salvage from this PR:** ~80% of the logic here — the constructors, `sql_literal`, `shell_arg`, `sanitize_request_url`, `enrich_provider_detail`, `to_plain_message`, `grpc_status_code`, and the full test suite — moves intact to `coral-engine` as the internal type. What goes away: the serde derives, `to_json_bytes` / `from_json_bytes`, the `SCHEMA_VERSION` constant, the `coral-app` re-export of the engine type, and the direct engine consumption from `coral-client`.

**Branch `pawelkapica/pr1-query-error-contract` is kept in place** as a reference for the V2 work — not deleted. Same for `pawelkapica/pr2-engine-structured-errors` on #64.

Thanks again for catching this before it merged — the V2 architecture is cleaner and the V1 → V2 rework is bounded (~1 day of focused work given how much logic salvages). I'll open the new PR1 once the plan update is on `main`.

PR #64, opened 2026-04-10
feat(engine): emit structured QueryError for provider failures

mystic123:
## Summary

Rewrites `provider_error_to_core` in `crates/coral-engine/src/runtime/query.rs` to emit `QueryError` at the failure origin instead of flattening `ProviderQueryError` into a `CoreError::InvalidInput(String)` / `CoreError::Unavailable(String)`. Callers that already look at `Status::code()` see no behavioural change; callers that decode `Status::details()` (or that the CLI and MCP renderers will grow) now get the full summary, detail, hint, retryable flag, and structured fields.

Stacked on #61 — this PR's diff only includes the engine rewrite; the `QueryError` type, wire format, and client-side decoder come from #61.

## Motivation

Today `provider_error_to_core` flattens every provider failure into an opaque string:

```rust
ProviderQueryError::MissingRequiredFilter { schema, table, field } => CoreError::FailedPrecondition(format!(
    "{schema}.{table} requires WHERE {field} = <constant>"
)),
```

The schema / table / field structure is thrown away, no hint is attached, and the 401 / 403 / 404 / 429 / 5xx distinctions collapse into two string-wrapped variants. CLI and MCP consumers can only display the raw message — they can't pattern-match the failure, surface a recovery suggestion, or tell a retryable error from a permanent one.

## What this adds

**`crates/coral-engine/src/contracts/error.rs`** — new `CoreError::Structured(Box<QueryError>)` variant.
- `status_code()` delegates to `QueryError::grpc_status_code()` so routing stays consistent with the legacy flat variants.
- Display impl uses `QueryError::to_plain_message()` (summary / detail / hint).
- Boxed because `QueryError` is ~256 bytes — an unboxed variant trips clippy's `result_large_err` across every `Result<T, CoreError>` in the crate.

**`crates/coral-engine/src/contracts/query_error.rs`** — new `QueryError::grpc_status_code()` method plus supporting helpers.

Mapping:
| `QueryErrorCode` | `StatusCode` |
|---|---|
| `Unknown` | `Internal` |
| `EmptyQuery` / `SqlError` | `InvalidArgument` |
| `MissingRequiredFilter` / `InvalidQueryShape` | `FailedPrecondition` |
| `UnknownField` / `TableNotFound` | `NotFound` |
| `ProviderRequestFailed` with `http_status ∈ {429, 500..600}` | `Unavailable` |
| `ProviderRequestFailed` with `http_status = 404` | `NotFound` |
| `ProviderRequestFailed` else | `FailedPrecondition` |

**`crates/coral-engine/src/runtime/query.rs`** — `provider_error_to_core` rewritten to wrap a typed `provider_error_to_query_error` helper. `MissingRequiredFilter` populates `{schema, table, field}` fields and ships with the standard discovery hint; `ApiRequest` passes `source_schema`, `table`, `status`, `method`, `url`, and `detail` through to `QueryError::provider_request`.

**`crates/coral-app/src/bootstrap/error.rs`** — `core_status` now dispatches on the new variant. `Structured` → `Status::with_details(grpc_code, to_plain_message, to_json_bytes)`. Flat variants → existing `Status::new` path, so non-provider error paths are unchanged.

## Secret redaction

The HTTP backend's `build_logged_url` in `crates/coral-engine/src/backends/http/client.rs:800` concatenates query parameters into the request URL, and `resolve_value_source` at line 677 resolves `ValueSourceSpec::Secret` query values from source secrets. A raw URL like `https://api.datadoghq.eu/api/v1/events?api_key=SECRET&app_key=ALSO_SECRET` is a real possibility, and attaching it to `QueryError.fields.url` would leak credentials to any consumer decoding `Status::details()`.

A new `sanitize_request_url` helper strips the query component, fragment, and userinfo while keeping scheme, host/port, and path. `provider_request` routes the raw URL through it before both `fields.url` and the detail string. A regression test (`provider_request_redacts_secret_query_params_from_fields_url`) feeds a URL with `?api_key=SECRET&app_key=ALSO_SECRET` and asserts `SECRET` appears in neither.

Path-level secrets (template-expanded path segments) are out of scope — no bundled source currently embeds secrets in the path.

## Preserving request context in the plain-text fallback

The pre-refactor 429 / 5xx branch appended `[{method}] {url}` to the plain error text in `Status::message()`. Callers that still render `Status::message()` without decoding `Status::details()` would otherwise lose the failing-endpoint context.

A new `enrich_provider_detail` helper appends `[{method}] {sanitized_url}` to the detail string for every HTTP provider status (not just 429 / 5xx — the reviewer's concern applies equally to 401 / 403 / 404 and consistency is easier to reason about than a status-gated rule). The enriched detail flows through `QueryError::to_plain_message()` → `Status::message()`, so existing consumers see the failing endpoint in their existing error rendering, while upgraded consumers still pattern-match on the structured `fields`.

## Reauth hints cover bundled and imported sources

`coral source add <name>` only works for bundled sources — imported YAML sources must be refreshed with `coral source import <path>`. The previous 401 hint assumed every failing source could be fixed with `coral source add`, which is wrong on the imported path.

The 401 `provider_request` hint and the `table_not_found` source-schema hint now mention both install paths:

> Credentials for this source are invalid or expired. Re-install it to refresh: `coral source add github` for bundled sources, or `coral source import <manifest-path>` for imported sources.

The immediately-copy-pasteable form for the common bundled case is preserved (with the existing `shell_arg` escaping), and imported-source users now get correct guidance.

## Test plan

`make validate` green end-to-end. 195 tests passing, 1 pre-existing ignored, 0 failures.

- [x] `cargo check --workspace --all-targets --all-features --locked`
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo test --workspace --all-targets --all-features --locked`
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --all-features --no-deps`

New / updated tests:

**`crates/coral-engine/src/contracts/query_error.rs`**
- [x] `grpc_status_code_routes_codes_to_expected_buckets` — unit tests every `QueryErrorCode` variant.
- [x] `grpc_status_code_provider_request_inspects_http_status` — tests 429, 500, 599, 404, 401, 403, and no-status dispatch.
- [x] `sanitize_request_url_strips_query_fragment_and_userinfo` — covers query, fragment, userinfo, host:port, path-only, and invalid inputs.
- [x] `provider_request_redacts_secret_query_params_from_fields_url` — regression: `?api_key=SECRET` must not appear in `fields.url` or `detail`.
- [x] `provider_request_detail_preserves_method_and_sanitized_url` — 5xx detail carries `[GET] https://…` with the query stripped.
- [x] `provider_request_detail_handles_missing_method_or_url` — all four present/absent combinations.
- [x] `provider_request_401_hint_covers_bundled_and_imported_paths` — asserts both install commands appear.
- [x] `table_not_found_hint_covers_bundled_and_imported_paths` — same, for the discovery hint.

**`crates/coral-app/src/bootstrap/error.rs`**
- [x] `core_status_flat_variant_uses_plain_status_new` — legacy path unchanged, empty details.
- [x] `core_status_structured_variant_attaches_query_error_details` — details round-trip through `to_json_bytes` / `from_json_bytes`.
- [x] `core_status_structured_provider_500_routes_to_unavailable` — 500 → `Code::Unavailable`.
- [x] `core_status_structured_provider_401_routes_to_failed_precondition` — 401 → `Code::FailedPrecondition`.

**`crates/coral-engine/tests/engine/http_tests.rs`**
- [x] `api_returns_500` updated — now asserts on `CoreError::Structured` with `code`, `retryable`, `http_status`, `source`, `table`, hint content.
- [x] `api_returns_401` updated — same, plus asserts the hint mentions `coral source add http_401`.
- [x] `missing_required_filter_surfaces_structured_error` — new. First coverage for the missing-filter path, asserting schema/table/field all reach the caller end-to-end through the engine.
- [x] `api_returns_malformed_json` unchanged — non-provider path still produces `CoreError::Internal`.

### DIFF
--- a/crates/coral-app/src/bootstrap/error.rs
+++ b/crates/coral-app/src/bootstrap/error.rs
@@ -1,6 +1,6 @@
 //! Defines bootstrap and application-management errors for the local app.
 
-use coral_engine::{CoreError, StatusCode};
+use coral_engine::{CoreError, QueryError, StatusCode};
 use tonic::{Code, Status};
 
 use crate::state::CredentialsError;
@@ -59,7 +59,25 @@ pub(crate) fn app_status(error: AppError) -> Status {
     reason = "used directly as a map_err adapter across tonic service handlers"
 )]
 pub(crate) fn core_status(error: CoreError) -> Status {
-    Status::new(grpc_code(error.status_code()), error.to_string())
+    match error {
+        CoreError::Structured(query_error) => status_with_query_error(&query_error),
+        other => Status::new(grpc_code(other.status_code()), other.to_string()),
+    }
+}
+
+/// Builds a `tonic::Status` whose `details()` payload carries a structured
+/// [`QueryError`] as JSON bytes.
+///
+/// The gRPC `Status::message()` is set to [`QueryError::to_plain_message`] so
+/// proxies or client version-skew that strip `Status::details()` still deliver
+/// the summary, detail, and hint to the caller in a single string.
+fn status_with_query_error(query_error: &QueryError) -> Status {
+    let details = query_error.to_json_bytes();
+    Status::with_details(
+        grpc_code(query_error.grpc_status_code()),
+        query_error.to_plain_message(),
+        details.into(),
+    )
 }
 
 fn grpc_code(status: StatusCode) -> Code {
@@ -89,5 +107,65 @@ fn app_code(error: &AppError) -> Code {
         | AppError::Transport(_)
         | AppError::TaskJoin(_)
         | AppError::Credentials(_) => Code::Internal,
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{CoreError, core_status};
+    use coral_engine::QueryError;
+    use tonic::Code;
+
+    #[test]
+    fn core_status_flat_variant_uses_plain_status_new() {
+        let status = core_status(CoreError::FailedPrecondition(
+            "missing config dir".to_string(),
+        ));
+        assert_eq!(status.code(), Code::FailedPrecondition);
+        assert!(status.message().contains("missing config dir"));
+        assert!(status.details().is_empty());
+    }
+
+    #[test]
+    fn core_status_structured_variant_attaches_query_error_details() {
+        let query_error = QueryError::missing_required_filter(
+            "github",
+            "issues",
+            "repo",
+            "missing required filter",
+        );
+        let status = core_status(CoreError::Structured(Box::new(query_error.clone())));
+        assert_eq!(status.code(), Code::FailedPrecondition);
+        // Message carries the fallback plain text (summary + detail + hint).
+        let message = status.message();
+        assert!(message.contains(&query_error.summary));
+        assert!(message.contains("Hint: "));
+        // Details carry the JSON payload — round-trip through from_json_bytes.
+        let decoded = QueryError::from_json_bytes(status.details()).expect("details should decode");
+        assert_eq!(decoded, query_error);
+    }
+
+    #[test]
+    fn core_status_structured_provider_500_routes_to_unavailable() {
+        let query_error = QueryError::provider_request(
+            "github",
+            "issues",
+            Some(500),
+            Some("GET".to_string()),
+            Some("https://api.github.com/repos/coral/coral/issues".to_string()),
+            "upstream boom",
+        );
+        let status = core_status(CoreError::Structured(Box::new(query_error)));
+        assert_eq!(status.code(), Code::Unavailable);
+        assert!(status.message().contains("upstream boom"));
+    }
+
+    #[test]
+    fn core_status_structured_provider_401_routes_to_failed_precondition() {
+        let query_error =
+            QueryError::provider_request("github", "issues", Some(401), None, None, "unauthorized");
+        let status = core_status(CoreError::Structured(Box::new(query_error)));
+        assert_eq!(status.code(), Code::FailedPrecondition);
+        assert!(status.message().contains("unauthorized"));
     }
 }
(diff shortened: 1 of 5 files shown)


mystic123:
## Superseded — closing as part of the V2 plan

This PR was stacked on #61, which has been closed as part of the V2 architectural refactor. See #61's closing comment for the full decision thread.

**Relevant to this PR specifically:** the engine-side logic here — the `provider_error_to_core` rewrite, the HTTP-status dispatch, the URL sanitisation, the `enrich_provider_detail` method/URL appending, the widened reauth hints covering bundled and imported sources, and the new `missing_required_filter_surfaces_structured_error` integration test — is all still correct and transplants almost verbatim into V2 PR2′. What changes in V2 is the **destination** of the structured error: instead of `coral-app::core_status` serialising it as JSON bytes with a `SCHEMA_VERSION` sentinel, it now calls `engine_query_error.to_transport()` to map to the protobuf `coral_api::v1::QueryErrorDetail`, wraps it in `StatusDetail { kind: QueryErrorV1(...) }`, and encodes via `prost::Message::encode_to_vec()` before attaching to `tonic::Status::with_details`.

The `CoreError::Structured(Box<QueryError>)` variant stays. The engine's `QueryError` becomes a private classification type rather than a public engine export. The conversion boundary moves to `coral-app`.

**Code salvage from this PR:** ~90%. The classification logic, the hint generation, the secret-redaction helper, and all the engine tests are unchanged. Only the serialisation path rewrites.

**Branch `pawelkapica/pr2-engine-structured-errors` is kept in place** as reference for the V2 work — not deleted.

Closing.

PR #69, opened 2026-04-12
chore: lint sources workflow

iaptsiauri:
## Summary

- Introduce a YAML lint pipeline for `sources/*/manifest.yaml` using [`ryl`](https://github.com/owenlamont/ryl) (Rust-native yamllint port),  a standalone `Source Lint` GitHub workflow, and a `make yaml-lint` / `make yaml-fix` entry point.
- Normalize all 13 bundled source manifests: consistent indentation, quoting, key ordering, and reduction of column/table descriptions to single coherent sentences.

  **Description rewriting** (4,792 + 755 changes)
- 4,792 `source/table/column descriptions` that were multi-sentence or > 150 chars were passed through 35 parallel Claude Haiku sub-agents and rewritten to single coherent sentences. Rewrite rules: ≤ 150 char cap, preserve technical identifiers, drop URLs/links/examples/historical notes, no hallucination.
- 755 additional descriptions had markdown formatting stripped (`**bold**`, `*italic*`, `[text](url)`, `` `code` `` → plain text).
- 5 descriptions containing literal `#` characters rewritten to "hash" phrasing (ryl false-positively flags `#` inside quoted strings as a comment marker).

  ## Test plan

  - [x] `make yaml-lint` — exits 0 locally
  - [x] `make yaml-fix` — no-op on a clean tree
  - [x] `Source Lint` CI workflow — green on push
  - [x] `make rust-checks` — still green
  - [ ] Smoke test an MCP client against stripe / github sources — descriptions still useful for table selection
  - [x] Spot-check rewritten descriptions in `coral.tables` via `SELECT name, description FROM coral.tables WHERE schema_name IN ('stripe', 'github', 'gitlab') LIMIT 20`


.github/workflows/lint-source.yml
```diff
@@ -0,0 +1,50 @@
+name: Lint Sources
```
antonmry: Should we make this part of the `validate.yml' workflow?

sources/otel/manifest.yaml
```diff
@@ -3,124 +3,112 @@ version: 0.1.0
 dsl_version: 3
```
antonmry: I think it would be good to remove the otel connector. We aren't planning to port the Otel receiver functionality in the short term and it can be confusing for the final user

Makefile
```diff
@@ -4,3 +4,27 @@ rust-checks:
...
+#   make yaml-fix    # apply safe fixes in place
+
+RYL_VERSION := 0.6.0
+YAMLLINT_FILES := $(wildcard sources/*/manifest.yaml sources/*/manifest.yml)
+
+ryl-ensure:
+	@if ! command -v ryl >/dev/null 2>&1 || \
+	  [ "$$(ryl --version 2>&1 | awk '{print $$NF}')" != "$(RYL_VERSION)" ]; then \
+		echo "Installing ryl $(RYL_VERSION)..."; \
+		cargo install ryl --locked --version $(RYL_VERSION); \
```
antonmry: This approach doesn't follow the pattern we use with other tools: the user has to install them if `make validate` fails, and we install them inside the validate.yml workflow. It would be good to be consistent on how we do it.

Check in `validate.yml` how is being used for ruff. A benefit of using that approach, it's that we could do `uv tool install ryl` (or similar) instead of having to compile the tool with cargo (it takes times)

antonmry (changes requested):
Great PR! I've added a couple of minor comments that could be potential improvements, just to be considered.

.github/workflows/lint-source.yml
iaptsiauri: Yes moved it to validate

antonmry (approved):

PR #71, opened 2026-04-13
docs(cli): expand `coral --help` with tagline

AlbertQM:
## Summary

Trying to give a warmer welcome to first time users, and point them to the `onboard`ing.

## Test plan

image
image

### DIFF
--- a/crates/coral-cli/src/main.rs
+++ b/crates/coral-cli/src/main.rs
@@ -23,8 +23,6 @@ use tonic::Request;
 #[derive(Debug, Parser)]
 #[command(name = "coral", version, arg_required_else_help = true)]
 /// A local-first SQL interface for APIs, files, and other data sources.
-///
-/// New here? Run `coral onboard` for an interactive setup wizard.
 struct Cli {
     #[command(subcommand)]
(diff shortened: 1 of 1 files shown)


jsummerfield (changes requested):
See comments

crates/coral-cli/src/main.rs
```diff
@@ -22,7 +22,9 @@ use tonic::Request;
 
 #[derive(Debug, Parser)]
 #[command(name = "coral", version, arg_required_else_help = true)]
-/// Query and manage local data sources
+/// A local-first SQL interface for APIs, files, and other data sources.
+///
+/// New here? Run `coral onboard` for an interactive setup wizard.
```
jsummerfield: This improves the first experience at the expense of every subsequent one. I could see this getting annoying after onboarding.

Let’s remove the line, keep the fist change, and make sure docs do the heavy lifting instead

AlbertQM: Fair point, will update. Thanks!

jsummerfield (approved):

PR #72, opened 2026-04-13
test(app): app integration tests - gRPC suite

sauldhernandez:
## Intent
This change replaces the old monolithic `local_mode` integration test with a structured app-level gRPC suite for `coral-app`, split into shared harness, source lifecycle, and resilience tests. The goal is to make app-surface behavior easier to extend and review as the gRPC contract grows.

## What It Enables
The new layout gives us focused coverage for source import, validation, deletion, bundled-source creation, rollback behavior, broken-source isolation, read-only SQL enforcement, and gRPC error mapping. It also adds explicit negative tests for missing, unknown, and repeated input bindings so source input and credential validation can evolve without hiding regressions inside one broad end-to-end test.

## Usage Examples
- Run the full app-level gRPC suite with `cargo test -p coral-app --test grpc`.
- Run a focused lifecycle case with `cargo test -p coral-app --test grpc source_lifecycle_tests::import_source_missing_required_secret_returns_invalid_argument`.
- Run the full repo validation on current `main` with `make rust-checks`.
- Add new app/gRPC integration tests under `crates/coral-app/tests/grpc/`.
- Use `GrpcHarness` from `harness.rs` to boot a real server and get typed source/query clients.

## Validation
Validated with `make rust-checks`.


crates/coral-app/tests/grpc/harness.rs
```diff
@@ -0,0 +1,289 @@
...
+                {"name": "id", "type": "Utf8"},
+            ],
+        }],
+    }))
+}
+
+pub(crate) fn unreachable_http_manifest_yaml() -> String {
+    let listener = TcpListener::bind("127.0.0.1:0").expect("bind closed port probe");
+    let base_url = format!("http://{}", listener.local_addr().expect("local addr"));
+    drop(listener);
```
jsummerfield: Seems a bit awkward, wonder if there's a nicer way. It also risks a port reuse race.

Something to consider instead:

```rust
struct FailingHttpFixture {
    base_url: String,
    _task: tokio::task::JoinHandle<()>,
}

impl FailingHttpFixture {
    async fn new() -> Self {
        let listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
        let addr = listener.local_addr().unwrap();
        let task = tokio::spawn(async move {
            loop {
                let (socket, _) = listener.accept().await.unwrap();
                drop(socket); // immediate connection reset/EOF
            }
        });

        Self {
            base_url: format!("http://{addr}"),
            _task: task,
        }
    }
}
```

jsummerfield (approved):

PR #79, opened 2026-04-13
feat(spec): add timestamp columns and parsed column templates

sauldhernandez:
## Summary

Fixes DATA-366 by exposing Slack message timestamps as real `Timestamp` values and adding message permalinks, while also formalizing the new source-spec column expression API.

## What changed

- add `Timestamp` as a source-spec column type
- add `format_timestamp` to normalize epoch `seconds` or `milliseconds` into UTC timestamp microseconds
- switch column-expression `template` rendering onto the shared parsed template system with `{{expr.*}}` and `{{filter.*}}` tokens
- add `replace` as a normal expression for string transforms, instead of introducing token-specific transform syntax
- validate column-expression templates at manifest load time, including unknown `expr` keys and unsupported `secret` / `variable` / `state` namespaces in row mapping
- update the Slack `messages` manifest so `ts` is a `Timestamp` and `permalink` is rendered from `{{filter.channel}}` plus a sanitized `ts_id`
- update the source-spec reference docs to cover `Timestamp`, `format_timestamp`, `replace`, and parsed column templates

## Examples

Timestamp column:

```yaml
- name: ts
  type: Timestamp
  expr:
    kind: format_timestamp
    input: seconds
    expr:
      kind: path
      path:
        - ts
```

Permalink column:

```yaml
- name: permalink
  type: Utf8
  expr:
    kind: template
    template: "https://slack.com/archives/{{filter.channel}}/p{{expr.ts_id}}"
    values:
      ts_id:
        kind: replace
        expr:
          kind: path
          path:
            - ts
        from: "."
        to: ""
```

## Test plan

- [x] `cargo test -p coral-spec`
- [x] `cargo test -p coral-engine`
- [x] `make rust-checks`


crates/coral-engine/src/backends/shared/mapping.rs
```diff
@@ -168,7 +177,67 @@ fn eval_expr(expr: &ExprSpec, row: &Value, filters: &HashMap<String, String>) ->
...
+            Some(eval_template(template, values, row, filters))
+        }
+    }
+}
+
+/// Evaluate a `FormatTimestamp` expression, returning epoch **microseconds** as
+/// a `Value::Number` suitable for an Arrow `TimestampMicrosecondArray`.
+#[allow(
+    clippy::cast_possible_truncation,
+    reason = "Epoch-second f64 is intentionally converted to i64 microseconds; sub-microsecond precision is acceptable to lose"
```
antonmry: Another option to evaluate:  avoid floats entirely for string inputs by parsing the integer and fractional parts separately:

```rust
  fn parse_epoch_micros(s: &str, input: &TimestampInput) -> Option<i64> {
      let (int_str, frac_str) = s.split_once('.').unwrap_or((s, ""));
      let whole: i64 = int_str.parse().ok()?;
      // Pad or truncate fractional part to 6 digits for seconds, 3 for millis
      let frac_width = match input {
          TimestampInput::Seconds => 6,
          TimestampInput::Milliseconds => 3,
      };
      let padded = format!("{frac_str:0<frac_width$}");
      let frac: i64 = padded[..frac_width].parse().ok()?;
      let multiplier = match input {
          TimestampInput::Seconds => 1_000_000,
          TimestampInput::Milliseconds => 1_000,
      };
      Some(whole * multiplier + frac)
  }
```

crates/coral-engine/src/backends/shared/mapping.rs
```diff
@@ -66,6 +67,14 @@ pub(crate) fn convert_items(
                     .collect();
                 arrays.push(Arc::new(array));
             }
+            DataType::Timestamp(_, _) => {
+                let array: TimestampMicrosecondArray = items
+                    .iter()
+                    .map(|row| to_i64(eval_expr(&expr, row, filters)))
+                    .collect();
+                let array = array.with_timezone("+00:00");
```
antonmry, line 77: This is fine, but if Timestamp is used in the future for other use cases, it might introduce some bugs. Consider to check it so it's enforced:

```rush
// ManifestDataType::Timestamp currently only produces Microsecond/UTC.
  // Match explicitly so that adding a different precision or timezone
  // later is a compile- or runtime-visible error, not a silent mismatch.
  DataType::Timestamp(TimeUnit::Microsecond, Some(tz)) if tz.as_ref() == "+00:00" => {
```

antonmry (changes requested):
I've added some minor comments, in particular, the possibility of reusing template is worth a look to see if it could be an improvement

sources/slack/manifest.yaml
```diff
@@ -153,13 +153,31 @@ tables:
...
+        type: Timestamp
+        nullable: false
+        description: Message timestamp formatted as ISO-8601 UTC
+        expr:
+          kind: format_timestamp
+          input: seconds
+          expr:
+            kind: path
+            path:
+              - ts
```
simonwhitaker, line 164: This may or may not matter, but it's worth pointing out that only the bit in front of the dot is the timestamp. The bit after the dot is a Slack-generated value to guarantee uniqueness, not a fraction of a second.

docs/reference/source-spec-reference.mdx
```diff
@@ -32,18 +32,70 @@ backend: http
...
+### `format_timestamp`
+
+Converts a numeric or string epoch value into microseconds for a `Timestamp` column.
+
+```yaml
+- name: created_at
+  type: Timestamp
+  expr:
+    kind: format_timestamp
+    input: seconds        # or "milliseconds"
```
simonwhitaker, line 72: I'm a bit confused by this. I _think_ it's saying that the input value must be an epoch value in either seconds or milliseconds, and (as per L60) the output will be an epoch value in microseconds. Is that right? What if an API only vends e.g. microseconds or nanoseconds?

simonwhitaker (approved):

docs/reference/source-spec-reference.mdx
sauldhernandez: Yeah wording was confusing. Clarified now. If an API vends microseconds or nanoseconds we can add support for that later on. Didn't want to overdo it here.

crates/coral-engine/src/backends/shared/mapping.rs
```diff
@@ -168,9 +177,113 @@ fn eval_expr(expr: &ExprSpec, row: &Value, filters: &HashMap<String, String>) ->
...
+        }
+        _ => None,
+    }?;
+    Some(Value::Number(micros.into()))
+}
+
+/// Parse a timestamp string into epoch microseconds without intermediate
+/// floating-point arithmetic, preserving full microsecond precision.
+fn parse_epoch_micros(s: &str, input: &TimestampInput) -> Option<i64> {
+    let (int_str, frac_str) = s.split_once('.').unwrap_or((s, ""));
```
antonmry, line 219: Copy&paste a suggestion after a chat with the code agent about extending coral-spec/src/template.rs for this, I think it's worth a try:

```
Extend the existing ParsedTemplate system. Concretely:

  1. Store the ExprSpec::Template template as a ParsedTemplate (using {{...}} syntax), so it gets parsed and validated at manifest load time
  2. Add a new namespace (e.g., expr) so sub-expressions can be referenced: {{expr.channel}}, {{expr.ts_id}}
  3. If remove is truly needed, extend TemplateToken to support transform pipes alongside default values — or handle it as a post-processing step on the
  resolved value
  4. Render using the existing render_template() pattern (walk parts, resolve tokens) rather than find/replace

  This gives you proper parsing, error reporting on unclosed tokens, manifest-time validation, and consistency with how every other template in the system
  works. 
```

antonmry (approved):
Great!

PR #80, opened 2026-04-13
feat(sources): add Attio source

kyracheng:
## Summary

- Adds a new `attio` source with 8 tables: `people`, `companies`, `deals`, `notes`, `tasks`, `lists`, `list_entries`, and `workspace_members`
- Adds `offset_body_path` support to offset pagination mode, fixing infinite pagination loops against APIs that only accept `offset` in the POST request body (not as a query param)

## What changed

**`sources/attio/manifest.yaml`** — new bundled source for Attio CRM. Covers the main record objects, activity (notes, tasks), list membership, and workspace membership. All record tables paginate at 500 records/request using body-based limit/offset.

**`crates/coral-spec`** — adds `offset_body_path: Vec<String>` to `PaginationSpec` and `OffsetPagination`. Mutually exclusive with `offset_param`; validator rejects manifests that define both.

**`crates/coral-engine`** — `apply_pagination_body_fields` now writes `state.offset` to the configured body path when `offset_body_path` is set. The query-param path skips offset when body path is used instead.

## Why

Attio's record query endpoints (`POST /v2/objects/*/records/query`) only accept `limit` and `offset` in the request body. With only `offset_param` available, the offset was sent as a query parameter which Attio silently ignored, causing every page request to return page 0 and pagination to loop indefinitely.

## Known limitations

- Attio supports custom objects beyond `people`, `companies`, and `deals` — these are not yet covered
- `list_entries` requires a `list_id` filter; cross-list querying would need a separate approach

## Test plan

- [ ] `cargo test --workspace` passes
- [ ] `coral source add attio` prompts for `ATTIO_API_KEY`
- [ ] `SELECT COUNT(*) FROM attio.people` returns full record count (not capped at 500)
- [ ] `SELECT * FROM attio.list_entries WHERE list_id = '<id>'` returns all entries
- [ ] Offset pagination does not loop: companies/people with >500 records paginate correctly

### DIFF
--- a/crates/coral-spec/src/common.rs
+++ b/crates/coral-spec/src/common.rs
@@ -257,6 +257,8 @@ pub struct PaginationSpec {
     #[serde(default)]
     pub offset_param: Option<String>,
     #[serde(default)]
+    pub offset_body_path: Vec<String>,
+    #[serde(default)]
     pub offset_start: i64,
     #[serde(default)]
     pub offset_step: Option<i64>,
@@ -278,6 +280,7 @@ impl Default for PaginationSpec {
             page_start: 0,
             page_step: default_page_step(),
             offset_param: None,
+            offset_body_path: Vec::new(),
             offset_start: 0,
             offset_step: None,
             link_header_require_results: false,
@@ -309,7 +312,10 @@ pub enum ValidatedPaginationMode {
 /// Validated typed offset-pagination settings.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct OffsetPagination {
-    pub param: String,
+    /// Query-parameter name for the offset (mutually exclusive with `body_path`).
+    pub param: Option<String>,
+    /// JSON body path for the offset (mutually exclusive with `param`).
+    pub body_path: Vec<String>,
     pub start: i64,
     step: OffsetStep,
 }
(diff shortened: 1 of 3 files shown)


crates/coral-spec/src/common.rs
```diff
@@ -257,6 +257,8 @@ pub struct PaginationSpec {
     #[serde(default)]
     pub offset_param: Option<String>,
     #[serde(default)]
+    pub offset_body_path: Vec<String>,
```
antonmry, line 260: It would be good to evaluate to add new tests covering:
  - Validation rejects manifests with both offset_param and offset_body_path
  - Validation accepts offset_body_path alone (without offset_param)
  - The engine correctly writes offset into the request body via apply_pagination_body_fields

sources/attio/manifest.yaml
```diff
@@ -0,0 +1,782 @@
...
+      path: /v2/objects/companies/records/query
+    response:
+      rows_path:
+        - data
+    pagination:
+      mode: offset
+      offset_body_path:
+        - offset
+      offset_start: 0
+      offset_step: 500
```
antonmry, line 170: Not totally sure about this one but worth a comment:

> The record tables (people, companies, deals, list_entries) set both offset_step: 500 and page_size: {default: 500, max: 500}. Since offset_step is
  explicit, it uses OffsetStep::Explicit(500) which is fixed regardless of resolved page size.
>  Compare to notes and tasks which omit offset_step and get OffsetStep::PageSize — the offset step tracks the actual page size sent to the API.
>  For full table scans the behavior is identical. But if page_size is reduced (e.g., via SQL LIMIT), explicit offset_step: 500 could cause the offset to
  jump ahead by 500 while only requesting limit: <smaller> records per page — potentially skipping records or fetching redundant pages. Removing
  offset_step: 500 from the record tables would be both simpler and more correct.

PR #81, opened 2026-04-13
fix(app): stop local gRPC PROTOCOL_ERROR on wide-manifest queries

AlbertQM:
## Summary

`coral sql` against wide manifests like `github.search_issues` surfaced as `h2 protocol error … PROTOCOL_ERROR` instead of a normal gRPC status. Two independent tonic 0.14 defaults on our in-process CLI ↔ server channel were too tight:

- **Response body:** `ExecuteSql` is unary and ships the full Arrow IPC stream in one message. Raised `QueryService` response message size limit to 64 MB (tonic default: 4 MB). Only the response direction is bumped — requests are small SQL strings.
- **Error trailers:** DataFusion's `"No field named X. Valid fields are …"` message enumerating ~391 field names overflowed HTTP/2 trailers (default `MAX_HEADER_LIST_SIZE` ~16 KiB), so the server emitted invalid trailers and the client's h2 stack reported `PROTOCOL_ERROR` before surfacing the intended status.

### Fixes

**Transport layer** (`coral-app`, `coral-client`, `coral-api`):

1. `QUERY_RESPONSE_MAX_MESSAGE_SIZE` (64 MB) — server encode only, client decode only.
2. `HTTP2_MAX_HEADER_LIST_SIZE` (128 KiB) — on server, production client, and test endpoints.
3. 4 KiB UTF-8-safe `truncate_status_detail` chokepoint in `app_status` and `core_status`.

**Semantic error mapping** (`coral-engine`):

4. `datafusion_to_core` now calls `DataFusionError::find_root()` to unwrap `Context`/`Shared` wrappers, so schema errors map to the correct `CoreError` variant instead of falling through to `Internal`.
5. `SchemaError` uses DataFusion's own `Display` — no custom formatting in the engine. The `truncate_status_detail` chokepoint in the app layer handles oversized messages. Presentation hints (discovery guidance, UX explanations) belong in the app/CLI/MCP layer, not the engine.

## Behavior changes

- Missing-column errors: `Code::Internal` → `Code::InvalidArgument` (via `find_root()` exposing the real `SchemaError`).
- Missing-table/schema errors: `Code::Internal` → `Code::InvalidArgument` (`Plan` errors now map to `InvalidInput`). The `NotFound` distinction for missing resources needs more thought and is deferred to a follow-up.
- Error detail strings from `Plan` errors no longer carry the `"Error during planning: "` prefix (matched on the variant directly).

## Before / after

| Query | Before | After |
|---|---|---|
| `SELECT bogus_column FROM github.search_issues LIMIT 0` | `h2 PROTOCOL_ERROR` | `InvalidArgument: No field named "bogus_column". Valid fields are …(truncated)` |
| `SELECT user_login FROM github.search_issues LIMIT 1` | `h2 PROTOCOL_ERROR` | Same clean error (hint can be added in app/MCP layer later) |
| `SELECT … user__login FROM github.search_issues WHERE q = 'NoSuchBucket' …` | `h2 PROTOCOL_ERROR` | Rows |
| `SELECT * FROM missing.table` | `Code::Internal` | `Code::InvalidArgument` |

## Known follow-up (not in this PR)

- Streaming `ExecuteSql` RPC to remove the 64 MB body ceiling entirely.
- App/MCP-layer "Did you mean?" hints and `coral.columns` discovery guidance for `InvalidArgument` schema errors.

## Test plan

- [x] Regression test: >4 MB `ExecuteSqlResponse` round-trips (asserts payload `> 4 * 1024 * 1024`).
- [x] Regression test: 600-column synthetic manifest + invalid column returns `Code::InvalidArgument` with `message.len() <= 4 KiB`, no PROTOCOL_ERROR.
- [x] Unit tests for `truncate_status_detail` (short passthrough, long ASCII + marker, UTF-8 boundary safety).
- [x] Unit test proving `find_root()` unwraps `Context`-wrapped `SchemaError` to `CoreError::InvalidInput`.
- [x] `cargo test --workspace` and `make rust-checks` green.

### DIFF
--- a/crates/coral-app/src/bootstrap/server.rs
+++ b/crates/coral-app/src/bootstrap/server.rs
@@ -6,7 +6,7 @@ use std::sync::Mutex;
 
 use coral_api::v1::query_service_server::QueryServiceServer;
 use coral_api::v1::source_service_server::SourceServiceServer;
-use coral_api::{HTTP2_MAX_HEADER_LIST_SIZE, QUERY_SERVICE_MAX_MESSAGE_SIZE};
+use coral_api::{HTTP2_MAX_HEADER_LIST_SIZE, QUERY_RESPONSE_MAX_MESSAGE_SIZE};
 use tokio::net::TcpListener;
 use tokio::sync::oneshot;
 use tokio::task::JoinHandle;
@@ -180,8 +180,7 @@ async fn start_server(
             .add_service(SourceServiceServer::new(source_service))
             .add_service(
                 QueryServiceServer::new(query_service)
-                    .max_encoding_message_size(QUERY_SERVICE_MAX_MESSAGE_SIZE)
-                    .max_decoding_message_size(QUERY_SERVICE_MAX_MESSAGE_SIZE),
+                    .max_encoding_message_size(QUERY_RESPONSE_MAX_MESSAGE_SIZE),
             )
             .serve_with_incoming_shutdown(TcpListenerStream::new(listener), async {
                 let _ = shutdown_rx.await;
@@ -201,7 +200,7 @@ mod tests {
     use coral_api::v1::query_service_client::QueryServiceClient;
     use coral_api::v1::source_service_client::SourceServiceClient;
     use coral_api::v1::{ExecuteSqlRequest, ImportSourceRequest, Workspace};
-    use coral_api::{HTTP2_MAX_HEADER_LIST_SIZE, QUERY_SERVICE_MAX_MESSAGE_SIZE};
+    use coral_api::{HTTP2_MAX_HEADER_LIST_SIZE, QUERY_RESPONSE_MAX_MESSAGE_SIZE};
     use coral_engine::QueryRuntimeContext;
     use tempfile::TempDir;
     use tonic::Request;
@@ -257,8 +256,7 @@ mod tests {
             .expect("connect");
         let mut source_client = SourceServiceClient::new(channel.clone());
         let mut query_client = QueryServiceClient::new(channel)
-            .max_decoding_message_size(QUERY_SERVICE_MAX_MESSAGE_SIZE)
-            .max_encoding_message_size(QUERY_SERVICE_MAX_MESSAGE_SIZE);
+            .max_decoding_message_size(QUERY_RESPONSE_MAX_MESSAGE_SIZE);
 
         source_client
             .import_source(Request::new(ImportSourceRequest {
@@ -332,8 +330,7 @@ tables:
             .await
             .expect("connect");
         let mut query_client = QueryServiceClient::new(channel)
-            .max_decoding_message_size(QUERY_SERVICE_MAX_MESSAGE_SIZE)
-            .max_encoding_message_size(QUERY_SERVICE_MAX_MESSAGE_SIZE);
+            .max_decoding_message_size(QUERY_RESPONSE_MAX_MESSAGE_SIZE);
 
         // No underscore separator — DataFusion's SQL parser is conservative
         // about numeric literal formats.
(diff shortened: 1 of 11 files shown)


crates/coral-app/src/bootstrap/server.rs
```diff
@@ -175,8 +176,13 @@ async fn start_server(
 
     let task = tokio::spawn(async move {
         Server::builder()
+            .http2_max_header_list_size(HTTP2_MAX_HEADER_LIST_SIZE)
             .add_service(SourceServiceServer::new(source_service))
```
antonmry, line 180: This is a bit confusing for me. I assume we are using `QUERY_SERVICE_MAX_MESSAGE_SIZE` both for the request and the response, but the problem only happens in the response (because queries should be pretty small). It would be good to consider:

- Apply this only to the responses and wait until we needed for the requests.
- Or, at least, define different constants for requests and responses to make it more clear

antonmry (approved):
I added a comment that's worth a review, but good in general

crates/coral-app/src/bootstrap/server.rs
AlbertQM: > Apply this only to the responses and wait until we needed for the requests.

YAGNI, will apply this suggestion. Thanks for raising! 🙏

crates/coral-engine/src/runtime/query.rs
```diff
@@ -90,27 +91,69 @@ fn read_only_sql_options() -> SQLOptions {
...
+    // Unwrap Context/Shared/Diagnostic wrappers so wrapped schema errors
+    // get classified by their root variant instead of all landing in the
+    // `Internal` bucket. Without `find_root()`, `SELECT bogus FROM wide`
+    // surfaces as `CoreError::Internal` because DataFusion wraps the
+    // SchemaError in `Context`/`Execution`, hiding the structured variant
+    // from the match arms below.
+    match error.find_root() {
         DataFusionError::SQL(detail, _) => CoreError::InvalidInput(detail.to_string()),
-        DataFusionError::Plan(detail) => CoreError::InvalidInput(detail),
+        DataFusionError::Plan(detail) => CoreError::InvalidInput(detail.clone()),
```
jsummerfield, line 102: Using `find_root()` here fixes the wrapped `FieldNotFound` case, but it also reclassifies genuine missing-resource failures like `SELECT * FROM missing.users` from `CoreError::NotFound` to `CoreError::InvalidInput`. That breaks the existing `CoreError` contract for missing resources and changes downstream MCP classification/hints, which currently distinguish `NotFound` from `InvalidArgument`. I think we should keep the root unwrapping for wrapped column errors, but preserve `NotFound` semantics for missing schema/table lookups.

crates/coral-engine/src/runtime/query.rs
```diff
@@ -90,27 +91,69 @@ fn read_only_sql_options() -> SQLOptions {
...
 }
 
+/// Format a `SchemaError::FieldNotFound` into a concise, hint-bearing
+/// `Status` detail. The full `valid_fields` list is deliberately *not*
+/// embedded in the message — on wide manifests (e.g. `github.search_issues`
+/// with ~561 columns) that list can exceed the HTTP/2 trailer size limit
+/// and trigger a `PROTOCOL_ERROR` before the CLI ever sees the status.
+/// Callers who want the full set of valid columns can query
+/// `coral.columns`.
+fn format_field_not_found(field: &Column, valid_fields: &[Column]) -> String {
```
jsummerfield: I think the transport-limit and trailer-truncation pieces in `coral-app` / `coral-client` are the right ownership seam, but this function feels like it pushes presentation policy too far down into `coral-engine`. In particular, embedding product guidance like `query coral.columns for the list` and the nested-field hint here couples the engine to one user-facing surface. It would be cleaner for the engine to preserve the semantic error shape, and for the app / CLI / MCP layers to decide how to render hints and discovery guidance.

AlbertQM: Strong point, removed this function altogether. Thanks!

jsummerfield:
One follow-up issue in the current version: `datafusion_to_core()` now maps any `DataFusionError::Plan(detail)` containing `"not found"` to `CoreError::NotFound`.

That preserves the missing table/schema case, but it also catches other planner failures that DataFusion phrases the same way, like unknown scalar/aggregate/window functions. Those are invalid SQL, not missing Coral resources, so this changes the transport contract and downstream MCP UX (`NotFound` becomes `resource_not_found` plus the source/schema/table hint).

I think this remap needs to be narrowed to the actual table/schema lookup cases, or keyed off a structured variant instead of matching the message text.

jsummerfield (approved):

PR #82, opened 2026-04-13
fix(app): load bundled sources from binary

simonwhitaker:
Bundled source installs currently persist a `manifest.yaml` copy under the workspace, so upgrading Coral can leave an installed bundled source pinned to an older embedded spec even when the binary ships fixes. That stale manifest also means newly bundled input requirements are invisible until the user manually re-adds the source.

This change makes bundled installs persist only source identity, origin, variables, and secrets, while imported installs continue to persist their user-owned manifest snapshot on disk. Query and validation now resolve bundled manifests from the current binary by name, derive the surfaced source version from the resolved manifest, and fail with clear missing variable or secret errors if a newer bundled spec introduces new required inputs.

### DIFF
--- a/crates/coral-app/src/query/manager.rs
+++ b/crates/coral-app/src/query/manager.rs
@@ -102,6 +102,7 @@ impl QueryManager {
         for source in self.config_store.list_workspace_sources(workspace)? {
             match self.load_query_source(&source) {
                 Ok((query_source, _version)) => query_sources.push(query_source),
+                Err(error @ AppError::FailedPrecondition(_)) => return Err(error),
                 Err(error) => {
                     tracing::warn!(
                         source = %source.name,
@@ -124,13 +125,24 @@ impl QueryManager {
             .secret_store
             .read_source_secrets_for(&source.workspace, &source.name)?;
         let mut resolved_secrets = BTreeMap::new();
-        for secret_name in source_spec.required_secret_names() {
-            let Some(value) = stored_secrets.get(&secret_name).cloned() else {
-                return Err(AppError::FailedPrecondition(format!(
-                    "source '{}' is missing secret '{}'",
-                    source.name, secret_name
-                )));
+        let missing_secrets: Vec<String> = source_spec
+            .required_secret_names()
+            .into_iter()
+            .filter(|name| !stored_secrets.contains_key(name))
+            .collect();
+        if let Some((first, rest)) = missing_secrets.split_first() {
+            let detail = if rest.is_empty() {
+                format!("secret '{first}'")
+            } else {
+                format!("secret '{first}' and {} other(s)", rest.len())
             };
+            return Err(AppError::FailedPrecondition(format!(
+                "source '{}' is missing {detail} — re-run `coral source add {}` to update",
+                source.name, source.name
+            )));
+        }
+        for secret_name in source_spec.required_secret_names() {
+            let value = stored_secrets[&secret_name].clone();
             resolved_secrets.insert(secret_name, value);
         }
         Ok((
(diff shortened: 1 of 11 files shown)


docs/reference/cli-reference.mdx
```diff
@@ -46,6 +46,7 @@ coral source add github
 ```
 
 Coral prompts for required variables or secrets interactively. If the source is already installed, running this command again replaces its stored credentials with the new values you provide.
+Bundled installs keep the source identity plus your configured variables and secrets, then resolve the bundled manifest from the current Coral binary at runtime.
```
AlbertQM: I'm not sure what this means, and don't think we use "Bundled installs" anywhere else?

Wasn't too sure what could be useful to tell users, maybe the update lifecycle? (Feel free to disregard and come up with a better copy)

```suggestion
Since these sources are bundled with the Coral binary, updating Coral will update the sources.
```

docs/reference/cli-reference.mdx
```diff
@@ -55,6 +56,8 @@ Add a custom source spec from a YAML file.
 coral source add --file ./local-messages.yaml
 ```
 
+Imported installs persist the manifest snapshot under Coral's config directory because that source spec is user-owned.
```
AlbertQM: Not sure if we want to share information about the lifecycle of these.

```suggestion
These user-owned sources persist the manifest snapshot under Coral's config directory. To update them, run `coral source add --file` with the latest YAML.
```

AlbertQM (approved):
Love it, thanks!
Two tiny nits for the docs

crates/coral-app/src/sources/manager.rs
```diff
@@ -216,7 +229,7 @@ impl SourceManager {
         let stored = ManagedSource {
             workspace: workspace.clone(),
             name: source_name.clone(),
-            version: request.available.version.clone(),
+            version: String::new(),
```
jsummerfield: This blanks the persisted `version` for every source, not just bundled ones. Because `render_config` now drops empty versions, imported installs also stop writing `version = "..."` to `config.toml`, which widens this change beyond the PR intent and changes a stable config surface. Could we keep this behavior scoped to bundled origins and continue persisting the imported-source version?

crates/coral-app/src/query/manager.rs
```diff
@@ -98,7 +101,8 @@ impl QueryManager {
         let mut query_sources = Vec::new();
         for source in self.config_store.list_workspace_sources(workspace)? {
             match self.load_query_source(&source) {
-                Ok(query_source) => query_sources.push(query_source),
+                Ok((query_source, _version)) => query_sources.push(query_source),
+                Err(error @ AppError::FailedPrecondition(_)) => return Err(error),
```
jsummerfield: This changes the query isolation policy as well as the bundled-manifest behavior. Before this, a broken source was warned-and-skipped here; now any `FailedPrecondition` aborts `load_query_sources`, so one stale source can block `list_tables` and unrelated queries against healthy schemas. That seems like a separate product decision from manifest sync, and I think it should be evaluated independently rather than coupled into this PR.

crates/coral-app/src/query/manager.rs
```diff
@@ -111,43 +115,39 @@ impl QueryManager {
...
+            .filter(|name| !stored_secrets.contains_key(name))
+            .collect();
+        if let Some((first, rest)) = missing_secrets.split_first() {
+            let detail = if rest.is_empty() {
+                format!("secret '{first}'")
+            } else {
+                format!("secret '{first}' and {} other(s)", rest.len())
             };
+            return Err(AppError::FailedPrecondition(format!(
+                "source '{}' is missing {detail} — re-run `coral source add {}` to update",
```
jsummerfield: I think this remediation is being generated too low in the stack. `QueryManager` can correctly detect "missing required secret/variable", but hardcoding `coral source add ...` here bakes CLI-specific guidance into shared app/query logic and loses origin-aware behavior for imported sources. Could we keep the lower-layer error neutral here and let a higher user-facing layer add recovery guidance when it has the right client and source-origin context?

simonwhitaker:
Addressed all three review comments from @jsummerfield:

**Comment 1 (version blanking):** `persist_source` now branches on origin — bundled sources get `version: String::new()` (resolved at runtime from the binary), while imported sources continue persisting `request.available.version` into `config.toml`.

**Comment 2 (query isolation):** Reverted `load_query_sources` to the original warn-and-skip behavior for all errors. A broken source no longer blocks healthy ones. Updated the resilience test to assert the healthy source still loads while the broken one is silently skipped.

**Comment 3 (CLI guidance in lower layer):** Removed the CLI-specific remediation text from both missing-secret and missing-variable error messages. They now just say e.g. `source 'foo' is missing secret 'bar'` — a higher-level user-facing layer can add origin-aware recovery guidance.

jsummerfield (approved):
Nice, thank you!

PR #91, opened 2026-04-14
fix(spec):  handle 403 rate limits & improve GitHub search table guides

iaptsiauri:
## Summary

1. GitHub returns 403 instead of 429 for some rate-limited requests. The HTTP client now also retries on 403 responses when Retry-After or X-RateLimit-Reset headers  are present — otherwise it still treats 403 as a normal forbidden error.

2. GitHub search tables: add guides and improve q filter descriptions
  - search_commits
  - search_issues
  - search_labels
  - search_repositories
  - search_topics
  - search_users

  Each guide documents scoping requirements (repo:/org:), the 5-operator limit on q, secondary rate limit warnings, common qualifiers, and usage examples.

## Validation

 | # | Test | Result |
  |---|------|--------|
  | 1 | `cargo build -p coral-app` compiles | PASS |
  | 2 | `cargo test -p coral-engine` — 91 passed, 1 ignored | PASS |
  | 3 | `search_issues` — scoped query `is:pr is:open author:iaptsiauri org:withcoral` | PASS |
  | 4 | `search_commits` — scoped query `repo:withcoral/coral author:iaptsiauri` | PASS |
  | 5 | `search_repositories` — scoped query `org:withcoral` | PASS |
  | 6 | `search_users` — query `iaptsiauri` | PASS |
  | 7 | `search_topics` — query `datafusion` | PASS |
  | 8 | 403 rate limit — only retries when `Retry-After` or `X-RateLimit-Reset` present (code review) | PASS |
  | 9 | No existing rate limit tests broken (0 matched, 92 filtered) | PASS |


crates/coral-engine/src/backends/http/client.rs
```diff
@@ -503,6 +499,31 @@ fn request_error(
...
 }
 
+fn check_rate_limit(
+    response: &reqwest::Response,
+    retries: usize,
+    method_label: &str,
+    logged_url: &str,
+) -> Result<Option<Duration>, DataFusionError> {
+    let is_rate_limited = response.status() == reqwest::StatusCode::TOO_MANY_REQUESTS
+        || (response.status() == reqwest::StatusCode::FORBIDDEN
```
simonwhitaker: This check is a bit Github-specific, according to the PR description. It might be worth adding a comment to explain why we're also allowing 403s to be treated as rate limit errors, and/or to expand this to other status codes, e.g. it can also be used for 503 errors during load shedding. (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After)

crates/coral-engine/src/backends/http/client.rs
```diff
@@ -503,6 +499,31 @@ fn request_error(
...
+            "source API request hit rate limit too many times for {method_label} {logged_url}"
+        )));
+    }
+    let wait_secs = rate_limit_wait_secs(response.headers(), SystemTime::now());
+    Ok(Some(Duration::from_secs(wait_secs)))
+}
+
+fn has_rate_limit_headers(headers: &HeaderMap) -> bool {
+    headers.contains_key("Retry-After") || headers.contains_key("X-RateLimit-Reset")
+}
```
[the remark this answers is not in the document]
simonwhitaker: `X-` headers are often proprietary (with [some exceptions](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers)), so it feels like these would be better defined in a source spec than in the code.

simonwhitaker (approved):

antonmry (approved):

.claude/scheduled_tasks.lock
```diff
@@ -0,0 +1 @@
+{"sessionId":"aaf494a3-5622-4ce3-8f31-24b301e42570","pid":6347,"acquiredAt":1776253555618}
```
AlbertQM: What's this?

sources/github/manifest.yaml
```diff
@@ -1,9 +1,13 @@
 name: github
-version: 1.1.4
+version: 1.1.6
```
AlbertQM: Why are we bumping the version by 2? 🤔

sources/github/README.md
```diff
@@ -1,6 +1,6 @@
 # GitHub Connector
 
-**Version:** 1.1.4
+**Version:** 1.1.6
```
AlbertQM: (same question about the +2 bump, just curious)

jsummerfield:
Bare 429s lose the old fallback retry path.

This refactor changes the default behavior for every HTTP source, not just GitHub. Before this patch, a plain 429 without retry headers still got a bounded fallback retry with a short default backoff; now it fails immediately because `retry_after` is `None`.

If that behavior change is intentional, it should be called out explicitly. Otherwise, consider preserving the old behavior by allowing a short fallback retry for bare 429s while still keeping the stricter handling for long waits and GitHub-specific 403 classification.

jsummerfield (approved):

PR #92, opened 2026-04-14
feat(engine): emit AIP-193 structured errors for provider failures

mystic123:
## Summary

- Introduces `StructuredQueryError` with first-class `summary`, `detail`, `hint` fields (not metadata conventions) and a `pub(crate)` constructor so only `coral-engine` can build them
- Rewrites `provider_error_to_core` to produce structured errors for HTTP provider failures (400/401/403/404/429/5xx) with actionable hints
- `coral-app` owns wire encoding: inserts fields into `ErrorInfo.metadata`, renders plain-text fallback for `Status::message()`, applies trailer-safe truncation
- `coral-client` decodes AIP-193 details and promotes `summary`/`detail`/`hint` out of metadata into first-class `CoralQueryError` fields
- `StructuredQueryError` implements `Display` with the full multi-line message so direct `CoreError` consumers retain detail and hint

### Ownership model

| Layer | Crate | Owns |
|---|---|---|
| Error semantics | `coral-engine` | `StructuredQueryError` with typed fields, no transport awareness |
| Wire encoding | `coral-app` | AIP-193 `ErrorInfo` packing, plain-text rendering, trailer truncation |
| Wire decoding | `coral-client` | `CoralQueryError` with first-class summary/detail/hint |
| Rendering | `coral-cli`, `coral-mcp` | Consume `CoralQueryError` fields directly (PRs #100, #102) |

**Stacked on:** #84 (merged)

## Test plan

- [x] 5 engine unit tests (missing filter, 401/500 provider errors, URL sanitization)
- [x] 15 engine integration tests (HTTP 500/401, missing required filter assertions via `sqe.summary()`/`sqe.detail()`/`sqe.hint()`)
- [x] 7 client decoder tests (structured decode, promotion out of metadata, plain fallback)
- [x] clippy + fmt clean

### DIFF
--- a/crates/coral-app/src/bootstrap/error.rs
+++ b/crates/coral-app/src/bootstrap/error.rs
@@ -95,22 +95,47 @@ pub(crate) fn app_status(error: AppError) -> Status {
 pub(crate) fn core_status(error: CoreError) -> Status {
     match error {
         CoreError::Structured(sqe) => {
-            let mut details: Vec<ErrorDetail> =
-                vec![ErrorDetail::ErrorInfo(tonic_types::ErrorInfo::new(
-                    sqe.reason(),
-                    CORAL_ERROR_DOMAIN,
-                    sqe.metadata().clone(),
-                ))];
+            let mut metadata = sqe.metadata().clone();
+            metadata.insert("summary".to_string(), sqe.summary().to_string());
+            if !sqe.detail().is_empty() {
+                metadata.insert("detail".to_string(), sqe.detail().to_string());
+            }
+            if let Some(hint) = sqe.hint() {
+                metadata.insert("hint".to_string(), hint.to_string());
+            }
+
+            let mut details: Vec<ErrorDetail> = vec![ErrorDetail::ErrorInfo(
+                tonic_types::ErrorInfo::new(sqe.reason(), CORAL_ERROR_DOMAIN, metadata),
+            )];
             if sqe.retryable() {
                 details.push(ErrorDetail::RetryInfo(tonic_types::RetryInfo::new(None)));
             }
-            Status::with_error_details_vec(grpc_code(sqe.status()), sqe.plain_message(), details)
+
+            let plain = render_plain_message(sqe.summary(), sqe.detail(), sqe.hint());
+            Status::with_error_details_vec(
+                grpc_code(sqe.status()),
+                truncate_status_detail(plain),
+                details,
+            )
         }
         other => Status::new(
             grpc_code(other.status_code()),
             truncate_status_detail(other.to_string()),
         ),
     }
+}
+
+fn render_plain_message(summary: &str, detail: &str, hint: Option<&str>) -> String {
+    let mut message = summary.to_string();
+    if !detail.is_empty() {
+        message.push('\n');
+        message.push_str(detail);
+    }
+    if let Some(hint) = hint {
+        message.push_str("\nHint: ");
+        message.push_str(hint);
+    }
+    message
 }
 
 fn grpc_code(status: StatusCode) -> Code {
--- a/crates/coral-app/src/bootstrap/error.rs
+++ b/crates/coral-app/src/bootstrap/error.rs
@@ -2,6 +2,7 @@
 
 use coral_engine::{CoreError, StatusCode};
 use tonic::{Code, Status};
+use tonic_types::{ErrorDetail, StatusExt as _};
 
 use crate::state::CredentialsError;
 
@@ -88,10 +89,24 @@ pub(crate) fn app_status(error: AppError) -> Status {
     reason = "used directly as a map_err adapter across tonic service handlers"
 )]
 pub(crate) fn core_status(error: CoreError) -> Status {
-    Status::new(
-        grpc_code(error.status_code()),
-        truncate_status_detail(error.to_string()),
-    )
+    match error {
+        CoreError::Structured(sqe) => {
+            let mut details: Vec<ErrorDetail> =
+                vec![ErrorDetail::ErrorInfo(tonic_types::ErrorInfo::new(
+                    sqe.reason(),
+                    "coral.withcoral.com",
+                    sqe.metadata().clone(),
+                ))];
+            if sqe.retryable() {
+                details.push(ErrorDetail::RetryInfo(tonic_types::RetryInfo::new(None)));
+            }
+            Status::with_error_details_vec(grpc_code(sqe.status()), sqe.plain_message(), details)
+        }
+        other => Status::new(
+            grpc_code(other.status_code()),
+            truncate_status_detail(other.to_string()),
+        ),
+    }
 }
 
 fn grpc_code(status: StatusCode) -> Code {
--- a/crates/coral-app/src/bootstrap/error.rs
+++ b/crates/coral-app/src/bootstrap/error.rs
@@ -98,7 +98,10 @@ pub(crate) fn core_status(error: CoreError) -> Status {
             let mut metadata = sqe.metadata().clone();
             metadata.insert("summary".to_string(), sqe.summary().to_string());
             if !sqe.detail().is_empty() {
-                metadata.insert("detail".to_string(), sqe.detail().to_string());
+                metadata.insert(
+                    "detail".to_string(),
+                    truncate_status_detail(sqe.detail().to_string()),
+                );
             }
             if let Some(hint) = sqe.hint() {
                 metadata.insert("hint".to_string(), hint.to_string());
--- a/crates/coral-app/src/bootstrap/error.rs
+++ b/crates/coral-app/src/bootstrap/error.rs
@@ -6,6 +6,10 @@ use tonic_types::{ErrorDetail, StatusExt as _};
 
 use crate::state::CredentialsError;
 
+/// Coral error domain for `google.rpc.ErrorInfo`.
+/// Matches `coral_client::CORAL_ERROR_DOMAIN`.
+const CORAL_ERROR_DOMAIN: &str = "coral.withcoral.com";
+
 /// Errors surfaced by the local application layer.
 #[derive(Debug, thiserror::Error)]
 pub enum AppError {
@@ -94,7 +98,7 @@ pub(crate) fn core_status(error: CoreError) -> Status {
             let mut details: Vec<ErrorDetail> =
                 vec![ErrorDetail::ErrorInfo(tonic_types::ErrorInfo::new(
                     sqe.reason(),
-                    "coral.withcoral.com",
+                    CORAL_ERROR_DOMAIN,
                     sqe.metadata().clone(),
                 ))];
             if sqe.retryable() {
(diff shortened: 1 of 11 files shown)


crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -0,0 +1,305 @@
+//! Engine-internal structured query error with user- and agent-facing hints.
+//!
+//! This type is **not** re-exported from the `coral-engine` crate root.
```
jsummerfield: Can you try to de-Claude some of these before submitting.

crates/coral-app/src/bootstrap/error.rs
```diff
@@ -59,7 +60,21 @@ pub(crate) fn app_status(error: AppError) -> Status {
     reason = "used directly as a map_err adapter across tonic service handlers"
 )]
 pub(crate) fn core_status(error: CoreError) -> Status {
-    Status::new(grpc_code(error.status_code()), error.to_string())
+    match error {
+        CoreError::Structured(sqe) => {
+            let mut details: Vec<ErrorDetail> =
+                vec![ErrorDetail::ErrorInfo(tonic_types::ErrorInfo::new(
+                    sqe.reason(),
+                    "coral.withcoral.com",
```
jsummerfield: Reuse the same const if possible.

crates/coral-engine/src/contracts/query_error.rs
mystic123: Done in 2421d19 — rewrote the module doc.

crates/coral-app/src/bootstrap/error.rs
mystic123: Added a local `CORAL_ERROR_DOMAIN` const in 2421d19. Can't import from `coral_client` directly — it's only a dev-dependency and promoting it would make the server depend on the client crate. Added a comment pointing to the canonical source.

crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -0,0 +1,304 @@
...
+            metadata,
+            status: StatusCode::FailedPrecondition,
+        }
+    }
+
+    #[allow(
+        clippy::needless_pass_by_value,
+        reason = "method and url are conditionally moved into the metadata HashMap"
+    )]
+    pub(crate) fn provider_request(
```
antonmry: Minor comment: coral also supports non-http backends but this assumes it's http. Not sure if it would be enough with a rename to `http_request` or `http_provider_request` so it doesn't collide with more request types in the future or this is an indicator of more generalization being needed here.

mystic123: Good point — renamed to `http_provider_request` in 0055646. Non-HTTP backends will get their own constructors when they need structured errors.

jsummerfield:
Stepping back, I think the larger concern is that the error model is getting useful but not yet cleanly owned.

After tracing PR #92 end to end and then looking at the CLI consumer in PR #100, I think the direction is right, but the current layering is still too split:

- `ProviderQueryError`
- engine-internal `QueryError`
- contract-level `StructuredQueryError`
- AIP-193 encoding in `coral-app`
- decode in `coral-client`
- rendering in CLI / MCP

That’s a lot of shapes for one concept, and PR #100 makes the main issue clearer: the downstream consumer is not really consuming a typed Coral error contract yet, it is reading ad hoc `summary` / `detail` / `hint` keys out of `ErrorInfo.metadata`.

So the structured-error idea looks good, but the contract still feels presentation-oriented and duplicated.

What I think would be cleaner:

- Keep structured errors.
- Collapse the engine-side duplication into one transport-neutral `CoreError` shape that can carry first-class semantic fields like `code`, `reason`, `summary`, `detail`, `hint`, `retryable`, and any real metadata.
- Do not store a pre-rendered `plain_message` in `coral-engine`; that belongs in `coral-app`, which owns transport encoding and trailer-size budgeting.
- Make `coral-client` decode into a normalized Coral error object with first-class `summary/detail/hint`, instead of asking each consumer to pull those out of metadata.
- Then have CLI and MCP render from that normalized object, rather than each adapter inventing its own interpretation policy.

So my recommendation is not “back out structured errors”; it is “tighten the ownership and make the structured contract truly canonical before more consumers build on it.” PR #100 actually increases confidence in the idea, but also makes it easier to see that the current wire contract is still effectively `metadata` conventions rather than a single normalized error model.

crates/coral-engine/src/runtime/query.rs
```diff
@@ -117,35 +120,42 @@ fn datafusion_to_core(error: &DataFusionError) -> CoreError {
...
+        } => QueryError::http_provider_request(
+            source_schema.clone(),
+            table.clone(),
+            *status,
+            method.clone(),
+            url.clone(),
+            detail.clone(),
+        ),
+    };
+    CoreError::Structured(Box::new(StructuredQueryError::new(
```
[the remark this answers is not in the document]
mystic123: Not applicable — this is an internal engine contract, not user-facing. Docs for the consumer-facing error shape were discussed and deferred on #100.

crates/coral-app/src/bootstrap/error.rs
```diff
@@ -88,10 +93,49 @@ pub(crate) fn app_status(error: AppError) -> Status {
...
-    Status::new(
-        grpc_code(error.status_code()),
-        truncate_status_detail(error.to_string()),
-    )
+    match error {
+        CoreError::Structured(sqe) => {
+            let mut metadata = sqe.metadata().clone();
+            metadata.insert("summary".to_string(), sqe.summary().to_string());
+            if !sqe.detail().is_empty() {
+                metadata.insert("detail".to_string(), sqe.detail().to_string());
```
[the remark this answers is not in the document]
mystic123: Good catch — fixed in 6122536. The detail value now goes through truncate_status_detail before insertion into ErrorInfo.metadata.

crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -0,0 +1,369 @@
+//! Structured query error type for provider failures.
+//!
+//! `StructuredQueryError` is the canonical error shape that crosses the
+//! engine boundary. The app layer reads it via getters and encodes it
+//! as AIP-193 error details for the wire.
```
jsummerfield: Nit, but no need to discuss other crates' responsibilities here.

crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -0,0 +1,369 @@
...
+            metadata,
+            status: StatusCode::FailedPrecondition,
+        }
+    }
+
+    #[allow(
+        clippy::needless_pass_by_value,
+        reason = "method and url are conditionally moved into the metadata HashMap"
+    )]
+    pub(crate) fn http_provider_request(
```
jsummerfield: There's too much backend specifics in this `contracts/` file. The intent of this file/folder is to capture the canonical contract/seam, and keep the details outside.

Right now `query_error.rs `also owns HTTP status dispatch, URL sanitization, shell quoting, and provider-detail enrichment.

Better would be to put that mapping next to the query/backend code, e.g.  `backends/http/error_mapping.rs`

crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -0,0 +1,369 @@
...
+        let raw_detail = detail.into();
+        let source_shell = shell_arg(&source);
+        let sanitized_url = url.and_then(|raw| sanitize_request_url(&raw));
+
+        let (reason, summary, hint, status) = match http_status {
+            Some(400) => (
+                "INVALID_QUERY_SHAPE",
+                "Source rejected the request".to_string(),
+                Some("Adjust the query filters or shape to match the target table's supported inputs.".to_string()),
+                StatusCode::FailedPrecondition,
```
jsummerfield: If 400 really means “caller asked for an unsupported shape,” keep `INVALID_QUERY_SHAPE` but use `InvalidArgument`

crates/coral-engine/src/contracts/query_error.rs
mystic123: Trimmed in 6893e8a — module doc now just says "Canonical structured query error contract."

crates/coral-engine/src/contracts/query_error.rs
mystic123: Moved in 6893e8a — HTTP status dispatch, URL sanitization, shell quoting, and detail enrichment now live in backends/http/error.rs next to ProviderQueryError. contracts/query_error.rs is just the struct + Display + getters + new().

crates/coral-engine/src/contracts/query_error.rs
mystic123: Fixed in 6893e8a — 400 now maps to InvalidArgument.

antonmry (approved):

crates/coral-engine/src/contracts/query_error.rs
mystic123: Moved the HTTP mapping to `crates/coral-engine/src/backends/http/error.rs` in 42c1a58 — the old `http_provider_request` constructor is now `ProviderQueryError::to_structured()` colocated with the HTTP backend, so the naming is explicitly scoped and won't collide with future non-HTTP backends. The `contracts/` seam stays backend-agnostic.

crates/coral-engine/src/contracts/query_error.rs
mystic123: Done in 42c1a58 — moved HTTP status dispatch, URL sanitization, shell quoting, and provider-detail enrichment to `crates/coral-engine/src/backends/http/error.rs`. `contracts/query_error.rs` now only holds the `StructuredQueryError` struct, getters, `Display`, and `new()`.

crates/coral-engine/src/runtime/query.rs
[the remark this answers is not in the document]
mystic123: Skipping the `/docs` update: the error shape is already self-describing through AIP-193 `ErrorInfo` (first-class `summary`/`detail`/`hint` + metadata), and we're deliberately not introducing a parallel developer-facing doc surface that would need to be kept in sync. Resolving.

jsummerfield:
Follow-up from review:

1. `CoreError::Structured` still reads too generically. The shape itself is fine, but the variant name makes the enum feel like a wrapper around another top-level error type. I'd rename it to `CoreError::QueryFailure(Box<StructuredQueryError>)` so the relationship is explicit: `CoreError` stays the engine-wide sum type, and this variant is the rich query-failure payload.

2. In the HTTP mapping code, prefer `reqwest::StatusCode` constants over magic numbers for the status dispatch. `UNAUTHORIZED`, `FORBIDDEN`, `NOT_FOUND`, `TOO_MANY_REQUESTS`, etc. will read better and make the intent harder to misread than raw `401/403/404/429` branches.

I would keep the overall architecture, just tighten those two seams.

jsummerfield (approved):
See PR comment

PR #93, opened 2026-04-14
test(cli): E2E coverage for non-interactive commands

sauldhernandez:
## Intent

Add black-box E2E test coverage for all non-interactive CLI commands using a configurable mock gRPC server. The mock is feature-gated behind `cli-test-server` and injected via `CORAL_ENDPOINT`.

## What it enables

19 tests across five categories:

**Command output** — `sql` (table + JSON), `source list`, `source discover`, `source test`, `source remove`

**Output shape** — multi-column/multi-row tables, empty result sets, JSON array formatting, null field omission

**Error surfaces** — gRPC server errors propagate to stderr with the right message

**Arg validation** — `source add` requires exactly one of `--name` or `--file` (clap enforces)

**Name validation** — `source test` and `source remove` reject names containing `/`

**Interactive gating** — `source add` fails with a TTY requirement error in non-interactive mode

## Usage

```
cargo test -p coral-cli --features cli-test-server
```


crates/coral-cli/tests/cli_e2e.rs
```diff
@@ -33,13 +49,453 @@ async fn source_list_renders_configured_sources() {
...
+    )
+    .await;
+
+    let assert = server
+        .cmd()
+        .args(["sql", "--format", "json", "select id, name from users"])
+        .assert()
+        .success();
+
+    let stdout = String::from_utf8_lossy(&assert.get_output().stdout);
```
jsummerfield, line 376: The test is checking raw substrings instead of parsing the JSON. If the formatter regresses to emit `"name": null` with different key ordering or whitespace, this still passes because it only rejects the exact fragment `"id":2,"name"`. Since the point here is to pin the JSON contract, deserialize the output and assert that the second object only contains `id`.

crates/coral-cli/tests/harness/mod.rs
```diff
@@ -69,58 +248,35 @@ impl QueryService for MockQueryService {
...
-            Arc::new(schema.clone()),
-            vec![Arc::new(Int64Array::from(vec![1_i64]))],
-        )
-        .expect("build record batch");
-
-        Ok(Response::new(ExecuteSqlResponse {
-            arrow_ipc_stream: encode_arrow_ipc_stream(&schema, &[batch]).expect("encode arrow ipc"),
-            row_count: 1,
-        }))
+        Ok(Response::new(
```
jsummerfield: These tests call the real binary, but the mock services still ignore the incoming `Request` and always return canned responses. That means the suite stays green if the CLI sends the wrong SQL, source name, or workspace, and `source_remove_reports_removed_source` would even pass if the command skipped the delete RPC and just printed the confirmation string. For a PR whose value is contract coverage, the harness needs to assert the request contents before replying.

jsummerfield (approved):

PR #95, opened 2026-04-14
refactor(app): add query runtime extension seams

jsummerfield:
## Summary
This change makes query runtime customization an application concern without pushing product-specific behavior into the engine or CLI adapters.

## What changed
- add an app-owned `EngineExtensionsProvider` and thread it through server bootstrap and query orchestration
- add neutral engine extension hooks around source registration using typed `QuerySource` context
- expose reusable `ServerBuilder`, `ClientBuilder`, and shared `coral_cli::run*` entrypoints for alternate local bootstraps
- keep the default OSS path as a no-op extension provider

## Why
`coral-app` remains the composition root, `coral-engine` stays generic, and sibling binaries get a supported way to customize runtime assembly without duplicating CLI logic or leaking product-specific concepts into OSS.

## Verification
- `make rust-checks`


crates/coral-app/src/query/extensions.rs
```diff
@@ -0,0 +1,19 @@
+//! App-owned selection of optional engine extensions for query runtime builds.
+
+use coral_engine::{EngineExtensions, QuerySource};
+
+/// App-layer provider that selects engine extensions for one runtime build.
+pub trait EngineExtensionsProvider: Send + Sync {
+    /// Returns the engine extensions that apply to the selected sources.
+    fn extensions_for(&self, sources: &[QuerySource]) -> EngineExtensions;
```
sauldhernandez: Could you explain the expected behavior of the trait here? Do the extensions returned here need to be applicable to all sources on the passed slice or only some of them?

crates/coral-cli/src/lib.rs
```diff
@@ -0,0 +1,246 @@
+//! CLI entrypoint logic for the local Coral app.
```
sauldhernandez: Not against this CLI change, but it should probably go in a separate PR

crates/coral-engine/src/composition.rs
```diff
@@ -0,0 +1,108 @@
...
+    /// Reports a selected source that failed during registration.
+    ///
+    /// Returning [`SourceFailurePolicy::Abort`] causes the original source
+    /// registration error to abort runtime construction.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`SourceDecoratorError`] if the decorator cannot process the
+    /// failure event.
+    fn source_failed(
```
sauldhernandez, line 92: Based on what registry.rs does, there are two steps that can yield errors: the call to r`egister_source` and the call to `catalog.register_schema`. Currently only the former calls this `source_failed` error, but the latter doesn't .Is that intentional? Does a call to `source_failed` imply the decorator did something wrong or is it meant more as a means to cleanup because an unrelated failure happened?

sauldhernandez (changes requested):

antonmry (approved):
The PR needs a rebase and there are a bunch of comments, but I like the approach. It definitely helps to see it with coral-cloud

crates/coral-cli/src/lib.rs
jsummerfield: Ack

crates/coral-cli/src/lib.rs
```diff
@@ -0,0 +1,238 @@
+//! Shared CLI command parsing and dispatch for Coral clients.
```
jsummerfield, line 1: @sauldhernandez Had to extract to a library in this PR to avoid duplicating logic

crates/coral-engine/src/composition.rs
jsummerfield: Fixed.

crates/coral-app/src/query/extensions.rs
jsummerfield: The intent is “extensions for this runtime build”, not “one extension must apply uniformly to every source in the slice”. The slice is the full selected source set for the query, and the provider can use that to decide which extensions to install for this runtime. An extension may only affect a subset of those sources, but it must be valid in the context of that exact selected set.

I've updated the docs to make that clearer.

crates/coral-cli/src/lib.rs
```diff
@@ -0,0 +1,238 @@
...
+) -> Result<(), anyhow::Error> {
+    let output = match format {
+        OutputFormat::Table => format_batches_table(batches)?,
+        OutputFormat::Json => format_batches_json(batches)?,
+    };
+    println!("{output}");
+    Ok(())
+}
+
+async fn run_source_add(app: &AppClient, args: SourceAddArgs) -> Result<(), anyhow::Error> {
```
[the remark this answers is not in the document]
jsummerfield: Ack - to address in a follow up PR

sauldhernandez (approved):

simonwhitaker (approved):

PR #96, opened 2026-04-14
refactor(app): make source and workspace identity explicit

jsummerfield:
## Summary
- introduce `WorkspaceName` and `SourceName` as app-owned identity types and keep them at the `coral-app` service, state, layout, and manager seams
- replace the flat config source list with a keyed `SourceCatalog`, preserving workspace/source uniqueness in the type shape instead of re-scanning vectors
- fold workspace validation into the checked identity types, remove the extra validator layer, and keep protobuf conversion at the transport edge

## Testing
- `cargo test -p coral-app`
- `make rust-checks`

## Intent
This refactor makes installed-source identity explicit inside `coral-app` instead of relying on raw strings, protobuf structs, and naming conventions to carry that invariant. The goal is to keep the app layer honest about what it owns, reduce stringly cross-layer seams, and make persistence and transport conversions easier to reason about.


crates/coral-app/src/sources/name.rs
```diff
@@ -0,0 +1,35 @@
...
+/// before app logic runs.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+#[serde(transparent)]
+pub(crate) struct SourceName(String);
+
+impl SourceName {
+    /// Wrap an already-validated source name for app-internal use.
+    #[must_use]
+    pub(crate) fn new(name: String) -> Self {
+        Self(name)
```
sauldhernandez: docstring says the name arg would already be validated. By whom?

crates/coral-app/src/workspaces/name.rs
```diff
@@ -0,0 +1,43 @@
...
+    /// boundaries that still operate on strings.
+    #[must_use]
+    pub(crate) fn as_str(&self) -> &str {
+        &self.0
+    }
+
+    /// Project the app-owned identity back into the gRPC transport type at the
+    /// service boundary.
+    #[must_use]
+    pub(crate) fn to_proto(&self) -> Workspace {
```
sauldhernandez: Proto conversion should be in the service layer, not here.

crates/coral-app/src/workspaces/name.rs
```diff
@@ -0,0 +1,43 @@
...
+/// `WorkspaceName` at the service edge, and only converted back when preparing
+/// transport responses.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub(crate) struct WorkspaceName(String);
+
+impl WorkspaceName {
+    /// Wrap an already-validated workspace name for app-internal use.
+    #[must_use]
+    pub(crate) fn new(name: String) -> Self {
+        Self(name)
```
sauldhernandez: Same question here. What's a valid name?

crates/coral-app/src/workspaces/validator.rs
```diff
@@ -25,24 +26,27 @@ impl WorkspaceManager {
...
+        &self,
+        workspace: Option<&Workspace>,
+    ) -> Result<WorkspaceName, AppError> {
         let workspace =
             workspace.ok_or_else(|| AppError::InvalidInput("missing workspace".to_string()))?;
         self.normalize(workspace)
     }
 
-    pub(crate) fn require(&self, workspace: Option<&Workspace>) -> Result<Workspace, Status> {
+    pub(crate) fn require(&self, workspace: Option<&Workspace>) -> Result<WorkspaceName, Status> {
```
sauldhernandez: If the goal is to make the domain side more agnostic to the API, why do the Status mapping here and not closer to the gRPC layer?

crates/coral-app/src/workspaces/validator.rs
```diff
@@ -25,24 +26,27 @@ impl WorkspaceManager {
...
         self.require_app(workspace).map_err(app_error_to_status)
     }
 
-    pub(crate) fn normalize(&self, workspace: &Workspace) -> Result<Workspace, AppError> {
-        Ok(Workspace {
-            name: self.validate_path_name("workspace name", &workspace.name)?,
-        })
+    pub(crate) fn normalize(&self, workspace: &Workspace) -> Result<WorkspaceName, AppError> {
```
sauldhernandez: Why not move this logic into the WorkspaceName struct directly? That way, if you have an instance of WorkspaceName, then you _know_ it's a valid one.

sauldhernandez (changes requested):
Small nit unrelated to the changes but relevant on the context of workspaces: coral-engine's lib.rs makes a reference to workspaces in its docstring, but that seems unnecessary since the engine is (correctly IMO) oblivious of workspaces.

jsummerfield:
@sauldhernandez thanks for the review, all good points. PTAL

sauldhernandez (approved):

antonmry (approved):

PR #97, opened 2026-04-14
feat(spec): top level input declaration

iaptsiauri:
## Summary
Instead of inferring variables and secrets from variable.* / secret.* references, HTTP manifests now declare inputs in a top-level inputs: block with `kind`, `optional variable default`, and `optional hint`. Runtime references use `input.KEY`, and the declared kind determines whether the value resolves from source variables or source secrets.

This also cleans up input handling end-to-end:

- removes auth.required_secrets so there is no parallel hidden install contract
- centralizes variable default materialization in coral-app during install/import
- keeps CLI prompting thin by surfacing hints but leaving default application to the server
- restricts inputs support to backend: http at the schema layer so parquet/jsonl do not advertise unsupported behavior

## Validation 

  | # | Test | Result |
  |---|---|---|
  | 1 | cargo check — workspace compiles with the new explicit inputs model | PASS |
  | 2 | make rust-checks  | PASS |
  | 3 | `./target/debug/coral sql --format json "select * from github.rate_limit limit 1"` | PASS |
  | 4 | `./target/debug/coral sql --format json "select number, title, comments from github.search_issues where q = 'repo:withcoral/coral is:pr' and sort = 'updated' and order = 'desc' limit 3"` | PASS |


AlbertQM:
Thanks Ilia!

I branched off this PR and added a hint for Slack, which needed more than a single line IMO:
image

My understanding is that we also want to use these hints to autogenerate the docs, to avoid drift.
I was initially thinking we'd need markdown, to render nicer docs (e.g. prettify that big Slack link above); but I think we can make it work with plaintext too, and maybe support for `[]()` links.

crates/coral-api/proto/coral/v1/sources.proto
```diff
@@ -67,6 +67,8 @@ message SourceInputSpec {
   bool required = 3;
   // Optional manifest-declared default value.
   string default_value = 4;
+  // Optional authored hint shown to the user while collecting this input.
+  string hint = 5;
```
antonmry: Would it make sense to use `optional string hint = 5` in the proto definition? proto3 optional fields track presence, so the generated Rust code gives you Option<String> directly. The conversion becomes a straight input.hint.

There's a bunch of logic `None vs Some(...)` or `unwrap_or_default` that might be eliminated with it.

antonmry (approved):

crates/coral-cli/src/source_ops.rs
```diff
@@ -318,9 +329,6 @@ pub(crate) fn finalize_input_value(
     if !value.is_empty() {
         return Ok(Some(value));
     }
-    if !input.default_value.is_empty() {
-        return Ok(Some(input.default_value.clone()));
-    }
     if input.required {
```
jsummerfield, line 332: [P2] This now drops authored defaults for optional inputs instead of materializing them. The CLI returns None for every empty optional answer here, but there is no matching server-side fallback in the current app path, so manifest defaults are lost. That includes valid cases like default: "".

crates/coral-engine/src/backends/http/client.rs
```diff
@@ -788,14 +774,15 @@ fn resolve_template_token(
 ) -> Result<String> {
     let default = token.default_value().map(ToString::to_string);
 
-    if token.namespace() == &TemplateNamespace::Secret {
+    if token.namespace() == &TemplateNamespace::Input {
         return source_secrets
             .get(token.key())
+            .or_else(|| source_variables.get(token.key()))
```
jsummerfield: [P2] The new manifest contract says the declared input kind decides which store supplies input.KEY, but this resolver still falls back from secrets to variables. If a source upgrade leaves stale state behind or flips an input from secret to variable, requests can silently use the wrong store instead of following the manifest.

jsummerfield:
[P3] This PR changes the user-facing source manifest contract to top-level inputs plus input.* references, but the docs still show the removed variable./secret. syntax. docs/reference/source-spec-reference.mdx and docs/guides/write-a-custom-source.mdx should be updated in the same change so users are not sent to manifests that no longer validate.

crates/coral-cli/src/source_ops.rs
iaptsiauri: Defaults are moved server side.

crates/coral-api/proto/coral/v1/sources.proto
jsummerfield: I've been leaning on Google's AIP when it comes to proto design. The [AGENTS.md](http://AGENTS.md) borrows from https://google.aip.dev/149:

\>Be careful with proto3 `optional`. Use presence tracking only when unset must  
be distinguished from the default value; do not add `optional` by default to  
strings or booleans just to signal "this may be absent".

Protobuf contracts should describe API semantics first, not Rust preferences first. We may have clients in other languages.  


In this case, is there a difference between `no hint authored` and `hint authored as ""`? If the client behaviour is just “show the hint when non-empty”, then presence is not meaningful and optional is just buying Rust Option\<String> at the wire level

crates/coral-cli/src/source_ops.rs
jsummerfield: I traced the current head and but I don’t see the server-side materialisation. finalize_input_value() now omits empty optional answers, but validate_bindings() only validates and persist_source() stores request.bindings.variables verbatim. I couldn’t find a path that inserts candidate.inputs[*].default_value before runtime. If I’m missing the new server-side hook, can you point me to it? Otherwise this still looks unresolved.

jsummerfield (approved):

simonwhitaker (approved):

PR #100, opened 2026-04-15
feat(cli): render structured query errors as Error/Detail/Hint

mystic123:
## Summary

- Intercepts gRPC status errors on the `coral sql` path and decodes AIP-193 details via `coral_client::decode_status_error`
- **Structured errors** render as labelled `Error:` / `Detail:` / `Hint:` lines on stderr, reading first-class fields from `CoralQueryError`
- **Plain/legacy errors** render as `Error (<gRPC code>): <message>` so distinct failure modes remain distinguishable
- Stdout stays empty on failure, preserving the scripting contract

### Verified output (structured — missing required filter)

```
$ coral sql "SELECT * FROM github.pulls"
Error: github.pulls requires `WHERE owner = <constant>`
Detail: github.pulls requires a constant equality filter on owner
Hint: Add a constant equality filter on `owner` or inspect `coral.columns` / `coral.tables` first.
```

### Verified output (plain fallback — empty SQL)

```
$ coral sql ""
Error (invalid argument): invalid input: SQL must not be empty
```

**Stacked on:** #92

## Test plan

- [x] 8 unit tests covering structured rendering, optional detail/hint omission, summary-missing fallback, gRPC code in plain path, multi-line preservation
- [x] All 19 coral-cli tests pass
- [x] clippy clean
- [x] Manual verification with live GitHub source

### DIFF
--- a/crates/coral-cli/src/main.rs
+++ b/crates/coral-cli/src/main.rs
@@ -123,7 +123,10 @@ async fn main() -> Result<(), anyhow::Error> {
                 .await
             {
                 Ok(response) => response.into_inner(),
-                Err(status) => query_error::render_query_error_and_exit(&status),
+                Err(status) => {
+                    eprint!("{}", query_error::render_query_error(&status));
+                    std::process::exit(1);
+                }
             };
             let result = decode_execute_sql_response(&response)?;
             print_batches(result.batches(), args.format)?;
--- a/crates/coral-cli/src/main.rs
+++ b/crates/coral-cli/src/main.rs
@@ -9,6 +9,7 @@
 mod bootstrap;
 mod branding;
 mod onboard;
+mod query_error;
 mod source_ops;
 
 use std::path::PathBuf;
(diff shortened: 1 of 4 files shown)


docs/reference/cli-reference.mdx
```diff
@@ -17,6 +17,24 @@ coral sql "<SQL>"
...
+Error: github.pulls requires `WHERE owner = <constant>`
+Detail: github.pulls requires a constant equality filter on owner
+Hint: Add a constant equality filter on `owner` or inspect `coral.columns` / `coral.tables` first.
+```
+
+Legacy errors include the gRPC status code:
+
+```text
+Error (not found): resource not found: github.nonexistent_table
+```
```
AlbertQM: Do we really need to document this? 🤔 Not sure if I see much value in adding this and having to maintain it in sync with code changes.

mystic123: Agreed — reverted in 8abc714. The error format is self-describing and not something users need to reference docs for.

crates/coral-cli/src/main.rs
```diff
@@ -118,14 +119,17 @@ async fn main() -> Result<(), anyhow::Error> {
...
                 .execute_sql(Request::new(ExecuteSqlRequest {
                     workspace: Some(default_workspace()),
                     sql: args.sql,
                 }))
-                .await?
-                .into_inner();
+                .await
+            {
+                Ok(response) => response.into_inner(),
+                Err(status) => query_error::render_query_error_and_exit(&status),
```
antonmry: nitpick: consider to move the `std::process::exit(1)` to main so we aren't exiting the app from different places and the full life-cycle always goes through main

antonmry (approved):

crates/coral-cli/src/main.rs
jsummerfield: Agree with Anton here, exiting in the helper is neater, but it's less testable and composable. Better to let main own the process termination.

jsummerfield (approved):

crates/coral-cli/src/main.rs
mystic123: Refactored in 00aab42 — `render_query_error` now returns a `String` with no side effects; `main` owns the `eprint!` + `std::process::exit(1)`.

PR #106, opened 2026-04-15
feat(docs): auto-generate bundled-sources.mdx from manifests

AlbertQM:
## Summary

- Introduces `coral-docgen`, a workspace binary that reads `sources/*/manifest.y{a,}ml` via `coral-spec` and regenerates `docs/reference/bundled-sources.mdx` so the new per-input `hint` strings surface in the docs instead of staying buried in YAML.
- Each source gets a deep-linkable `### \`<name>\`` sub-section (e.g. `/reference/bundled-sources#github`) listing its inputs with `(required|optional)`, optional default, and the authored hint. The top-of-page table links rows to those anchors.
- `make docs-check` (and a new `docs-freshness` CI job) fail the build when the generated page drifts from the manifests; `make docs-generate` regenerates.
- Also updated Slack hint to be more comprehensive, to test how larger hints would play both in CLI and in docs.

## Test plan

image

image


crates/coral-docgen/src/render.rs
```diff
@@ -0,0 +1,359 @@
...
+pub(crate) fn escape_mdx(input: &str) -> String {
+    let mut out = String::with_capacity(input.len());
+    let mut chars = input.chars().peekable();
+    let mut in_fence = false;
+    // Track whether we're at the start of a line for fence detection.
+    let mut at_line_start = true;
+
+    while let Some(ch) = chars.next() {
+        // Detect a ``` fence opening/closing at the start of a line.
+        if at_line_start && ch == '`' {
```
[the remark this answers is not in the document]
AlbertQM: On a second thought, don't think we need to support fenced code blocks at all, and favour prose instead, to guide users. I'll update the PR and remove all this unnecessary logic, thanks!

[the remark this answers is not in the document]
AlbertQM: Huh? Don't think this was worthy of a new learning 🤷

simonwhitaker (approved):
Nice!

xtask/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
```
jsummerfield, line 1: This binary looks like repo-only tooling: it walks `sources/`, reuses `coral-spec` for validation, then renders Mintlify files. That seam is real, but it is not a new runtime subsystem. Consider moving it to a workspace-level `xtask`-style tool (or similar internal tooling package) rather than a first-class crate under `crates/`, so the main crate graph stays focused on Coral product ownership boundaries.

jsummerfield (commented):
Left three comments: two correctness issues and one architecture note about keeping repo tooling out of the product crate graph.

xtask/src/nav.rs
```diff
@@ -0,0 +1,86 @@
...
+//! this module now only removes stale generator entries and leaves every
+//! hand-authored nav entry in place.
+
+use anyhow::{Context, Result};
+use serde_json::Value;
+
+/// Returns an updated `docs.json` body with any `reference/sources/*`
+/// entries stripped from the Reference group. All other navigation entries
+/// are preserved in-place.
+pub(crate) fn update_docs_json(existing: &str) -> Result<String> {
```
jsummerfield, line 19: `update_docs_json` only strips `reference/sources/*` entries; it never restores or validates `reference/bundled-sources` itself. I reproduced this by removing `reference/bundled-sources` from a copy of `docs/docs.json` and running `cargo run -p coral-docgen -- --check`, which still exited 0. That means the new freshness check can pass while the generated page is missing from Mintlify navigation.

crates/coral-docgen/src/render.rs
```diff
@@ -0,0 +1,313 @@
...
+            other => {
+                out.push(other);
+                prev_space = false;
+            }
+        }
+    }
+    out
+}
+
+/// Escape MDX-hostile characters (`{`, `}`, `<`, `>`) in plain prose.
```
jsummerfield: `ManifestInputSpec::hint` is freeform authored markdown, but this escape pass now explicitly declines fenced blocks and treats every backtick run as inline code. A hint containing a fenced example like ```yaml ... <token> {{input.KEY}} ... ``` gets rewritten with backslashes inside the block, so the generator silently mangles valid code examples instead of preserving them verbatim.

AlbertQM: I've pulled in a library to support markdown instead of hand rolling some logic, thanks!

xtask/Cargo.toml
AlbertQM: much better, thanks!

antonmry (approved):

jsummerfield (approved):
Thanks, looks good

PR #107, opened 2026-04-15
feat(cli): run source-spec test queries during source test

lazybean:
## Summary

- run top-level `test_queries` during `coral source test` and plumb structured query-test results through the spec, engine, app, API, and CLI layers
- keep `coral source add` / onboarding validation warn-only while `coral source test` stays strict, and add CLI regression coverage for both failing and passing query-test output
- document the warn-vs-strict behavior and add a lightweight `SELECT COUNT(*) AS channel_count FROM slack.channels` test query to the bundled Slack source

## Testing

- `make rust-checks`
- `cargo run --quiet -- source lint sources/slack/manifest.yaml``  
``  
`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,4 +27,3 @@ target
 # Local task files
 tasks/
 /.github/scripts/__pycache__
-.omx/
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ target
 # Local task files
 tasks/
 /.github/scripts/__pycache__
+.omx/
--- a/crates/coral-api/proto/coral/v1/sources.proto
+++ b/crates/coral-api/proto/coral/v1/sources.proto
@@ -180,11 +180,11 @@ message ValidateSourceResponse {
   // Validation-query results declared by the source spec.
   repeated QueryTestResult query_tests = 3;
   // Total number of declared validation queries.
-  uint32 declared_query_count = 4;
+  uint64 declared_query_count = 4;
   // Number of validation queries that passed.
-  uint32 passed_query_count = 5;
+  uint64 passed_query_count = 5;
   // Number of validation queries that failed.
-  uint32 failed_query_count = 6;
+  uint64 failed_query_count = 6;
   // Whether every declared validation query passed.
   bool all_query_tests_passed = 7;
 }
--- a/crates/coral-api/proto/coral/v1/sources.proto
+++ b/crates/coral-api/proto/coral/v1/sources.proto
@@ -159,16 +159,29 @@ message ValidateSourceRequest {
   string name = 2;
 }
 
+// Success details for one validation query.
+message QueryTestSuccess {
+  // Row count captured for the successful query.
+  uint64 row_count = 1;
+}
+
+// Failure details for one validation query.
+message QueryTestFailure {
+  // Error message captured for the failed query.
+  string error_message = 1;
+}
+
 // One validation query executed for a source.
 message QueryTestResult {
   // SQL text declared by the source spec.
   string sql = 1;
-  // Whether the query executed successfully.
-  bool passed = 2;
-  // Row count captured for the query result; zero when not reported.
-  uint64 row_count = 3;
-  // Error message captured for failed queries; empty when none.
-  string error_message = 4;
+  // Outcome of the validation query.
+  oneof outcome {
+    // Successful query execution metadata.
+    QueryTestSuccess success = 2;
+    // Failed query execution metadata.
+    QueryTestFailure failure = 3;
+  }
 }
 
 // Returns source details plus the tables exposed during validation.
--- a/crates/coral-api/proto/coral/v1/sources.proto
+++ b/crates/coral-api/proto/coral/v1/sources.proto
@@ -159,12 +159,34 @@ message ValidateSourceRequest {
   string name = 2;
 }
 
+// One validation query executed for a source.
+message QueryTestResult {
+  // SQL text declared by the source spec.
+  string sql = 1;
+  // Whether the query executed successfully.
+  bool passed = 2;
+  // Optional row count captured for successful queries.
+  optional uint64 row_count = 3;
+  // Optional error message captured for failed queries.
+  optional string error_message = 4;
+}
+
 // Returns source details plus the tables exposed during validation.
 message ValidateSourceResponse {
   // The validated source resource.
   Source source = 1;
   // Tables exposed by the validated source.
   repeated Table tables = 2;
+  // Validation-query results declared by the source spec.
+  repeated QueryTestResult query_tests = 3;
+  // Total number of declared validation queries.
+  uint32 declared_query_count = 4;
+  // Number of validation queries that passed.
+  uint32 passed_query_count = 5;
+  // Number of validation queries that failed.
+  uint32 failed_query_count = 6;
+  // Whether every declared validation query passed.
+  bool all_query_tests_passed = 7;
 }
 
 // Source registration and validation APIs.
--- a/crates/coral-api/proto/coral/v1/sources.proto
+++ b/crates/coral-api/proto/coral/v1/sources.proto
@@ -165,10 +165,10 @@ message QueryTestResult {
   string sql = 1;
   // Whether the query executed successfully.
   bool passed = 2;
-  // Optional row count captured for successful queries.
-  optional uint64 row_count = 3;
-  // Optional error message captured for failed queries.
-  optional string error_message = 4;
+  // Row count captured for the query result; zero when not reported.
+  uint64 row_count = 3;
+  // Error message captured for failed queries; empty when none.
+  string error_message = 4;
 }
 
 // Returns source details plus the tables exposed during validation.
--- a/crates/coral-api/proto/coral/v1/sources.proto
+++ b/crates/coral-api/proto/coral/v1/sources.proto
@@ -192,14 +192,6 @@ message ValidateSourceResponse {
   repeated Table tables = 2;
   // Validation-query results declared by the source spec.
   repeated QueryTestResult query_tests = 3;
-  // Total number of declared validation queries.
-  uint64 declared_query_count = 4;
-  // Number of validation queries that passed.
-  uint64 passed_query_count = 5;
-  // Number of validation queries that failed.
-  uint64 failed_query_count = 6;
-  // Whether every declared validation query passed.
-  bool all_query_tests_passed = 7;
 }
 
 // Source registration and validation APIs.
--- a/crates/coral-app/src/sources/service.rs
+++ b/crates/coral-app/src/sources/service.rs
@@ -17,7 +17,9 @@ use crate::sources::manager::SourceManager;
 use crate::sources::model::{
     CandidateSource, CandidateSourceInput, CandidateSourceInputKind, InstalledSource, SourceOrigin,
 };
-use crate::transport::{query_status, table_to_proto, workspace_to_proto};
+use crate::transport::{
+    query_status, query_test_result_to_proto, table_to_proto, workspace_to_proto,
+};
 use crate::workspaces::WorkspaceName;
 
 #[derive(Clone)]
@@ -181,13 +183,25 @@ impl SourceServiceApi for SourceService {
             .await
             .map_err(query_status)?;
         let tables = result
+            .outcome
             .tables
             .into_iter()
             .map(|table| table_to_proto(&workspace_name, table))
             .collect::<Vec<_>>();
+        let query_tests = result
+            .outcome
+            .query_tests
+            .into_iter()
+            .map(query_test_result_to_proto)
+            .collect();
         Ok(Response::new(ValidateSourceResponse {
             source: Some(installed_source_to_proto(&workspace_name, result.source)),
             tables,
+            query_tests,
+            declared_query_count: result.outcome.declared_query_count,
+            passed_query_count: result.outcome.passed_query_count,
+            failed_query_count: result.outcome.failed_query_count,
+            all_query_tests_passed: result.outcome.all_query_tests_passed,
         }))
     }
 }
--- a/crates/coral-app/src/sources/service.rs
+++ b/crates/coral-app/src/sources/service.rs
@@ -195,7 +195,7 @@ impl SourceServiceApi for SourceService {
             .map(|table| table_to_proto(&workspace_name, table))
             .collect::<Vec<_>>();
         let query_tests = query_tests
-            .into_iter()
+            .iter()
             .map(query_test_result_to_proto)
             .collect::<Vec<_>>();
         Ok(Response::new(ValidateSourceResponse {
--- a/crates/coral-app/src/sources/service.rs
+++ b/crates/coral-app/src/sources/service.rs
@@ -182,26 +182,30 @@ impl SourceServiceApi for SourceService {
             .validate_source(&workspace_name, &source_name)
             .await
             .map_err(query_status)?;
-        let tables = result
-            .outcome
-            .tables
+        let coral_engine::SourceValidationOutcome {
+            tables,
+            query_tests,
+            declared_query_count,
+            passed_query_count,
+            failed_query_count,
+            all_query_tests_passed,
+        } = result.outcome;
+        let tables = tables
             .into_iter()
             .map(|table| table_to_proto(&workspace_name, table))
             .collect::<Vec<_>>();
-        let query_tests = result
-            .outcome
-            .query_tests
+        let query_tests = query_tests
             .into_iter()
             .map(query_test_result_to_proto)
-            .collect();
+            .collect::<Vec<_>>();
         Ok(Response::new(ValidateSourceResponse {
             source: Some(installed_source_to_proto(&workspace_name, result.source)),
             tables,
             query_tests,
-            declared_query_count: result.outcome.declared_query_count,
-            passed_query_count: result.outcome.passed_query_count,
-            failed_query_count: result.outcome.failed_query_count,
-            all_query_tests_passed: result.outcome.all_query_tests_passed,
+            declared_query_count,
+            passed_query_count,
+            failed_query_count,
+            all_query_tests_passed,
         }))
     }
 }
--- a/crates/coral-app/src/sources/service.rs
+++ b/crates/coral-app/src/sources/service.rs
@@ -17,9 +17,7 @@ use crate::sources::manager::SourceManager;
 use crate::sources::model::{
     CandidateSource, CandidateSourceInput, CandidateSourceInputKind, InstalledSource, SourceOrigin,
 };
-use crate::transport::{
-    query_status, query_test_result_to_proto, table_to_proto, workspace_to_proto,
-};
+use crate::transport::{query_status, validate_source_response_to_proto, workspace_to_proto};
 use crate::workspaces::WorkspaceName;
 
 #[derive(Clone)]
@@ -182,31 +180,13 @@ impl SourceServiceApi for SourceService {
             .validate_source(&workspace_name, &source_name)
             .await
             .map_err(query_status)?;
-        let coral_engine::SourceValidationOutcome {
-            tables,
-            query_tests,
-            declared_query_count,
-            passed_query_count,
-            failed_query_count,
-            all_query_tests_passed,
-        } = result.outcome;
-        let tables = tables
-            .into_iter()
-            .map(|table| table_to_proto(&workspace_name, table))
-            .collect::<Vec<_>>();
-        let query_tests = query_tests
-            .iter()
-            .map(query_test_result_to_proto)
-            .collect::<Vec<_>>();
-        Ok(Response::new(ValidateSourceResponse {
-            source: Some(installed_source_to_proto(&workspace_name, result.source)),
-            tables,
-            query_tests,
-            declared_query_count,
-            passed_query_count,
-            failed_query_count,
-            all_query_tests_passed,
-        }))
+        let crate::query::manager::ValidatedSource { source, report } = result;
+        let source = installed_source_to_proto(&workspace_name, source);
+        Ok(Response::new(validate_source_response_to_proto(
+            source,
+            &workspace_name,
+            report,
+        )))
     }
 }
 
--- a/crates/coral-cli/src/onboard.rs
+++ b/crates/coral-cli/src/onboard.rs
@@ -158,6 +158,7 @@ async fn run_installed_source_menu(
                 app,
                 &source.name,
                 source_ops::TableDisplayLimit::DEFAULT,
+                source_ops::ValidationSeverityMode::Strict,
             )
             .await?;
         }
@@ -175,6 +176,7 @@ async fn run_installed_source_menu(
                 app,
                 &result.name,
                 source_ops::TableDisplayLimit::DEFAULT,
+                source_ops::ValidationSeverityMode::WarnOnly,
             )
             .await?;
         }
@@ -196,7 +198,13 @@ async fn run_add_bundled_source(
     let (variables, secrets) = source_ops::prompt_for_inputs(&inputs)?;
     let result = source_ops::add_bundled_source(app, &source.name, variables, secrets).await?;
     println!("Added source {}", result.name);
-    source_ops::validate_and_print(app, &result.name, source_ops::TableDisplayLimit::DEFAULT).await
+    source_ops::validate_and_print(
+        app,
+        &result.name,
+        source_ops::TableDisplayLimit::DEFAULT,
+        source_ops::ValidationSeverityMode::WarnOnly,
+    )
+    .await
 }
 
 async fn run_next_steps(
--- a/crates/coral-cli/src/onboard.rs
+++ b/crates/coral-cli/src/onboard.rs
@@ -154,11 +154,10 @@ async fn run_installed_source_menu(
 
     match selection.map(|i| actions[i]) {
         Some(InstalledSourceAction::Validate) => {
-            source_ops::validate_and_print(
+            source_ops::validate_and_warn(
                 app,
                 &source.name,
                 source_ops::TableDisplayLimit::DEFAULT,
-                source_ops::ValidationSeverityMode::Strict,
             )
             .await?;
         }
@@ -172,11 +171,10 @@ async fn run_installed_source_menu(
             let result =
                 source_ops::add_bundled_source(app, &source.name, variables, secrets).await?;
             println!("Reconfigured source {}", result.name);
-            source_ops::validate_and_print(
+            source_ops::validate_and_warn(
                 app,
                 &result.name,
                 source_ops::TableDisplayLimit::DEFAULT,
-                source_ops::ValidationSeverityMode::WarnOnly,
             )
             .await?;
         }
@@ -198,13 +196,7 @@ async fn run_add_bundled_source(
     let (variables, secrets) = source_ops::prompt_for_inputs(&inputs)?;
     let result = source_ops::add_bundled_source(app, &source.name, variables, secrets).await?;
     println!("Added source {}", result.name);
-    source_ops::validate_and_print(
-        app,
-        &result.name,
-        source_ops::TableDisplayLimit::DEFAULT,
-        source_ops::ValidationSeverityMode::WarnOnly,
-    )
-    .await
+    source_ops::validate_and_warn(app, &result.name, source_ops::TableDisplayLimit::DEFAULT).await
 }
 
 async fn run_next_steps(
--- a/crates/coral-cli/tests/source.rs
+++ b/crates/coral-cli/tests/source.rs
@@ -146,7 +146,103 @@ tables:
         "expected query-test summary in stdout, got: {stdout}"
     );
     assert!(
+        stdout.contains("SELECT * FROM local_messages.missing"),
+        "expected failing query text in stdout, got: {stdout}"
+    );
+    assert!(
         stderr.contains("1 of 1 validation query failed"),
         "expected strict failure in stderr, got: {stderr}"
     );
 }
+
+#[test]
+fn source_test_succeeds_when_query_tests_pass() {
+    let config_dir = tempdir().expect("failed to create temp dir");
+    let source_dir = config_dir
+        .path()
+        .join("workspaces")
+        .join("default")
+        .join("sources")
+        .join("local_messages");
+    std::fs::create_dir_all(source_dir.join("fixture-data")).expect("create source dir");
+    std::fs::write(
+        source_dir.join("fixture-data/messages.jsonl"),
+        r#"{"type":"user","text":"hello"}
+"#,
+    )
+    .expect("write fixture data");
+    std::fs::write(
+        source_dir.join("manifest.yaml"),
+        r#"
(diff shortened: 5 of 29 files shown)


lazybean:
* **#107** <a href="https://app.graphite.com/github/pr/withcoral/coral/107?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 10  dependent PRs ([#133](https://github.com/withcoral/coral/pull/133) <a href="https://app.graphite.com/github/pr/withcoral/coral/133?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#135](https://github.com/withcoral/coral/pull/135) <a href="https://app.graphite.com/github/pr/withcoral/coral/135?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#136](https://github.com/withcoral/coral/pull/136) <a href="https://app.graphite.com/github/pr/withcoral/coral/136?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 7 others) 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/107?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.gitignore
lazybean, line 1: no no no

crates/coral-engine/src/contracts/query.rs
```diff
@@ -65,6 +65,94 @@ impl QuerySource {
...
+    pub failed_query_count: u32,
+    /// Whether every declared validation query passed.
+    pub all_query_tests_passed: bool,
+}
+
+impl SourceValidationOutcome {
+    #[must_use]
+    /// Builds one structured source-validation outcome.
+    pub fn new(tables: Vec<super::TableInfo>, query_tests: Vec<QueryTestResult>) -> Self {
+        let declared_query_count = u32::try_from(query_tests.len()).unwrap_or(u32::MAX);
```
antonmry: Nitpick: consider to use uint64 at the proto level, so we can't avoid the cast and the unwrap_or

crates/coral-engine/tests/engine/harness.rs
```diff
@@ -58,6 +58,22 @@ pub(crate) fn assert_row_count(execution: &QueryExecution, expected: usize) {
     assert_eq!(execution_to_rows(execution).len(), expected);
 }
 
+#[allow(
+    dead_code,
+    reason = "retained for engine tests that may assert internal errors"
```
antonmry: Consider to remove the clippy exception or the function.

crates/coral-engine/tests/engine/harness.rs
```diff
@@ -58,6 +58,22 @@ pub(crate) fn assert_row_count(execution: &QueryExecution, expected: usize) {
...
+    match error {
+        CoreError::Internal(detail) => assert_eq!(detail, expected_detail),
+        other => panic!("expected CoreError::Internal, got {other:?}"),
+    }
+}
+
+#[allow(
+    dead_code,
+    reason = "retained for other engine tests that may assert invalid input"
+)]
```
antonmry, line 64: Consider to remove the clippy exception or the function.

docs/guides/write-a-custom-source.mdx
```diff
@@ -75,6 +81,29 @@ coral sql "
...
+When you add `test_queries` to a source spec, treat them as a lightweight validation contract for that source.
+
+1. Run `coral source lint ./my-source.yaml` before installing the source.
+2. Add or update the source with `coral source add --file ./my-source.yaml`.
+3. Run `coral source test <source_name>` and confirm:
+   - the source still exposes the expected tables
+   - each declared `test_queries` entry runs in manifest order
+   - successful queries show a pass status and optional row count
+   - failing queries surface the SQL text plus a readable error message
+4. Run one or two real `coral sql` queries against the new tables to confirm the shape is useful beyond the validation checks.
```
antonmry: nitpick: consider to remove `one or two`

antonmry (approved):
I've added some minor comments but the engine looks good to me.

docs/guides/write-a-custom-source.mdx
lazybean: Addressed in 00fa45c: I removed the "one or two" wording to make the guidance less fuzzy.

crates/coral-engine/tests/engine/harness.rs
lazybean: Addressed in 00fa45c: I removed the unused helper instead of keeping the dead_code allowance.

crates/coral-engine/src/contracts/query.rs
lazybean: Good call. I left this PR scoped to the conservative overflow fix because switching these counters to uint64 would widen the contract surface (including the proto) beyond this follow-up, but I agree it could be a sensible separate cleanup.

crates/coral-cli/tests/source.rs
```diff
@@ -71,3 +71,174 @@ fn source_test_errors_when_required_secret_is_missing() {
...
+        .join("default")
+        .join("sources")
+        .join("local_messages");
+    std::fs::create_dir_all(source_dir.join("fixture-data")).expect("create source dir");
+    std::fs::write(
+        source_dir.join("fixture-data/messages.jsonl"),
+        r#"{"type":"user","text":"hello"}
+"#,
+    )
+    .expect("write fixture data");
```
sauldhernandez: This test could use the mock gRPC service in the harness instead of trying to build the whole stack by itself

crates/coral-cli/tests/source.rs
```diff
@@ -71,3 +71,174 @@ fn source_test_errors_when_required_secret_is_missing() {
...
+    std::fs::write(
+        source_dir.join("manifest.yaml"),
+        r#"
+name: local_messages
+version: 0.1.0
```
sauldhernandez: Same here, you can use the mock gRPC service in the test harness and avoid setting up all of this.

crates/coral-engine/src/lib.rs
```diff
@@ -129,10 +129,64 @@ impl CoralQuery {
...
+    }
+
+    /// Validates one source and then executes any declared validation queries.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`CoreError`] if the source cannot be initialized or enumerated.
+    /// Query-test failures are reported in the returned outcome instead of
+    /// failing the whole call.
+    pub async fn validate_source_with_tests(
```
sauldhernandez: I'm not entirely sure if this should live in the engine crate. The code reads more like a user of the engine and not as part of the engine itself. Maybe move it to the app crate? In the end, validating a source sounds more of a source lifecycle concern, not a query engine concern.

crates/coral-engine/src/runtime/query.rs
```diff
@@ -149,6 +164,12 @@ fn provider_error_to_core(error: &ProviderQueryError) -> CoreError {
     }
 }
 
+pub(crate) fn is_non_read_only_sql_error(detail: &str) -> bool {
```
sauldhernandez: This doesn't look like it belongs in this file, the only caller is in a different file and the function doesn't use anything related to this file.

crates/coral-api/proto/coral/v1/sources.proto
```diff
@@ -157,12 +157,34 @@ message ValidateSourceRequest {
...
+// One validation query executed for a source.
+message QueryTestResult {
+  // SQL text declared by the source spec.
+  string sql = 1;
+  // Whether the query executed successfully.
+  bool passed = 2;
+  // Optional row count captured for successful queries.
+  optional uint64 row_count = 3;
+  // Optional error message captured for failed queries.
+  optional string error_message = 4;
```
sauldhernandez: This shouldn't be marked as `optional`, having the default empty string value is enough to convey there was no error

crates/coral-api/proto/coral/v1/sources.proto
```diff
@@ -157,12 +157,34 @@ message ValidateSourceRequest {
...
 }
 
+// One validation query executed for a source.
+message QueryTestResult {
+  // SQL text declared by the source spec.
+  string sql = 1;
+  // Whether the query executed successfully.
+  bool passed = 2;
+  // Optional row count captured for successful queries.
+  optional uint64 row_count = 3;
```
sauldhernandez: This might also be worth not marking as optional. Since we have `passed` to differentiate when the test passed or not, having the default value when tests fail might be enough.

sauldhernandez (changes requested):

crates/coral-api/proto/coral/v1/sources.proto
antonmry: I'm seeing some complex validations in Rust code to deal with defaults in optional fields. What's the benefit of that approach vs. marking it as optional?

crates/coral-engine/src/contracts/query.rs
antonmry: I don't fully understand the comment, declared_query_count and some of the other fields are created in this PR, why is it better to do a separate cleanup?

crates/coral-api/proto/coral/v1/sources.proto
lazybean: 0 row could be a successful result no?

crates/coral-api/proto/coral/v1/sources.proto
lazybean: Empty string is more ambiguous no?  
  
 It could mean:

- no error
- unknown error
- bug in formatting
- forgot to fill it in  
  
While it is fair to say, it should only mean 'no error', this mostly push the responsability away.

crates/coral-engine/src/lib.rs
lazybean: I can explore that.  
  
My fear is that  the implementation is mostly “run queries and interpret results”, so we would need to expose lots more engine internals upward so app can stitch them together.

crates/coral-api/proto/coral/v1/sources.proto
sauldhernandez: Yes, 0 can be a valid successful row count. My point is you don't need to rely on that to know if it was successful since you have the `passed` boolean

crates/coral-api/proto/coral/v1/sources.proto
sauldhernandez: I agree empty string should mean "no error". My point was that if you put it as optional you have more states that represent the same thing? Because you could have (with optional):

- Value missing
- Empty string
- Some value

vs with no optional which only allows

- Empty string
- Some value

Do we really need three states here? I think 2 is more than enough, specially since we got `passed`.



See https://google.aip.dev/149 for more arguments about it

crates/coral-engine/src/lib.rs
sauldhernandez: Which internals would need exposure exactly?

sauldhernandez: Actually it does make sense to keep it here. `build_runtime` is private and it's also more efficient to use the same runtime when executing multiple queries

docs/guides/write-a-custom-source.mdx
```diff
@@ -41,6 +41,8 @@ name: local_messages
 version: 0.1.0
 dsl_version: 3
 backend: jsonl
+test_queries:
+  - SELECT COUNT(*) AS n FROM local_messages.messages
```
simonwhitaker: nit: I wonder if we should avoid normalising `COUNT(*)` in `test_queries`. (Imagine using `SELECT COUNT(*) FROM github.commits` as the test query for Github, for example; it would take a long time to complete on large repos.)

It's fine for this local_messages example, but a better canonical example might be something like `SELECT * FROM schema.table LIMIT 1`.

docs/guides/write-a-custom-source.mdx
```diff
@@ -58,13 +60,17 @@ tables:
 
 Replace `/absolute/path/to/demo-data/` with the real absolute path on your machine.
 
+The optional top-level `test_queries` field lets you declare read-only SQL checks that Coral runs during `coral source test`. Coral preserves the order you write, allows duplicates, and reports each query result individually.
```
simonwhitaker: It might be worth mentioning that the intention here is to test the mapping logic and (in the case of http sources), the API connection, so you should prefer `SELECT` queries over `DESCRIBE` queries, etc.

docs/guides/write-a-custom-source.mdx
```diff
@@ -75,6 +81,29 @@ coral sql "
...
 
+## How to verify and validate
+
+When you add `test_queries` to a source spec, treat them as a lightweight validation contract for that source.
+
+1. Run `coral source lint ./my-source.yaml` before installing the source.
+2. Add or update the source with `coral source add --file ./my-source.yaml`.
+3. Run `coral source test <source_name>` and confirm:
+   - the source still exposes the expected tables
+   - each declared `test_queries` entry runs in manifest order
```
simonwhitaker: Why should a source author validate this? This expectation (that tests run in manifest order) is implemented in coral, not in their source.

docs/guides/write-a-custom-source.mdx
```diff
@@ -75,6 +81,29 @@ coral sql "
...
+   - successful queries show a pass status and optional row count
+   - failing queries surface the SQL text plus a readable error message
```
simonwhitaker: Same question with these two lines. I'd suggest that at this point, a source spec author should be running `coral source test <source_name>` and confirming only that the source exposes the expected tables, and test queries run successfully.

docs/reference/source-spec-reference.mdx
```diff
@@ -19,6 +19,8 @@ name: my_source
 version: 0.1.0
 dsl_version: 3
 backend: http
+test_queries:
+  - SELECT COUNT(*) AS n FROM my_source.messages
```
simonwhitaker: Same comment as above. For an http backend source we definitely shouldn't encourage `SELECT COUNT(*)`.

docs/reference/source-spec-reference.mdx
```diff
@@ -27,6 +29,28 @@ backend: http
...
+
+## Validation queries
+
+Use the optional top-level `test_queries` field when you want Coral to run one
+or more read-only SQL checks after the source validates and its tables become
+queryable.
+
+```yaml
+test_queries:
+  - SELECT COUNT(*) AS n FROM my_source.messages
```
simonwhitaker: Please remove. (Same reason as above.)

docs/reference/source-spec-reference.mdx
```diff
@@ -27,6 +29,28 @@ backend: http
...
+  - SELECT * FROM my_source.messages LIMIT 1
+```
+
+Notes:
+
+- Coral preserves the order exactly as you write it.
```
simonwhitaker: Just curious: why does preserving the order matter? We don't allow mutating queries, and we don't short-circuit fail on the first failure (as far as I can tell), so the ordering shouldn't be important. (We might even want to parallelise these test queries at some point.)

simonwhitaker (approved):

crates/coral-api/proto/coral/v1/sources.proto
lazybean: unoptionalized

crates/coral-api/proto/coral/v1/sources.proto
lazybean: unoptionalized

crates/coral-engine/src/contracts/query.rs
lazybean: moved to uint64. Thank you!

crates/coral-engine/src/lib.rs
lazybean: thank you!

docs/guides/write-a-custom-source.mdx
lazybean: 100% right. thank you!

docs/getting-started/quickstart.mdx
```diff
@@ -28,6 +28,8 @@ coral source add github
 
 Coral prompts for any required variables or secrets. Once connected, the source's data is available as SQL tables. To update a source's credentials later, run the same command again.
 
+After install, Coral runs a validation pass automatically. If the connection check or any declared `test_queries` fail, Coral keeps the source installed and prints warnings so you can fix the configuration and rerun `coral source test github`.
+
```
AlbertQM: Don't think this is necessary here -- I'd like to keep the quickstart as lean as possible. Ideally users won't bump into this issue, and if they do, the error they see locally will unblock them (e.g. auto-re-prompt to add the invalid key)

```suggestion
```

docs/guides/write-a-custom-source.mdx
```diff
@@ -75,6 +81,29 @@ coral sql "
 "
 ```
 
+## How to verify and validate
+
+When you add `test_queries` to a source spec, treat them as a lightweight validation contract for that source.
+
+1. Run `coral source lint ./my-source.yaml` before installing the source.
```
AlbertQM: Do we need to run the linter? (just curious)

docs/guides/write-a-custom-source.mdx
```diff
@@ -75,6 +81,29 @@ coral sql "
...
+2. Add or update the source with `coral source add --file ./my-source.yaml`.
+3. Run `coral source test <source_name>` and confirm:
```
AlbertQM: If we `coral source add --file ./my-source.yaml`, wouldn't we also run the validation? Why do we have step 3 that says to run `coral source test`?

crates/coral-api/proto/coral/v1/sources.proto
```diff
@@ -159,12 +159,34 @@ message ValidateSourceRequest {
   string name = 2;
 }
 
+// One validation query executed for a source.
+message QueryTestResult {
```
sauldhernandez, line 175: @lazybean @antonmry as an alternative approach, what do you think of doing the result message like this?

```
// One validation query executed for a source.
message QueryTestSuccess {
  // Row count captured for the successful query.
  uint64 row_count = 1;
}

// Failure details captured for one validation query.
message QueryTestFailure {
  // Error message captured for the failed query.
  string error_message = 1;
}

// One validation query executed for a source.
message QueryTestResult {
  // SQL text declared by the source spec.
  string sql = 1;
  oneof outcome {
    // Successful query execution metadata.
    QueryTestSuccess success = 2;
    // Failed query execution metadata.
    QueryTestFailure failure = 3;
  }
}
```

That way we get rid of a whole bunch of invalid representable states and clients can use pattern matching to execute their logic.

docs/guides/write-a-custom-source.mdx
```diff
@@ -131,4 +160,24 @@ For the full set of HTTP response, pagination, auth, and column fields, see [HTT
 - **Lint before installing.** Run `coral source lint ./my-source.yaml` to catch schema and semantic errors without touching your workspace.
 - **Start small.** Begin with one table and a few columns. Validate with `coral source test`, check `coral.columns`, then expand.
 - **Use `coral.tables` and `coral.columns`.** They show you exactly what Coral sees after adding a source — including required filters and column metadata.
+<<<<<<< HEAD
```
AlbertQM: Bad merge? 😄
I can't see this new paragraph in the docs preview -- will re-review once it's fixed. Thank you!

docs/reference/cli-reference.mdx
```diff
@@ -58,7 +58,7 @@ Add a custom source spec from a YAML file.
 coral source add --file ./local-messages.yaml
 ```
 
-Like `source add`, this automatically validates and prints discovered tables after install.
+Like `source add`, this automatically validates, prints discovered tables, and warns about post-install validation issues instead of rolling the source back.
```
AlbertQM: Don't think we need to mention that to users

```suggestion
Like `source add`, this automatically validates, prints discovered tables.
```

docs/reference/cli-reference.mdx
```diff
@@ -48,7 +48,7 @@ coral source add github
 ```
 
 Coral prompts for required variables or secrets interactively. If the source is already installed, running this command again replaces its stored credentials with the new values you provide.
-After installation, it automatically validates the connection and prints the discovered tables.
+After installation, it automatically validates the connection, prints the discovered tables, and runs any optional top-level `test_queries` declared in the source spec. Any post-install validation issue is reported as a warning here so the source remains installed and re-runnable.
```
AlbertQM: Few things: when I wrote the docs, I used "validates the connection", even though we didn't validate the connection. In this PR, you're now adding this feature (thank you!)

Does that make sense/is that how you interpret those words too? If so, feel free to reword to e.g. `After installation, it automatically validates the connection and prints the discovered tables.` (ie no changes from main)

Otherwise, I think we can get rid of the warming bit, that's should hopefully be self explanatory from the cli errors/warnings

```suggestion
After installation, it automatically validates the connection, prints the discovered tables, and runs any optional top-level `test_queries` declared in the source spec.
```

docs/reference/cli-reference.mdx
```diff
@@ -70,13 +70,15 @@ coral source lint ./my-source.yaml
 
 ### `coral source test <NAME>`
 
-Validate that an installed source can initialize and expose tables.
+Validate that an installed source can initialize, expose tables, and run any optional top-level `test_queries` declared in its source spec. The command exits non-zero when one or more declared query tests fail.
```
AlbertQM:
```suggestion
Validate that an installed source can initialize, expose tables, and run any optional top-level `test_queries` declared in its source spec. The command errors when one or more declared query tests fail.
```

docs/reference/cli-reference.mdx
```diff
@@ -70,13 +70,15 @@ coral source lint ./my-source.yaml
...
 
 ```shellscript
 coral source test github
 coral source test local_messages
 ```
 
+Unlike `source add`, this command is strict: it returns a non-zero exit code when any declared query test fails so you can use it in verification and CI workflows.
+
```
AlbertQM: Don't think we should add this text, until we actually use this command for those reasons (and iron out any issues)

docs/reference/cli-reference.mdx
```diff
@@ -93,6 +95,8 @@ Run the guided source setup flow.
 coral onboard
 ```
 
+The onboarding wizard uses the same warn-only post-install validation behavior as `coral source add`: Coral validates the source after install, runs any declared `test_queries`, and keeps the source installed if that validation reports warnings. Use the wizard's explicit **Validate** action or `coral source test <NAME>` when you want strict pass/fail validation.
+
```
AlbertQM: Don't think this is necessary 😄 

```suggestion
```

crates/coral-api/proto/coral/v1/sources.proto
antonmry: I like it more than not using optional, it's a good middleground solution to comply with AIP-149 :+1:

docs/guides/write-a-custom-source.mdx
```diff
@@ -75,6 +81,29 @@ coral sql "
 "
 ```
 
+## How to verify and validate
```
AlbertQM: What's the difference between verify and validate?

docs/reference/source-spec-reference.mdx
```diff
@@ -27,6 +29,28 @@ backend: http
...
+
+Use the optional top-level `test_queries` field when you want Coral to run one
+or more read-only SQL checks after the source validates and its tables become
+queryable.
+
+```yaml
+test_queries:
+  - SELECT * FROM my_source.messages LIMIT 1
+  - SELECT * FROM my_source.messages LIMIT 1
+```
```
AlbertQM, line 43: Either use two different examples, or just one? 😄

crates/coral-api/proto/coral/v1/sources.proto
lazybean: :rocket:

docs/reference/source-spec-reference.mdx
```diff
@@ -27,6 +29,28 @@ backend: http
 | `version`     | Source spec version string                         |
 | `dsl_version` | Coral source spec DSL version (currently `3`)      |
 | `backend`     | How data is fetched: `http`, `jsonl`, or `parquet` |
+| `test_queries` | Optional read-only SQL checks Coral runs during `coral source test` |
+
+## Validation queries
```
AlbertQM: I'd avoid adding 2 names to refer to 1 thing 😄 

```suggestion
## Test queries
```

sauldhernandez (approved):

docs/reference/source-spec-reference.mdx
```diff
@@ -27,6 +29,28 @@ backend: http
...
+```
+
+Notes:
+
+- Coral allows duplicate entries.
+- Each query result is reported individually during `coral source test`.
+- `coral source add` and `coral onboard` surface post-install validation issues as warnings; `coral source test` is the strict pass/fail command.
+- Prefer cheap read-only checks that validate connectivity and mapping logic. For example, `SELECT * FROM schema.table LIMIT 1` is usually a better fit than `SELECT COUNT(*)`.
+- Keep these queries read-only. Statements such as `CREATE`, `COPY`, `INSERT`,
+  `UPDATE`, `DELETE`, or `SET` are reported as failed validation queries.
```
AlbertQM: Why do we allow/want duplicate entries?

```suggestion 
<Note>
Prefer cheap read-only checks that validate connectivity and mapping logic. For example, `SELECT * FROM schema.table LIMIT 1` is usually a better fit than `SELECT COUNT(*)`.
</Note>
```

docs/guides/write-a-custom-source.mdx
lazybean: You do not need to build or have keys to run the linter. That is faster, and the agent do it without you (no key to input)

docs/guides/write-a-custom-source.mdx
lazybean: Agree. We do not need to run `coral source test`at that point.

docs/reference/cli-reference.mdx
lazybean: well , it actaully does not validate the connection if it does not run the query. So Ill try to rephrase

maybe mov the `validadet the connection' alongside the test_queries.

"and runs any optional top test_queries to validate the connection"

```suggestion
After installation, it prints the discovered tables, and validates the connection by running the optional top-level `test_queries` declared in the source spec. Any post-install validation issue is reported as a warning here so the source remains installed and re-runnable.a
```

crates/coral-api/proto/coral/v1/sources.proto
```diff
@@ -159,12 +159,47 @@ message ValidateSourceRequest {
...
+  // Validation-query results declared by the source spec.
+  repeated QueryTestResult query_tests = 3;
+  // Total number of declared validation queries.
+  uint64 declared_query_count = 4;
+  // Number of validation queries that passed.
+  uint64 passed_query_count = 5;
+  // Number of validation queries that failed.
+  uint64 failed_query_count = 6;
+  // Whether every declared validation query passed.
+  bool all_query_tests_passed = 7;
```
jsummerfield: These feel low-value compared with just returning query_tests and deriving counts at the edge. Unless I've missed a reason to keep them, please trim the public API back.

crates/coral-cli/src/onboard.rs
```diff
@@ -196,7 +198,13 @@ async fn run_add_bundled_source(
     let (variables, secrets) = source_ops::prompt_for_inputs(&inputs)?;
     let result = source_ops::add_bundled_source(app, &source.name, variables, secrets).await?;
     println!("Added source {}", result.name);
-    source_ops::validate_and_print(app, &result.name, source_ops::TableDisplayLimit::DEFAULT).await
+    source_ops::validate_and_print(
```
jsummerfield: This onboarding path still propagates unary `validate_source` failures because it does `validate_and_print(..., WarnOnly).await?` directly. That differs from `run_source_add`, which wraps the same call in `if let Err(err) = ...` and downgrades any validation error to a warning. As written, a source that installs successfully but then hits a validation RPC error (missing secret, registration failure, source never became queryable, etc.) will still abort the wizard instead of remaining installed and re-runnable.

crates/coral-engine/src/contracts/query.rs
```diff
@@ -65,6 +65,122 @@ impl QuerySource {
...
+    #[must_use]
+    /// Returns the outcome metadata for this query test.
+    pub fn outcome(&self) -> &QueryTestOutcome {
+        &self.outcome
+    }
+}
+
+/// Structured outcome for validating one source and its optional test queries.
+#[derive(Debug, Clone)]
+pub struct SourceValidationOutcome {
```
jsummerfield: Stepping back, I think the feature direction is right, but this contract shape is less idiomatic than it needs to be. `QueryTestOutcome` vs `QueryTestResult` vs `SourceValidationOutcome` makes the reader decode three near-synonyms across two levels. I would consider a simpler model like `SourceValidationReport { tables, query_tests }` plus one per-query type, and then use Rust's normal `Result<Success, Failure>` shape inside that per-query record instead of inventing both `Outcome` and `Result`. Separately, the aggregate count/boolean fields here look derivable from `query_tests`, so they feel more transport-shaped than engine-shaped.

crates/coral-app/src/sources/service.rs
```diff
@@ -180,14 +182,30 @@ impl SourceServiceApi for SourceService {
             .validate_source(&workspace_name, &source_name)
             .await
             .map_err(query_status)?;
-        let tables = result
-            .tables
+        let coral_engine::SourceValidationOutcome {
```
jsummerfield: This works, but it feels a bit unidiomatic for the service layer. The handler is now doing both orchestration and an in-place projection from the engine contract into the transport contract, including destructuring `SourceValidationOutcome` field-by-field. I think the cleaner shape here would be to keep the service method thin and move this mapping into one transport helper (alongside `table_to_proto` / `query_test_result_to_proto`), so the service just calls `queries.validate_source(...)` and then `validate_source_response_to_proto(...)`.

crates/coral-engine/src/lib.rs
```diff
@@ -129,10 +129,61 @@ impl CoralQuery {
...
```
jsummerfield: This API feels like it wants to be the primary validation operation, so I think it would read more idiomatically as `validate_source(...)` returning something like `SourceValidationResult`. The current `validate_source_with_tests` / `SourceValidationOutcome` / `QueryTestOutcome` naming stack makes the reader decode too many near-synonyms. I would also consider using Rust's normal `Result<Success, Failure>` shape inside the per-query record instead of introducing a separate `QueryTestOutcome` enum.

jsummerfield (changes requested):

crates/coral-api/proto/coral/v1/sources.proto
lazybean: Good call. I trimmed the derived count/boolean fields from ValidateSourceResponse and now return the primary validation facts plus query_tests. The CLI derives the summary counts at the edge instead of carrying redundant state in the wire contract.

simonwhitaker (approved):

lazybean:
@jsummerfield ping

jsummerfield (approved):
Thanks for pushing through, looking good now!

PR #113, opened 2026-04-16
feat(engine): add SQL-ready identifier fields to catalog

sauldhernandez:
## Intent

Agents (and humans) who query Coral through `coral.tables` / `coral.columns` currently have to reconstruct quoted identifiers themselves. When table or column names use mixed case (e.g. `Master`, `playerID`), DataFusion normalizes unquoted identifiers to lowercase and the lookup fails. The discovery surface exposes raw names that are not safe to paste verbatim.

## What it enables

- `coral.tables` gains `sql_table_ref` (e.g. `hockey.\"Master\"`).
- `coral.columns` gains `sql_column_ref` and `sql_qualified_column_ref` (e.g. `\"playerID\"` and `hockey.\"Master\".\"playerID\"`).
- The MCP guide, initial instructions, and user docs now tell agents to copy these fields verbatim instead of reconstructing quoted identifiers.

## Usage examples

\`\`\`sql
SELECT schema_name, table_name, sql_table_ref FROM coral.tables ORDER BY 1, 2;

SELECT column_name, sql_column_ref, sql_qualified_column_ref
FROM coral.columns
WHERE schema_name = 'hockey' AND table_name = 'Master'
ORDER BY ordinal_position;
\`\`\`


mystic123 (changes requested):
Nice!
Could you please wire the table-not-found hint path from #162  to use `TableInfo.sql_table_ref` for rendered suggestions?

#162 currently has its own `format_schema_table` / dotted-identifier helper so hints like `Did you mean stripe.accounts?` and `"foo.bar"."Items"` are SQL-safe. Once this PR makes `sql_table_ref` the canonical SQL-ready table reference, that error path should probably consume it instead of duplicating identifier formatting.

I think the matching should stay exactly as in #162: compare against raw `table_name`, then render the selected candidate via `sql_table_ref`. Also worth updating the fallback catalog hint to include `sql_table_ref`, e.g. `SELECT schema_name, table_name, sql_table_ref FROM coral.tables`.

antonmry (approved):

PR #114, opened 2026-04-16
feat(engine): add coral.source_variables table

simonwhitaker:
## Summary

- add `coral.source_variables`, a new system catalog table that exposes effective non-secret source variables alongside defaulting, required-ness, and manifest version metadata
- thread manifest input specs through `QuerySource` and backend registration so HTTP, JSONL, and Parquet sources all populate the new catalog metadata consistently
- cover the new catalog surface with engine tests and update MCP guide text and initial instructions so clients discover `coral.source_variables` with the existing `coral.tables` and `coral.columns` metadata tables

## Why

`coral.tables` and `coral.columns` expose table- and column-level metadata, but there was no catalog view for source-level configuration. This makes it hard to discover effective values like site domains or org slugs when constructing source-specific queries. `coral.source_variables` fills that gap without exposing secrets.

## Testing

- existing GitHub PR checks are passing

### DIFF
--- a/crates/coral-engine/src/backends/common.rs
+++ b/crates/coral-engine/src/backends/common.rs
@@ -3,10 +3,12 @@
 use std::collections::{BTreeMap, HashMap};
 use std::sync::Arc;
 
-use crate::QueryRuntimeContext;
+use crate::{QueryRuntimeContext, QuerySourceOrigin};
 use async_trait::async_trait;
 use coral_spec::backends::file::PartitionColumnSpec;
-use coral_spec::{ColumnSpec, FilterSpec, ManifestDataType, TableCommon};
+use coral_spec::{
+    ColumnSpec, FilterSpec, ManifestDataType, ManifestInputKind, ManifestInputSpec, TableCommon,
+};
 use datafusion::arrow::datatypes::{DataType, Field, Schema, SchemaRef, TimeUnit};
 use datafusion::datasource::TableProvider;
 use datafusion::error::DataFusionError;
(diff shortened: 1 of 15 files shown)


iaptsiauri:
Take a look at https://github.com/withcoral/coral/pull/97 as it restructures the variable/secret declaration and introduces inputs concept same way as you do in your PR.

crates/coral-engine/src/backends/common.rs
```diff
@@ -117,6 +131,38 @@ pub(crate) fn build_registered_table(
...
+                        variable_key: input.key.clone(),
+                        variable_value: input.default_value.clone(),
+                        is_defaulted: true,
+                        is_required: false,
+                        default_value: input.default_value.clone(),
+                    })
+                })
+        })
+        .collect::<Vec<_>>();
+    rows.sort_by(|left, right| left.variable_key.cmp(&right.variable_key));
```
antonmry, line 162: Is this sort required or is it overrided later with `build_source_variables_table` re-sorting all rows by (schema_name, variable_key)?

crates/coral-engine/tests/engine/harness.rs
```diff
@@ -37,8 +37,10 @@ pub(crate) fn build_source_with_inputs(
     variables: BTreeMap<String, String>,
     secrets: BTreeMap<String, String>,
 ) -> QuerySource {
-    let manifest = parse_source_manifest_value(value).expect("manifest should parse");
-    QuerySource::new(manifest, variables, secrets)
+    let raw = value.to_string();
+    std::mem::drop(value);
+    let (manifest, inputs) = parse_manifest_and_inputs(&raw).expect("manifest should parse");
```
antonmry, line 42: Check this comment, the drop in particular

>  This converts an already-parsed serde_json::Value back to a JSON string just to re-parse it through the YAML/JSON path. The std::mem::drop(value) is also
  non-idiomatic — it's explicit where implicit drop would do. If the goal is to get inputs from an already-parsed value, collect_source_inputs_value(&value)
   (the internal function in coral-spec) could be called directly, or parse_source_manifest_value + declared_inputs() would avoid the round-trip entirely.

simonwhitaker:
Going to close this PR and redo this once #97 has landed, otherwise we're going to be introducing two different variations on a concept of source variables.

PR #120, opened 2026-04-17
feat(engine): structured TABLE_NOT_FOUND and UNKNOWN_FIELD errors

mystic123:
## Summary

Promote two DataFusion error shapes into `StructuredQueryError` so agents and CLI users get actionable hints for the most common query-time failures.

| Input | Before | After |
|---|---|---|
| `SELECT * FROM hockey.Master` (table `Master` registered) | `InvalidArgument: table 'datafusion.hockey.master' not found` | `NotFound` / `TABLE_NOT_FOUND` + hint: `Unquoted identifiers are lowercased. Try \`hockey.\"Master\"\`.` |
| `ON g.playerID = m.playerID` on a `playerID` column | `InvalidArgument: No field named g.playerid. Use double quotes: \"g.playerid\"` (wrong case) | `InvalidArgument` / `UNKNOWN_FIELD` + hint: `Unquoted identifiers are lowercased. Try \`g.\"playerID\"\`.` |
| `SELECT * FROM nba.games` (source absent) | `InvalidArgument: table 'datafusion.nba.games' not found` | `NotFound` + hint: `Schema \`nba\` is not currently registered. Run \`coral source list\` ...` |
| `SELECT * FROM hockey.gamers` (typo) | opaque | `NotFound` + Levenshtein hint: `Did you mean \`hockey.games\`?` |

## What changed

- `coral-engine/contracts/query_error.rs` — new `unknown_field` / `table_not_found` constructors on `StructuredQueryError`, with case-aware quoting (via `datafusion::common::utils::quote_identifier`) and Levenshtein fallback via `strsim`.
- `coral-engine/runtime/query.rs` — `datafusion_to_core` threads `&[TableInfo]` through so hint builders can do case-insensitive catalog lookups; extracts `SchemaError::FieldNotFound` and `Plan(\"table '...' not found\")` into the structured path.
- Wire-stable reason-code consts (`UNKNOWN_FIELD_REASON`, `TABLE_NOT_FOUND_REASON`) — crate-private, guarded by a wire-contract test.
- Three pre-existing negative-path tests updated: a missing source is semantically `NotFound`, not `InvalidArgument` (catalog / jsonl / parquet `missing_file_returns_error`, plus `coral-app`'s `broken_source_does_not_block_healthy_sources`).

## Relation to #113 (sauldhernandez)

#113 adds `sql_table_ref` / `sql_column_ref` / `sql_qualified_column_ref` to `coral.tables` / `coral.columns`. Once it lands on main, a follow-up should switch `table_not_found_hint`'s match and Levenshtein branches from reconstructing via `quote_identifier` to reading `hit.sql_table_ref` directly. The unknown-field hint builder should likely stay on reconstruction (see open question below).

## Open questions / assumptions

1. **Alias preservation for unknown-field hints.** DataFusion's `valid_fields` arrive shaped like `[\"g.id\", \"g.playerID\"]` when the query uses an alias, or `[\"hockey.master.id\", ...]` when not. The hint preserves the query's qualifier form rather than always showing the fully-qualified catalog ref, on the theory that it's closer to the code the user wrote and matches DataFusion's own suggestion shape. Flag if you'd prefer always showing the catalog's `sql_qualified_column_ref`.
2. **DataFusion 52.5 message-format coupling.** `extract_table_not_found` matches the literal text `\"table 'X' not found\"`, and `split_table_ref` assumes the 3-part `datafusion.schema.table` form. Both are verified on datafusion 52.5 directly. A DataFusion upgrade that changes these strings needs a matching edit here.
3. **No hint when schema exists but no close match.** Today if `hockey.foo` fails and no table in `hockey` is within Levenshtein 0.5 of `foo`, the hint is omitted. Listing the schema's tables is the alternative; I picked brevity. Flag if the noisier option is preferred.

## Test plan

- [x] `cargo test --workspace --locked` — green (new: 11 unit tests in `query_error.rs`, 3 in `runtime/query.rs`, 5 integration tests in `structured_error_tests.rs`)
- [x] `cargo clippy --workspace --all-targets --locked -- -D warnings` — clean
- [x] End-to-end wire verified by integration tests: `structured_error_tests::unknown_table_in_installed_schema_suggests_case_preserved_quoted_name` and `unknown_column_on_aliased_join_suggests_case_preserved_quoted_name` assert the plan's exact `hockey."Master"` and `g."playerID"` hint shapes through the real `DataFusion` → `CoreError` → `StructuredQueryError` path.

## History

- #92 (PR2) merged: structured-error plumbing now on main.
- #100 (PR3) merged: CLI renders these hints on stderr via the now-side-effect-free `render_query_error` helper.

## Depends on

Merged: #92.

### DIFF
--- a/crates/coral-engine/src/contracts/query_error.rs
+++ b/crates/coral-engine/src/contracts/query_error.rs
@@ -2,7 +2,70 @@
 
 use std::collections::HashMap;
 
+use datafusion::common::utils::quote_identifier;
+
+use super::catalog::TableInfo;
 use super::error::StatusCode;
+
+/// Wire-stable `reason` code for unknown-column errors.
+pub(crate) const UNKNOWN_COLUMN_REASON: &str = "UNKNOWN_COLUMN";
+
+/// Wire-stable `reason` code for table-not-found errors.
+pub(crate) const TABLE_NOT_FOUND_REASON: &str = "TABLE_NOT_FOUND";
+
+/// Structure-preserving column reference used by `unknown_column`.
+///
+/// Keeping the qualifier components separate from the bare name means the
+/// hint builder can render a `"player.id"`-style identifier (literal dot
+/// inside the name) without any downstream logic having to guess whether a
+/// given `.` is a qualifier separator or part of the identifier.
+#[derive(Debug, Clone)]
+pub(crate) struct ColumnParts {
+    /// 0, 1, 2, or 3 qualifier segments (alias / schema.table /
+    /// catalog.schema.table) — each stored as a bare identifier.
+    pub relation: Vec<String>,
+    /// The bare column name, exactly as registered in the schema
+    /// (case-preserving, may contain embedded dots).
+    pub name: String,
+}
+
+impl ColumnParts {
+    /// Renders the reference as SQL, quoting each component individually.
+    fn quoted(&self) -> String {
+        let mut out = String::new();
+        for part in &self.relation {
+            out.push_str(&quote_identifier(part));
+            out.push('.');
+        }
+        out.push_str(&quote_identifier(&self.name));
+        out
+    }
+
+    /// Joins all components with dots, lowercased. Used for
+    /// structure-preserving case-insensitive equality: two references with
+    /// the same dotted form AND the same number of components match.
+    fn joined_lower(&self) -> String {
+        let mut out = String::new();
+        for part in &self.relation {
+            out.push_str(&part.to_lowercase());
+            out.push('.');
+        }
+        out.push_str(&self.name.to_lowercase());
+        out
+    }
+
+    /// Renders without quoting — for the user-facing summary line where
+    /// readability matters more than round-trip SQL safety.
+    fn flat_display(&self) -> String {
+        let mut out = String::new();
+        for part in &self.relation {
+            out.push_str(part);
+            out.push('.');
+        }
+        out.push_str(&self.name);
+        out
+    }
+}
 
 /// Structured query failure with first-class semantic fields.
 #[derive(Debug, Clone)]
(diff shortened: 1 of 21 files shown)


crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -38,6 +47,83 @@ impl StructuredQueryError {
...
+
+        let mut metadata = HashMap::new();
+        metadata.insert("field".to_string(), missing.to_string());
+
+        let detail = if valid_fields.is_empty() {
+            format!("No column matching `{missing}` is in scope.")
+        } else {
+            let preview: Vec<String> = valid_fields.iter().take(10).cloned().collect();
+            format!(
+                "No column `{missing}` is in scope. Valid fields include: {}.",
```
sauldhernandez: field/column used interchangeably. Stick just to column since we're in SQL land?

crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -38,6 +47,83 @@ impl StructuredQueryError {
...
+            false,
+            StatusCode::InvalidArgument,
+            metadata,
+        )
+    }
+
+    /// Builds a structured `TABLE_NOT_FOUND` error from a missing table
+    /// reference and the currently registered catalog.
+    ///
+    /// In `datafusion` 52.5 the plan error text always resolves to the
```
antonmry: It would be good to avoid the specific datafusion version here. There's a risk of this message being outdated. It's unlikely that future version changes this behaviour, and it does it, we should be able to detect it via tests.

crates/coral-engine/tests/engine/structured_error_tests.rs
```diff
@@ -0,0 +1,184 @@
...
+    let hint = sqe.hint().expect("hint should be present");
+    assert!(
+        hint.contains("hockey.games"),
+        "hint should suggest closest table, got: {hint}"
+    );
+}
+
+#[tokio::test]
+async fn unknown_column_on_aliased_join_suggests_case_preserved_quoted_name() {
+    // Reproduces PR5 plan test case 2: `ON g.playerID = m.playerID` on a
```
antonmry: what's PR5 plan test case 2?

crates/coral-engine/src/contracts/query_error.rs
mystic123: Renamed in 68ff74d. Wire reason is now `UNKNOWN_COLUMN`; metadata key is `column`; constructor is `unknown_column`; parameter is `valid_columns`. User-facing messages say column throughout.

crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -93,3 +179,309 @@ impl std::fmt::Display for StructuredQueryError {
...
+
+    // Fall back to Levenshtein over the bare names. Candidates arrive
+    // qualified (e.g. `hockey.master.id`), but comparing only the trailing
+    // component lets an unqualified typo like `id2` still match `id`.
+    let (best, _score) = valid_fields
+        .iter()
+        .filter_map(|candidate| {
+            let score = strsim::normalized_levenshtein(
+                bare_field_name(candidate),
+                bare_field_name(missing),
```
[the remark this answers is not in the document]
mystic123: Fixed in 68ff74d. Both sides are now lowercased before `strsim::normalized_levenshtein`, matching the `table_not_found_hint` convention.

crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -93,3 +179,309 @@ impl std::fmt::Display for StructuredQueryError {
...
+        .collect();
+
+    if tables_in_schema.is_empty() {
+        // `known_tables` only contains successfully-registered sources, so an
+        // empty schema here could mean either "not installed" or "configured
+        // but failed to register". Don't prescribe a remediation — point at
+        // `coral source list` which surfaces both cases.
+        return Some(format!(
+            "Schema `{schema}` is not currently registered. \
+             Run `coral source list` to see installed sources and any registration errors."
```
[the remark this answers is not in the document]
mystic123: Fixed in 68ff74d. Replaced `coral source list` with a SQL catalog pointer (`Query \`SELECT DISTINCT schema_name FROM coral.tables\``) so the engine hint is transport-neutral. CLI / MCP / API adapters can enrich the hint with surface-specific commands at their own layer.

crates/coral-engine/src/contracts/query_error.rs
mystic123: Dropped the `datafusion 52.5` pin in 68ff74d. The `tests/engine/structured_error_tests.rs` integration tests exercise the plan-error format end-to-end, so any upstream wording change surfaces as a concrete failure.

crates/coral-engine/tests/engine/structured_error_tests.rs
mystic123: Clarified in 68ff74d. Rewrote the comment to describe the scenario directly (agent discovers `playerID` in `coral.columns`, writes a self-join with unquoted `ON g.playerID = m.playerID`, DataFusion lowercases both sides, the hint must suggest `g."playerID"`).

crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -93,3 +236,500 @@ impl std::fmt::Display for StructuredQueryError {
...
+        .find(|candidate| candidate.name.to_lowercase() == missing_name_lower)
+    {
+        return Some(case_sensitive_hint(&exact.quoted()));
+    }
+
+    // Levenshtein over bare names for typos.
+    let (best, _score) = valid_fields
+        .iter()
+        .filter_map(|candidate| {
+            let score = strsim::normalized_levenshtein(&candidate.name, &missing.name);
```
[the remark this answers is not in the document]
mystic123: Addressed in 68ff74d — both sides are now lowercased before `strsim::normalized_levenshtein`, matching the `table_not_found_hint` convention.

sauldhernandez (approved):
There are a few more field references still in the PR, but I'll leave it to your judgement on which ones should remain or not.

mystic123:
@sauldhernandez addressed the remaining `field` references in `cfaee48`. The HTTP-provider `MissingRequiredFilter` error still named its column payload `field` and used `"field"` as the metadata key — same SQL-column concept, so I finished the rename: struct field `column: String`, metadata key `"column"`, format strings reference `{column}`. Left the `coral-spec` `key_field` / `value_field` alone since those are JSON field names in dict-mapping, not SQL columns.

antonmry (approved):

jsummerfield (approved):
Good to see the structured errors paying off.

PR #121, opened 2026-04-17
feat(engine): expose source config via coral.inputs

simonwhitaker:
## Summary
- Adds a `coral.inputs` virtual table (`schema_name`, `key`, `kind`, `value`, `default_value`, `hint`, `required`, `is_set`) so agents can introspect per-source config like Datadog's `DD_SITE` or Sentry's `SENTRY_ORG` and compose absolute URLs or account-scoped identifiers.
- Secret values are never exposed — secret rows always have `value IS NULL`, with `is_set` reporting whether the secret is configured; the canary test locks this invariant in.
- `is_set` for variables is derived from key presence (matching the app layer's required-variable validation), so an explicitly-configured empty string is correctly reported as set.
- Updates the MCP `coral://guide` template and the user-facing `docs/guides/use-coral-over-mcp.mdx`.

## Test plan
- [x] `cargo test --workspace`
- [ ] Manual: `coral sql "SELECT * FROM coral.inputs WHERE schema_name='datadog'"` shows `DD_SITE` value and `DD_API_KEY` with NULL value

### DIFF
--- a/crates/coral-engine/src/backends/common.rs
+++ b/crates/coral-engine/src/backends/common.rs
@@ -38,7 +38,9 @@ pub(crate) struct RegisteredInput {
     pub(crate) key: String,
     pub(crate) kind: ManifestInputKind,
     pub(crate) required: bool,
-    pub(crate) default_value: Option<String>,
+    /// Mirrors [`ManifestInputSpec::default_value`]: empty string means
+    /// "no default declared". The catalog layer maps empty to SQL `NULL`.
+    pub(crate) default_value: String,
     pub(crate) hint: Option<String>,
     /// Resolved value for variables. Unconditionally `None` for secrets.
     pub(crate) resolved_value: Option<String>,
@@ -132,23 +134,19 @@ pub(crate) fn build_registered_inputs(
     declared
         .iter()
         .map(|input| {
-            let default_value = if input.default_value.is_empty() {
-                None
-            } else {
-                Some(input.default_value.clone())
-            };
             let (resolved_value, is_set) = match input.kind {
                 ManifestInputKind::Variable => {
-                    let resolved = variables
-                        .get(&input.key)
-                        .cloned()
-                        .or_else(|| default_value.clone());
+                    let explicit = variables.get(&input.key).cloned();
+                    let has_default = !input.default_value.is_empty();
+                    let resolved = explicit
+                        .clone()
+                        .or_else(|| has_default.then(|| input.default_value.clone()));
                     // Variable is "set" if the user explicitly configured the
                     // key (even with an empty string — HTTP input resolution
                     // and required-variable validation both treat the key's
                     // presence as authoritative) or the manifest provides a
                     // non-empty default.
-                    let is_set = variables.contains_key(&input.key) || default_value.is_some();
+                    let is_set = explicit.is_some() || has_default;
                     (resolved, is_set)
                 }
                 ManifestInputKind::Secret => (None, secret_keys.contains(&input.key)),
@@ -161,7 +159,7 @@ pub(crate) fn build_registered_inputs(
                 key: input.key.clone(),
                 kind: input.kind,
                 required: input.required,
-                default_value,
+                default_value: input.default_value.clone(),
                 hint: input.hint.clone(),
                 resolved_value,
                 is_set,
--- a/crates/coral-engine/src/backends/common.rs
+++ b/crates/coral-engine/src/backends/common.rs
@@ -1,12 +1,14 @@
 //! Shared internal backend contracts and registry-visible metadata.
 
-use std::collections::{BTreeMap, HashMap};
+use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::sync::Arc;
 
 use crate::QueryRuntimeContext;
 use async_trait::async_trait;
 use coral_spec::backends::file::PartitionColumnSpec;
-use coral_spec::{ColumnSpec, FilterSpec, ManifestDataType, TableCommon};
+use coral_spec::{
+    ColumnSpec, FilterSpec, ManifestDataType, ManifestInputKind, ManifestInputSpec, TableCommon,
+};
 use datafusion::arrow::datatypes::{DataType, Field, Schema, SchemaRef, TimeUnit};
 use datafusion::datasource::TableProvider;
 use datafusion::error::DataFusionError;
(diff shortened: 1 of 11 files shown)


crates/coral-engine/src/backends/jsonl/mod.rs
```diff
@@ -287,6 +287,7 @@ impl CompiledBackendSource for JsonlCompiledSource {
             source: RegisteredSource {
                 schema_name: self.manifest.common.name.clone(),
                 tables: table_infos,
+                inputs: Vec::new(),
```
sauldhernandez: Wouldn't it be better to future-proof here by mimicing the behavior done in the http backend? Even if today on practical terms the inputs would always be an empty Vec?

crates/coral-engine/src/backends/parquet/mod.rs
```diff
@@ -239,6 +239,7 @@ impl CompiledBackendSource for ParquetCompiledSource {
             source: RegisteredSource {
                 schema_name: self.manifest.common.name.clone(),
                 tables: table_infos,
+                inputs: Vec::new(),
```
sauldhernandez: Wouldn't it be better to future-proof here by mimicing the behavior done in the http backend? Even if today on practical terms the inputs would always be an empty Vec?

docs/guides/use-coral-over-mcp.mdx
```diff
@@ -202,6 +202,13 @@ Once your client is connected, ask the agent to list available tables or run a s
 
 The agent should call `list_tables` or query `coral.tables` and return the schemas from your installed sources.
 
+Agents can also read per-source configuration (for example, the Datadog site or Sentry organization slug) from the virtual `coral.inputs` table to compose absolute URLs and account-scoped identifiers. Secret values are never exposed — secret rows always have `value IS NULL`, but `is_set` reports whether the secret is configured.
+
+```sql
+SELECT value FROM coral.inputs
+WHERE schema_name = 'datadog' AND kind = 'variable' AND key = 'DD_SITE';
+```
+
```
AlbertQM: Hmm not sure this should live in the coral MCP guide?
How about in http://localhost:3000/reference/source-spec-reference#source-inputs?

AlbertQM (approved):

sauldhernandez:
Codex also surfaced these gotchas with required/default values:

1. `inputs.required` is currently a no-op. The schema now advertises the field in crates/coral-spec/src/schema/source_manifest.schema.json, but the parser in crates/coral-spec/src/inputs.rs never reads it and instead derives `required` solely from whether `default` is present. I confirmed this with a throwaway `parse_source_manifest_yaml` repro: a secret declared with `required: false` still surfaced as `API_TOKEN true`, so install-time prompting and `coral.inputs.required` will be wrong for any manifest that uses the schema-supported flag.
2. `default: ""` is accepted and then silently treated as “no default” later. The schema allows any string default in crates/coral-spec/src/schema/source_manifest.schema.json, and the parser preserves `""` in crates/coral-spec/src/inputs.rs, but the app/runtime/catalog layers all collapse empty strings back to `None` via `is_empty()`in crates/coral-app/src/sources/manager.rs, crates/coral-engine/src/backends/common.rs, and crates/coral-engine/src/backends/http/client.rs. A manifest author can express an empty-string default today, but Coral won’t persist it, won’t show it in `coral.inputs.default_value`, and won’t use it during template resolution.
3. Bundled-source defaults are now frozen into config, so later bundled default changes will not reach existing installs. crates/coral-app/src/sources/manager.rs writes manifest defaults into persisted `source.variables`, and crates/coral-engine/src/backends/http/client.rs always prefers stored variables over the current manifest default. That directly conflicts with the new README claim in README.md that bundled manifest upgrades “pick up newer bundled manifests without re-adding the source.”

crates/coral-engine/src/backends/common.rs
```diff
@@ -31,10 +33,25 @@ pub(crate) struct RegisteredTable {
...
 
+#[derive(Debug, Clone)]
+pub(crate) struct RegisteredInput {
+    pub(crate) key: String,
+    pub(crate) kind: ManifestInputKind,
+    pub(crate) required: bool,
+    /// Mirrors [`ManifestInputSpec::default_value`]: empty string means
+    /// "no default declared". The catalog layer maps empty to SQL `NULL`.
+    pub(crate) default_value: String,
+    pub(crate) hint: Option<String>,
```
antonmry, line 44: This will need to change once this PR is merged https://github.com/withcoral/coral/pull/122/

antonmry (approved):

crates/coral-engine/src/backends/jsonl/mod.rs
simonwhitaker: I went ahead and had Claude implement `inputs` support for JSONL and parquet backends. But then I wondered whether that was the right approach. Maybe we're better sticking with `Vec::new()` until we actually have a need for `inputs` support here?

@iaptsiauri was there a reason why `inputs` was only implemented for http?

simonwhitaker:
> Codex also surfaced these gotchas with required/default values:
> 
> 1. `inputs.required` is currently a no-op. The schema now advertises the field in crates/coral-spec/src/schema/source_manifest.schema.json, but the parser in crates/coral-spec/src/inputs.rs never reads it and instead derives `required` solely from whether `default` is present. I confirmed this with a throwaway `parse_source_manifest_yaml` repro: a secret declared with `required: false` still surfaced as `API_TOKEN true`, so install-time prompting and `coral.inputs.required` will be wrong for any manifest that uses the schema-supported flag.
> 2. `default: ""` is accepted and then silently treated as “no default” later. The schema allows any string default in crates/coral-spec/src/schema/source_manifest.schema.json, and the parser preserves `""` in crates/coral-spec/src/inputs.rs, but the app/runtime/catalog layers all collapse empty strings back to `None` via `is_empty()`in crates/coral-app/src/sources/manager.rs, crates/coral-engine/src/backends/common.rs, and crates/coral-engine/src/backends/http/client.rs. A manifest author can express an empty-string default today, but Coral won’t persist it, won’t show it in `coral.inputs.default_value`, and won’t use it during template resolution.
> 3. Bundled-source defaults are now frozen into config, so later bundled default changes will not reach existing installs. crates/coral-app/src/sources/manager.rs writes manifest defaults into persisted `source.variables`, and crates/coral-engine/src/backends/http/client.rs always prefers stored variables over the current manifest default. That directly conflicts with the new README claim in README.md that bundled manifest upgrades “pick up newer bundled manifests without re-adding the source.”

@sauldhernandez these are mostly issues that existed before this PR, right?

sauldhernandez:
@simonwhitaker They are, but they affect the outcome of this PR since they're input for what you produce here. Ok not to tackle them here but bear in mind that the output of these new tables might be off until those are fixed

mystic123 (approved):

AlbertQM (approved):

crates/coral-engine/src/backends/jsonl/mod.rs
iaptsiauri: I skipped it on purpose as there was no need for inputs on file based backends in any of our manifests

sauldhernandez (approved):

PR #124, opened 2026-04-20
ci: adopt release-please for Coral CLI releases

simonwhitaker:
## Summary

Replaces the bespoke release machinery (`prepare_release.py` + the `generate-release-pr` skill) with [googleapis/release-please](https://github.com/googleapis/release-please). Release-please now owns:

- opening and maintaining the **release PR** (version bump + regenerated `CHANGELOG.md`)
- creating the **git tag** (`vX.Y.Z`) when the release PR merges
- creating the **GitHub release** with generated notes

`release.yml` is repurposed: it no longer detects releases itself. It now triggers on `release: published` and just builds cross-platform artifacts, uploads them to the release release-please created, and dispatches the Homebrew tap bump — exactly as before for everything downstream of the tag.

## Why this over the status quo

The existing release flow works, but it has costs that compound every release:

- **Releases require human ceremony.** Today, someone has to invoke the `generate-release-pr` skill (or hand-craft the PR) to kick off each release. With release-please, the release PR is opened automatically the first time a releasable commit lands on `main`, and is **updated in place** as subsequent commits land. When it's time to ship, the release PR already exists, already has the right version bump, already has an up-to-date changelog. You just merge it.
- **Less custom code to maintain.** This PR deletes ~460 lines of Python (the `prepare_release.py` release-metadata script, its unit + integration tests, `pyproject.toml`, and the `.python-version` pin) and a ~60-line skill, replacing them with ~80 lines of declarative config + workflow. Net –564 / +113.
- **Fewer CI moving parts.** Removing the Python release scripts also removes the `python-gha-helper-tests` job in `validate.yml` — no more Python toolchain (`setup-python`, `uv`, `ruff`) on every PR just to lint and test a helper script.
- **Version bump logic is no longer ours to own.** The skill had to encode conventional-commit parsing and the 0.x bump rules in prose for an LLM to follow. Release-please implements the same rules natively (`bump-minor-pre-major`, `bump-patch-for-minor-pre-major`) and applies them deterministically.
- **We get a CHANGELOG for free.** We've never had one. Release-please generates `CHANGELOG.md` from conventional commits and appends to it on every release, so the history of what changed version-to-version lives in the repo rather than scattered across GitHub release pages.
- **Release notes can't drift from reality.** They're derived from the commits that actually landed, not composed by hand in a PR body where it's easy to miss something.
- **Standard, maintained tooling.** Release-please is widely used (Google, Cloud SDKs, Homebrew) and actively maintained. Our bespoke script was a small surface but still one we owned.

Downstream of the tag, nothing changes: `release.yml` still builds the same 4 cross-platform tarballs with `--locked`, still generates checksums, still dispatches the Homebrew tap bump via the same GitHub App token. The replacement is scoped to the "how does a release get cut" layer.

## Flow (before → after)

**Before**
1. Human / skill opens a PR bumping `Cargo.toml` + writing release notes in the PR body.
2. PR merges → `release.yml` fires on `paths: [Cargo.toml]` and runs `prepare_release.py` to compute metadata.
3. Build, checksums, `gh release create`, dispatch Homebrew bump.

**After**
1. Release-please watches `push: main` and maintains a release PR (`chore(main): release X.Y.Z`) based on conventional commits since the last tag.
2. PR merges → release-please tags `vX.Y.Z` and creates the GitHub release.
3. `release.yml` fires on `release: published`, builds cross-platform artifacts, **uploads** them to the release release-please created (`gh release upload --clobber`), and dispatches the Homebrew bump.

## Files changed

- **New**: \`release-please-config.json\`, \`.release-please-manifest.json\`, \`.github/workflows/release-please.yml\`.
- **Modified**: \`.github/workflows/release.yml\` (trigger + job structure), \`.github/workflows/validate.yml\` (drop the python-scripts lint/test job), \`Cargo.toml\` (annotate the workspace version line).
- **Deleted**: \`.github/scripts/\` (entire dir — \`prepare_release.py\`, its tests, \`pyproject.toml\`, \`.python-version\`), \`.agents/skills/generate-release-pr/\`.

Net: +113 / -564 lines.

## The road not taken: what we learned about release-please + Rust workspaces

The final config uses \`release-type: \"simple\"\` with an \`x-release-please-version\` annotation on the version line in \`Cargo.toml\`. That looks unusual for a Rust project, and it's the result of two dead ends worth documenting so future maintainers don't retrace them.

### Dead end 1: \`release-type: \"rust\"\` + \`cargo-workspace\` plugin

The canonical setup for a Rust monorepo. The \`cargo-workspace\` plugin walks each workspace member's \`Cargo.toml\` and bumps \`[package].version\` on each one, plus propagates across \`[dependencies]\` entries.

In our workspace, every inner crate inherits via \`version.workspace = true\`:

\`\`\`toml
[package]
name = \"coral-api\"
version.workspace = true
\`\`\`

The plugin fails on this structure:

\`\`\`
ConfigurationError: cargo-workspace: package manifest at crates/coral-api/Cargo.toml has an invalid [package.version]
    at CargoWorkspace.buildAllPackages (.../plugins/cargo-workspace.js:76:23)
\`\`\`

It expects a literal string for \`[package].version\`, not an inheritance directive. To use this plugin we'd have to stop inheriting and duplicate the literal version into all 7 crates on every release — giving up the single-source-of-truth declaration at the workspace root.

### Dead end 2: \`release-type: \"rust\"\` without the plugin

Without the plugin, the \`rust\` strategy tries to update \`Cargo.toml\` as a single package manifest. Our root \`Cargo.toml\` has \`[workspace]\` / \`[workspace.package]\` but no top-level \`[package]\` block, so:

\`\`\`
Error: is not a package manifest (might be a cargo workspace)
    at CargoToml.updateContent (.../updaters/rust/cargo-toml.js:39:19)
\`\`\`

The built-in Cargo.toml updater has no path for workspace roots.

### What we landed on

\`release-type: \"simple\"\` with an \`extra-files\` entry for \`Cargo.toml\` and a comment annotation on the version line:

\`\`\`toml
[workspace.package]
edition = \"2024\"
version = \"0.1.3\" # x-release-please-version
\`\`\`

The generic updater finds the annotation, replaces the literal, and leaves the rest of the file alone. The manifest (\`.release-please-manifest.json\`) is the bookkeeping source of truth; the annotated line in \`Cargo.toml\` is what cargo actually reads.

### \`Cargo.lock\` handling

\`simple\` release-type doesn't know how to regenerate \`Cargo.lock\`. Our \`make rust-checks\` uses \`--locked\`, so a release PR with a bumped \`Cargo.toml\` but a stale \`Cargo.lock\` would fail CI.

The release-please workflow therefore has a follow-up step: when release-please creates/updates a release PR, we check out the PR branch, run \`cargo update --workspace\` to bump the workspace-member entries in the lockfile, and push the lockfile update back to the PR branch as a separate commit from \`release-please[bot]\`. The step is gated on \`steps.rp.outputs.pr\` so it's a no-op on regular pushes.

## Config choices worth calling out

- \`bump-minor-pre-major: true\` + \`bump-patch-for-minor-pre-major: true\` — matches our current 0.x semantics (breaking → minor bump, feat → patch bump). Mirrors what the \`generate-release-pr\` skill encoded.
- \`include-component-in-tag: false\` — keeps tag shape as \`vX.Y.Z\`, not \`coral-vX.Y.Z\`.
- \`release-please-config.json\` (no leading dot) — that's release-please-action's default path, so the config-file/manifest-file inputs can be omitted from the workflow. (The manifest default \`.release-please-manifest.json\` _does_ have a leading dot; inconsistent but matches upstream.)
- No \`plugins: [cargo-workspace]\` — see dead end #1.

## GitHub App token

\`release-please.yml\` uses the same \`RELEASE_APP_ID\` / \`RELEASE_APP_PRIVATE_KEY\` App credentials that \`release.yml\` already uses. This is deliberate: the default \`GITHUB_TOKEN\` cannot trigger other workflows from its events, so if release-please created the release with the default token, \`release.yml\` wouldn't fire on \`release: published\`. The App token also authenticates the \`cargo update\` push back to the release PR branch.

## Dry-run evidence

Ran \`release-please release-pr --dry-run\` against this branch. Predicted next release against the 12 non-trivial commits since \`v0.1.3\`:

- **Title**: \`chore(main): release 0.1.4\`
- **Version bump**: 0.1.3 → 0.1.4 (1 feat + 4 fixes, no breaking — lands on patch under the 0.x rules above)
- **Files updated**: \`CHANGELOG.md\` (created), \`Cargo.toml\` (annotated line bumped), \`.release-please-manifest.json\`. \`Cargo.lock\` then gets synced by the follow-up step.
- **Changelog sections**: Features (1), Bug Fixes (4). Refactor/docs/test/chore entries are filtered out by default.

## First-run expectations after merge

1. Merging this PR to \`main\` triggers \`release-please.yml\`.
2. Release-please opens \`chore(main): release 0.1.4\` with the changes above; the follow-up step commits the synced \`Cargo.lock\`.
3. Validate CI runs on the release PR. Should be green (lockfile is synced).
4. Merge the release PR → release-please tags \`v0.1.4\` and creates the GitHub release → \`release.yml\` fires, builds artifacts, uploads, dispatches Homebrew.
5. Subsequent commits to \`main\` update the open release PR in place.

## Rollback

Closing the release PR aborts the release — no tag/release is created until merge. Reverting this PR restores the prior \`prepare_release.py\` flow since nothing about Homebrew or artifact layout changes downstream of the tag.

## Test plan

- [ ] Merge to \`main\` and confirm release-please opens a release PR.
- [ ] Confirm the \`Sync Cargo.lock to new workspace version\` step commits a lockfile update to the PR branch.
- [ ] Confirm validate CI passes on the release PR (i.e. \`--locked\` checks succeed).
- [ ] Merge the release PR and confirm \`v0.1.4\` tag + GitHub release are created.
- [ ] Confirm \`release.yml\` fires on \`release: published\`, uploads 4 platform tarballs + checksums, and dispatches the Homebrew tap bump.
- [ ] Confirm Homebrew tap formula gets updated to 0.1.4.


simonwhitaker:
Both review comments addressed in 4fc1e7f86ba8bd3a61702f736031a36f03d6ac39:

- **cubic (release.yml)**: you were right. The existing `v0.1.3` release assets are `coral-{target}.tar.gz` / `checksums.sha256` with no version in the filename, and the deleted `prepare_release.py` had `artifact_prefix: \"coral\"` confirming that. Embedding the tag in the filename would have broken the Homebrew tap's fixed download URLs. Reverted to the unversioned naming throughout: `archive=`, `upload-artifact` name/path, the `sha256sum` glob, and `gh release upload`.
- **zizmor (release-please.yml)**: added `environment: release` to the release-please job, matching what `publish-version` in release.yml already uses. The `RELEASE_APP_PRIVATE_KEY` secret is now scoped to the same deployment environment in both workflows.

release-please-config.json
```diff
@@ -0,0 +1,14 @@
...
+  "include-component-in-tag": false,
+  "bump-minor-pre-major": true,
+  "bump-patch-for-minor-pre-major": true,
+  "changelog-path": "CHANGELOG.md",
+  "packages": {
+    ".": {
+      "package-name": "coral",
+      "extra-files": ["Cargo.toml"]
+    }
+  }
```
jsummerfield: Seeding the manifest version is only half of the bootstrap. On the first manifest-mode run, `release-please` also needs a `bootstrap-sha` to know where changelog collection should start; without it, the initial release PR will mine pre-`v0.1.3` history and replay already-shipped entries into the first generated `CHANGELOG.md` / release notes.

.github/workflows/release.yml
```diff
@@ -141,35 +113,30 @@ jobs:
...
         with:
           github-token: ${{ steps.github-app-token.outputs.token }}
           script: |
+            const version = process.env.RELEASE_VERSION.replace(/^v/, "");
             await github.rest.repos.createDispatchEvent({
               owner: "withcoral",
               repo: "homebrew-tap",
               event_type: "coral-release",
-              client_payload: { version: process.env.RELEASE_VERSION },
+              client_payload: { version },
```
jsummerfield, line 141: `workflow_dispatch` is described above as a rebuild/re-upload path for an existing tag, but this dispatch still runs unconditionally. Re-running the workflow for an already-published tag will fire `coral-release` again and can create duplicate Homebrew bump attempts for the same version; I’d gate this step to the `release` event (or otherwise skip it for manual rebuilds).

jsummerfield (commented):
Leaving the two remaining findings we discussed. I’m not carrying forward the `validate.yml` point.

jsummerfield (approved):

release-please-config.json
simonwhitaker: Good catch — added in commit f01434d06ed2cbcacb2f18232b5e26777b0d1ff8.

You're right that the manifest alone is insufficient as a guarantee. In my earlier dry-run, release-please did correctly bound the changelog on v0.1.3 (it discovered the release via the GitHub API: `Found release for path ., v0.1.3 ... sha: 51de7ec4...`), but that's incidental — any disruption to release discovery (renamed tag, deleted release, migration) would send it mining pre-v0.1.3 history. `bootstrap-sha: \"51de7ec4e927d627e02ada30d258921eb258d8e8\"` now pins the lower bound explicitly. Re-ran the dry-run with it in place — same 0.1.4 proposal, same changelog (Features × 1, Bug Fixes × 4).

simonwhitaker: In the absence of `bootstrap-sha`, release-please queries the GH releases API to find the SHA of the previous release. But we can add a bootstrap-sha as a cheap belt-and-braces.

release-please-config.json
```diff
@@ -0,0 +1,15 @@
...
+  "bump-minor-pre-major": true,
+  "bump-patch-for-minor-pre-major": true,
+  "changelog-path": "CHANGELOG.md",
+  "packages": {
+    ".": {
+      "package-name": "coral",
+      "bootstrap-sha": "51de7ec4e927d627e02ada30d258921eb258d8e8",
+      "extra-files": ["Cargo.toml"]
+    }
+  }
```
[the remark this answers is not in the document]
simonwhitaker: Valid — fixed in 49709bbf749db174415a58e1e7dbcf82b69b6ad5. I verified against release-please source: `manifest.ts` declares `'bootstrap-sha'?: string` on `ManifestConfig` alongside (not inside) `packages`, and the option is read via `config['bootstrap-sha']` at line 1467. My nested placement was silently ignored. Moved to the top level as suggested. Dry-run still produces the same 0.1.4 proposal.

.github/workflows/release.yml
simonwhitaker: This is fine; this runs as the final step of the process. The only time we need to run workflow dispatch is if one of the previous steps fails (e.g. the code fails to compile) -- in which case, we do want to send the coral-release event over to the homebrew-tap repo.

PR #126, opened 2026-04-20
docs: reframe intro subtitle around agent use case

AlbertQM:
## Summary
- Update the docs landing page subtitle to frame Coral as an SQL interface for agents, positioned against direct MCP/CLI/API-wrapper usage.

## Test plan
- [x] Docs site preview renders the new subtitle under the "Introduction to Coral" header.

### DIFF
--- a/docs/index.mdx
+++ b/docs/index.mdx
@@ -1,9 +1,9 @@
 ---
 title: "Introduction to Coral"
-description: "Coral is a local-first SQL interface for APIs, files, and other data sources."
+description: "Coral is a single SQL interface for APIs, files, and other data sources."
 ---
 
-Coral gives agents and developers one local-first SQL interface over APIs, files, and other data sources. It has support for a number of [popular data sources](/reference/bundled-sources) bundled in, and you can easily extend it to accommodate others by writing your own [source specs](/reference/source-spec-reference). You can run SQL from the [CLI](/reference/cli-reference) or through [MCP](/guides/use-coral-over-mcp). And everything is local; your data, credentials and usage history never leave your machine.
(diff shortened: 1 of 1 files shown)


docs/index.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 title: "Introduction to Coral"
-description: "Coral is a local-first SQL interface for APIs, files, and other data sources."
+description: "Coral is an SQL interface for agents to query across APIs and other data sources more efficiently than direct MCP or CLI tools or API wrappers."
```
simonwhitaker: [Sorry....](https://x.com/iamdevloper/status/397664295875805184)

- "an SQL" vs "a SQL" -- whichever you choose will sound weird to the other 50% who pronounce SQL differently. (I pronounce it "sequel", so "an SQL" sounds instantly weird to me.) Let's find a way to avoid that. (e.g. keep `"a local-first SQL interface...")
- I think we should keep this description short. We can give more context in the prose below. So I'd actually suggest we keep what we have for now.
- If we want the longer description, this one doesn't quite scan. "for agents to query" isn't quite right, and "X or Y or Z" sounds clunky. So maybe something like this instead: "Coral gives agents a single SQL interface for querying APIs and other data sources. It's more efficient than using individual MCP servers, CLI tools or other API wrappers." (But then I want to link "more efficient" to our evidence to back up this claim, and this already feels like this description line is doing too much work.)

AlbertQM: This is great feedback, I forgot to mention in the PR summary that this was Matt asking me to update it to exactly those words 🙈

simonwhitaker (approved):

PR #128, opened 2026-04-20
feat: typed auth system with AWS SigV4 signer

iaptsiauri:
## Summary

Rework HTTP-backed source authentication from a single freeform headers into new AuthSpec shape:
  - BasicAuthSpec — username + password templates, base64-encoded at runtime  
  - HeaderAuthSpec — freeform list of headers for sources that don't fit a single-token shape (e.g. Datadog's dual API keys)
  - CustomAuth — dispatches to a compiled-in signer via authenticator: tag; first impl is aws_sigv4

New top-level request_headers: — generic per-source HTTP headers not conveyed by auth (Accept, X-API-Version, etc.). Applied to every outbound request.

Runtime User-Agent — set once on the reqwest client in HttpSourceClient::from_manifest.

## Test plan

  | Check | Status |
  |-------|--------|
  | `cargo test --workspace` | PASS|
  | `make rust-checks` | PASS |
  | Manual End-to-end verification against GitHub + OpenObserve during development | PASS |
  | Manual check: SigV4 against a live AWS endpoint | PASS |

Breaking changes
- OpenObserve source config: users must re-enter credentials as `OPENOBSERVE_USERNAME` + `OPENOBSERVE_PASSWORD` (plaintext) instead of pre-encoded `OPENOBSERVE_BASIC_AUTH`.

### DIFF
--- a/crates/coral-engine/Cargo.toml
+++ b/crates/coral-engine/Cargo.toml
@@ -12,6 +12,8 @@ workspace = true
 [dependencies]
 async-trait.workspace = true
 arrow.workspace = true
+aws-credential-types.workspace = true
+aws-sigv4.workspace = true
 base64.workspace = true
 bytes.workspace = true
 datafusion.workspace = true
--- a/crates/coral-engine/src/backends/http/client.rs
+++ b/crates/coral-engine/src/backends/http/client.rs
@@ -9,7 +9,7 @@ use reqwest::header::HeaderMap;
 use serde_json::{Map, Value, json};
 
 use crate::backends::http::ProviderQueryError;
-use crate::backends::http::auth::{AuthContext, Authenticator, resolve_auth_headers};
+use crate::backends::http::auth::{Authenticator, resolve_auth_headers};
 use crate::backends::http::rate_limit::{RateLimitDecision, check_rate_limit};
 use crate::backends::shared::json_path::get_path_value;
 use crate::backends::shared::template::{
@@ -352,22 +352,15 @@ fn validate_source_scoped_http_config(
         })?;
     }
 
-    // Auth is source-scoped — `Authenticator::auth` always runs against
-    // empty filters/state at runtime too, so signing a synthetic request
-    // here exercises the same template-resolution contract as real fetches.
-    let validation_url = reqwest::Url::parse("http://preflight.invalid/")
-        .expect("static preflight URL must parse");
-    let request = reqwest::Request::new(reqwest::Method::GET, validation_url);
-    let ctx = AuthContext {
-        request: &request,
-        resolved_inputs,
-    };
-    let _ = manifest.auth.auth(&ctx).map_err(|error| {
-        DataFusionError::Execution(format!(
-            "{} source auth could not be resolved: {error}",
-            manifest.common.name
-        ))
-    })?;
+    manifest
+        .auth
+        .validate_inputs(resolved_inputs)
+        .map_err(|error| {
+            DataFusionError::Execution(format!(
+                "{} source auth could not be resolved: {error}",
+                manifest.common.name
+            ))
+        })?;
     Ok(())
 }
 
--- a/crates/coral-engine/src/backends/http/client.rs
+++ b/crates/coral-engine/src/backends/http/client.rs
@@ -15,7 +15,7 @@ use crate::backends::shared::json_path::get_path_value;
 use crate::backends::shared::template::{render_template, resolve_value_source, value_to_string};
 use coral_spec::backends::http::{HttpSourceManifest, HttpTableSpec, RateLimitSpec};
 use coral_spec::{
-    AuthSpec, HeaderSpec, HttpMethod, ManifestInputKind, PageSizeSpec, ParsedTemplate, RowStrategy,
+    AuthSpec, HeaderSpec, HttpMethod, PageSizeSpec, ParsedTemplate, RowStrategy,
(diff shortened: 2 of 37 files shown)


crates/coral-engine/src/backends/http/client.rs
```diff
@@ -86,29 +86,8 @@ impl HttpSourceClient {
...
-                &resolved_inputs,
-            )?;
-            if resolved.is_none() {
-                return Err(DataFusionError::Execution(format!(
-                    "{} source request header '{}' could not be resolved",
-                    manifest.common.name, header.name
-                )));
-            }
-        }
+        let resolved_inputs =
```
sauldhernandez, line 92: Removing the eager auth/request-header resolution here changes `CoralQuery::test_source` semantics in a user-visible way: a source with a required auth input but no secret now registers successfully, so `coral source test` passes and the failure is deferred until the first query (`missing source input ...`). That contradicts both the engine contract (`https://github.com/withcoral/coral/blob/801db3088665ccab112213379574151d91adefd0/crates/coral-engine/src/lib.rs#L122-L127`) and the CLI docs (`https://github.com/withcoral/coral/blob/801db3088665ccab112213379574151d91adefd0/docs/reference/cli-reference.mdx#L71-L77`), which both describe `test_source` as validating initialization/queryability. Can we keep a preflight validation step for source-scoped auth and `request_headers` so misconfigured HTTP sources still fail during registration?

docs/reference/source-spec-reference.mdx
```diff
@@ -208,6 +207,91 @@ In this minimal example:
...
+```yaml
+auth:
+  type: CustomHeadersAuth
+  headers:
+    - name: DD-API-KEY
+      from: input
+      key: DD_API_KEY
+    - name: DD-APPLICATION-KEY
+      from: input
+      key: DD_APPLICATION_KEY
```
simonwhitaker, line 240: I think `CustomHeadersAuth` and `ApiKeyAuth` should be the same shape, and quite possibly should be the same type. `ApiKeyAuth` is mostly the special case of `CustomHeadersAuth` where there is only one header. The fact that one has "header" and "api_token" keys, and the other has "name" and "key" keys (from a "from" key) is likely to confuse people. It looks like we should standardise on the `ValueSourceSpec` shape for both.

simonwhitaker (approved):

crates/coral-engine/src/backends/http/client.rs
iaptsiauri: Restored pre flight validation

crates/coral-engine/Cargo.toml
```diff
@@ -12,6 +12,8 @@ workspace = true
 [dependencies]
 async-trait.workspace = true
 arrow.workspace = true
+aws-credential-types.workspace = true
+aws-sigv4.workspace = true
```
antonmry, line 16: I've been thinking about a potential way to avoid this dependency for the `coral-engine`

This is a potential plan based on #95 :

## Context

PR #128 adds typed auth variants (Basic, ApiKey, CustomHeaders, CustomAuth/aws_sigv4)
directly inside `coral-engine`. The SigV4 implementation pulls AWS SDK dependencies
(`aws-sigv4`, `aws-credential-types`, `aws-smithy-*`) into the engine crate.

PR #95 establishes a pattern where `coral-engine` exposes generic traits
(`SourceDecorator`) and the app layer provides concrete implementations. The engine
stays generic; `coral-app` is the composition root.

## Proposal

Apply the same trait-based pattern from #95 to custom auth in #128.

### What stays in the engine

Basic, ApiKey, and CustomHeaders are simple header manipulation with no external
dependencies. They stay in `coral-engine`.

### What moves out

The `CustomAuth` dispatch and the SigV4 signer move behind a trait:

```rust
// coral-engine: generic contract
pub trait RequestAuthenticator: Send + Sync {
    fn name(&self) -> &str;
    fn authenticate(
        &self,
        request: reqwest::RequestBuilder,
        resolved_inputs: &BTreeMap<String, String>,
    ) -> Result<reqwest::RequestBuilder>;
}
```

When the engine encounters `authenticator: aws_sigv4` in a manifest, it looks up a
`RequestAuthenticator` by name from a registry provided by the app layer. The SigV4
implementation lives in a separate crate (e.g. `coral-auth-aws`) or directly in
`coral-app`.

### Wiring

PR #95's `EngineExtensions` struct is the natural place to carry the authenticator
registry alongside source decorators:

```rust
#[derive(Default)]
pub struct EngineExtensions {
    pub source_decorators: Vec<Box<dyn SourceDecorator>>,
    pub authenticators: Vec<Box<dyn RequestAuthenticator>>,
}
```

The app layer registers authenticators during bootstrap, same as it does for
decorators.

## Benefits

- **Engine stays dependency-light.** No AWS SDK (or future OAuth2, GCP, etc.)
  dependencies in `coral-engine`.
- **Open for extension.** New auth mechanisms are added by implementing a trait and
  registering it at the app layer. No engine changes needed.
- **Consistent architecture.** Both PRs follow the same pattern: engine defines
  contracts, app provides implementations.

## Trade-offs

- Slightly more indirection for auth dispatch vs the current compiled-in match.
- The two PRs need to coordinate on the `EngineExtensions` shape.

docs/reference/source-spec-reference.mdx
iaptsiauri: Done

PR #129, opened 2026-04-20
ci(build): prune stale Rust caches after build timings

antonmry:
## Summary
- change the build timings workflow to emit a warning instead of failing when GitHub Actions cache usage reaches 80% or more
- keep cache pressure visible in CI while avoiding hard failures from the usage threshold alone

## Test plan
- [ ] Run the `build-timings` workflow and confirm cache usage is still reported in the logs
- [ ] Verify the workflow emits a `::warning::` annotation, not an error, when cache usage is 80% or higher
- [ ] Verify stale and least-recently-used Rust cache cleanup steps run as expected in CI

### DIFF
--- a/.github/workflows/build-timings.yml
+++ b/.github/workflows/build-timings.yml
@@ -63,45 +63,6 @@ jobs:
           path: build-output.txt
           retention-days: 30
 
-      - name: Evict least-recently-used caches when quota is high
-        if: always()
-        env:
-          GH_TOKEN: ${{ github.token }}
-        run: |
-          set -euo pipefail
-          limit=$((10 * 1024 * 1024 * 1024))  # 10 GiB
-          target=$((limit * 70 / 100))         # evict down to 70%
-
-          usage=$(gh api "repos/$GITHUB_REPOSITORY/actions/cache/usage" \
-            --jq '.active_caches_size_in_bytes')
-          pct=$((usage * 100 / limit))
-          echo "Cache usage before eviction: $((usage / 1024 / 1024)) MiB (${pct}%)"
-
-          if [ "$usage" -le "$target" ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/build-timings.yml
```diff
@@ -63,6 +63,28 @@ jobs:
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          for prefix in coral-build coral-validate coral-source-lint; do
+            echo "--- Pruning $prefix ---"
+            stale=$(gh cache list --repo "$GITHUB_REPOSITORY" \
+              --key "v0-rust-${prefix}-" \
+              --sort created_at --order desc \
+              --json key --jq '.[1:][].key')
```
simonwhitaker: Do we need to prune this aggressively? This means that on the day after someone has merged a change to Cargo.lock, any PRs whose merge base doesn't have the new Cargo.lock will lose their caches, so actions running against any new commits on those branches will miss the cache. And because we only save to the cache when the workflows run on main, those caches will never get repopulated. Someone who's actively working on a branch that pre-dates the latest cache will end up spending a lot of unnecessary GHA minutes.

If there's a reason why we need to manage the cache (I don't think the 80% usage warning in this workflow is, on its own, a sufficient reason) then we should find a way to clean out just the least recently used cache entries, rather than everything except the one that was written last.

antonmry: We had issues for operating at 100% of cache capacity in the past, mostly because of a combination of docker cache usage and many active branches. In any case, I agree we can wait for the problem to happen, I've changed the PR to simply move the full cache error to a warning.

simonwhitaker (approved):

PR #137, opened 2026-04-21
fix(sources/jira): align Jira ID column types

jsummerfield:
## Summary
Jira source tables should expose related identifier columns with consistent SQL types so users can join them without extra casts.

This change normalizes the bundled Jira manifest so project and category identifiers are all surfaced as `Utf8`, matching the API payload shape used elsewhere in the source.

## Verification
- `target/debug/coral source lint sources/jira/manifest.yaml`

### DIFF
--- a/sources/jira/manifest.yaml
+++ b/sources/jira/manifest.yaml
@@ -55,7 +55,7 @@ tables:
         query_param: maxResults
     columns:
       - name: id
-        type: Int64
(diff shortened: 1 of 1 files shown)


jsummerfield:
@lazybean @iaptsiauri worth thinking about how to catch this in validation

antonmry (approved):

PR #150, opened 2026-04-22
feat: confluence

AlbertQM:
## Summary

Add a bundled Confluence Cloud source covering spaces, pages, blog posts, comments (page + blog post, footer + inline), attachments, and labels via the Confluence REST v2 API. Auth mirrors Jira: Basic auth via CONFLUENCE_BASIC_AUTH (Base64 of email:token) with a confluence-auth.sh helper. Uses link_header pagination.

## Test plan

  Built and exercised the new bundled confluence source via local CLI (cargo run -p coral-cli -- ...) against a live Confluence Cloud tenant.

  ### 1) Source registration and schema surfacing

  SELECT table_name
  FROM coral.tables
  WHERE schema_name = 'confluence'
  ORDER BY table_name;

  Expected:

  - Returns all 10 tables:
      - blog_post_footer_comments
      - blog_post_inline_comments
      - blog_posts
      - labels
      - page
      - page_attachments
      - page_footer_comments
      - page_inline_comments
      - pages
      - spaces

  ### 2) Core list endpoints

  SELECT id, key, name, type, status
  FROM confluence.spaces
  LIMIT 5;

  Expected:

  - Returns space rows with non-null id/key/name.
  - type/status populated when present.

  SELECT id, title, space_id, status
  FROM confluence.pages
  LIMIT 5;

  Expected:

  - Returns page rows with non-null id/title.
  - space_id mapped from payload spaceId.
  - status populated.

  SELECT id, name, prefix
  FROM confluence.labels
  LIMIT 10;

  Expected:

  - Returns labels with non-null id/name.
  - prefix populated from payload (real column, not virtual-only).

  SELECT COUNT(*) AS spaces_count
  FROM confluence.spaces;

  Expected:

  - Returns a positive count.

  ### 3) Pagination / higher limit behavior

  SELECT id, title
  FROM confluence.pages
  LIMIT 200;

  Expected:

  - Query succeeds with link_header pagination enabled.
  - Returns up to available rows without pagination errors.

  ### 4) Single-row endpoint + body format extraction

  SELECT id, title, status, body_storage
  FROM confluence.page
  WHERE id = '<PAGE_ID>'
    AND body_format = 'storage';

  Expected:

  - Returns one row for an existing page.
  - body_storage populated with storage/XHTML content.
  - Confirms body_format request shaping works.

  ### 5) Page comments endpoints

  SELECT id, title, page_id, parent_comment_id
  FROM confluence.page_footer_comments
  WHERE page_id = '<PAGE_ID>'
  LIMIT 20;

  Expected:

  - Succeeds; returns 0..N rows depending on seeded footer comments.

  SELECT id, title, page_id, status, body_storage
  FROM confluence.page_footer_comments
  WHERE page_id = '<PAGE_ID>'
    AND body_format = 'storage'
  LIMIT 20;

  Expected:

  - For pages with footer comments, returns rows with body_storage populated.

  SELECT id, page_id, inline_marker_ref
  FROM confluence.page_inline_comments
  WHERE page_id = '<PAGE_ID>'
  LIMIT 20;

  Expected:

  - Succeeds; returns 0..N rows depending on seeded inline comments.

  ### 6) Blog post endpoints

  SELECT id, title, space_id, status
  FROM confluence.blog_posts
  LIMIT 20;

  Expected:

  - Returns published/current blog posts (may be empty if only drafts exist).

  SELECT id, title, status, space_id
  FROM confluence.blog_posts
  WHERE status = 'current'
  LIMIT 20;

  Expected:

  - Succeeds; returns 0..N current posts.

  ### 7) Blog post comments endpoints

  SELECT id, title, blog_post_id, status
  FROM confluence.blog_post_footer_comments
  WHERE blog_post_id = '<BLOG_POST_ID>'
  LIMIT 20;

  Expected:

  - Succeeds; returns 0..N rows depending on seeded footer comments.

  SELECT id, title, blog_post_id, status, inline_marker_ref
  FROM confluence.blog_post_inline_comments
  WHERE blog_post_id = '<BLOG_POST_ID>'
  LIMIT 20;

  Expected:

  - For blog posts with inline comments, returns rows with inline_marker_ref populated.


sources/confluence/confluence-auth.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/zsh
+
+set -euo pipefail
+
+usage() {
+  cat <<'EOF' >&2
+Usage:
+  ./sources/confluence/confluence-auth.sh <email> [api_token]
```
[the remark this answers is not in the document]
AlbertQM: docs for sources are now autogenerated

antonmry (approved):

sources/confluence/manifest.yaml
```diff
@@ -0,0 +1,1781 @@
...
+name: confluence
+version: 0.1.0
+description: |
+  Query spaces, pages, blog posts, comments, attachments, labels, and
+  related metadata from Confluence Cloud.
+backend: http
+inputs:
+  CONFLUENCE_BASE_URL:
+    kind: variable
+    default: https://your-domain.atlassian.net/wiki
```
simonwhitaker: Should we be providing a known-broken default? Is this better than no default at all?

sources/confluence/manifest.yaml
```diff
@@ -0,0 +1,1781 @@
...
+    hint: |
+      Base URL of your Confluence Cloud site (for example
+      `https://acme.atlassian.net/wiki`).
```
simonwhitaker: It feels weird making `wiki` part of the base URL here. I think I'd expect base URL to be `https://mycompany.atlassian.net`, and have the path be fully represented in the `path` arg for the various tables.

Put a different way: the path to the api is `/wiki/api/v2`, and we've split that path with the first part here and the other parts in the `path` args for tables. I think we should not do that.

sources/confluence/confluence-auth.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/zsh
```
simonwhitaker: This is cool, but I wonder if we need it. It's another thing to maintain and keep in sync, and I wonder how many users will find this any easier than executing the hint you've given in the manifest, i.e. `echo foo@bar:secret | base64`.

simonwhitaker: If we're keeping this, you generally want to use bash or just `/bin/sh` for shell scripts you share with others. Not every system has `zsh` these days. My usual advice is to go with `#!/bin/sh` (most portable), and use [`shellcheck`](https://www.shellcheck.net/) to help you spot errors and bashisms.

simonwhitaker (changes requested):

sources/confluence/confluence-auth.sh
AlbertQM: Happy to remove it, I was following the JIRA script that is already on main. Will ping @lazybean and see if it's OK to remove that one too (I updated both docs with the oneliner to get the token, which I've tested and works)

sources/confluence/manifest.yaml
AlbertQM: I'll clean this up, thanks!

sources/confluence/confluence-auth.sh
AlbertQM: Fixed JIRA too 😄  thank you!

simonwhitaker (approved):

PR #159, opened 2026-04-22
fix(spec): recognize inputs block for parquet and jsonl manifests

jamesaud:
## Summary

Parquet- and JSONL-backed source manifests silently ignored any declared `inputs:` block. `ValidatedSourceManifest::required_secret_names()` and `declared_inputs()` were hardcoded to return empty for those backends, so secrets written to `<config>/workspaces/<ws>/sources/<name>/secrets.env` were filtered out of the resolved set in `load_query_source` before ever reaching the backend. For parquet S3 sources, `build_object_store` then failed to find credentials like `aws_access_key_id`, and source registration failed with only a `tracing::warn!` line — leaving the schema missing from the DataFusion catalog.

This PR:

- Teaches `ParquetSourceManifest` and `JsonlSourceManifest` to parse the `inputs:` block via the shared `collect_source_inputs_value` helper (same path HTTP already uses).
- Adds `declared_inputs` + `required_secret_names()` on both file-backed manifests.
- Dispatches `ValidatedSourceManifest::required_secret_names()` / `declared_inputs()` to those methods instead of returning empty.

The fix is backend-agnostic: any secret name declared in `inputs:` now flows into `QuerySource.secrets` → `source_secrets: BTreeMap<String, String>` verbatim. The parquet backend's S3 code was already reading AWS-named keys from that map; it's unchanged.

## Files changed

- `crates/coral-spec/src/backends/file.rs` — parse `inputs:`, expose `declared_inputs` + `required_secret_names()` on both parquet and jsonl manifests; 3 new unit tests.
- `crates/coral-spec/src/parser.rs` — route `required_secret_names()` and `declared_inputs()` through to the per-backend implementations.
- `crates/coral-engine/tests/engine/parquet_tests.rs` — new regression test `parquet_manifest_with_declared_secret_inputs_registers_and_queries`: builds a parquet source that declares two secret inputs, populates the secret map, asserts the schema appears in `information_schema.schemata`, and that rows come back from the registered table.

## Test plan

- [x] `cargo test -p coral-spec` — 44 passed
- [x] `cargo test -p coral-engine` — 46 passed (including the new regression test)
- [x] `cargo test -p coral-app` — 37 passed (no regressions in the query/load path)
- [x] `cargo clippy -p coral-spec -p coral-engine --tests` — clean
- [x] Manual sanity check: author a parquet manifest with `inputs:` declaring secrets, install it, populate `secrets.env`, and confirm the schema is visible via `SELECT schema_name FROM information_schema.schemata`.

### DIFF
--- a/crates/coral-spec/src/parser.rs
+++ b/crates/coral-spec/src/parser.rs
@@ -87,14 +87,12 @@ impl ValidatedSourceManifest {
 
     /// Returns the set of source secrets required to compile or authenticate
     /// the source spec.
-    ///
-    /// File-backed source specs do not currently declare secret inputs at the
-    /// source level, so they return an empty set here.
     #[must_use]
     pub fn required_secret_names(&self) -> BTreeSet<String> {
         match &self.inner {
             ValidatedManifestKind::Http(manifest) => manifest.required_secret_names(),
-            ValidatedManifestKind::Parquet(_) | ValidatedManifestKind::Jsonl(_) => BTreeSet::new(),
+            ValidatedManifestKind::Parquet(manifest) => manifest.required_secret_names(),
+            ValidatedManifestKind::Jsonl(manifest) => manifest.required_secret_names(),
         }
     }
 
@@ -103,7 +101,8 @@ impl ValidatedSourceManifest {
     pub fn declared_inputs(&self) -> &[ManifestInputSpec] {
         match &self.inner {
             ValidatedManifestKind::Http(manifest) => &manifest.declared_inputs,
-            ValidatedManifestKind::Parquet(_) | ValidatedManifestKind::Jsonl(_) => &[],
+            ValidatedManifestKind::Parquet(manifest) => &manifest.declared_inputs,
+            ValidatedManifestKind::Jsonl(manifest) => &manifest.declared_inputs,
         }
     }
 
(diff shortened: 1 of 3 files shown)


crates/coral-spec/src/parser.rs
```diff
@@ -87,14 +87,12 @@ impl ValidatedSourceManifest {
...
-    ///
-    /// File-backed source specs do not currently declare secret inputs at the
-    /// source level, so they return an empty set here.
     #[must_use]
     pub fn required_secret_names(&self) -> BTreeSet<String> {
         match &self.inner {
             ValidatedManifestKind::Http(manifest) => manifest.required_secret_names(),
-            ValidatedManifestKind::Parquet(_) | ValidatedManifestKind::Jsonl(_) => BTreeSet::new(),
+            ValidatedManifestKind::Parquet(manifest) => manifest.required_secret_names(),
+            ValidatedManifestKind::Jsonl(manifest) => manifest.required_secret_names(),
```
antonmry, line 95: Now that all three backends implement declared_inputs and required_secret_names() identically, could we lift declared_inputs: Vec<ManifestInputSpec> into SourceManifestCommon and put required_secret_names() there too? That would remove these matches entirely and avoid needing to update them for future  backends.

antonmry (approved):

crates/coral-spec/src/parser.rs
jamesaud: sounds good, I'll merge this in first to unblock other work immediately, and will make a follow up PR addressing this

PR #160, opened 2026-04-22
feat(engine): add JSON manifest type and query functions

sauldhernandez:
## Intent

Agents using PostHog (and other JSON-heavy sources) had no way to know that columns like `events.properties` are queryable as JSON, and the engine had no JSON SQL functions registered. This adds first-class support for JSON columns end-to-end.

## What it enables

- New `Json` manifest column type — stored as `Utf8` containing valid JSON, but signals to users, tooling, and `coral.columns` that the column is JSON-queryable.
- The engine registers the JSON query functions on its `SessionContext`, so queries can use `json_get`, `json_get_str`, `json_get_int`, `json_get_float`, `json_get_bool`, `json_get_array`, `json_get_json`, `json_contains`, `json_length`, `json_as_text` against any `Utf8` or `Json` column.
- JSON access is function-only. `->`, `->>`, and `?` are intentionally unsupported.
- PostHog's `events.properties` and `persons.properties` are now typed `Json` with descriptions pointing at the new functions.
- Docs updated (`source-spec-reference.mdx`) with the type and a querying example; new `engine/json_tests.rs` covers the JSON UDFs against both `Json` and plain `Utf8` columns.

## Usage examples

```sql
SELECT id, json_get_str(properties, '$browser') AS browser
FROM posthog.events
WHERE json_get_int(properties, 'count') > 5;

SELECT id, json_get_str(properties, 'country') AS country
FROM posthog.persons;
```


antonmry (commented):
nice one! Is it worth to update the Skill? I'm not sure if agents will use it  https://github.com/withcoral/skills/blob/main/coral/SKILL.md

antonmry (approved):

docs/reference/source-spec-reference.mdx
```diff
@@ -54,19 +54,32 @@ Notes:
...
-| `Boolean`   | Boolean true/false                       |
-| `Timestamp` | UTC timestamp (microsecond precision)    |
+| Type        | Description                                                                                        |
+| ----------- | -------------------------------------------------------------------------------------------------- |
+| `Utf8`      | UTF-8 string                                                                                       |
+| `Int64`     | 64-bit signed integer                                                                              |
+| `Float64`   | 64-bit floating point                                                                              |
+| `Boolean`   | Boolean true/false                                                                                 |
+| `Timestamp` | UTC timestamp (microsecond precision)                                                              |
+| `Json`      | UTF-8 string containing JSON. Hint that the column is queryable with `json_get`, `->`, `->>`, etc. |
```
simonwhitaker: [nit] I might just add a reference to the "Querying JSON columns" section here, for the benefit of those who haven't scrolled that far yet. Something like...

```suggestion
| `Json`      | UTF-8 string containing JSON. Hint that the column is queryable with JSON accessors. See "Querying JSON columns" below. |
```

docs/reference/source-spec-reference.mdx
```diff
@@ -54,19 +54,32 @@ Notes:
...
+
+`Json`-typed columns (and any `Utf8` column whose values happen to be JSON) can be queried with the built-in JSON functions:
+
+```sql
+SELECT id, properties->>'$browser' AS browser
+FROM posthog.events
+WHERE json_get_int(properties, 'count') > 5
+```
+
+Available: `json_get`, `json_get_str`, `json_get_int`, `json_get_float`, `json_get_bool`, `json_get_array`, `json_get_json`, `json_contains`, `json_length`, `json_as_text`, plus the `->`, `->>`, and `?` operators.
```
simonwhitaker: Can we link here to the docs for these functions? (e.g. [this](https://github.com/datafusion-contrib/datafusion-functions-json/blob/main/README.md).) Some of them are self-descriptive, but it's very unclear what `->` and `->>` do, for example.

simonwhitaker (approved):

sources/posthog/manifest.yaml
```diff
@@ -9011,9 +9011,11 @@ tables:
           kind: from_filter
           key: person_id
       - name: properties
-        type: Utf8
+        type: Json
         nullable: true
         virtual: false
+        description: Event properties as a JSON object. Query with `properties->>'$.key'`
```
jsummerfield: The `->>` examples here look off. The registered operators take keys/path segments, not a JSONPath string, so `properties->>$.key` will look up the literal key `$.key`. This should use the actual field name form instead, otherwise `coral.columns` will point users at a query that appears supported but does not hit the intended property.

crates/coral-engine/src/backends/shared/mapping.rs
```diff
@@ -638,4 +638,34 @@ mod tests {
 
         assert_eq!(rendered, Some(Value::String("untitled".to_string())));
     }
+
+    #[test]
+    fn json_type_serializes_objects_as_valid_json_strings() {
```
jsummerfield: This test only proves object-valued JSON round-trips. `Json` still lowers to the generic `Utf8` mapper path, so a scalar string like `{ "properties": "hello" }` would be stored as `hello` instead of valid JSON `"hello"`. Can we add a scalar-string case here and fix the mapper so `Json` preserves JSON encoding for all JSON values?

crates/coral-engine/src/runtime/query.rs
```diff
@@ -22,17 +22,18 @@ pub(crate) async fn build_runtime(
     sources: &[QuerySource],
     runtime: &dyn QueryRuntimeProvider,
 ) -> Result<QueryRuntimeAdapter, CoreError> {
-    let session_config = SessionConfig::new().with_information_schema(true);
+    let session_config = SessionConfig::new()
+        .with_information_schema(true)
+        .set_str("datafusion.sql_parser.dialect", "postgres");
```
jsummerfield: This is a broad engine-wide parser choice made in service of one feature. Let's chat quickly f2f.

sauldhernandez: This wasn't supposed to make it to the PR. Reverted it.

Cargo.toml
```diff
@@ -31,6 +31,7 @@ bytes = "1"
 chrono = { version = "0.4", default-features = false, features = ["clock", "serde", "std"] }
 clap = { version = "4.5", features = ["derive", "env"] }
 datafusion = "52"
+datafusion-functions-json = "0.52"
```
jsummerfield: Bump to .53

jsummerfield (approved):

Cargo.toml
sauldhernandez: Doing it right now. Will need to drop operator support. DF 53 has lambdas which clash with `->` so I'll just drop them and just rely on functions.

AlbertQM (approved):

mystic123 (approved):

PR #162, opened 2026-04-22
feat(engine): suggest schema-qualified name for unqualified table misses

mystic123:
## Summary

Follow-up to #120. PR #120 promoted unqualified `FROM account` into a structured `TABLE_NOT_FOUND` (via `parse_table_ref` detecting DataFusion's synthetic `public` schema), but the hint for the unqualified case was still the generic catalog pointer:

> List available tables with `SELECT schema_name, table_name FROM coral.tables`.

This PR extends `table_not_found_hint` so the unqualified branch scans every known `TableInfo` across every schema by normalized Levenshtein (same `>= 0.5` threshold already used for the qualified branch) and, when a close match exists, returns `Did you mean \`stripe.accounts\`?`. So:

- `FROM account` against a catalog containing `stripe.accounts` now suggests `stripe.accounts`.
- `FROM account` against a catalog with no close match still falls back to the generic `coral.tables` pointer.

Metadata on the structured error stays unqualified (`schema` key absent, `table` reflects what the user wrote) — the wire contract for consumers keying off parsed user intent is unchanged; only the human-readable `hint` gets smarter.

## Test plan

- [x] Unit: `table_not_found_unqualified_cross_schema_levenshtein_suggests_match` — `datafusion.public.account` with catalog `[stripe/accounts, github/issues]` hints at `stripe.accounts`; metadata has no `schema`, `table == "account"`.
- [x] Unit: `table_not_found_unqualified_no_close_match_falls_back_to_catalog_pointer` — same unqualified path with a distant catalog entry still surfaces the `coral.tables` pointer.
- [x] Unit: existing `table_not_found_missing_schema_points_at_coral_tables_catalog` still passes (schema is non-empty, so the unqualified branch doesn't trigger).
- [x] Unit: `table_not_found_synthetic_public_treated_as_unqualified` updated — the synthetic-`public` path now rides the same smarter hint, so the assertion follows suit (metadata semantics unchanged).
- [x] Unit: `table_not_found_unqualified_falls_back_to_catalog_pointer` updated — uses a distant catalog entry so the fallback branch is exercised.
- [x] Integration: `unqualified_table_suggests_schema_qualified_name` registers a real JSONL source `stripe.accounts` and runs `SELECT * FROM account` end-to-end; asserts hint contains `stripe.accounts`.
- [x] `cargo fmt --all -- --check` clean.
- [x] `cargo clippy --workspace --all-targets --locked -- -D warnings` clean.
- [x] `cargo test --workspace --locked` — all coral-engine tests pass; 2 pre-existing unrelated `coral-cli` test failures (`source_test_succeeds_when_query_tests_pass`, `source_test_exits_non_zero_when_query_tests_fail`) are present on the PR #120 base without this change and are out of scope.

### DIFF
--- a/crates/coral-engine/src/contracts/query_error.rs
+++ b/crates/coral-engine/src/contracts/query_error.rs
@@ -13,6 +13,16 @@ pub(crate) const UNKNOWN_COLUMN_REASON: &str = "UNKNOWN_COLUMN";
 /// Wire-stable `reason` code for table-not-found errors.
 pub(crate) const TABLE_NOT_FOUND_REASON: &str = "TABLE_NOT_FOUND";
 
+/// Minimum `strsim::normalized_levenshtein` score for a "did you mean?"
+/// suggestion to surface.
+///
+/// Matches the value `datafusion-common` uses in its own `FieldNotFound`
+/// Display impl (`filter(|s| normalized_levenshtein(s, field_name) >= 0.5)`),
+/// so a user sees the same forgiveness whether the error routes through
+/// our structured enrichment path or `DataFusion`'s raw error text. Change
+/// this only if we have data justifying divergence from that baseline.
+const DID_YOU_MEAN_SIMILARITY: f64 = 0.5;
+
 /// Structure-preserving column reference used by `unknown_column`.
 ///
 /// Keeping the qualifier components separate from the bare name means the
@@ -302,7 +312,7 @@ fn unknown_column_hint(missing: &ColumnParts, valid_columns: &[ColumnParts]) -> 
         .filter_map(|candidate| {
             let score =
                 strsim::normalized_levenshtein(&candidate.name.to_lowercase(), &missing_name_lower);
-            (score >= 0.5).then_some((candidate, score))
+            (score >= DID_YOU_MEAN_SIMILARITY).then_some((candidate, score))
         })
         .max_by(|left, right| {
             left.1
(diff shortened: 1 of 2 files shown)


crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -779,13 +818,54 @@ mod tests {
         // currently always emits a 3-part path, but the constructor
         // accepts this shape defensively — if an upstream change ever
         // drops the catalog prefix, this test still guards the fallback.
-        let tables = vec![table("hockey", "games")];
+        // Catalog entry is intentionally far enough from the missing name
+        // that the Levenshtein cross-schema pass yields no suggestion,
+        // so the generic catalog pointer still surfaces.
+        let tables = vec![table("hockey", "zzzzzzzzz")];
```
antonmry, line 799: I appreciate comments but this part of the code is becoming a bit too verbose. I wonder if would be better to add a comment per test, instead of a code per line?

antonmry (approved):

crates/coral-engine/src/contracts/query_error.rs
mystic123: Good call — condensed in 174c140. Each test now carries at most one short header comment at the top of the body; mid-function comments are gone where the assert messages already communicate intent. Net diff for the test module: -29 lines. Specifically collapsed the multi-paragraph blocks on the tests I added (synthetic-public, dotted-source, unqualified fallback, cross-schema Levenshtein). Let me know if you want any of them trimmed further.

PR #170, opened 2026-04-22
fix(sources): restore truncated column descriptions

sauldhernandez:
## Intent

Column descriptions in several bundled source manifests were capped around 120–130 characters by an LLM-driven regeneration step. The result: descriptions that end mid-sentence, mid-link, or mid-clause — giving callers (and agents reading these manifests) an incomplete picture of the underlying column. Linear: SOURCE-465.

## What it enables

- **Clean bundled-source descriptions.** GitHub (13), Stripe (11), and PagerDuty (328) entries are restored to their full upstream text, HTML-stripped in the same style as 8c22b56.
- **A guard against regression.** Adds a `detect-truncations` subcommand to the workspace `xtask` binary (`xtask/src/detect.rs`). It flags likely-truncated entries based on mid-sentence endings, unbalanced brackets, stopword tails, and a 120–130-char suspicious-length window. The prior classifier self-tests now live as `#[test]` cases that run under `cargo test -p xtask`.
- The detector reports 5 remaining hits in PagerDuty, all confirmed false positives: upstream's own short descriptions (`"…the Incident Workflow does"`, `"…License can have"`) matching a stopword-tail pattern. Left as-is; addressing would require rule tuning that overfits to the current manifests.

## Usage examples

Scan `sources/` (the default):

```sh
cargo run -p xtask -- detect-truncations
```

Scan a specific manifest:

```sh
cargo run -p xtask -- detect-truncations sources/pagerduty/manifest.yaml
```

Run the classifier unit tests:

```sh
cargo test -p xtask detect
```

## Test plan

- [x] `ryl sources` passes (yamllint clean)
- [x] `cargo test -p xtask` passes (20 tests, including 10 in the new `detect` module)
- [x] `cargo clippy -p xtask --all-targets --all-features -- -D warnings` clean
- [x] `cargo run -p xtask -- detect-truncations` against the three fixed manifests reports only the known false positives
- [x] `make docs-check` still succeeds under the new `generate-docs` subcommand
- [x] Reviewer spot-check on 2–3 restored descriptions in each manifest to confirm text matches upstream intent


scripts/detect-truncated-descriptions.py
```diff
@@ -0,0 +1,551 @@
+#!/usr/bin/env python3
```
antonmry: This could be a good candidate for the `xtasks` crate with an entry in the Makefile. Nothing wrong with using Python for this, it's just a way to follow an existing pattern in the code base and centralize tooling in a single place.

antonmry (approved):

PR #172, opened 2026-04-22
docs(sources/datadog): correct live events filter guidance

lazybean:
This PR keeps a manifest-only scope and updates only `sources/datadog/manifest.yaml`.

It rewrites the Datadog table guides so each one:
- says what the table is for
- names only the highest-value filters
- explains how to narrow large result sets
- includes one real caveat or quirk
- stays concise and query-oriented
- avoids the phrase "No required filters"

Follow-up fix from live validation:
- `datadog.events` now correctly says `start` and `end` use Unix epoch seconds, not ISO 8601

Verification:
- `./target/debug/coral sql` live query against `datadog.events` with Unix-second `start`/`end` filters
- `./target/debug/coral sql "SELECT table_name, guide FROM coral.tables WHERE schema_name = 'datadog' AND table_name = 'events'"`
- `cargo run -p coral-cli -- source lint sources/datadog/manifest.yaml`
- `cargo run -p xtask -- --sources-dir sources --index docs/reference/bundled-sources.mdx --docs-json docs/docs.json --check`
- `rg -n "No required filters" sources/datadog/manifest.yaml`
- `git diff --cached --check`
- `git diff --cached --name-only`
- `gh pr diff 172 --name-only`

Constraint: Final branch and PR must remain manifest-only
Constraint: Guides should stay query-oriented and only mention filters or caveats supported by the manifest schema
Rejected: Keep partial wording fixes on only the previously filterless tables | left the Datadog guide set inconsistent
Rejected: Keep the ISO 8601 note for `events` after the live query failure | the endpoint rejects that format with HTTP 400
Rejected: Regenerate or edit docs in this branch | violates the requested manifest-only PR scope
Confidence: high
Scope-risk: narrow
Reversibility: clean
Directive: Keep Datadog guides purpose-first and prefer the filter format proven by a live query when endpoint conventions differ
Not-tested: `make lint-sources` | `ryl` is unavailable in this environment

### DIFF
--- a/sources/datadog/manifest.yaml
+++ b/sources/datadog/manifest.yaml
@@ -47,8 +47,8 @@ tables:
   - name: monitors
     description: Datadog monitors with status and alert configuration.
     guide: >
-      No required filters — returns all monitors. Filter by status or type in SQL. Status values: Alert, Warn, No Data, OK,
-      Ignored, Skipped.
+      Use this table for full monitor inventory and current alert state. Start with status for active noise, then narrow by
+      type or name; status values are Alert, Warn, No Data, OK, Ignored, and Skipped.
     request:
       method: GET
       path: /api/v1/monitor
@@ -138,6 +138,11 @@ tables:
             - tags
   - name: monitor_search
     description: Datadog monitor search results
+    guide: >
+      Use this table for targeted monitor lookups when you already know a Datadog search query. Let the required query do
+      the big cut, then narrow the smaller result in SQL by status or type; use monitors instead when you need a full
+      inventory.
+
     request:
       method: GET
       path: /api/v1/monitor/search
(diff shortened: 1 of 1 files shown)


lazybean:
* **#172** <a href="https://app.graphite.com/github/pr/withcoral/coral/172?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/172?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sources/datadog/manifest.yaml
```diff
@@ -224,8 +229,9 @@ tables:
   - name: events
     description: Datadog events within a time range.
     guide: >
-      Requires start and end filters in ISO 8601 format (e.g. '2024-01-15T00:00:00Z'). date_happened is Unix epoch seconds.
```
lazybean, line 227: we should keep the format info

lazybean: Kept the ISO 8601 format guidance in the events guide while keeping the checklist wording concise.

antonmry (approved):

sources/datadog/manifest.yaml
lazybean: Live validation showed the events endpoint needs Unix epoch seconds for start/end, so I updated the guide accordingly. My earlier ISO 8601 reply was incorrect.

PR #173, opened 2026-04-22
feat: custom authenticators

iaptsiauri:
## Summary

Introduce a generic request authenticator extension point for source HTTP requests.

This lets Coral support authentication schemes that cannot be represented as static manifest headers alone. Frst implementation is aws_sigv4.

AWS SigV4 lives in a separate `coral-auth-aws` crate so AWS-specific dependencies and signing logic stay isolated from the core crates.

Extension registration now happens in coral-cli, coral-app keeps a noop default provider and remains extension-free by default; the CLI decides which extensions to install for its runtime.

`template.rs` was moved out of its previous location so auth and source parsing can share the same template model without tying that logic to one caller. 
`This is an intermediate step`: the template-related pieces are expected to move into coral-spec later, where source manifest parsing and validation should live.

  ## Manifest Auth Shape
HTTP-backed source auth was reworked from single freeform headers into a structured AuthSpec model:

  - BasicAuthSpec: username + password templates, base64-encoded at runtime.
  - HeaderAuthSpec: freeform header list for sources that do not fit a single-token shape, for example Datadog’s dual API keys.
  - CustomAuth: dispatches to a compiled-in signer via an authenticator: tag.
  - Top-level request_headers:: generic per-source HTTP headers not conveyed by auth, such as Accept or X-API-Version.
  - Runtime User-Agent: set once on the reqwest client in HttpSourceClient::from_manifest.


crates/coral-engine/src/backends/shared/template.rs
```diff
@@ -0,0 +1,183 @@
+//! Backend-agnostic template and value-source rendering.
```
antonmry, line 1: to be discussed. The move makes sense — template rendering is genuinely backend-agnostic logic, not HTTP-specific. But it's worth checking: could this live in coral-spec

antonmry (commented):
Nice! I've added some comments but it seems a good direction

Cargo.toml
```diff
@@ -27,6 +27,9 @@ assert_cmd = "2"
 anyhow = "1"
 arrow = "57.3.0"
 async-trait = "0.1"
+aws-credential-types = "1.2"
+aws-sigv4 = "1.4"
```
antonmry: we could move this to coral-auth-aws crate

Cargo.toml
```diff
@@ -69,6 +72,7 @@ coral-app = { path = "crates/coral-app" }
 coral-cli = { path = "crates/coral-cli" }
 coral-client = { path = "crates/coral-client" }
 coral-engine = { path = "crates/coral-engine" }
```
antonmry, line 72: we could use crates/auth/coral-aws or crates/auth/aws for crates that are extensions

crates/coral-app/src/query/extensions.rs
```diff
@@ -1,5 +1,8 @@
 //! App-owned selection of optional engine extensions for query runtime builds.
 
+use std::sync::Arc;
+
+use coral_auth_aws::AwsSigV4Authenticator;
```
antonmry, line 5: to be discussed. I think we should do this in `coral-cli` or `coral-cli-ee` to make coral-app agnostic.

crates/coral-engine/src/backends/shared/template.rs
iaptsiauri: I tried to do that in my previous PRs. There is a way to do that but I think not in this iteration as the change is substantial. Overall I think everything spec related should be moved to coral-spec

antonmry (approved):

crates/coral-spec/src/inputs.rs
```diff
@@ -41,6 +41,30 @@ pub struct ManifestInputSpec {
     pub hint: Option<String>,
 }
 
+/// Merge user-provided secrets and variables with manifest defaults into one
+/// runtime-ready input map.
+#[must_use]
+pub fn resolve_inputs(
```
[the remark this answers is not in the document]
iaptsiauri: This is intentional

crates/coral-cli/src/bootstrap.rs
```diff
@@ -30,14 +36,30 @@ pub(crate) async fn bootstrap() -> Result<Bootstrap, BootstrapError> {
         });
     }
 
-    let server = ServerBuilder::new().start().await?;
+    let server = ServerBuilder::new()
+        .with_engine_extensions_provider(Arc::new(AwsEngineExtensionsProvider))
```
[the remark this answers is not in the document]
iaptsiauri: This is intentional

crates/coral-cli/src/bootstrap.rs
```diff
@@ -24,14 +30,30 @@ pub(crate) async fn bootstrap() -> Result<Bootstrap, BootstrapError> {
...
```
jsummerfield: This puts AWS-specific app composition in the CLI adapter: `coral-cli` now knows about `AwsSigV4Authenticator`, the `aws_sigv4` dispatch key, and the engine extension provider types. That breaks the thin-wrapper boundary and also forces `coral_client::local` to re-export extension internals just so the CLI can wire the server. The cleaner ownership is to register built-in authenticators in `coral-app` as the local composition root, optionally behind a feature if the AWS dependency should be optional, and keep the CLI at `ServerBuilder::new().start().await?`.

docs/reference/source-spec-reference.mdx
```diff
@@ -230,6 +231,169 @@ In this minimal example:
...
+| ----- | -------- | ----- |
+| `service` | yes | AWS service code (`logs`, `monitoring`, `cloudtrail`, `execute-api`, …) |
+| `region` | yes | AWS region; templated from inputs |
+| `access_key_id` | yes | AWS access key ID |
+| `secret_access_key` | yes | AWS secret access key |
+| `session_token` | no | Omit for long-term credentials; include for STS/assumed-role sessions |
+
+**S3 note:** `service: s3` and `service: s3-outposts` switch to S3-specific signing (single percent-encoding, path normalization disabled, `X-Amz-Content-Sha256` enabled). This is required to avoid `SignatureDoesNotMatch` on bucket operations and is handled automatically.
+
+##### Registering additional custom authenticators
```
jsummerfield: Let's not make this a public interface yet, let's give it more time to settle.

docs/reference/source-spec-reference.mdx
```diff
@@ -230,6 +231,169 @@ In this minimal example:
...
+
+```yaml
+auth:
+  type: CustomAuth
+  authenticator: aws_sigv4
+  service: logs
+  region: "{{input.AWS_REGION}}"
+  access_key_id: "{{input.AWS_ACCESS_KEY_ID}}"
+  secret_access_key: "{{input.AWS_SECRET_ACCESS_KEY}}"
+  session_token: "{{input.AWS_SESSION_TOKEN}}"   # optional, for STS/role-assumed credentials
```
jsummerfield: The example marks `session_token` as optional, but this exact line is not optional once copied into a manifest: `{{input.AWS_SESSION_TOKEN}}` must be declared under `inputs`, and secret inputs cannot have defaults, so installs will require a session token. I would omit `session_token` from the default example and add a separate STS/assumed-role snippet showing the extra declared secret users need when they actually have temporary credentials.

crates/coral-app/src/bootstrap/server.rs
```diff
@@ -43,7 +43,7 @@ impl ServerConfig {
     pub(crate) fn new() -> Self {
         Self {
             config_dir: None,
-            engine_extensions_provider: Arc::new(NoopEngineExtensionsProvider),
+            engine_extensions_provider: Arc::new(BuiltinEngineExtensionsProvider),
```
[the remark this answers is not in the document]
iaptsiauri: For now this is fine

docs/reference/source-spec-reference.mdx
```diff
@@ -230,6 +231,117 @@ In this minimal example:
...
+
+The top-level `auth` block declares how Coral authenticates outgoing requests. Pick one of three `type` values:
+
+| `type` | Purpose |
+| ------ | ------- |
+| `HeaderAuth` | Attach one or more declarative auth headers (Bearer tokens, API keys, pre-signed headers). |
+| `BasicAuth` | HTTP Basic. Coral base64-encodes `username:password` into `Authorization: Basic ...`. |
+| `CustomAuth` | Dispatch request signing to a built-in Coral authenticator (for schemes that must observe the built request, such as AWS SigV4). |
+
+All three resolve their inputs from declared source inputs (`{{input.KEY}}` templates or `from: input`). Auth is source-scoped — it does not see SQL filters or pagination state.
```
AlbertQM: Not 100% sure what this means (and I'm guessing other users mind think the same), could we explain it in a different way?

AlbertQM (commented):
Thanks Ilia! The source-spec-reference is becoming quite heavy to digest, we might want to re-think how these concepts are presented (not for now).

In the meantime, maybe adding a more user friendly paragraph/table in the `write-a-custom-source` might work nicely

crates/coral-app/src/bootstrap/server.rs
```diff
@@ -59,7 +59,10 @@ impl ServerConfig {
         mut self,
         engine_extensions_provider: Arc<dyn EngineExtensionsProvider>,
     ) -> Self {
-        self.engine_extensions_provider = engine_extensions_provider;
+        self.engine_extensions_provider = compose_engine_extensions_providers(
+            Arc::clone(&self.engine_extensions_provider),
+            engine_extensions_provider,
```
sauldhernandez: Small nit - this forces me to use the built-in providers - even if I don't want to. No use case needs that for now, but worth either documenting or offering a escape hatch

crates/coral-app/src/query/extensions.rs
```diff
@@ -21,3 +24,135 @@ impl EngineExtensionsProvider for NoopEngineExtensionsProvider {
...
+}
+
+pub(crate) fn compose_engine_extensions_providers(
+    base: Arc<dyn EngineExtensionsProvider>,
+    extra: Arc<dyn EngineExtensionsProvider>,
+) -> Arc<dyn EngineExtensionsProvider> {
+    Arc::new(CompositeEngineExtensionsProvider { base, extra })
+}
+
+struct CompositeEngineExtensionsProvider {
```
sauldhernandez: This could be generalized to take a Vec or slice instead of just two providers

crates/coral-engine/src/backends/http/auth.rs
```diff
@@ -0,0 +1,148 @@
...
+use coral_spec::{AuthSpec, BasicAuthSpec, HeaderAuthSpec};
+
+use crate::RequestAuthenticator;
+use crate::backends::shared::template::{
+    EMPTY_MAP, render_template, resolve_value_source, value_to_string,
+};
+
+/// Built-in auth variants resolve their headers from resolved inputs only;
+/// they do not need access to the fully built request.
+trait BuiltinAuth {
```
sauldhernandez, line 20: Why do we need to treat these in a special way? Why don't they become just one more auth extension?

crates/coral-engine/src/backends/http/client.rs
```diff
@@ -80,28 +91,16 @@ impl HttpSourceClient {
...
-            }
-        }
+        let resolved_inputs =
+            coral_spec::resolve_inputs(&manifest.declared_inputs, source_secrets, source_variables);
+        validate_source_scoped_http_config(manifest, request_authenticators, &resolved_inputs)?;
 
         let request_timeout = Duration::from_secs(DEFAULT_HTTP_REQUEST_TIMEOUT_SECS);
         let http = reqwest::Client::builder()
             .timeout(request_timeout)
+            .user_agent("coral")
```
sauldhernandez: nit: would be nice to add the coral version here as well

crates/coral-engine/src/backends/http/client.rs
```diff
@@ -333,6 +336,169 @@ impl HttpSourceClient {
     }
 }
 
+fn validate_source_scoped_http_config(
+    manifest: &HttpSourceManifest,
+    request_authenticators: &HashMap<String, Arc<dyn RequestAuthenticator>>,
```
sauldhernandez, line 342: Why do we need a hashmap of authenticators at this point? Shouldn't the caller to the engine have already resolved which authenticator to use?

docs/reference/source-spec-reference.mdx
```diff
@@ -230,6 +231,117 @@ In this minimal example:
...
 
+### HTTP authentication
+
+The top-level `auth` block declares how Coral authenticates outgoing requests. Pick one of three `type` values:
+
+| `type` | Purpose |
+| ------ | ------- |
+| `HeaderAuth` | Attach one or more declarative auth headers (Bearer tokens, API keys, pre-signed headers). |
+| `BasicAuth` | HTTP Basic. Coral base64-encodes `username:password` into `Authorization: Basic ...`. |
+| `CustomAuth` | Dispatch request signing to a built-in Coral authenticator (for schemes that must observe the built request, such as AWS SigV4). |
```
sauldhernandez: From a user's POV, all three of these are built-in Coral authenticators. This distinction here is confusing

sauldhernandez (changes requested):

crates/coral-app/src/query/extensions.rs
iaptsiauri: Right now the builder API is incremental:
```
  ServerBuilder::new()
      .with_engine_extensions_provider(a)
      .with_engine_extensions_provider(b)
```

`CompositeEngineExtensionsProvider { base, extra }` matches that call pattern directly. Generalizing to `vec` or `slice` would mostly change the internal representation not the external capability.

sauldhernandez: Nice! I didn't tie those 2 together, makes sense.

crates/coral-engine/src/backends/http/auth.rs
iaptsiauri: imho `BasicAuth` and `HeaderAuth` should stay separate from auth extensions, because they are fully represented in the manifest and don’t need any additional runtime registration to work. `RequestAuthenticator`is the extension hook for `CustomAuth`.

crates/coral-engine/src/backends/http/client.rs
iaptsiauri: Good catch

crates/coral-engine/src/backends/http/client.rs
iaptsiauri: I agree with the end state, `coral-spec` should own validation and the engine should consume a validated/resolved auth model. I don’t think this PR is the right place to do that, because it would mix spec-layer validation changes, runtime resolution, and engine execution refactoring in one step. For now this remains an intermediate step, so I’d keep the registry lookup here and move it once we do the broader compiled-auth cleanup.

Slightly hinted this in the summary as well

simonwhitaker (approved):

crates/coral-engine/src/backends/http/auth.rs
sauldhernandez: I see. What if custom auth contributes to the schema? That way not only you can treat BasicAuth and HeaderAuth in the same way as any other custom auth, but custom auth gets schema validation to catch issues early on.

This last bit is quite important. With the current PR proposal we're validating at runtime (likely multiple times) instead of a single time when the source is added via schema validation.

crates/coral-engine/src/backends/http/client.rs
sauldhernandez: So the end-goal is to move validation to the spec crate, with enhancable schemas provided by extensions?

iaptsiauri: Yeap exactly

crates/coral-engine/src/backends/http/auth.rs
iaptsiauri: I do agree on validation. But validation we can address differently. So if validation is concern there are two options. We can make schema changes to the manifest where we do not do a generic `CustomAuth` with `additionalProperties: allowed`, but have a schema definition for each `CustomAuth` type. This is one quick and slightly dirty option. 

Another would be to extend `RequestAuthenticator::validate_config` which will be called during registration time and this would solve issue you are describing

crates/coral-app/src/bootstrap/server.rs
iaptsiauri: Added docs for it

docs/guides/write-a-custom-source.mdx
AlbertQM, line 1: I'm surprised that this file doesn't have at least some information about the new auth -- I'd expect at least ~1 paragraph describing the auth options, and a link to the source-spec-reference.mdx for more information. WDYT? 😄

AlbertQM (approved):

crates/coral-engine/src/backends/http/auth.rs
sauldhernandez: I like the latter option, but instead of doing the validation itself it could return the partial json schema for that custom auth, so the main schema validation flow can merge it and validate

crates/coral-app/src/bootstrap/server.rs
```diff
@@ -87,7 +91,12 @@ impl ServerBuilder {
...
 
     #[must_use]
-    /// Overrides the engine extensions provider used for query runtime builds.
+    /// Adds an engine extensions provider used for query runtime builds.
+    ///
+    /// Providers are composed in call order, so extensions from this provider
+    /// are merged with extensions from any provider already configured,
+    /// including providers added by [`Self::with_builtin_extensions`]. Built-in
+    /// providers are only included when explicitly added.
     pub fn with_engine_extensions_provider(
```
jsummerfield: with_ suggests replace. add_ would be clearer imo.

crates/coral-engine/src/backends/http/auth.rs
iaptsiauri: I would not make it part of this pr though to be honest. This needs little bit more thinking and it will explode already massive scope for this one

jsummerfield (approved):

sauldhernandez (approved):

PR #179, opened 2026-04-23
refactor(cli): centralize CLI env access

jsummerfield:
## Summary

Centralizes CLI-owned process environment reads behind a `coral-cli` env boundary so client-side env-backed workflows do not rely on local `std::env` exceptions.

This keeps PR #164's env-based `coral source add` flow in the CLI layer while preserving Clippy enforcement that raw process env reads happen only through crate-owned seams.

## Changes

- Add `coral_cli::env` as the CLI-owned process environment module.
- Route `coral source add` env input collection and the `CORAL_ENDPOINT` test hook through that module.
- Update Clippy disallowed-method guidance to cover all raw `std::env` reads.
- Document env ownership in root and crate AGENTS guidance.

## Verification

- `make rust-checks`

### DIFF
--- a/crates/coral-cli/src/env.rs
+++ b/crates/coral-cli/src/env.rs
@@ -1,25 +1,17 @@
 //! CLI-owned process environment accessors.
 //!
 //! `coral-cli` is allowed to read process environment when the CLI surface
-//! explicitly defines an env-backed workflow, such as test-only bootstrap
-//! overrides today and install-time input collection for source setup.
-
-#![allow(
-    clippy::disallowed_methods,
-    reason = "coral-cli owns CLI-surface process environment access in this module."
-)]
-
-/// Reads a CLI-owned UTF-8 environment variable.
-#[must_use]
-pub fn var(key: &str) -> Option<String> {
-    std::env::var(key).ok()
-}
+//! explicitly defines an env-backed workflow.
 
 #[cfg(feature = "cli-test-server")]
 const CORAL_ENDPOINT_ENV: &str = "CORAL_ENDPOINT";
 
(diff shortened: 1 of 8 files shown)


crates/coral-cli/src/env.rs
```diff
@@ -0,0 +1,28 @@
+//! CLI-owned process environment accessors.
+//!
+//! `coral-cli` is allowed to read process environment when the CLI surface
+//! explicitly defines an env-backed workflow, such as test-only bootstrap
+//! overrides today and install-time input collection for source setup.
+
+#![allow(
+    clippy::disallowed_methods,
```
simonwhitaker: This will suppress any future disallowed methods anywhere in this file. Can we be more targetted?

crates/coral-cli/src/env.rs
```diff
@@ -0,0 +1,28 @@
...
+    reason = "coral-cli owns CLI-surface process environment access in this module."
+)]
+
+/// Reads a CLI-owned UTF-8 environment variable.
+#[must_use]
+pub fn var(key: &str) -> Option<String> {
```
simonwhitaker: I don't get it. IIUC, the clippy rule exists to prevent arbitrary code from reading from the environment. Any code that genuinely needs to read from the environment has to explicitly declare that intent via an `allow` directive.

This method effectively adds a blessed backdoor for calling `std::env::var`, so I can now read from the environment anywhere I like in this crate without having to declare my intent. How is this an improvement?

jsummerfield: That's fair - reverted back to explicitly allowing in source_ops.

simonwhitaker (approved):

PR #181, opened 2026-04-23
refactor(spec): lift declared_inputs and required_secret_names to SourceManifestCommon

jamesaud:
## Summary
- Lift `declared_inputs: Vec<ManifestInputSpec>` and `required_secret_names()` from each backend manifest (Http / Parquet / Jsonl) into `SourceManifestCommon`, so there is one source of truth.
- Add `ValidatedSourceManifest::common()` and route the existing accessors (`schema_name`, `source_version`, `description`, `test_queries`, `declared_inputs`, `required_secret_names`) through it, collapsing six identical three-arm matches in `parser.rs`.
- Update the two direct consumers of the old field (`coral-engine/src/backends/http/{mod,client}.rs`) to read `manifest.common.declared_inputs`.

Follow-up to #159 — addresses [this review comment](https://github.com/withcoral/coral/pull/159#discussion_r3129998810) on the match arms that got touched incidentally there.

Public API is unchanged: all existing `ValidatedSourceManifest` accessors keep the same signatures. The only removed surface is the per-backend `HttpSourceManifest::required_secret_names` / `ParquetSourceManifest::required_secret_names` / `JsonlSourceManifest::required_secret_names` methods and the public `declared_inputs` fields on those structs; there are no external callers of those methods, and the two field reads in `coral-engine` are updated in this PR.

## Test plan
- [x] `make rust-checks` (fmt + check + clippy -D warnings + tests + doc -D warnings)
- [x] `cargo test -p coral-spec` — 44 passing
- [x] `cargo test -p coral-engine` — passing
- [ ] Two pre-existing flakes in `coral-cli --test source` (`source_test_succeeds_when_query_tests_pass`, `source_test_exits_non_zero_when_query_tests_fail`) also fail on a clean `main` checkout — unrelated to this change.

### DIFF
--- a/crates/coral-spec/src/parser.rs
+++ b/crates/coral-spec/src/parser.rs
@@ -11,7 +11,7 @@ use serde_json::Value;
 use crate::backends::file::{JsonlSourceManifest, ParquetSourceManifest};
 use crate::backends::http::HttpSourceManifest;
 use crate::schema::validate_manifest_schema;
-use crate::{ManifestError, ManifestInputSpec, Result, SourceBackend};
+use crate::{ManifestError, ManifestInputSpec, Result, SourceBackend, SourceManifestCommon};
 
 /// Validated top-level source spec for one registered source.
 ///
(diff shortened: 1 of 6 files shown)


crates/coral-spec/src/parser.rs
```diff
@@ -46,64 +46,50 @@ impl ValidatedSourceManifest {
...
     #[must_use]
-    /// Returns the source-spec `name`, which is also the stable SQL schema name.
-    pub fn schema_name(&self) -> &str {
+    /// Returns the backend-agnostic source metadata shared by every manifest.
+    pub fn common(&self) -> &SourceManifestCommon {
         match &self.inner {
-            ValidatedManifestKind::Http(manifest) => &manifest.common.name,
-            ValidatedManifestKind::Parquet(manifest) => &manifest.common.name,
-            ValidatedManifestKind::Jsonl(manifest) => &manifest.common.name,
+            ValidatedManifestKind::Http(manifest) => &manifest.common,
```
antonmry, line 52: The six matches collapsed into one is a nice cleanup, but common() itself still pattern-matches three arms that all do &manifest.common. Since every
  backend embeds the same SourceManifestCommon type, could we store it at the ValidatedSourceManifest level instead of inside each variant?

```rust
   pub struct ValidatedSourceManifest {
       pub common: SourceManifestCommon,
       inner: ValidatedManifestKind,
   }
```

That way common is a direct field access with no match at all, and adding a new backend can't forget to wire it up.

PR #188, opened 2026-04-23
feat(cli): add `source info` command

simonwhitaker:
## Summary
- Adds `coral source info <name>` which shows whether a source is installed, along with version, description, and declared inputs.
- `-v/--verbose` reveals each input's hint text; hints are hidden by default to keep output terse.
- Uses the existing `DiscoverSources` RPC; errors cleanly with `unknown source '<name>'` when no match is found.

## Test plan
- [x] `cargo test -p coral-cli --features cli-test-server` (new `source_info_*` integration tests cover installed, available, verbose, and unknown-source cases)
- [x] `cargo clippy -p coral-cli --features cli-test-server --all-targets`

### Regular

CleanShot 2026-04-24 at 9  40 50@2x

### Verbose

CleanShot 2026-04-24 at 9  41 06@2x

### DIFF
--- a/crates/coral-api/proto/coral/v1/sources.proto
+++ b/crates/coral-api/proto/coral/v1/sources.proto
@@ -71,8 +71,8 @@ message SourceInputSpec {
   string hint = 5;
 }
 
-// One bundled or imported source available for installation.
-message AvailableSource {
+// One bundled or imported source metadata record.
+message SourceInfo {
   // Canonical source name.
   string name = 1;
   // Human-readable description.
@@ -83,7 +83,7 @@ message AvailableSource {
   repeated SourceInputSpec inputs = 4;
   // Whether a source with this name is already configured in the workspace.
   bool installed = 5;
-  // Where this available source came from.
+  // Where this source came from.
   SourceOrigin origin = 6;
 }
 
@@ -93,10 +93,10 @@ message DiscoverSourcesRequest {
   Workspace workspace = 1;
 }
 
-// Returns all bundled available sources.
+// Returns all bundled source metadata records.
 message DiscoverSourcesResponse {
   // Bundled sources compiled into this Coral build.
-  repeated AvailableSource sources = 1;
+  repeated SourceInfo sources = 1;
 }
 
 // Installs one bundled source by name.
@@ -140,6 +140,14 @@ message GetSourceRequest {
   // Owning workspace resource.
   Workspace workspace = 1;
   // Bare source name to fetch.
+  string name = 2;
+}
+
+// Fetches source metadata by identifier.
+message GetSourceInfoRequest {
+  // Owning workspace resource.
+  Workspace workspace = 1;
+  // Bare source name to fetch metadata for.
   string name = 2;
 }
 
@@ -202,6 +210,8 @@ service SourceService {
   rpc ListSources(ListSourcesRequest) returns (ListSourcesResponse);
   // Fetches one registered source.
   rpc GetSource(GetSourceRequest) returns (Source);
+  // Fetches metadata for one installed or bundled source.
+  rpc GetSourceInfo(GetSourceInfoRequest) returns (SourceInfo);
   // Installs one bundled source.
   rpc CreateBundledSource(CreateBundledSourceRequest) returns (Source);
   // Imports one user-provided source manifest.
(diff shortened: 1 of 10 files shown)


AlbertQM (approved):
Docs look good. Though as Ilia suggested, I think `coral source add <name> --help` would also be a good place for this; less API surface 😄

simonwhitaker:
> Docs look good. Though as Ilia suggested, I think `coral source add <name> --help` would also be a good place for this; less API surface 😄

I'm coming down against this idea, for a couple of reasons:

1. In `coral source add <name>`, name is an argument, not a subcommand. So this means we'd have two different outputs for the `--help` option for the `coral source add` subcommand, depending on whether or not an argument was pased. This is unorthodox, and IMO confusing.
2. Learning about a source is a separate concern from adding it. Making this the `--help` option for `coral source add <name>` feels like an obvious place _if you want to know which env vars to set when adding a source_, but a non-obvious place if you want to learn about that source for any other reason.

jsummerfield (commented):
Found one CLI contract issue to tighten before this surface settles.

crates/coral-cli/src/source_ops.rs
```diff
@@ -134,6 +134,64 @@ pub(crate) fn load_validated_manifest_file(
...
+    app: &AppClient,
+    name: &str,
+    verbose: bool,
+) -> Result<(), anyhow::Error> {
+    let canonical = source_name_arg(Some(name))?;
+    let sources = discover_sources(app).await?;
+    let source = sources
+        .into_iter()
+        .find(|source| source.name == canonical)
+        .ok_or_else(|| anyhow::anyhow!("unknown source '{canonical}'"))?;
```
jsummerfield: The CLI help exposes `coral source info <NAME>` as metadata for a generic source, but this implementation only searches `discover_sources`, which is the bundled-source catalog. An installed custom/imported source shown by `coral source list` will return `unknown source`. I think we should either scope the generated help/error text explicitly to bundled sources, or include installed-source lookup semantics now before this CLI surface settles.

simonwhitaker: Done

```sh
$ cargo run --bin coral -- source info unifi_site_manager
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.70s
     Running `target/debug/coral source info unifi_site_manager`
unifi_site_manager
  Status:      installed
  Origin:      imported
  Version:     0.1.0

  Configured inputs
    UNIFI_SITE_MANAGER_API_KEY (secret)
```

crates/coral-cli/src/source_ops.rs
```diff
@@ -134,6 +134,105 @@ pub(crate) fn load_validated_manifest_file(
...
+            println!("      {}", style(&input.hint).dim());
+        }
+    }
+}
+
+fn print_installed_source_info(source: &Source) {
+    println!("{}", style(&source.name).bold());
+    println!("  Status:      {}", style("installed").green());
+    println!("  Origin:      {}", source_origin_label(source.origin));
+    println!("  Version:     {}", source.version);
```
jsummerfield: This duplication is a symptom of a deeper problem - the CLI is doing too much by composing two views over sources (discover_ and list_sources). This goes against the "keep adapters thin" rule.

Better would be for coral-app/API to return a single “source info” shape, backed by the effective installed manifest when available. Then coral-cli becomes a thin renderer and this duplication disappears naturally.

Worth thinking whether we should extend `DiscoverSources` or have a separate endpoint.

simonwhitaker: DiscoverSources is documented as being a view over bundled manifests, so I'd say we want a separate endpoint.

simonwhitaker: Codex thinks the same. Letting it do its thing...

crates/coral-api/proto/coral/v1/sources.proto
```diff
@@ -71,8 +71,8 @@ message SourceInputSpec {
   string hint = 5;
 }
 
-// One bundled or imported source available for installation.
-message AvailableSource {
+// One bundled or imported source metadata record.
+message SourceInfo {
```
jsummerfield, line 75: Nit - `SourceDefinition`? `SourceMetadata`?

`SourceInfo` is a bit too close to `Source` imo.

jsummerfield (approved):

PR #193, opened 2026-04-24
fix: centralize local name validation invariants

jsummerfield:
## Summary
Keep local source and workspace name validation in one app-owned place so the same invariants are enforced consistently across Coral surfaces.

## Why
The app already owns `SourceName` and `WorkspaceName`, but their validation logic had drifted into duplicate implementations. That made small rule changes easy to miss and let the CLI reject some invalid names later than necessary.

## Validation
- `make rust-checks docs-check`

### DIFF
--- a/crates/coral-app/src/identity.rs
+++ b/crates/coral-app/src/identity.rs
@@ -1,0 +1,52 @@
+//! Shared validation helpers for app-owned identifiers.
+
+use crate::bootstrap::AppError;
+
+pub(crate) fn parse_path_segment(kind: &str, value: &str) -> Result<String, AppError> {
+    let trimmed = value.trim();
+    if trimmed.is_empty() {
+        return Err(AppError::InvalidInput(format!("missing {kind} name")));
+    }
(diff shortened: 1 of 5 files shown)


crates/coral-app/src/identity.rs
```diff
@@ -0,0 +1,52 @@
+//! Shared validation helpers for app-owned identifiers.
+
+use crate::bootstrap::AppError;
+
+pub(crate) fn parse_path_segment(kind: &str, value: &str) -> Result<String, AppError> {
```
sauldhernandez, line 5: I get this PR is a lift of existing logic, but based on the impl we're safeguarding against invalid FS names, but shouldn't we also safeguard against SQL-invalid names?

sauldhernandez (approved):

crates/coral-app/src/identity.rs
jsummerfield: will follow up with a PR in `coral-spec`

PR #199, opened 2026-04-24
feat(slack): read thread replies

jsummerfield:
## Summary

- adds a Slack `thread_replies` table so Coral can read replies from Slack threads through `conversations.replies`
- exposes `channel` and `thread_ts` as required filters, with optional `oldest` and `latest` bounds
- updates bundled source docs so Slack advertises thread reply support

## Verification

- `coral source lint sources/slack/manifest.yaml`
- `make docs-check`
- `cargo run -p coral-cli -- sql "SELECT COUNT(*) AS n FROM slack.thread_replies WHERE channel = 'C061S280LM9' AND thread_ts = '1773682377.310009'"` returned 25 rows

### DIFF
--- a/sources/slack/manifest.yaml
+++ b/sources/slack/manifest.yaml
@@ -1,9 +1,9 @@
 name: slack
-version: 2.0.0
+version: 2.1.0
 dsl_version: 3
 backend: http
 description: >-
-  Query channels, messages, and users from your Slack workspace.
+  Query channels, messages, thread replies, and users from your Slack workspace.
 inputs:
   SLACK_TOKEN:
     kind: secret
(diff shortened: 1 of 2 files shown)


sources/slack/manifest.yaml
```diff
@@ -248,11 +240,138 @@ tables:
...
+  - name: thread_replies
+    description: Slack thread replies for a specific channel and parent message
+    request:
+      method: GET
+      path: /api/conversations.replies
+      query:
+        - name: channel
+          from: filter
+          key: channel
+        - name: ts
```
simonwhitaker, line 252: Why does this have name `ts` but key `thread_ts`? Is there a good reason not to use the same in both places?

simonwhitaker (approved):

antonmry (approved):

sources/slack/manifest.yaml
jsummerfield: Good question. It's to work around a quirk of Slack's API:

In the API request, `ts` means: “which thread should I fetch?”

```
conversations.replies?channel=C...&ts=1773682377.310009
```

In the API response, every returned message also has its own ts:

```json
{ "thread_ts": "1773682377.310009", "ts": "1773682393.126099" }
```

That response `ts` means: “when was this particular message/reply sent?”

Since `ts` means two things, Coral incorrectly tries to filter the individual results with the parent's`ts`.

PR #200, opened 2026-04-24
feat: add cloudwatch source

iaptsiauri:
## Summary

Adds two new sources `cloudwatch_logs` and `cloudwatch_metrics`

To support the CloudWatch metrics API, the HTTP client now builds JSON request bodies from manifest paths that include array indexes. That lets a manifest express requests like:
```
  body:
    - path: [Statistics, "0"]
      from: literal
      value: Average
    - path: [Statistics, "1"]
      from: literal
      value: Minimum
    - path: [Dimensions, "0", Name]
      from: filter
      key: dimension_name
    - path: [Dimensions, "0", Value]
      from: filter
      key: dimension_value
```
  which now render to the AWS-compatible JSON body:
```
  {
    "Statistics": ["Average", "Minimum"],
    "Dimensions": [
      {
        "Name": "ClusterName",
        "Value": "titaness"
      }
    ]
  }
```
Before this change, numeric path segments were treated like object keys, so the same manifest would have produced invalid object-shaped JSON instead of arrays.

## Validation:

- cargo build
- cargo test -p coral-spec --locked
- live queries with ./target/debug/coral against a configured cloudwatch_logs source for log_groups, log_streams, log_events, and prefix-filtered log_streams

### DIFF
--- a/sources/cloudwatch_logs/manifest.yaml
+++ b/sources/cloudwatch_logs/manifest.yaml
@@ -124,7 +124,7 @@ tables:
           key: order_by
           default: LogStreamName
         - path: [descending]
-          from: filter
+          from: filter_bool
           key: descending
           default: true
         - path: [limit]
@@ -157,7 +157,7 @@ tables:
         description: Ordering mode filter.
         expr: {kind: from_filter, key: order_by}
       - name: descending
-        type: Utf8
+        type: Boolean
         nullable: true
         virtual: true
         description: Descending sort filter.
--- a/sources/cloudwatch_logs/manifest.yaml
+++ b/sources/cloudwatch_logs/manifest.yaml
@@ -1,0 +1,294 @@
+dsl_version: 3
+name: cloudwatch_logs
+version: 0.1.0
+backend: http
+description: >-
+  Query Amazon CloudWatch Logs groups, streams, and events.
+inputs:
+  AWS_REGION:
+    kind: variable
+    default: us-east-1
+    hint: |
+      AWS region for CloudWatch Logs API requests, for example `us-east-1`.
+  AWS_ENDPOINT_SUFFIX:
+    kind: variable
+    default: amazonaws.com
+    hint: |
+      AWS endpoint DNS suffix. Keep `amazonaws.com` for standard AWS
+      regions. Use the appropriate suffix for other partitions.
+  AWS_ACCESS_KEY_ID:
+    kind: secret
+    hint: |
+      AWS access key ID with CloudWatch Logs read permissions.
+  AWS_SECRET_ACCESS_KEY:
+    kind: secret
+    hint: |
+      AWS secret access key for SigV4 signing.
+base_url: https://logs.{{input.AWS_REGION}}.{{input.AWS_ENDPOINT_SUFFIX}}
+auth:
+  type: CustomAuth
+  authenticator: aws_sigv4
+  service: logs
+  region: "{{input.AWS_REGION}}"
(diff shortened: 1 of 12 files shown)


simonwhitaker (approved):
Docs look good.

sources/cloudwatch_metrics/manifest.yaml
```diff
@@ -0,0 +1,769 @@
...
+        type: Utf8
+        nullable: false
+        description: Dashboard name.
+        expr: {kind: path, path: [DashboardName]}
+      - name: dashboard_arn
+        type: Utf8
+        nullable: true
+        description: Dashboard ARN.
+        expr: {kind: path, path: [DashboardArn]}
+      - name: last_modified
```
sauldhernandez, line 578: This maps `last_modified` directly to a Coral `Timestamp`. Coral interprets `Timestamp` values as epoch microseconds in crates/coral-engine/src/backends/shared/mapping.rs, but AWS documents `DashboardEntry.LastModified` as milliseconds since epoch. That will shift dashboard timestamps about 1000x too early; at minimum this field is wrong, and the same raw timestamp mapping pattern appears elsewhere in this manifest.

sources/cloudwatch_metrics/manifest.yaml
```diff
@@ -0,0 +1,769 @@
...
+        - path: [Namespace]
+          from: filter
+          key: namespace
+        - path: [MetricName]
+          from: filter
+          key: metric_name
+        - path: [RecentlyActive]
+          from: filter
+          key: recently_active
+        - path: [IncludeLinkedAccounts]
```
sauldhernandez, line 71: This uses `from: filter` for `IncludeLinkedAccounts`. In this engine, `filter` values are always serialized as JSON strings by crates/coral-engine/src/backends/shared/template.rs, so `WHERE include_linked_accounts = true` becomes `"IncludeLinkedAccounts":"true"`. CloudWatch expects a boolean here, so cross-account `metrics` queries using this filter will fail.

sources/cloudwatch_logs/manifest.yaml
```diff
@@ -0,0 +1,294 @@
...
+          from: filter
+          key: log_group_name
+        - path: [logStreamNamePrefix]
+          from: filter
+          key: log_stream_name_prefix
+        - path: [orderBy]
+          from: filter
+          key: order_by
+          default: LogStreamName
+        - path: [descending]
```
sauldhernandez, line 126: Same here, boolean will be parsed into a JSON string, which will not be the expected shape by AWS

sauldhernandez (changes requested):

antonmry (approved):

sources/cloudwatch_logs/manifest.yaml
iaptsiauri: Done

sources/cloudwatch_metrics/manifest.yaml
iaptsiauri: Done

sources/cloudwatch_metrics/manifest.yaml
iaptsiauri: Fixed

iaptsiauri: This is a good example we should add more guidance to the agent regarding the Timestamps. Will make a ticket for it.

sauldhernandez (approved):

PR #209, opened 2026-04-27
feat(cli): friendly not-found error for `coral source remove`

AlbertQM:
## Summary

- Render `coral source remove <name>` missing-source failures as a friendly CLI error that keeps the user-provided source name and suggests `coral source list`.
- Reuse the same structured source-not-found discriminator for `coral source test <name>` so only AIP-193 `SOURCE_NOT_FOUND` statuses get normalized; unrelated `Code::NotFound` errors continue to surface raw.
- Emit `SOURCE_NOT_FOUND` `ErrorInfo` from `app_status` for `AppError::SourceNotFound` and centralize the Coral structured-error domain/reason/metadata constants in `coral-api`.

## Test plan

When trying to removing a source that's not installed:

Before
Error: code: 'Some requested entity was not found', message: "source 'default:hn' not found"

After
Error: source 'hn' was not found. Run coral source list to see installed sources.

### DIFF
--- a/crates/coral-api/src/lib.rs
+++ b/crates/coral-api/src/lib.rs
@@ -69,6 +69,9 @@ pub const CORAL_ERROR_DOMAIN: &str = "coral.withcoral.com";
 /// Canonical default workspace name used across local Coral surfaces.
 pub const DEFAULT_WORKSPACE_ID: &str = "default";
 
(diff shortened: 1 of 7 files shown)


jsummerfield (approved):
Sorry for the delay!

PR #210, opened 2026-04-28
feat(api): filter listed tables by schema

mystic123:
Stack: 1 of 3 for SOURCE-459.

## Summary
- add exact `schema_filter` to `ListTablesRequest`
- normalize the filter at the app service edge and pass it through the query manager
- add engine and app gRPC coverage for filtered, missing, and unfiltered table listing

## Tests
- `cargo test -p coral-engine --test engine catalog_tests::list_tables_filters_by_schema`
- `cargo test -p coral-app --test grpc -- list_tables_filters_by_schema`
- `make rust-checks` on the final stacked branch


mystic123:
Closing per team decision: follow-up benchmark metrics were flat, and this is not the direction we want to take for SOURCE-459.

PR #211, opened 2026-04-28
feat(mcp): expose schema filter on list_tables

mystic123:
Stack: 2 of 3 for SOURCE-459. Base: `pawel/source-459-list-tables-schema-filter`.

## Summary
- add optional MCP `list_tables.schema` input with exact-match semantics
- map MCP `schema` to proto `schema_filter` explicitly in the adapter
- update MCP/CLI docs and tests for the new argument

## Tests
- `cargo test -p coral-mcp`
- `make rust-checks` on the final stacked branch


mystic123:
Closing per team decision: follow-up benchmark metrics were flat, and this is not the direction we want to take for SOURCE-459.

PR #213, opened 2026-04-28
docs: swap demo gif for Claude Code example

AlbertQM:
## Summary
- Replace `coral-sql-join.gif` with `claude-query-example.gif` in README and docs index
- Update lead-in to "Here's Claude Code using Coral to answer the question:"

## Test plan

`npx mint dev`, look at the gif on the preview


docs/index.mdx
```diff
@@ -77,11 +77,11 @@ Full [benchmark report](https://withcoral.com/benchmarks).
 
 Coral sits between you and your data sources: you (or your agent) write SQL, and Coral translates it into API calls or file reads, then returns a single query result.
 
-For example, here's Coral answering the question:
+For example, here's Claude Code using Coral to answer the question:
 
 > Which Sentry errors create the most noise in Slack, and who is working on them?
 
-![coral sql demo](/images/coral-sql-join.gif)
+![coral sql demo](/images/claude-query-example.gif)
```
simonwhitaker: I didn't love the SQL join GIF because it felt a bit like a wall of text and rendering it as a GIF just made it harder to read. This change takes that objection and dials it up to 11! 😆 

I just don't get what the value of this GIF is. It's so dense, I find it hard to parse, beyond a general vibe that "ah yes, that's Claude doing... something".

simonwhitaker (changes requested):

docs/index.mdx
simonwhitaker: At this point in the docs, the narrative we're telling is:

> Coral sits between you and your data sources: you (or your agent) write SQL, and Coral translates it into API calls or file reads, then returns a single query result.

Given that, I think the current example is fine. But I think we should replace the GIF with a plain old screenshot, or simply formatted text. I don't think the animation adds anything useful.

AlbertQM: >I just don't get what the value of this GIF is. It's so dense, I find it hard to parse, beyond a general vibe that "ah yes, that's Claude doing... something".

Fair point! 
"One terminal gif is worth more than a thousand words" imo -- The better looking README usually have a gif at the top, targeting people that hate reading (e.g. https://github.com/anthropics/claude-code#readme), and/or will only engage with a project if the first 5 seconds are interesting enough, so I think we should have a gif that shows concretely what `coral` is.
Agreed that the current wall of text gif isn't good enough, will craft a much neater version and see if that works better

For context: the original motivation to add a recording of `coral` in action was that the README alone was too abstract. what is `coral`? "A single SQL interface that sits between you and your data" means very little to most people I spoke to

AlbertQM: Given there are 2 ways of using coral, I landed on having an image for agent usage (easier to make it look clean), and the old gif for manual sql usage

image

jsummerfield (approved):

simonwhitaker (approved):

PR #214, opened 2026-04-28
fix(engine): validate regex-style query patterns

antonmry:
## Summary
- Add a runtime pattern validator rewrite that rejects `%` and `_` in `SIMILAR TO` and rejects `%` in regex operators with clear guidance to use regex syntax or `LIKE`/`ILIKE`.
- Register the validator when building the DataFusion session so invalid regex-style patterns fail before execution.
- Add unit and end-to-end engine tests covering invalid wildcard usage, valid regex patterns, and unchanged `LIKE` behavior.

Check https://linear.app/withcoral/issue/SOURCE-475/show-an-error-for-unsupported-regex-sql-operators

## Test plan
- [x] Run the `coral-engine` unit tests for `runtime::pattern_validator`.
- [x] Run the `coral-engine` integration tests including `pattern_error_tests`.
- [x] Verify `SIMILAR TO '(Slack|Weekly)%'` and `~ '(Slack|Weekly)%'` return `InvalidInput` errors with the new guidance.
- [x] Verify valid regex-style queries and `LIKE` queries still return the expected matching rows.

image

### DIFF
--- a/crates/coral-engine/src/runtime/pattern_validator.rs
+++ b/crates/coral-engine/src/runtime/pattern_validator.rs
@@ -1,0 +1,267 @@
+//! Query-pattern validation for regex-style operators with common LIKE-wildcard
+//! mistakes.
+
+use std::sync::Arc;
+
+use datafusion::common::config::ConfigOptions;
+use datafusion::common::tree_node::Transformed;
+use datafusion::common::{DFSchema, Result as DataFusionResult, ScalarValue, plan_err};
+use datafusion::execution::FunctionRegistry;
+use datafusion::logical_expr::Operator;
+use datafusion::logical_expr::expr::{BinaryExpr, Expr, Like};
+use datafusion::logical_expr::expr_rewriter::FunctionRewrite;
+
+pub(crate) fn register_pattern_validator(
+    registry: &mut dyn FunctionRegistry,
+) -> DataFusionResult<()> {
+    registry.register_function_rewrite(Arc::new(PatternValidator))?;
+    Ok(())
+}
+
+#[derive(Debug)]
+struct PatternValidator;
+
+impl FunctionRewrite for PatternValidator {
+    fn name(&self) -> &'static str {
+        "PatternValidator"
+    }
+
+    fn rewrite(
+        &self,
+        expr: Expr,
+        _schema: &DFSchema,
+        _config: &ConfigOptions,
+    ) -> DataFusionResult<Transformed<Expr>> {
+        validate_expr(&expr)?;
+        Ok(Transformed::no(expr))
+    }
+}
+
+fn validate_expr(expr: &Expr) -> DataFusionResult<()> {
+    match expr {
+        Expr::SimilarTo(like) => validate_similar_to(like),
+        Expr::BinaryExpr(binary) => validate_regex_binary(binary),
+        _ => Ok(()),
+    }
+}
+
+fn validate_similar_to(like: &Like) -> DataFusionResult<()> {
+    let Some(pattern) = extract_string_literal(&like.pattern) else {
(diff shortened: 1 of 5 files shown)


crates/coral-engine/src/runtime/pattern_validator.rs
```diff
@@ -0,0 +1,267 @@
...
+}
+
+fn validate_similar_to(like: &Like) -> DataFusionResult<()> {
+    let Some(pattern) = extract_string_literal(&like.pattern) else {
+        return Ok(());
+    };
+
+    if contains_like_wildcards(&pattern) {
+        return plan_err!(
+            "SIMILAR TO pattern '{pattern}' contains `%` or `_` which are literal characters in SIMILAR TO, not wildcards. Use `.*` instead of `%`, `.` instead of `_`, or use LIKE for wildcard matching."
```
simonwhitaker, line 55: Is this correct for SIMILAR TO? Assuming SIMILAR TO in DF follows the same conventions as SIMILAR TO in Postgres, you can absolutely have `%` in a SIMLAR TO query:

https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-SIMILARTO-REGEXP

simonwhitaker: I might be looking in the wrong place, but it looks like SIMILAR TO in DF expects a LIKE expression: https://github.com/apache/datafusion/blob/22bb4e6b752c7a62b677d94a63bcf08b68e8d5ec/datafusion/expr/src/expr.rs#L334

simonwhitaker (changes requested):

mystic123 (approved):

simonwhitaker (approved):

crates/coral-engine/src/runtime/pattern_validator.rs
antonmry: Looking at  https://github.com/apache/datafusion/blob/eae7bf4fa1c037c0a065d1f36d0669f5bb97a9cf/datafusion/physical-expr/src/expressions/binary.rs#L970-L983 , it seems that SIMILAR TO is different in DF: it accepts regex, but not Like expressions. We could add that to DF if they are open to it.

This is the result of a script I vibecoded to show it in action:

image

https://gist.github.com/antonmry/fd3d6103d1c99036a91ae923fb5a8d49

crates/coral-engine/tests/engine/pattern_error_tests.rs
```diff
@@ -0,0 +1,153 @@
...
+    .await
+    .expect_err("SIMILAR TO wildcard mismatch should fail");
+
+    let detail = invalid_input_detail(error);
+    assert!(detail.contains("SIMILAR TO pattern '(Slack|Weekly)%'"));
+    assert!(detail.contains("Use `.*` instead of `%`"));
+}
+
+#[tokio::test]
+async fn regex_match_with_like_wildcard_returns_clear_error() {
```
sauldhernandez, line 84: What happens if the query legitimately wants to use `%` as part of it? should we support escaping it so it goes through?

sauldhernandez (approved):

crates/coral-engine/tests/engine/pattern_error_tests.rs
antonmry: The error approach for SIMILAR TO is an interim safety net. The right long-term fix is either upstream in DataFusion or a Coral-level FunctionRewrite that converts the wildcards, making SIMILAR TO work as users expect. Adding escape support on top of the error is complexity for a temporary workaround. I will explore the option to convert the wildcards upstream to see if that might make sense to them

PR #233, opened 2026-04-30
feat(cli): bootstrap local web UI

sauldhernandez:
## Intent
Add a local browser entrypoint for Coral that can serve bundled UI assets from the CLI, with the current UI surface intentionally limited to a placeholder until the real app is wired in.

## What changed
- Adds the `coral ui` command behind the `embedded-ui` feature.
- Serves embedded UI assets from the local Coral server and opens the UI in the browser by default.
- Adds `coral ui --no-open` to start the same embedded UI server without launching a browser; embedded assets are still required and served by Coral.
- Replaces independent server transport/static/bind builder knobs with explicit `ServerMode` choices: native gRPC or embedded UI.
- Removes the separate `coral server` development command path; the UI development script now uses `coral ui --no-open`.
- The UI currently renders a placeholder page: `Coral UI Coming soon`.
- Adds the UI `proto:gen` script and protobuf-ts dependencies so gRPC-Web clients can be generated from the Coral protos during UI development.
- Documents the user-facing `coral ui` command and local bind address behavior.

## Validation
- `make rust-checks`
- `cargo test -p coral-cli --all-features --locked`
- `npm run build --prefix ui`


docs/reference/cli-reference.mdx
```diff
@@ -112,6 +112,22 @@ Run the guided source setup flow.
 coral onboard
 ```
 
+## `coral ui`
```
AlbertQM, line 117: `cargo run -p coral-cli -- --help` doesn't show this command. Is that intended?

crates/coral-cli/src/embedded_ui.rs
```diff
@@ -0,0 +1,31 @@
+//! Embedded Coral UI assets compiled into the binary.
+
+use std::borrow::Cow;
+
+use coral_app::{StaticAsset, StaticAssetsProvider};
+use rust_embed::Embed;
+
+#[derive(Embed)]
+#[folder = "../../ui/dist/"]
+#[allow_missing = true]
```
AlbertQM, line 10: Does this mean that this check passes even if `dist/ui` is missing? I.e. I `coral ui`, and get a 404?

docs/reference/cli-reference.mdx
sauldhernandez: Not intended, just missed it. I'll add it

crates/coral-cli/src/embedded_ui.rs
sauldhernandez: Yes, but you do get a  proper CLI error instead of a 404.  
  
The idea here is to allow to build the CLI without the UI, so devs or CI working on non-UI things are not forced to set it up

docs/reference/cli-reference.mdx
```diff
@@ -112,6 +112,22 @@ Run the guided source setup flow.
...
+
+Start the local gRPC-Web server with the embedded Coral UI bundled into the same port and open it in the browser.
+
+```shellscript
+coral ui
+```
+
+The server listens on `http://127.0.0.1:1457` by default, serving the SPA from `/` and gRPC-Web on the same origin, then opens the loopback URL in the default browser. The command keeps running until you press `Ctrl-C`. This subcommand is only present when the binary was built with the `embedded-ui` Cargo feature, which expects the UI bundle in `ui/dist/` at compile time.
+
+Use `--addr` to bind a different loopback address:
```
AlbertQM: Don't think we're enforcing only loopback address in code? Should we?

sauldhernandez: Good question. I'm more on the side of not enforcing it, but then I probably should change doc wording to align with that.

In a more general sense, having reasonable defaults (i.e. this binds to loopback by default) while allowing the user to override that feels like a good policy.

docs/reference/cli-reference.mdx
sauldhernandez: Double-checked. It does show in the help if you enable the `embedded-ui` feature

ui/.gitignore
```diff
@@ -0,0 +1,5 @@
+node_modules/
+dist/
+src/generated/
```
lazybean, line 3: I think I prefer to have the generated code checked in.  
This will help with TS running in CI and avoiding us breaking things

AlbertQM:
Thanks for putting this together and getting us started!

What was the thinking behind putting gRPC-Web and static asset serving directly in `coral-app` rather than adding a separate `coral-ui` adapter that forwards to coral-client/native gRPC?

When I was looking at this, I thought we could have a UI adapter that owns browser-specific stuff

```
  CLI adapter:
  terminal command -> coral-client -> coral-app

  MCP adapter:
  MCP tool call -> coral-client -> coral-app

  UI adapter:
  browser gRPC-Web -> coral-client -> coral-app
```

sauldhernandez:
@AlbertQM the UI adapter can't use the existing gRPC web server that served as a boundary for the CLI and MCP.

So the browser has to talk gRPC-Web. On which case either:

- You change the existing gRPC surface to be gRPC-Web for this use case (what's in the PR right now)
- You offer a gRPC-Web surface and bridge it to the existing gRPC surface via coral-client (what you state in the comment)

I thought the latter was a bit overkill and offered the former, but I agree that it breaks the clean-cut crate responsibilities we had before it. Bridging gRPC-Web<->gRPC respects the boundaries we have, but now you have two open ports for the same thing when running the UI.

I don't have a strong answer here, we can discuss it.

lazybean:
* **#233** <a href="https://app.graphite.com/github/pr/withcoral/coral/233?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#235](https://github.com/withcoral/coral/pull/235) <a href="https://app.graphite.com/github/pr/withcoral/coral/235?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#275](https://github.com/withcoral/coral/pull/275) <a href="https://app.graphite.com/github/pr/withcoral/coral/275?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/233?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ui/.gitignore
sauldhernandez: I'm not fond of checking in auto generated code. CI can regenerate it with no issues. Any particular problem you've encountered that makes you prefer to check it in?

docs/reference/cli-reference.mdx
sauldhernandez: Updated docs so it's clear we're not enforcing loopback.

ui/.gitignore
sauldhernandez: I'll leave the autogenerated stuff out for now. If we find any hurdles we can discuss it again.

docs/reference/cli-reference.mdx
```diff
@@ -114,6 +114,22 @@ Run the guided source setup flow.
...
+```shellscript
+coral ui
+```
+
+The UI opens at `http://127.0.0.1:1457` by default, which only accepts connections from your machine. The command keeps running until you press `Ctrl-C`.
+
+Use `--addr` to bind a different address:
+
+```shellscript
+coral ui --addr 127.0.0.1:1458
```
simonwhitaker: Same comment as last time; this should accept hostnames as well as IP addresses. (e.g. `localhost:1457`), or you should document clearly here that we've chosen not to support hostnames.

simonwhitaker (approved):

martinrajdl (commented):
@sauldhernandez sorry for lack of context here, but I'm not really understand why the protobuf-ts is needed, is that a requirement to use the grpc-web?

sauldhernandez:
> @sauldhernandez sorry for lack of context here, but I'm not really understand why the protobuf-ts is needed, is that a requirement to use the grpc-web?



Yes, grpc-web relies on protobuf

crates/coral-app/src/bootstrap/server.rs
```diff
@@ -42,24 +79,43 @@ impl Default for ServerConfig {
...
         self
     }
 
-    #[must_use]
+    pub(crate) fn with_bind_addr(mut self, bind_addr: SocketAddr) -> Self {
+        self.bind_addr = bind_addr;
+        self
+    }
+
+    pub(crate) fn with_grpc_web(mut self) -> Self {
```
jsummerfield: Non-blocking follow-up: this builder is still a bag of transport knobs, which lets callers express combinations that are not really product modes (`with_grpc_web()` without UI assets, `with_static_assets()` implicitly changing transport, or a non-loopback bind through `with_bind_addr()`). I would eventually make the mode explicit, e.g.

```rust
enum ServerMode {
    NativeGrpc,
    GrpcWebDev { port: u16 },
    EmbeddedUi { port: u16, assets: Arc<dyn StaticAssetsProvider> },
}
```

or expose named constructors like `native_grpc()`, `grpc_web_loopback(port)`, and `embedded_ui_loopback(port, assets)`. Not a blocker for this PR now that `coral ui` itself is port-only, but this would keep the native CLI/MCP path and browser UI path as distinct contracts instead of independent toggles.

jsummerfield: Related dev-flow alternative: I don’t think this needs a separate `coral server` / `start_dev_server` path. `coral ui --no-open` could handle a missing or empty `ui/dist/` gracefully by starting the same loopback gRPC-Web API, skipping static assets when unavailable, and printing something like `Embedded UI assets were not found; run npm run dev --prefix ui and open http://localhost:5173`.

Then Vite can serve local dev assets and proxy API requests to the same `coral ui --no-open --port 1457` backend. Release/normal `coral ui` can still require embedded assets and open the embedded UI, but dev does not need a second command mode.

jsummerfield (approved):

PR #237, opened 2026-04-30
feat(engine): advertise source-scoped table functions

Bradley-Butcher:
## What changed

Adds runtime catalog metadata for manifest-declared source-scoped table functions. This does not execute functions yet; it only carries the function shape from validated source manifests into `coral.table_functions`.

## Scope

- Add `RegisteredTableFunction` metadata to registered sources
- Populate HTTP source function metadata from manifest `functions`
- Add `coral.table_functions` with `schema_name`, `function_name`, description, arguments JSON, and result columns JSON
- Keep non-HTTP backends explicit with no table functions
- Add catalog coverage for function discovery

## Not in this PR

- DataFusion UDTF registration
- source-scoped function planning
- executing `github.search_issues(...)`
- GitHub/Datadog manifest rollout

## Validation

- `cargo test -p coral-engine --test engine catalog_tests::coral_table_functions_lists_source_functions -- --nocapture`
- `make rust-checks`

### DIFF
--- a/crates/coral-engine/src/backends/common.rs
+++ b/crates/coral-engine/src/backends/common.rs
@@ -7,7 +7,8 @@ use crate::{QueryRuntimeContext, RequestAuthenticator};
 use async_trait::async_trait;
 use coral_spec::backends::file::PartitionColumnSpec;
 use coral_spec::{
-    ColumnSpec, FilterSpec, ManifestDataType, ManifestInputKind, ManifestInputSpec, TableCommon,
+    ColumnSpec, FilterSpec, ManifestDataType, ManifestInputKind, ManifestInputSpec,
+    SourceTableFunctionSpec, TableCommon,
 };
 use datafusion::arrow::datatypes::{DataType, Field, Schema, SchemaRef, TimeUnit};
 use datafusion::datasource::TableProvider;
@@ -34,6 +35,17 @@ pub(crate) struct RegisteredTable {
 }
 
 #[derive(Debug, Clone)]
+pub(crate) struct RegisteredTableFunction {
+    pub(crate) schema_name: String,
+    pub(crate) function_name: String,
+    pub(crate) public_name: String,
+    pub(crate) kind: String,
+    pub(crate) description: String,
+    pub(crate) arguments_json: String,
+    pub(crate) result_columns_json: String,
+}
+
+#[derive(Debug, Clone)]
 pub(crate) struct RegisteredInput {
     pub(crate) key: String,
     pub(crate) kind: ManifestInputKind,
@@ -51,6 +63,7 @@ pub(crate) struct RegisteredInput {
 pub(crate) struct RegisteredSource {
     pub(crate) schema_name: String,
     pub(crate) tables: Vec<RegisteredTable>,
+    pub(crate) table_functions: Vec<RegisteredTableFunction>,
     pub(crate) inputs: Vec<RegisteredInput>,
 }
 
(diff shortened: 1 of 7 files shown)


Bradley-Butcher:
* **#237** <a href="https://app.graphite.com/github/pr/withcoral/coral/237?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/237?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

antonmry (commented):
It's looking great, I've added some minor in-line comments in the PR.

crates/coral-engine/src/runtime/catalog.rs
```diff
@@ -45,6 +50,47 @@ pub(crate) fn register(ctx: &SessionContext, active_sources: &[RegisteredSource]
     Ok(())
 }
 
+fn build_table_functions_table(active_sources: &[RegisteredSource]) -> Result<MemTable> {
+    let schema = Arc::new(Schema::new(vec![
+        Field::new("schema_name", DataType::Utf8, false),
+        Field::new("public_name", DataType::Utf8, false),
```
antonmry: `coral.tables` exposes `(schema_name, table_name, ...)` as separate columns, but `coral.table_functions` only exposes `public_name = "<schema>.<func>"`. Since `RegisteredTableFunction` already carries `function_name`, could we add it as a separate column too? Consumers can filter or group without stripping the schema prefix, and the virtual tables stay symmetric. Keeping `public_name` as a convenience is fine.

crates/coral-engine/tests/engine/catalog_tests.rs
```diff
@@ -310,6 +363,43 @@ fn build_demo_source(variables: &[(&str, &str)], secrets: &[(&str, &str)]) -> Qu
...
+        )
+        .await
+        .expect("table function catalog query should succeed"),
+    );
+
+    assert_eq!(rows.len(), 1);
+    let row = &rows[0];
+    assert_eq!(row["schema_name"], "searchy");
+    assert_eq!(row["public_name"], "searchy.search_issues");
+    assert_eq!(row["kind"], "search");
```
antonmry: This test covers an explicit `kind: search`, but it does not exercise the omitted-kind/default path above. Could we add a second function without a `kind` field and assert it appears as `table`? That would pin both shapes.

crates/coral-engine/src/backends/common.rs
```diff
@@ -183,6 +196,49 @@ pub(crate) fn build_registered_table(
...
+                "description": column.description,
+            })
+        })
+        .collect::<Vec<_>>();
+
+    RegisteredTableFunction {
+        schema_name: schema_name.to_string(),
+        function_name: function.name.clone(),
+        public_name,
+        kind: if function.kind.is_empty() {
```
antonmry: Could we move this default into `coral-spec` instead of the engine catalog metadata builder? `SourceTableFunctionSpec.kind` currently deserializes omitted `kind` to `""`, then this renderer canonicalizes it to `"table"`. Future catalog consumers should be able to rely on a canonical spec value, so this can reduce to `kind: function.kind.clone()` after spec-side defaulting or normalization.

crates/coral-engine/src/backends/common.rs
```diff
@@ -7,7 +7,8 @@ use crate::{QueryRuntimeContext, RequestAuthenticator};
 use async_trait::async_trait;
```
iaptsiauri, line 7: I think we should rename this file accordingly. It should be named `registry` rather then  `common`. skimming through the file it does only registration management.
I would also  separate table and function registries, but  maybe later

crates/coral-engine/src/backends/common.rs
iaptsiauri: Canonical spec should be driver of the spec representation, engine shouldn't be responsible for it imho

crates/coral-engine/src/runtime/catalog.rs
Bradley-Butcher: Fair. This has changed since the comment: `coral.table_functions` now exposes `schema_name` and `function_name` separately, and we dropped `public_name` rather than keeping both.

Current shape is:

schema_name
function_name
description
arguments_json
result_columns_json

So consumers can filter/group symmetrically with `coral.tables` without parsing a dotted name.

crates/coral-engine/tests/engine/catalog_tests.rs
Bradley-Butcher: This is stale now. We removed table-function `kind` entirely, so there is no omitted-kind/default path to test.

`SourceTableFunctionSpec` no longer has `kind`, the schema no longer accepts it under `functions`, and `coral.table_functions` no longer exposes a `kind` column.

crates/coral-engine/src/backends/common.rs
Bradley-Butcher: Agreed with the direction, but this is stale now: we removed `kind` from table functions entirely.

There is no longer an engine-side canonicalization from empty string to `table`; the spec type does not carry `kind`, and the catalog does not expose it. See PR1 for drop reasoning.

crates/coral-engine/src/backends/common.rs
Bradley-Butcher: I'd say this is OOS for this PR, I agree it may want a cleaner split

antonmry (approved):

iaptsiauri (approved):

PR #241, opened 2026-05-01
docs(mcp): clarify query pattern guidance

jsummerfield:
## Summary
- Update the Coral MCP guide template so agents use LIKE/ILIKE for SQL wildcards.
- Clarify that SIMILAR TO expects regex-shaped patterns and that regex operators treat `%` and `_` literally.

## Verification
- cargo test -p coral-mcp guide_content

### DIFF
--- a/crates/coral-mcp/src/guide_template.md
+++ b/crates/coral-mcp/src/guide_template.md
@@ -50,4 +50,6 @@ FROM source.events;
 - Fully qualify tables in SQL, for example `slack.messages`.
 - Check `coral.tables.required_filters` and `coral.columns.is_required_filter` before querying tables that depend on filter-only inputs.
 - Cross-source joins work with standard SQL after source scans complete.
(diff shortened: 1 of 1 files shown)


Bradley-Butcher (approved):
@mystic123 are the benchmarks in a shape we can be happy this does not worsen performance?
 
Lets make sure the skill is updated too. Someone should think how to stop divergence skill-mcp instructions more generally

[Added a linear ticket.](https://linear.app/withcoral/issue/SOURCE-496/develop-method-of-enforcing-parity-between-mcp-skill-instructions)

mystic123 (approved):

mystic123:
> @mystic123 are the benchmarks in a shape we can be happy this does not worsen performance?
> 
> Lets make sure the skill is updated too. Someone should think how to stop divergence skill-mcp instructions more generally
> 
> [Added a linear ticket.](https://linear.app/withcoral/issue/SOURCE-496/develop-method-of-enforcing-parity-between-mcp-skill-instructions)

I don't expect this to influence heavily the perf, but we should see some movement on the daily vakra.

PR #244, opened 2026-05-01
feat(sources): roll out source-scoped table functions

Bradley-Butcher:
## Summary

Rolls out source-scoped table functions in the bundled GitHub and Datadog manifests, and removes the old table surfaces where the function is now the canonical shape.

This PR is the source rollout for the engine work below it in the stack. The engine PRs add the execution/planning support; this PR decides which bundled source endpoints should use it.

## Heuristic

We used this rule:

> If `SELECT * FROM source.thing LIMIT 10` is not meaningful without provider request inputs, and those inputs define the remote result set, model it as a source-scoped table function.

Tables remain the right shape for inventory/discovery collections:

```sql
SELECT * FROM datadog.monitors LIMIT 10;
SELECT * FROM github.user_repos LIMIT 10;
```

Functions are the right shape for provider-native search/query operations:

```sql
SELECT * FROM github.search_issues(q => 'repo:withcoral/coral is:pr is:open');
SELECT * FROM datadog.logs(start => 'now-1h', end => 'now', query => 'service:web');
```

Function arguments map to provider request inputs. `WHERE` still filters the returned rows locally; it is not a second way to pass request args.

## What Changed

GitHub search endpoints are now functions, and the old search tables are removed:

- `github.search_code(q, sort, order)`
- `github.search_commits(q, sort, order)`
- `github.search_issues(q, mode, advanced, sort, order)`
- `github.search_labels(repository_id, q, sort, order)`
- `github.search_repositories(q, sort, order)`
- `github.search_topics(q)`
- `github.search_users(q, sort, order)`

Datadog query/search endpoints are now functions, and the old overlapping tables are removed:

- `datadog.search_monitors(query)`
- `datadog.events(start, end)`
- `datadog.logs(start, end, query)`
- `datadog.spans(start, end, query)`
- `datadog.metrics(query, start, end)`

Kept as tables: inventory/discovery surfaces like `datadog.monitors`, `datadog.incidents`, `datadog.dashboards`, `datadog.hosts`, `datadog.services`, and `datadog.metric_names`.

## GitHub Scope

GitHub has a much larger required-filter surface than the search endpoints. We intentionally limited this PR to `/search/*` because the rest of GitHub is mostly generated endpoint coverage and needs a broader source-model pass before converting cleanly.

In particular, a full GitHub migration should first address:

- function result columns that expose function args, equivalent to table `from_filter` virtual columns
- function request routes, equivalent to table `requests` / `when_filters`
[...]

PR #243, opened 2026-05-01
feat(engine): plan source-scoped table functions

Bradley-Butcher:
## Summary

Adds public source-scoped table-function planning.

After this PR, manifest-declared functions can be called as normal SQL relations:

```sql
SELECT title, score
FROM github.find_issues(q => 'repo:withcoral/coral flaky')
LIMIT 10;
```

The previous PR registers the internal UDTFs. This PR adds the generic planner that recognizes `source.function(...)` and lowers it to those internal functions.

## What Changed

- registers a generic DataFusion `RelationPlanner` when source functions exist
- resolves two-part relation names like `github.find_issues(...)`
- requires named arguments
- normalizes unquoted schema, function, and argument identifiers consistently with DataFusion
- lowers named args into manifest argument order before calling the internal UDTF
- pads omitted optional args with `NULL`, letting the internal UDTF enforce required-arg validation
- returns a loud planning error for unknown functions in a known source schema

## Behavior

The planner only rewrites the table-function relation. It leaves normal DataFusion behavior in charge of projection, aliases, `WHERE`, ordering, and limits.

Function arguments are provider request inputs. `WHERE` predicates filter the returned rows; they are not pushed down as additional provider request args.

SQL `LIMIT` still flows through to execution and bounds HTTP pagination/fetching.

## Not in This PR

- internal UDTF execution substrate; that is the previous PR
- bundled GitHub/Datadog function rollout
- MCP guide updates

## Validation

- `cargo test -p coral-engine --test engine source_scoped_table_function -- --nocapture`
- `make rust-checks`

PR #245, opened 2026-05-01
feat(spec): add source-scoped table function specs

Bradley-Butcher:
## What changed

Adds the source-spec shape for source-scoped table functions. This lets HTTP source manifests declare provider-native operations that return rows, without pretending the operation is a backing table.

A function adds invocation arguments, but still defines the same HTTP execution shape as a table: `request`, `response`, `pagination`, and `columns`.

```yaml
functions:
  - name: search_issues
    args:
      - name: q
        required: true
        bind:
          arg: q
      - name: mode
        values: [lexical, semantic, hybrid]
        bind:
          arg: search_type
    request:
      method: GET
      path: /search/issues
      query:
        - name: q
          from: arg
          key: q
        - name: search_type
          from: arg
          key: search_type
    response:
      rows_path: [items]
    pagination:
      mode: page
      page_size:
        default: 30
        max: 100
        query_param: per_page
    columns:
      - name: title
        type: Utf8
```

## Scope

- Add manifest models for source-scoped table functions
- Add function args with explicit `bind.arg` request-arg mapping
- Add `from: arg`, `from: arg_int`, and `from: arg_bool` request value sources
- Validate function names, duplicate args, enum values, duplicate request-arg bindings, request templates, pagination, and result columns
- Reject table-only filter value sources inside function requests
- Reject function-only arg value sources inside table requests
- Update the source manifest JSON schema
- Document the source-spec shape

## Not in this PR

- Runtime catalog discovery through `coral.table_functions`
- DataFusion UDTF registration
- Executing `source.function(arg => value)` SQL
- GitHub/Datadog manifest rollout
- Semantic GitHub sugar such as `repo => ...`
- Fixed/hidden function request bindings

## Validation

- `cargo test -p coral-spec`
- `make rust-checks`

### DIFF
--- a/crates/coral-spec/src/common.rs
+++ b/crates/coral-spec/src/common.rs
@@ -150,8 +150,6 @@ pub struct FilterSpec {
 pub struct SourceTableFunctionSpec {
     pub name: String,
     #[serde(default)]
-    pub kind: String,
-    #[serde(default)]
     pub description: String,
     #[serde(default)]
     pub fetch_limit_default: Option<usize>,
--- a/crates/coral-spec/src/common.rs
+++ b/crates/coral-spec/src/common.rs
@@ -145,6 +145,45 @@ pub struct FilterSpec {
     pub mode: FilterMode,
 }
 
+/// Declarative source-scoped table-valued function.
+#[derive(Debug, Clone, Deserialize)]
+pub struct SourceTableFunctionSpec {
+    pub name: String,
+    #[serde(default)]
+    pub kind: String,
+    #[serde(default)]
+    pub description: String,
+    #[serde(default)]
+    pub fetch_limit_default: Option<usize>,
+    #[serde(default)]
+    pub args: Vec<TableFunctionArgSpec>,
+    #[serde(default)]
+    pub request: RequestSpec,
+    #[serde(default)]
+    pub response: ResponseSpec,
+    #[serde(default)]
+    pub pagination: PaginationSpec,
+    #[serde(default)]
+    pub columns: Vec<ColumnSpec>,
+}
+
+/// One argument accepted by a source-scoped table-valued function.
+#[derive(Debug, Clone, Deserialize)]
+pub struct TableFunctionArgSpec {
+    pub name: String,
+    #[serde(default)]
+    pub required: bool,
+    #[serde(default)]
+    pub values: Vec<String>,
+    pub bind: FunctionArgBinding,
+}
+
+/// How a table function argument contributes to the provider request.
+#[derive(Debug, Clone, Deserialize, PartialEq, Eq)]
+pub struct FunctionArgBinding {
+    pub arg: String,
+}
+
 /// The base request template for one HTTP table or request route.
 #[derive(Debug, Clone, Deserialize, Serialize, Default)]
 pub struct RequestSpec {
@@ -322,6 +361,21 @@ pub enum ValueSourceSpec {
         key: String,
         separator: String,
         part: usize,
+    },
+    Arg {
+        key: String,
+        #[serde(default)]
+        default: Option<Value>,
+    },
+    ArgInt {
+        key: String,
+        #[serde(default)]
+        default: Option<i64>,
+    },
+    ArgBool {
+        key: String,
+        #[serde(default)]
+        default: Option<bool>,
     },
     Input {
         key: String,
--- a/crates/coral-spec/src/schema/source_manifest.schema.json
+++ b/crates/coral-spec/src/schema/source_manifest.schema.json
@@ -211,7 +211,6 @@
       "required": ["name", "request"],
       "properties": {
         "name": { "type": "string", "minLength": 1 },
-        "kind": { "type": "string", "minLength": 1 },
         "description": { "type": "string" },
         "fetch_limit_default": { "type": "integer", "minimum": 1 },
         "args": {
--- a/crates/coral-spec/src/schema/source_manifest.schema.json
+++ b/crates/coral-spec/src/schema/source_manifest.schema.json
@@ -28,6 +28,10 @@
       "minItems": 1,
       "items": { "$ref": "#/$defs/table" }
     },
+    "functions": {
+      "type": "array",
+      "items": { "$ref": "#/$defs/table_function" }
+    },
     "onboarding": {
       "type": "object",
       "additionalProperties": false,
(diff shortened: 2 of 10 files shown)


Bradley-Butcher:
* **#245** <a href="https://app.graphite.com/github/pr/withcoral/coral/245?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/245?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-spec/src/common.rs
```diff
@@ -145,6 +145,45 @@ pub struct FilterSpec {
...
+    #[serde(default)]
+    pub kind: String,
+    #[serde(default)]
+    pub description: String,
+    #[serde(default)]
+    pub fetch_limit_default: Option<usize>,
+    #[serde(default)]
+    pub args: Vec<TableFunctionArgSpec>,
+    #[serde(default)]
+    pub request: RequestSpec,
```
Bradley-Butcher, line 159: duplicate functionality until a capability primitive is implemented

crates/coral-spec/src/schema/source_manifest.schema.json
```diff
@@ -371,6 +419,29 @@
           },
           "required": ["key"]
         },
+        {
+          "properties": {
+            "from": { "const": "arg" },
```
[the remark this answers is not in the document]
Bradley-Butcher: Thanks bot. This is valid at the JSON Schema layer: `arg` / `arg_int` / `arg_bool` are currently part of the shared request value-source schema, so schema-only validation can accept them in table requests.

Coral’s actual manifest validation still rejects this. Table requests go through `validate_value_source`, which errors on function arg sources with “uses function argument ... outside a function request”. Function requests go through separate validation that allows `arg*` and rejects table `filter*` sources.

Alternative considered: splitting request/value-source types into table-scoped and function-scoped variants, e.g. generic `RequestSpec<V>` with `TableValueSourceSpec` and `FunctionValueSourceSpec`. That would make the distinction unrepresentable in Rust and let the JSON Schema mirror it with separate table/function request defs.

I’m not doing that in this PR because the capability/source-spec model is about to be refactored, and that split would be a broader mechanical change across spec parsing, HTTP request resolution, and the later execution PRs. For now this stays as a shared wire shape with Coral-level semantic validation enforcing the context rules.

docs/reference/source-spec-reference.mdx
```diff
@@ -273,6 +273,76 @@ In this minimal example:
...
+    columns:
+      - name: title
+        type: Utf8
+      - name: html_url
+        type: Utf8
+      - name: score
+        type: Float64
+```
+
+The function is queried as:
```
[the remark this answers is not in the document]
Bradley-Butcher: fair - removed

crates/coral-spec/src/schema/source_manifest.schema.json
```diff
@@ -25,9 +25,12 @@
     "rate_limit": { "$ref": "#/$defs/rate_limit" },
     "tables": {
       "type": "array",
-      "minItems": 1,
       "items": { "$ref": "#/$defs/table" }
     },
+    "functions": {
```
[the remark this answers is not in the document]
Bradley-Butcher: Simplified for now - removed the minItems change s.t. we always require a table. If we encounter a source later with only table-function shaped things we can deal with it then.

simonwhitaker (approved):
Docs look good 👍

crates/coral-spec/src/validate.rs
```diff
@@ -51,6 +51,82 @@ pub(crate) fn validate_http_table(
     pagination.validate(schema, table_name)
 }
 
+pub(crate) fn validate_http_function_names(
```
iaptsiauri, line 70: function names are only checked against other functions not table names. If table functions share the same SQL namespace as tables, a manifest can declare both a table and a function with the same name. 
Currently the whole validation is outside scope of scope of this PR but I think we should have a chat to setup a desired state for it. As this might get quite messy sooner then later

crates/coral-spec/src/validate.rs
```diff
@@ -51,6 +51,82 @@ pub(crate) fn validate_http_table(
...
+                "source '{source_name}' function '{}' is declared more than once",
+                function.name
+            )));
+        }
+    }
+
+    Ok(())
+}
+
+pub(crate) fn validate_http_function(
```
iaptsiauri, line 103: function column expressions are not validated, 

`let known_filters = validate_filters_and_column_exprs(filters, columns, schema, table_name)?; ` 

This is a code snippet from table validation but I think it can be reused for the function as well.

docs/reference/source-spec-reference.mdx
```diff
@@ -414,13 +473,25 @@ Request query params, body fields, and headers support these `from` values:
...
+| `arg_bool` | Read a source function argument captured from the function call and serialize it as a JSON boolean |
 | `input` | Read a manifest-declared source input (variable/secret) by key |
 | `state` | Read pagination/runtime state |
 | `now_epoch_minus_seconds` | Emit the current Unix epoch seconds minus the configured offset |
 
+`filter*` and `arg*` differ by where the value comes from. Table requests use
+`filter`, `filter_int`, and `filter_bool` because their request values come from
+SQL predicates such as `WHERE id = 123`. Source function requests use `arg`,
+`arg_int`, and `arg_bool` because their request values come from named function
+arguments such as `github.search_issues(query => 'flaky')`.
```
iaptsiauri: nit: doc manifest defines function arg as `q` but example uses `query`

crates/coral-spec/src/validate.rs
Bradley-Butcher: Function names are now checked against table names as well as other functions, since they share the source SQL namespace.

crates/coral-spec/src/validate.rs
Bradley-Butcher: Yep, valid. Fixed by running function columns through the same expression validation path tables use.

docs/reference/source-spec-reference.mdx
Bradley-Butcher: Fixed. The example now uses q => 'flaky' to match the declared function arg. This was an artefact of a refactor I'd performed earlier. Sorry about that.

sauldhernandez:
Could I get a bit more context about this change? Why is it being proposed?

Bradley-Butcher:
Two relevant RFCs:  
https://www.notion.so/phoebeai/Source-Scoped-Table-Functions-351e4c1e4f5380c68fc9f4e1c56f9302?source=copy_link  
https://www.notion.so/phoebeai/RFC-Search-Retrieval-Semantics-for-Coral-Source-Specs-351e4c1e4f538157a9c2db2d1dc890b7?source=copy_link

crates/coral-spec/src/common.rs
```diff
@@ -145,6 +145,45 @@ pub struct FilterSpec {
     pub mode: FilterMode,
 }
 
+/// Declarative source-scoped table-valued function.
+#[derive(Debug, Clone, Deserialize)]
+pub struct SourceTableFunctionSpec {
+    pub name: String,
+    #[serde(default)]
```
antonmry, line 152: Could we make the default/canonical value for `kind` live in `coral-spec` here? Right now the schema rejects an explicit empty `kind`, but this `#[serde(default)] String` still turns an omitted field into `""`, so validated manifests can carry a non-canonical empty kind. Downstream consumers then need to know to treat empty as `table`. A spec-side default such as `#[serde(default = "default_table_function_kind")]` (or making the field required/validated) would keep the catalog/runtime surface canonical.

Bradley-Butcher: Agreed with the concern, but this is stale now: I removed `kind` from table functions entirely. This was an early discoverability optimisation that is additional complexity, but no signal it's needed yet.

`SourceTableFunctionSpec` no longer has a `kind` field, and the JSON Schema no longer accepts one under `functions`. So there’s no omitted-field `""` case and no downstream “empty means table” convention anymore.

antonmry (approved):

PR #246, opened 2026-05-01
docs: add Coral source model design spec

simonwhitaker:
## Summary

Adds `specs/coral-source-model-spec.md`, a design spec for a capability-centric source model that sits between provider specs (OpenAPI, GraphQL) and agent-facing surfaces (SQL, tools, codemode). The spec defines the layered model (sources, specs, capabilities, entities, bindings, type registry), describes how multiple specs per provider are reconciled, and lays out provenance, aliasing, support, projection hints, file layout, and lint rules.

## Test plan

- [ ] Skim spec for clarity and gather feedback from the team

### DIFF
--- a/specs/coral-source-model-spec.md
+++ b/specs/coral-source-model-spec.md
@@ -1,0 +1,406 @@
+# Coral source model — design spec
+
+A capability-centric model for Coral sources, sitting between provider specs (OpenAPI, GraphQL) and agent-facing surfaces (SQL, tools, codemode).
+
+## Pre-read
+
+James’s [Manifest-o](https://www.notion.so/phoebeai/Manifest-o-350e4c1e4f5380a7bea2f4e536940c43) describes the problems with our current source spec manifests that this spec is attempting to solve.
+
+## Goals
+
+- Bootstrap sources from standards (OpenAPI, GraphQL) instead of hand-authoring.
+- Keep provider truth, Coral's interpretation, and the agent-facing surface as separate concerns rather than collapsing them into one manifest.
+- Support multiple specs per provider (e.g. REST v3 + REST v4 + GraphQL).
+- Project the same internal model into multiple agent surfaces.
+- Make every Coral decision traceable back to its source evidence.
+
+## Out of scope
+
+- Webhooks and other server-initiated events. The model below covers reads, writes, and actions; async event delivery (push, fan-out, dead-lettering, signature verification) is a separate concern.
+
+## Layered model
+
+```
+┌─────────────────────────────────────────────────────────────┐
+│ Source: github                                              │
+│                                                             │
+│   Specs (snapshotted): rest_v3, rest_v4_preview, graphql    │
+│                                                             │
+│   Capabilities ──────────────► Entities                     │
+│   ├─ inputs/output                ├─ fields                 │
+│   ├─ kind, side_effects           ├─ aliases                │
+│   ├─ idempotency                  ├─ related                │
+│   ├─ aliases, provenance          └─ primitive_mappings ─┐  │
+│   └─ bindings ──► (refer up to specs)                    │  │
+│                                                          │  │
+└──────────────────────────────────────────────────────────┼──┘
+              │                                            │
+              ▼                                            ▼
+       Projections                                  Type registry
+       (SQL · tools · codemode)                     (global, additive)
+```
+
+The source is the unit. Specs sit inside it as snapshotted inputs. Capabilities and entities are siblings, not nested. Bindings live inside capabilities and reference specs. The two outbound edges from the model are projections (consume capabilities) and the type registry (referenced by entities for cross-provider correlation).
+
+## Source
+
+Top-level container for one provider. Carries metadata, the snapshotted specs it was compiled from, and the lifecycle policy.
+
+```yaml
+source:
+  id: github
+  namespace: github
+  specs:
+    - id: rest_v3
+      kind: openapi
+      url: …/api.github.com.yaml
+      version_hash: sha256:8f3a…
+      status: stable
+      preferred_for: [reads, writes]
+    - id: rest_v4_preview
+      kind: openapi
+      url: …/api.github.com.v4-preview.yaml
+      version_hash: sha256:c12d…
+      status: incomplete
+      preferred_for: []
+    - id: graphql
+      kind: graphql
+      url: …/schema.graphql
+      version_hash: sha256:a771…
+      status: stable
+      preferred_for: [field_selection, graph_traversal]
+  lifecycle:
+    strategy: replace_not_migrate
+  rate_limits:
+    classes:
+      core: { limit: 5000, window: 1h, applies_to: [rest_v3, rest_v4_preview] }
+      search: { limit: 30, window: 1m, applies_to: [rest_v3] }
+      graphql:
+        { limit: 5000, window: 1h, currency: points, applies_to: [graphql] }
+```
+
+### Lifecycle
+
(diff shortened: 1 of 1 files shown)


specs/coral-source-model-spec.md
```diff
@@ -0,0 +1,420 @@
...
+
+`scope` exists to keep cross-provider terminology bridges from polluting global resolution. The `list_merge_requests` example lives on `github.pull_request.list` so a user fluent in GitLab vocabulary can still hit the right capability when they've chosen GitHub. Without scope, that alias would also be a candidate when the agent is working against a connected GitLab source — competing with the native `gitlab.merge_request.list`. Scoping to `github.*` constrains the alias to fire only when resolution is already inside the GitHub namespace, so:
+
+- In a GitHub-only context, the alias is helpful.
+- In a GitLab-only or mixed context, the alias is silent and the native term wins.
+- Authors can attach cross-provider aliases to the providers they bridge to without making those aliases globally ambient.
+
+The same mechanism applies to internal disambiguation: a high-confidence short alias (`pr`) might be safe inside `github.*` but ambiguous globally.
+
+Aliases grow over time. New ones can be learned from agent usage; low-confidence ones can be demoted or removed when they cause collisions.
```
AlbertQM, line 307: > Aliases grow over time.

How do you imagine this happen? Ie what's the lifecycle of a source spec?

specs/coral-source-model-spec.md
```diff
@@ -0,0 +1,420 @@
...
+- `supported` — usable
+- `unsupported_in_runtime` — capability exists in provider, Coral can't execute it yet
+- `deprecated` — provider has marked deprecated
+- `blocked` — policy or governance prevents use
+
+Plus `support_reason` — a string explaining why. Unsupported things stay visible in the model (so agents and humans can see they exist) but are not materialised into projections.
+
+## Projection hints, not decisions
+
+Capabilities and bindings can express preferences:
```
AlbertQM, line 345: Can we use heuristics to determine the preference?
Would this actually help with the projection? (i.e. do we need it now?)

specs/coral-source-model-spec.md
```diff
@@ -0,0 +1,420 @@
...
+projection_hints:
+  sql: table_function # owner/repo are required path params
+  tool: list_github_pull_requests
+```
+
+These are advisory. The projection layer is free to override based on its own rules, the agent's request, or org policy. The capability stays descriptive; projection stays prescriptive.
+
+## Policy anchors, not policy
+
+Destructive capabilities expose hooks for the policy engine:
```
AlbertQM, line 357: Do we need this now, given there's no policy engine yet?

simonwhitaker: Good question. We don't need it today, I think it's useful that it's included here so I can imagine how it will look as and when we do have a "policy engine".

specs/coral-source-model-spec.md
simonwhitaker: I'm not sure we need these hints now, or indeed at all. But if we do have them, we will want some heuristics for sure. e.g. for SQL, use table functions instead of tables if (...insert conditions here).

(The conditions here might be along the lines of "if the capability requires input parameters that don't naturally belong in the output data", something like that.)

specs/coral-source-model-spec.md
simonwhitaker: It's a good question. The spec says earlier on that we're doing strictly "replace, not migrate", i.e. we don't attempt to version specs, we generate a new one any time the backing source (e.g. the third party API) changes. I'm not sure how we reconcile that with "aliases grow over time... learned from agent usage". I could see that working if e.g. we were gathering telemetry from users and using the things their agents tried to get coral to do to guide future iterations of a spec. But a) we're not doing that, and b) I think I'd want something more user-specific too.

In short, I think we should drop this idea of aliases growing over time (and maybe the idea of aliases, completely) for now.

specs/coral-source-model-spec.md
```diff
@@ -0,0 +1,420 @@
...
+
+## Pre-read
+
+James’s [Manifest-o](https://www.notion.so/phoebeai/Manifest-o-350e4c1e4f5380a7bea2f4e536940c43) describes the problems with our current source spec manifests that this spec is attempting to solve.
+
+## Goals
+
+- Bootstrap sources from standards (OpenAPI, GraphQL) instead of hand-authoring.
+- Keep provider truth, Coral's interpretation, and the agent-facing surface as separate concerns rather than collapsing them into one manifest.
+- Support multiple specs per provider (e.g. REST v3 + REST v4 + GraphQL).
```
jsummerfield, line 13: This is worth discussing. What problem(s) are you trying to solve here?

One UX impact is >1 auth mechanism per provider, e.g. Slack API + token and Slack MCP + oauth.

It also complicates the downstream planner now that there are >1 way to do something. Not nec a bad thing, but do we need this complexity now?

specs/coral-source-model-spec.md
```diff
@@ -0,0 +1,420 @@
...
+  auth:
+    schemes: [oauth, pat, github_app]
+    scopes_any_of: [repo, public_repo]
+  support: supported
+  aliases:
+    - { name: list_prs, source: heuristic, confidence: 0.9 }
+    - { name: list_pull_requests, source: openapi_tag, confidence: 1.0 }
+  projection_hints:
+    sql: table_function
+    tool: list_github_pull_requests
```
jsummerfield, line 138: This feels like the wrong layer. It embeds concrete SQL/tool surface choices in the semantic model. Even if advisory, this makes projection tuning part of the intermediate model instead of a separate model-to-projection artifact.

specs/coral-source-model-spec.md
```diff
@@ -0,0 +1,420 @@
...
+- `auth` requirements (logical scopes; binding may translate to spec-specific scheme)
+- `support` (supported, unsupported_in_runtime, deprecated, blocked) with `support_reason`
+- `policy_anchors` (risk_level, confirmation, audit) — hooks for the policy engine, not policy itself
+- `aliases` — multiple names with provenance and confidence (see "Aliases")
+- `projection_hints` — descriptive hints to projections, not prescriptive decisions
+- `provenance` — primary source the capability was derived from
+
+### What does NOT live at the capability level
+
+- **Pagination** — pagination is wire-shaped. Cursors minted by one binding are meaningless to another, so there's no cross-binding contract worth preserving. Lives entirely in the binding.
```
jsummerfield, line 161: I don't think it can live entirely in the binding. Facts like bounded/list/streaming/pageable matter to SQL limits, progressive discovery, and projection quality for example. We'll need a way to express this at the model level.

specs/coral-source-model-spec.md
simonwhitaker: Couple of concrete examples where we need this:

1. Datadog (v1 vs v2) — Bidirectional split. v1 still owns dashboards, monitors, screenboards, synthetics, and the original metrics submission endpoint. v2 introduced and exclusively hosts Incidents, Service Definitions / Software Catalog, RUM applications, Logs Archives, Restriction Policies, Cloud Cost Management, and Sensitive Data Scanner. You routinely end up using both in the same client.
2. GitHub (REST vs GraphQL) — GraphQL-only: Projects v2 (the rebuilt projects with custom fields), Discussions queries, sponsors data, and efficient nested traversals over PRs/issues/reviews. REST-only (or REST-much-better): Actions workflow admin, traffic statistics, most Git data plumbing (creating refs, blobs, trees), audit log streaming, and a fair chunk of enterprise/admin endpoints.

specs/coral-source-model-spec.md
simonwhitaker: Agreed, I'll remove it.

PR #248, opened 2026-05-02
feat(mcp): add opt-in feedback tool

jsummerfield:
## Intent

Agents can now report when they are blocked without making feedback collection part of the default MCP surface. The feedback path stays app-owned for persistence while MCP exposes it only when the user explicitly opts in.

## Changes

- Adds a `FeedbackService` transport contract and app-owned JSONL persistence for blocked-agent reports.
- Exposes a `feedback` MCP tool behind `coral mcp-stdio --enable-feedback`.
- Keeps default MCP tools limited to `sql` and `list_tables`.
- Updates client wiring, CLI/MCP tests, and docs for the optional feedback tool.

## Verification

- `cargo test -p coral-app feedback`
- `cargo test -p coral-mcp`
- `cargo test -p coral-cli --features cli-test-server mcp_stdio`

### DIFF
--- a/crates/coral-mcp/src/surface/tools.rs
+++ b/crates/coral-mcp/src/surface/tools.rs
@@ -54,28 +54,28 @@ pub(crate) fn list_tables_tool(tables: &[Table]) -> Tool {
 pub(crate) fn feedback_tool() -> Tool {
     Tool::new(
         "feedback",
-        "Submit feedback when the agent is blocked or stuck in an unproductive pattern.",
+        "Submit feedback when you are blocked or stuck in an unproductive pattern",
         json_object_schema(&json!({
             "type": "object",
             "required": ["trying_to_do", "tried", "stuck"],
             "properties": {
                 "trying_to_do": {
                     "type": "string",
-                    "description": "What the agent was attempting to do."
+                    "description": "What you were trying to do."
                 },
                 "tried": {
                     "type": "string",
-                    "description": "What the agent already tried."
+                    "description": "What you already tried."
                 },
                 "stuck": {
                     "type": "string",
-                    "description": "Where the agent got blocked."
+                    "description": "Where you got blocked."
                 }
             }
         })),
     )
     .with_annotations(
-        ToolAnnotations::with_title("Submit Feedback")
+        ToolAnnotations::with_title("Store Feedback Report")
             .read_only(false)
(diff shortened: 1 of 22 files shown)


Saurabhtbj1201 (approved):

docs/guides/use-coral-over-mcp.mdx
```diff
@@ -16,15 +16,17 @@ Before setting up a client, make sure you have [connected at least one source](/
 | -------- | ---------------- | ------------------------------------------------------- |
 | Tool     | `sql`            | Run a SQL query                                         |
 | Tool     | `list_tables`    | List queryable fully qualified tables                   |
+| Tool     | `feedback`       | Submit a blocked-agent feedback report when enabled     |
```
simonwhitaker: nit: "Submit" suggests to me that this posts the report somewhere remote (and that's also what I would expect such a feature to do). So it might be clearer to refer to this feature as "storing" a feedback report.

(Longer term, we should definitely make it easy for people to share those reports with us.)

simonwhitaker (approved):

docs/guides/use-coral-over-mcp.mdx
mystic123: I agree, how about making it more verbose: "generate_feedback_report" ?

crates/coral-mcp/src/surface/tools.rs
```diff
@@ -51,6 +51,38 @@ pub(crate) fn list_tables_tool(tables: &[Table]) -> Tool {
     )
 }
 
+pub(crate) fn feedback_tool() -> Tool {
+    Tool::new(
+        "feedback",
+        "Submit feedback when the agent is blocked or stuck in an unproductive pattern.",
```
Bradley-Butcher: The wording here is suboptimal - the agent is the one performing the act. May be better oriented as "What you were trying to do" - same goes for arg descriptions below.

Bradley-Butcher (approved):
MCP looks good - minor comment on wording

sauldhernandez (approved):

jsummerfield:
### Merge activity

* **May 5, 3:27 PM UTC**: @jsummerfield merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/248).

PR #265, opened 2026-05-04
feat(mcp): paginate list_tables discovery

mystic123:
## Summary

Implements PR 1 from the progressive MCP discovery plan: bounded API-backed MCP `list_tables`.

- Adds shared `PaginationRequest` and `PaginationResponse` protobuf messages using unsigned pagination fields, and uses them from `ListTablesRequest` / `ListTablesResponse`.
- Preserves legacy gRPC behavior when callers omit pagination: unbounded table list with columns.
- Routes MCP `list_tables` through bounded query API calls with default `limit: 50`, max `200`, and summary rows only.
- Keeps `coral://tables` as the explicit all-table summary resource while avoiding column serialization.
- Updates MCP docs and tests for page envelopes and validation errors.

## Before / After Output

Both versions were run against the same connected GitHub catalog:

```json
[{"schema_name":"github","table_count":369}]
[{"column_count":424}]
```

### `tools/list`

Before:

```json
{
  "name": "list_tables",
  "inputSchema": {
    "type": "object",
    "properties": {}
  }
}
```

After:

```json
{
  "name": "list_tables",
  "inputSchema": {
    "type": "object",
    "properties": {
      "schema": {"type": "string"},
      "limit": {"type": "integer"},
      "offset": {"type": "integer"}
    }
  }
}
```

### `list_tables {}`

Before returned all 369 tables and no pagination metadata:

```json
{
  "tables": [
    {
      "name": "github.accepted_assignments",
      "description": "List accepted assignments for an assignment",
      "required_filters": ["assignment_id"]
    },
    {
      "name": "github.access",
      "description": "Get the level of access for workflows outside of the repository",
      "required_filters": ["owner", "repo"]
    }
  ]
}
```

Actual payload: `369` tables, about `65,926` text chars.

After returns the first page only:

```json
{
  "tables": [
    {
      "schema_name": "github",
      "table_name": "accepted_assignments",
      "name": "github.accepted_assignments",
      "description": "List accepted assignments for an assignment",
      "required_filters": ["assignment_id"]
    }
  ],
  "total": 369,
  "limit": 50,
  "offset": 0,
  "has_more": true,
  "next_offset": 50
}
```

Actual payload: `50` tables, about `12,046` text chars.

### `list_tables {"schema":"github","limit":50,"offset":50}`

Before ignored args and returned the full first dump again:

```json
{
  "tables_returned": 369,
  "first": "github.accepted_assignments"
}
```

After returns page 2:

```json
{
  "tables_returned": 50,
  "total": 369,
  "limit": 50,
  "offset": 50,
  "has_more": true,
  "next_offset": 100,
  "first": "github.content_exclusion"
}
```

### Invalid Limit

Before `{"limit":0}` was ignored and returned all `369` tables.

After:

```json
{
  "error": "argument 'limit' must be between 1 and 200"
}
```

## Validation

- `cargo test -p coral-api`
- `cargo test -p coral-app --test grpc -- list_tables`
- `cargo test -p coral-mcp`
- `cargo test -p coral-engine --test engine catalog_tests`
- `cargo build`
- `make rust-checks`

Compiled-app validation against local connected sources:

- `target/debug/coral source list` showed `crag` imported and `github` bundled.
- `target/debug/coral sql --format json "SELECT schema_name, COUNT(*) AS table_count FROM coral.tables GROUP BY schema_name ORDER BY schema_name"` returned `github = 369` tables.
- `target/debug/coral sql --format json "SELECT COUNT(*) AS column_count FROM coral.columns WHERE schema_name = 'github' AND table_name = 'pulls'"` returned `424` columns.
- `target/debug/coral sql --format json "SELECT id, number, state, title FROM github.pulls WHERE owner = 'withcoral' AND repo = 'coral' LIMIT 1"` returned a live GitHub pull row.
- Drove `target/debug/coral mcp-stdio` directly over JSON-RPC: `tools/list` returned `sql` and `list_tables`; `list_tables {}` returned 50 of 369 with `next_offset: 50`; `list_tables {"schema":"github","limit":50,"offset":50}` returned page 2; unknown schema returned an empty page; `limit: 0` returned an MCP invalid-params error.

### DIFF
--- a/crates/coral-api/proto/coral/v1/query.proto
+++ b/crates/coral-api/proto/coral/v1/query.proto
@@ -21,16 +21,48 @@ message ExecuteSqlResponse {
   int64 row_count = 2;
 }
 
+// Page request parameters for list-style APIs.
+message PaginationRequest {
+  // Maximum number of items to return. Zero means unbounded for compatibility.
+  uint32 limit = 1;
+  // Number of matching items to skip before returning results.
+  uint32 offset = 2;
+}
+
+// Page metadata returned by list-style APIs.
+message PaginationResponse {
+  // Matching item count before pagination.
+  uint32 total_count = 1;
+  // Applied result limit. Zero means unbounded.
+  uint32 limit = 2;
+  // Applied result offset.
+  uint32 offset = 3;
+  // Whether another page is available.
+  bool has_more = 4;
+  // Offset to request for the next page when `has_more` is true.
+  uint32 next_offset = 5;
+}
+
 // Lists all query-visible tables in one workspace.
 message ListTablesRequest {
   // Workspace whose current query-visible tables should be listed.
   Workspace workspace = 1;
+  // Optional exact schema/source name.
+  string schema_name = 2;
+  // Optional page request. Missing or zero limit keeps the legacy unbounded behavior.
+  PaginationRequest pagination = 3;
+  // Whether table columns should be omitted from the response.
+  bool omit_columns = 4;
 }
 
 // Returns the query-visible tables in one workspace.
 message ListTablesResponse {
-  // Query-visible tables for the workspace.
+  // Full query-visible tables for legacy callers.
   repeated Table tables = 1;
+  // Page metadata for the table list.
+  PaginationResponse pagination = 2;
+  // Query-visible table summaries for callers that do not need columns.
+  repeated TableSummary table_summaries = 3;
 }
 
 // SQL query execution APIs.
--- a/crates/coral-app/src/query/service.rs
+++ b/crates/coral-app/src/query/service.rs
@@ -4,13 +4,17 @@ use arrow::datatypes::SchemaRef;
 use arrow::ipc::writer::StreamWriter;
 use arrow::record_batch::RecordBatch;
 use coral_api::v1::query_service_server::QueryService as QueryServiceApi;
-use coral_api::v1::{ExecuteSqlRequest, ExecuteSqlResponse, ListTablesRequest, ListTablesResponse};
+use coral_api::v1::{
+    ExecuteSqlRequest, ExecuteSqlResponse, ListTablesRequest, ListTablesResponse,
+    PaginationResponse,
+};
 use tonic::{Request, Response, Status};
 
 use crate::bootstrap::core_status;
 use crate::query::manager::QueryManager;
 use crate::transport::{
-    grpc_span, instrument_grpc, query_status, table_to_proto, workspace_name_from_proto,
+    grpc_span, instrument_grpc, query_status, table_summary_to_proto, table_to_proto,
+    workspace_name_from_proto,
 };
 
 #[derive(Clone)]
@@ -36,15 +40,54 @@ impl QueryServiceApi for QueryService {
         let queries = self.queries.clone();
         instrument_grpc(span, async move {
             let request = request.into_inner();
+            let pagination = request.pagination.unwrap_or_default();
             let workspace_name = workspace_name_from_proto(request.workspace.as_ref())?;
+            let schema_name = request.schema_name.trim();
+            let schema_name = if schema_name.is_empty() {
+                None
+            } else {
+                Some(schema_name)
+            };
             let tables = queries
-                .list_tables(&workspace_name)
+                .list_tables(&workspace_name, schema_name)
                 .await
-                .map_err(query_status)?
-                .into_iter()
-                .map(|table| table_to_proto(&workspace_name, table))
-                .collect();
-            Ok(Response::new(ListTablesResponse { tables }))
+                .map_err(query_status)?;
+            let total = tables.len();
+            let offset = pagination.offset as usize;
+            let limit = pagination.limit as usize;
+            let page = paginate_tables(tables, offset, limit);
+            let returned_count = page.len();
+            let has_more = pagination.limit != 0 && offset.saturating_add(returned_count) < total;
+            let (tables, table_summaries) = if request.omit_columns {
+                (
+                    Vec::new(),
+                    page.into_iter()
+                        .map(|table| table_summary_to_proto(&workspace_name, table))
+                        .collect(),
+                )
+            } else {
+                (
+                    page.into_iter()
+                        .map(|table| table_to_proto(&workspace_name, table))
+                        .collect(),
+                    Vec::new(),
+                )
+            };
+            Ok(Response::new(ListTablesResponse {
+                tables,
+                table_summaries,
+                pagination: Some(PaginationResponse {
+                    total_count: count_to_u32(total),
+                    limit: pagination.limit,
+                    offset: pagination.offset,
+                    has_more,
+                    next_offset: if has_more {
+                        count_to_u32(offset.saturating_add(returned_count))
+                    } else {
+                        0
+                    },
+                }),
+            }))
         })
         .await
     }
@@ -75,6 +118,23 @@ impl QueryServiceApi for QueryService {
         })
         .await
     }
+}
+
+fn paginate_tables(
+    tables: Vec<coral_engine::TableInfo>,
+    offset: usize,
+    limit: usize,
+) -> Vec<coral_engine::TableInfo> {
+    let iter = tables.into_iter().skip(offset);
+    if limit == 0 {
+        iter.collect()
+    } else {
+        iter.take(limit).collect()
+    }
+}
+
+fn count_to_u32(count: usize) -> u32 {
+    u32::try_from(count).unwrap_or(u32::MAX)
 }
 
 fn encode_arrow_ipc_stream(
--- a/crates/coral-app/src/transport.rs
+++ b/crates/coral-app/src/transport.rs
@@ -3,7 +3,7 @@
 use std::future::Future;
 
 use coral_api::v1::{
-    Column, QueryTestFailure, QueryTestResult, QueryTestSuccess, Source, Table,
+    Column, QueryTestFailure, QueryTestResult, QueryTestSuccess, Source, Table, TableSummary,
     ValidateSourceResponse, Workspace, query_test_result,
 };
 use opentelemetry::propagation::Extractor;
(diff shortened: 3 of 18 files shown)


crates/coral-api/proto/coral/v1/query.proto
```diff
@@ -21,16 +21,46 @@ message ExecuteSqlResponse {
...
+  uint32 next_offset = 5;
+}
+
 // Lists all query-visible tables in one workspace.
 message ListTablesRequest {
   // Workspace whose current query-visible tables should be listed.
   Workspace workspace = 1;
+  // Optional exact schema/source filter.
+  string schema_filter = 2;
+  // Optional page request. Missing or zero limit keeps the legacy unbounded behavior.
```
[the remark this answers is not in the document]
mystic123: Thanks. I think the intended contract split here is:

- MCP discovery is default-bounded. The MCP `list_tables {}` path always sends `PaginationRequest { limit: 50, offset: 0 }`, and it rejects `limit: 0` as invalid params before a protobuf request is constructed.
- The lower-level gRPC API keeps missing pagination / `limit=0` as legacy unbounded behavior for typed callers that already use `ListTablesRequest { workspace }` and expect the previous full response.

So empty discovery calls through MCP do return a bounded first page; only legacy gRPC callers retain the old unbounded behavior.

crates/coral-api/proto/coral/v1/query.proto
```diff
@@ -21,16 +21,46 @@ message ExecuteSqlResponse {
...
+// Page request parameters for list-style APIs.
+message PaginationRequest {
+  // Maximum number of items to return. Zero means unbounded for compatibility.
+  uint32 limit = 1;
+  // Number of matching items to skip before returning results.
+  uint32 offset = 2;
+}
+
+// Page metadata returned by list-style APIs.
+message PaginationResponse {
```
sauldhernandez, line 33: Is the intention to make this message reusable in the future? or just for the table listing?

simonwhitaker (approved):
Docs look good

sauldhernandez (commented):
Did tests around this corroborate that discovery for agents improved?

crates/coral-api/proto/coral/v1/query.proto
mystic123: It can be reusable, but I’m intentionally keeping the semantics simple here.

For table discovery I don’t expect more than ~1-2k results, so `limit`/`offset` is good enough and keeps the local MCP/API path straightforward. If we were paginating real table rows or a much larger/changing result set, I’d expect a different contract, likely cursor-based pagination or provider-native pagination tokens.

So the message is reusable for small list-style metadata APIs, but I wouldn’t treat it as the general pagination abstraction for high-volume data rows.

mystic123:
> Did tests around this corroborate that discovery for agents improved?

The issue we observed was mostly mechanical: the default discovery output was large enough to exceed the MCP call output limit. It got saved to a file, but agents tended not to recover cleanly from that and instead tried to brute-force around discovery.

This PR fixes the first part by making `list_tables {}` bounded by default. In the local smoke, current `main` returned all 369 GitHub tables in a ~66KB payload with no pagination metadata; this branch returns 50 summaries in ~12KB with `has_more` /`next_offset`, and supports `schema`, `limit`, and `offset`.

I wouldn’t claim this proves end-to-end task success improved by itself. The intended improvement is that the MCP discovery primitive is now bounded and resumable. Once this and the follow-up PRs land, agents will also have more focused discovery tools, and the skill/CLI guidance will tell them to use `limit`/`offset` when doing direct SQL metadata discovery.

sauldhernandez (approved):

mystic123:
* **#283** <a href="https://app.graphite.com/github/pr/withcoral/coral/283?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#282** <a href="https://app.graphite.com/github/pr/withcoral/coral/282?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#281** <a href="https://app.graphite.com/github/pr/withcoral/coral/281?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#265** <a href="https://app.graphite.com/github/pr/withcoral/coral/265?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/265?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-mcp/src/surface/tools.rs
```diff
@@ -33,13 +37,26 @@ pub(crate) fn sql_tool(sources: &[Source], tables: &[Table]) -> Tool {
...
+        list_tables_description(visible_table_count),
         json_object_schema(&json!({
             "type": "object",
-            "properties": {}
+            "properties": {
+                "schema": {
+                    "type": "string",
+                    "description": "Optional exact schema/source name to list."
+                },
+                "limit": {
```
jsummerfield, line 51: The runtime parser enforces `limit` as 1..=200 and defaults `limit`/`offset`, but the MCP tool JSON schema only says `type: integer` and leaves those constraints in prose. Since the tool schema is part of the agent-facing contract, please add `minimum`, `maximum`, and `default` fields so clients and models can avoid invalid calls before hitting server-side validation.

crates/coral-app/src/transport.rs
```diff
@@ -78,12 +78,29 @@ pub(crate) fn table_to_proto(
...
+    table_to_proto_with_columns(workspace_name, table, ColumnMapping::Omit)
+}
+
+#[derive(Clone, Copy)]
+enum ColumnMapping {
+    Include,
+    Omit,
+}
+
+fn table_to_proto_with_columns(
```
jsummerfield, line 153: This is an awkward shortcut. The `Table` proto was intended to represent a table in full. Reusing it with empty columns is confusing.  
  
Could you add a `TableSummary` proto and use that instead.

crates/coral-app/src/query/service.rs
```diff
@@ -35,15 +40,50 @@ impl QueryServiceApi for QueryService {
...
+                Some(schema_filter)
+            };
             let tables = queries
-                .list_tables(&workspace_name)
+                .list_tables(&workspace_name, schema_filter)
                 .await
-                .map_err(query_status)?
+                .map_err(query_status)?;
+            let total = tables.len();
+            let offset = usize::try_from(pagination.offset)
```
jsummerfield: This conversion is effectively infallible on Coral’s supported targets because the protobuf fields are `uint32`; clients cannot send negative values. The current `expect` message points at non-negativity even though the only theoretical failure would be an unsupported pointer width, and panics in request handlers are a sharp default. Prefer a plain `as usize` here, or map the impossible conversion without panicking if we want to preserve theoretical portability.

crates/coral-api/proto/coral/v1/query.proto
```diff
@@ -21,16 +21,46 @@ message ExecuteSqlResponse {
...
+  // Matching item count before pagination.
+  uint32 total = 1;
```
jsummerfield: Prefer `total_count` - more explicit and matches `row_count `elsewhere

crates/coral-api/proto/coral/v1/query.proto
```diff
@@ -21,16 +21,46 @@ message ExecuteSqlResponse {
...
+  // Offset to request for the next page when `has_more` is true.
+  uint32 next_offset = 5;
+}
+
 // Lists all query-visible tables in one workspace.
 message ListTablesRequest {
   // Workspace whose current query-visible tables should be listed.
   Workspace workspace = 1;
+  // Optional exact schema/source filter.
+  string schema_filter = 2;
```
jsummerfield: Prefer `schema_name` to match the catalog and response field. Filter sounds broader than an exact match.

crates/coral-mcp/src/server.rs
```diff
@@ -58,19 +62,50 @@ impl CoralMcpServer {
...
+                omit_columns,
             }))
             .await?
-            .into_inner()
+            .into_inner())
+    }
+
+    async fn load_all_table_summaries(&self) -> Result<Vec<Table>, tonic::Status> {
+        Ok(self
+            .load_tables(None, LIST_TABLES_UNBOUNDED_LIMIT, 0, true)
```
jsummerfield: This works, but it could be more self-documenting.  e.g.

```rust
struct LoadTablesParams<'a> {
    schema_filter: Option<&'a str>,
    pagination: PaginationRequest,
    omit_columns: bool,
}

...

self.load_tables(LoadTablesParams {
    schema_filter: None,
    pagination: PaginationRequest {
        limit: LIST_TABLES_COUNT_LIMIT,
        offset: 0,
    },
    omit_columns: true,
})
.await
```

crates/coral-mcp/src/surface/tools.rs
mystic123: Done in v8: `list_tables` now exposes `minimum`, `maximum`, and `default` for `limit` and `offset` in the MCP JSON schema.

crates/coral-app/src/transport.rs
mystic123: Done in v8: added a `TableSummary` proto and `list_tables` summary mode now returns `table_summaries` instead of `Table` objects with empty columns.

crates/coral-api/proto/coral/v1/query.proto
mystic123: Done in v8: renamed the proto/request field to `schema_name` to match the catalog and response shape.

crates/coral-mcp/src/server.rs
mystic123: Done in v8: `load_tables` now takes a `LoadTablesParams` struct with `schema_name`, `pagination`, and `omit_columns`.

crates/coral-api/proto/coral/v1/query.proto
mystic123: Done in v8: renamed `PaginationResponse.total` to `total_count`.

crates/coral-app/src/query/service.rs
mystic123: Done in v8: changed the `uint32` pagination conversions to direct `as usize` casts.

crates/coral-api/proto/coral/v1/query.proto
```diff
@@ -21,16 +21,62 @@ message ExecuteSqlResponse {
...
+  // Applied result offset.
+  uint32 offset = 3;
+  // Whether another page is available.
+  bool has_more = 4;
+  // Offset to request for the next page when `has_more` is true.
+  uint32 next_offset = 5;
+}
+
+// Table metadata without full column definitions.
+message TableSummary {
```
jsummerfield: Move to catalog.proto

jsummerfield (approved):

crates/coral-api/proto/coral/v1/query.proto
mystic123: Done in v9: moved `TableSummary` to `catalog.proto`.

Bradley-Butcher (approved):

mystic123:
### Merge activity

* **May 12, 3:55 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/265).

PR #272, opened 2026-05-05
refactor(engine): introduce HTTP fetch target

Bradley-Butcher:
## Summary

Extracts `HttpFetchTarget`, the engine-local HTTP row source used by `HttpSourceClient`.

This is not a new source-spec model. It is an execution-ready view built after a SQL-facing surface has already resolved its inputs.

## Why

The following PRs add HTTP-backed table functions. Tables and functions bind SQL inputs differently:

- tables bind `WHERE` filters and choose request routes
- functions bind call arguments

After binding, both need the same HTTP execution path: render a request, paginate, decode the response, and map rows into Arrow batches.

## What Changed

- `HttpSourceClient::fetch` now takes `HttpFetchTarget` instead of `HttpTableSpec`
- `HttpSourceTableProvider` still owns table-specific filter extraction and route selection
- `HttpFetchTarget::from_resolved_table_request(...)` makes the handoff explicit
- stable target metadata is shared behind `Arc`s; each scan only swaps in the resolved request
- generic HTTP execution uses `request_values` instead of table-specific `filters`
- `HttpFetchPlan` and `http_json_exec` remain private on this branch

## Not in This PR

- table-function execution
- source-scoped SQL syntax
- bundled GitHub/Datadog function rollout

## Validation

- `make rust-checks`

### DIFF
--- a/crates/coral-engine/src/backends/http/target.rs
+++ b/crates/coral-engine/src/backends/http/target.rs
@@ -1,0 +1,43 @@
+//! Engine-local request shape shared by HTTP tables and table functions.
+
+use coral_spec::backends::http::HttpTableSpec;
+use coral_spec::{ColumnSpec, PaginationSpec, RequestSpec, ResponseSpec};
+
+/// The HTTP request/response description needed to fetch rows.
+///
+/// Tables and table functions are distinct manifest concepts, but once their
+/// SQL-facing inputs have been resolved they share the same HTTP execution path.
+#[derive(Debug, Clone)]
+pub(crate) struct HttpRequestTarget {
+    name: String,
+    columns: Vec<ColumnSpec>,
+    fetch_limit_default: Option<usize>,
+    pub(crate) request: RequestSpec,
+    pub(crate) response: ResponseSpec,
+    pub(crate) pagination: PaginationSpec,
+}
+
+impl HttpRequestTarget {
+    pub(crate) fn from_table(table: &HttpTableSpec, request: RequestSpec) -> Self {
+        Self {
+            name: table.name().to_string(),
+            columns: table.columns().to_vec(),
+            fetch_limit_default: table.fetch_limit_default(),
+            request,
+            response: table.response.clone(),
+            pagination: table.pagination.clone(),
+        }
+    }
+
+    pub(crate) fn name(&self) -> &str {
+        &self.name
+    }
+
+    pub(crate) fn columns(&self) -> &[ColumnSpec] {
+        &self.columns
+    }
+
+    pub(crate) fn fetch_limit_default(&self) -> Option<usize> {
+        self.fetch_limit_default
+    }
+}
--- a/crates/coral-engine/src/backends/http/target.rs
+++ b/crates/coral-engine/src/backends/http/target.rs
@@ -1,4 +1,6 @@
-//! Engine-local request shape shared by HTTP tables and table functions.
+//! Engine-local fetch shape shared by HTTP tables and table functions.
+
+use std::sync::Arc;
 
 use coral_spec::backends::http::HttpTableSpec;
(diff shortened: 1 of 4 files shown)


Bradley-Butcher:
* **#272** <a href="https://app.graphite.com/github/pr/withcoral/coral/272?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/272?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-engine/src/backends/http/target.rs
```diff
@@ -0,0 +1,68 @@
...
+        table: &HttpTableSpec,
+        resolved_request: RequestSpec,
+    ) -> Self {
+        Self {
+            name: table.name().to_string(),
+            columns: table.columns().to_vec(),
+            fetch_limit_default: table.fetch_limit_default(),
+            request: resolved_request,
+            response: table.response.clone(),
+            pagination: table.pagination.clone(),
```
antonmry: Per-scan cloning regression worth confirming as intentional.

Previously `HttpFetchPlan` held an `Arc<HttpTableSpec>`, so a scan was effectively zero-copy. Here every scan deep-clones `columns` (`to_vec`) plus `response` and `pagination`, and the caller in `provider.rs` already cloned the resolved `RequestSpec` a line earlier. Likely negligible in practice, but the previous design avoided it.

Two cheap mitigations if it ever matters:
- store `Arc<…>` for the heavier fields, or
- carry `Arc<HttpTableSpec>` plus the resolved `RequestSpec` separately, so functions can plug in their own without going through a table.

Not blocking — just want to make sure the regression is deliberate.

crates/coral-engine/src/backends/http/target.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    pub(crate) fn columns(&self) -> &[ColumnSpec] {
+        &self.columns
+    }
+
+    pub(crate) fn fetch_limit_default(&self) -> Option<usize> {
+        self.fetch_limit_default
+    }
+
```
antonmry, line 69: Nit: consider renaming this accessor to `resolved_request()`.

The field stores the *post-route-resolution* request, but the accessor is just `.request()`. Readers in `client.rs` can't tell from the call site whether they're seeing the table's default request or the one chosen by `resolve_request`. `resolved_request()` makes the contract self-documenting and will read more naturally once table functions land, where there's no route concept to begin with.

antonmry (approved):
Clean refactor — extracts the post-binding HTTP execution shape cleanly and moves route selection up to the right layer. Left two non-blocking inline notes (per-scan cloning, `request()` accessor naming).

crates/coral-engine/src/backends/http/target.rs
Bradley-Butcher: Good catch. I tightened this up.

HttpFetchTarget now stores the stable table metadata behind Arcs (name, columns, response, pagination). HttpSourceTableProvider builds the reusable base target once in new, and scan now only overlays the route-specific resolved_request with with_resolved_request(...).

So scans no longer deep-clone columns/response/pagination each time. The only per-scan request clone left is the resolved RequestSpec, which is the bit that actually changes based on pushed-down filters.

crates/coral-engine/src/backends/http/target.rs
Bradley-Butcher: Agreed. Renamed this to resolved_request().

The target field is now also named resolved_request, so the call site in client.rs makes the contract explicit: route selection has already happened before the HTTP client sees the target. This should read better for both table scans and the upcoming table-function path.

PR #273, opened 2026-05-05
feat(engine): add DataFusion query tracing instrumentation

antonmry:
## Summary

- Integrates the `datafusion-tracing` crate to instrument DataFusion query execution with OpenTelemetry trace spans.
- Replaces direct `SessionContext` construction with a `SessionStateBuilder` pipeline that attaches physical optimizer and rule instrumentation.
- Emits metrics and trace spans under the `coral_engine::datafusion` target for both execution and optimizer rule phases.

Datafusion tracing is disabled by default, it can be enabled with:

```
[otel]
  trace_filter = "coral_app=trace,coral_engine=trace,coral_engine::datafusion=trace"
```

HTTP tracing can be disabled with:

```
 [otel]
  trace_filter = "coral_app=trace,coral_engine=trace,coral_engine::http=off"
```

Before:

image

After 4412de6468e0d6b7f501c090a3ec909dafdee57c :

image

After f62359662fbccec73911d88175f3a0323e2bdf72

image

## Test plan

- [x] Run the existing engine test suite and confirm no regressions.
- [x] Execute a query and verify trace spans appear under `coral_engine::datafusion` in the connected tracing backend.
- [x] Confirm `record_metrics` data is captured alongside spans during query execution.
- [x] Check that optimizer rule instrumentation (`phase_only`) produces spans for each planning phase.

### DIFF
--- a/crates/coral-engine/src/backends/http/client.rs
+++ b/crates/coral-engine/src/backends/http/client.rs
@@ -770,6 +770,7 @@ async fn execute_request(
                         status: Some(status.as_u16()),
                         method: Some(method_label.to_string()),
                         url: Some(logged_url.clone()),
+                        filters: filters.clone(),
                         detail: body,
                     },
                 ))));
--- a/crates/coral-engine/src/backends/http/client.rs
+++ b/crates/coral-engine/src/backends/http/client.rs
@@ -666,6 +666,9 @@ async fn execute_request(
             url.full = %traced_url,
             error = field::Empty,
             exception.message = field::Empty,
+            coral.http.error.timeout = field::Empty,
+            coral.http.error.connect = field::Empty,
+            coral.http.error.request = field::Empty,
         );
 
         let response = http
@@ -674,7 +677,10 @@ async fn execute_request(
             .await
             .map_err(|error| {
                 request_span.record("error", true);
-                request_span.record("exception.message", field::display(&error));
+                request_span.record("exception.message", trace_reqwest_error(&error));
+                request_span.record("coral.http.error.timeout", error.is_timeout());
+                request_span.record("coral.http.error.connect", error.is_connect());
+                request_span.record("coral.http.error.request", error.is_request());
                 request_error(
                     source_schema,
                     table_name,
@@ -859,6 +865,18 @@ fn request_error(
         detail,
         timed_out: error.is_timeout(),
     })
+}
+
+fn trace_reqwest_error(error: &reqwest::Error) -> &'static str {
+    if error.is_timeout() {
+        "source API request timed out"
+    } else if error.is_connect() {
+        "source API connection failed"
+    } else if error.is_request() {
+        "source API request failed before a response was received"
+    } else {
+        "source API request failed"
+    }
 }
 
 fn decode_error(
--- a/crates/coral-engine/src/backends/http/client.rs
+++ b/crates/coral-engine/src/backends/http/client.rs
@@ -753,7 +753,11 @@ async fn execute_request(
                 .instrument(response_span.clone())
                 .await
                 .unwrap_or_default();
-            response_span.record("exception.message", field::display(&body));
+            response_span.record("http.response.body.size", body.len());
+            response_span.record(
+                "exception.message",
+                field::display(response_error_summary(status, &body)),
+            );
             return Err(DataFusionError::External(Box::new(
                 ProviderQueryError::ApiRequest {
                     source_schema: source_schema.to_string(),
@@ -877,6 +881,14 @@ fn trace_reqwest_error(error: &reqwest::Error) -> &'static str {
     } else {
         "source API request failed"
     }
+}
+
+fn response_error_summary(status: reqwest::StatusCode, body: &str) -> String {
+    format!(
+        "upstream returned HTTP {}; body_bytes={}",
+        status.as_u16(),
+        body.len()
+    )
 }
 
 fn decode_error(
--- a/crates/coral-engine/src/backends/http/client.rs
+++ b/crates/coral-engine/src/backends/http/client.rs
@@ -568,6 +568,11 @@ async fn execute_request(
     request_timeout: Duration,
     request: RequestSpec<'_>,
 ) -> Result<Option<(Value, Option<String>)>> {
+    enum ResponseOutcome {
+        Done(Result<Option<(Value, Option<String>)>>),
+        Retry(Duration),
+    }
+
     let RequestSpec {
         auth,
         request_headers,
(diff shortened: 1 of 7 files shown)


sauldhernandez (approved):

crates/coral-engine/src/backends/http/client.rs
```diff
@@ -643,31 +647,74 @@ async fn execute_request(
...
-                &error,
-            )
-        })?;
+        let response = http
+            .execute(built)
+            .instrument(request_span.clone())
+            .await
+            .map_err(|error| {
+                request_span.record("error", true);
+                request_span.record("exception.message", field::display(&error));
```
jsummerfield: `field::display(&error)` can still leak the unsanitized URL. `reqwest::Error` display output appends the full request URL, including query parameters, so a timeout or connection failure on a request like `...?api_key=secret` bypasses the sanitized `url.full` field.

A safer shape is to record a bounded request-stage summary instead of formatting the full error:

```rust
fn trace_reqwest_error(error: &reqwest::Error) -> &'static str {
    if error.is_timeout() {
        "source API request timed out"
    } else if error.is_connect() {
        "source API connection failed"
    } else if error.is_request() {
        "source API request failed before a response was received"
    } else {
        "source API request failed"
    }
}

let response = {
    let request_span = tracing::info_span!(
        target: "coral_engine::http",
        "http.request",
        url.full = %traced_url,
        error = field::Empty,
        exception.message = field::Empty,
    );

    match http.execute(built).instrument(request_span.clone()).await {
        Ok(response) => response,
        Err(error) => {
            request_span.record("error", true);
            request_span.record("exception.message", trace_reqwest_error(&error));
            return Err(request_error(
                source_schema,
                table_name,
                method_label,
                &logged_url,
                request_timeout,
                &error,
            ));
        }
    }
};
```

If lower-level detail is needed, prefer explicit bounded fields such as `timeout=true`, `connect=true`, and the already sanitized `url.full`, not the formatted `reqwest::Error`.

crates/coral-engine/src/backends/http/client.rs
```diff
@@ -680,19 +727,26 @@ async fn execute_request(
...
-            let status = response.status();
-            let body = response.text().await.unwrap_or_default();
+        if !status.is_success() {
+            response_span.record("error", true);
+            let body = response
+                .text()
+                .instrument(response_span.clone())
+                .await
+                .unwrap_or_default();
+            response_span.record("exception.message", field::display(&body));
```
jsummerfield: This records the complete upstream error body into an exported span attribute. That is a new telemetry sink for provider payloads, and it is unbounded; many APIs echo request parameters, account identifiers, or other sensitive context in error bodies.

I would keep the full body in the existing provider error path if that is intentional for user-facing diagnostics, but record only a bounded summary in the span:

```rust
fn response_error_summary(status: reqwest::StatusCode, body: &str) -> String {
    format!(
        "upstream returned HTTP {}; body_bytes={}",
        status.as_u16(),
        body.len(),
    )
}

let body = response
    .text()
    .instrument(response_span.clone())
    .await
    .unwrap_or_default();

response_span.record("error", true);
response_span.record(
    "exception.message",
    field::display(response_error_summary(status, &body)),
);

return Err(DataFusionError::External(Box::new(
    ProviderQueryError::ApiRequest {
        source_schema: source_schema.to_string(),
        table: table_name.to_string(),
        status: Some(status.as_u16()),
        method: Some(method_label.to_string()),
        url: Some(logged_url),
        detail: body,
    },
)));
```

If you want more trace detail, add explicit bounded fields such as `http.response.body.size`; do not use the raw body as `exception.message`.

crates/coral-engine/src/backends/http/client.rs
```diff
@@ -680,19 +727,26 @@ async fn execute_request(
             }
         }
 
-        if response.status().is_server_error() && server_error_retries < 2 {
+        if status.is_server_error() && server_error_retries < 2 {
+            response_span.record("error", true);
+            response_span.record("exception.message", "server error; retrying");
             server_error_retries += 1;
```
jsummerfield: The `.instrument(request_span.clone())` call only enters a clone while `execute` is polled. The original `request_span` local remains alive until the loop iteration ends, so the exported `http.request` span can include response decoding. The response span has the same problem on retry: it remains open during `tokio::time::sleep`.

I would make the span handles lexical and make retry sleep happen outside the response-span scope:

```rust
let response = {
    let request_span = tracing::info_span!(/* existing http.request fields */);

    match http.execute(built).instrument(request_span.clone()).await {
        Ok(response) => response,
        Err(error) => {
            request_span.record("error", true);
            request_span.record("exception.message", trace_reqwest_error(&error));
            return Err(request_error(
                source_schema,
                table_name,
                method_label,
                &logged_url,
                request_timeout,
                &error,
            ));
        }
    }
}; // request_span closes here

enum ResponseOutcome {
    Done(Result<Option<(Value, Option<String>)>>),
    Retry(Duration),
}

let outcome = {
    let response_span = tracing::info_span!(/* existing http.response fields */);

    match check_rate_limit(status, response.headers(), rate_limit, throttle_retries) {
        RateLimitDecision::Retry(wait) => {
            response_span.record("error", true);
            response_span.record("exception.message", "rate limited; retrying");
            throttle_retries += 1;
            ResponseOutcome::Retry(wait)
        }
        RateLimitDecision::Fail(error) => {
            response_span.record("error", true);
            response_span.record("exception.message", field::display(&error));
            ResponseOutcome::Done(Err(/* existing rate-limit error */))
        }
        RateLimitDecision::Continue => {
            if status.is_server_error() && server_error_retries < 2 {
                response_span.record("error", true);
                response_span.record("exception.message", "server error; retrying");
                server_error_retries += 1;
                ResponseOutcome::Retry(Duration::from_secs(2))
            } else {
                // existing 404 / error body / decode handling
                ResponseOutcome::Done(/* existing result */)
            }
        }
    }
}; // response_span closes here, before any backoff

match outcome {
    ResponseOutcome::Retry(wait) => {
        tokio::time::sleep(wait).await;
        continue;
    }
    ResponseOutcome::Done(result) => return result,
}
```

Explicit `drop(response_span)` before every `continue` also works, but the lexical shape makes it much harder to accidentally include future backoff or response work in the wrong span.

jsummerfield (approved):

crates/coral-engine/src/runtime/query.rs
```diff
@@ -32,7 +34,26 @@ pub(crate) async fn build_runtime(
...
+        .with_runtime_env(runtime_env)
+        .with_default_features()
+        .with_physical_optimizer_rule(instrument_rule)
+        .build();
+    let rule_options = RuleInstrumentationOptions::phase_only();
+    let session_state = datafusion_tracing::instrument_rules_with_trace_spans!(
+        target: "coral_engine::datafusion",
+        options: rule_options,
+        state: session_state
+    );
```
jsummerfield: I would be careful enabling this rule-phase instrumentation as-is. `datafusion-tracing` implements phase spans with thread-local toggle sentinels: the first sentinel opens `PLANNING_CONTEXT`, and the trailing sentinel closes it. If a rule returns an error before the trailing sentinel runs, the context remains entered on that worker thread and the next query can close or reuse the previous phase span.

Lowest-risk fix is to keep execution-plan node instrumentation and omit rule-phase instrumentation until upstream has an error/drop guard:

```rust
let session_state = SessionStateBuilder::new()
    .with_config(session_config)
    .with_runtime_env(runtime_env)
    .with_default_features()
    .with_physical_optimizer_rule(instrument_rule)
    .build();

// Skip instrument_rules_with_trace_spans! for now.
let mut ctx = SessionContext::new_with_state(session_state);
```

If we keep it, I would want a regression test that installs a subscriber, runs a planning failure followed by a successful query on the same runtime, and asserts the second query does not close or inherit a stale `analyze_logical_plan` / `optimize_logical_plan` / `optimize_physical_plan` span.

crates/coral-engine/src/backends/http/client.rs
```diff
@@ -643,31 +647,74 @@ async fn execute_request(
         });
 
         let built = resolve_auth_headers(auth, request, request_authenticators, resolved_inputs)?;
+        let request_id = NEXT_HTTP_REQUEST_ID.fetch_add(1, Ordering::Relaxed);
+        let attempt = server_error_retries + throttle_retries + 1;
+        let traced_url = sanitize_trace_url(&logged_url);
+        let request_span = tracing::info_span!(
```
jsummerfield: This PR adds new exported trace targets, but the docs do not explain the OTEL config surface yet. Since these spans are enabled by the default `coral_engine=trace` target whenever an endpoint is configured, users need examples for enabling and narrowing them.

I would add a telemetry section with something like:

```toml
# ~/.config/coral/config.toml
[otel]
endpoint = "http://localhost:4318"
service_name = "coral"
headers = "authorization=Bearer <token>,x-scope=example"
log_filter = "coral_app=info,coral_engine=info"

# Default behavior enables Coral app/engine traces.
trace_filter = "coral_app=trace,coral_engine=trace"

# Narrow noisy engine spans if needed.
# trace_filter = "coral_app=trace,coral_engine=trace,coral_engine::http=off"
# trace_filter = "coral_app=trace,coral_engine=trace,coral_engine::datafusion=off"
```

Also call out what `coral_engine::http` and `coral_engine::datafusion` contain, because these can be high-cardinality and high-volume spans.

crates/coral-engine/src/backends/http/client.rs
antonmry: Fixed with 692d667da611cb3627aac26936a1716af55828ce

crates/coral-engine/src/backends/http/client.rs
antonmry: Fixed with 17569b0d0721f09973c41cf30eda029c76259193

crates/coral-engine/src/backends/http/client.rs
antonmry: Implemented with 760a56b3680d648f42808b0e54e0483db48e391e

crates/coral-engine/src/runtime/query.rs
antonmry: This one is a surprise. [I built a regression test](https://gist.github.com/antonmry/c596f9c47dfae32e25435ebd5de6bca1) and it seems real. I might report it upstream. It's disabled in the meantime with b87dbdd3a3744a2663303d3a4b1e347fab7ec492

Very impressed with the agent being able to detect this.

crates/coral-engine/src/backends/http/client.rs
antonmry: I will add the doc in a separate PR. In the meantime, I've disabled datafusion tracing by default, but keeping HTTP enabled a76870d7cdb8e3e3e7098346c2340dba62fa0672

crates/coral-engine/src/runtime/query.rs
antonmry: Reported upstream here https://github.com/datafusion-contrib/datafusion-tracing/issues/49

PR #274, opened 2026-05-05
feat(ui): add minimal Coral shell + WAX design system

lazybean:
## Summary

This PR brings the reusable frontend foundation from PR #247 into the embedded `ui/` app:

- Adds the Wax design system primitives: theme contract, palettes, typography, core components, utilities, and font assets.
- Replaces the placeholder `ui/` app with a minimal Coral shell.
- Adds the Coral navbar/sidebar with logo and brand text.
- Adds a small placeholder view: "Placeholder" / "Coral UI coming soon".
- Keeps `ui` as the canonical app used by both `npm run dev --prefix ui` and embedded `coral ui` assets.

## Intentionally excluded

- Schema explorer
- grpc-web transport/client usage
- Rust embed changes beyond using the existing #233 `ui/dist` path
- Product/API UI behavior beyond generated protobuf compatibility

## Build scripts

Preserves protobuf generation before UI checks/builds:

- `npm run type-check --prefix ui` runs `proto:gen && tsc -b`
- `npm run build --prefix ui` runs `proto:gen && tsc -b && vite build`

## Validation

- `npm run type-check --prefix ui`
- `npm run build --prefix ui`
- `npm run dev --prefix ui`
- `cargo run -p coral-cli --features embedded-ui --bin coral -- ui`
- `playwright-cli -h`
- Playwright snapshots confirmed dev UI and embedded UI render the same Coral shell/navbar plus placeholder view.

PR #235, opened 2026-04-30
No UI yet: only query in dev tools

lazybean:

PR #275, opened 2026-05-05
feat(app): add trace storage and inspection API

sauldhernandez:
## Summary

- add a JSONL-backed local span exporter and trace reader for internal tracing
- enable local trace inspection by default with configurable seven-day retention
- expose `TraceService` with `ListTraces` and `GetTrace` over gRPC and gRPC-Web
- record HTTP request/response bodies in the internal trace store by default while filtering those body attributes from external OTLP export
- use a fixed internal trace filter that includes DataFusion rule spans while keeping `[otel].trace_filter` scoped to OTLP export
- document local trace inspection in the observability guide and wire UI protobuf generation for the new trace API

## Follow-up Split

SQL plan recreation moved to stacked PR #305 as `QueryService.PlanSql(workspace, sql)`, so trace fetching stays independent of live query planning.

## Validation

- `make docs-check`
- `RUSTC_WRAPPER= cargo test -p coral-app telemetry --locked`
- `RUSTC_WRAPPER= cargo test -p coral-app trace_service --locked`
- `RUSTC_WRAPPER= cargo test -p coral-app config_persists_across_rebuilds_without_local_trace_state --locked`
- `RUSTC_WRAPPER= make rust-checks` (`RUSTC_WRAPPER=` bypasses a local `sccache` dep-info write failure in this workspace)
- Manual HyperDX/local trace test:
  - ran HyperDX locally with OTLP/HTTP available at `http://127.0.0.1:4318`
  - installed a local `trace_probe` HTTP source under an isolated `CORAL_CONFIG_DIR` with `[otel].enable_internal_tracing = true`, `[otel].record_internal_http_bodies = true`, and `[otel].endpoint = "http://127.0.0.1:4318"`
  - ran `coral sql --format json "SELECT id, message, echoed_query, echoed_marker FROM trace_probe.probes LIMIT 1"` against a local mock POST endpoint and confirmed the result echoed the request sentinel plus the response sentinel
  - confirmed internal JSONL trace files under `.context/trace-test/config/telemetry/traces/` include `coral.http.request.body` and `coral.http.response.body`
  - confirmed HyperDX ClickHouse `default.otel_traces` contains the exported `http.request` and `http.response` spans for service `coral-pr-body-filter-test`, but `coral.http.request.body`, `coral.http.request.body.truncated`, `coral.http.response.body`, and `coral.http.response.body.truncated` are absent from OTEL span attributes
  - searched HyperDX span/event attributes for the request and response sentinels and got `0` matching rows


sauldhernandez:
* **#275** <a href="https://app.graphite.com/github/pr/withcoral/coral/275?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 4  dependent PRs ([#274](https://github.com/withcoral/coral/pull/274) <a href="https://app.graphite.com/github/pr/withcoral/coral/274?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#279](https://github.com/withcoral/coral/pull/279) <a href="https://app.graphite.com/github/pr/withcoral/coral/279?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#305](https://github.com/withcoral/coral/pull/305) <a href="https://app.graphite.com/github/pr/withcoral/coral/305?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 1 other) 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/275?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-api/proto/coral/v1/traces.proto
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// Trace detail for timeline and diagnostic views.
+message GetTraceResponse {
+  // Summary for the trace.
+  TraceSummary summary = 1;
+  // Spans ordered by start time.
+  repeated TraceSpan spans = 2;
+  // Current-state plan recreation for query traces, when available.
+  RecreatedQueryPlan recreated_plan = 3;
```
antonmry: Could we land plan recreation as a separate PR? It's a self-contained feature — `plan_sql` is already a clean engine contract that takes `(workspace, sql)` and returns a `QueryPlan`, completely independent of traces. The only thing tying it in here is that `service.rs::query_span_inputs` happens to pull those two strings off the `coral.query` span.

Splitting would:
- keep this PR focused on what it advertises (local trace store + `ListTraces`/`GetTrace`), which is already a meaningful chunk on its own
- let plan recreation ship as e.g. `PlanService.PlanSql(workspace, sql)`, which is reusable beyond the trace detail page (ad-hoc "explain this query" UX, etc.)
- make each PR easier to review

The trace detail UI could then call `GetTrace` + `PlanSql` in parallel — plan failures stay isolated from trace fetch failures, and trace-store tests don't need a `QueryManager`.

Happy to keep the engine-side `plan_sql` contract here if pulling it out is too disruptive, but moving the `RecreatedQueryPlan` proto + service wiring to a follow-up seems like a clean cut.

sauldhernandez: I think that's fair. I'll move it out

antonmry:
Following up on my inline comment on `traces.proto`, with a bit more context now that #300 is in. The full design I'd like to propose breaks this into smaller independent pieces:

**1. Two trace filters, not one.** Today there's a single `trace_filter` gating both OTLP export and the local UI store. They have different audiences and different cost profiles:

- OTLP filter → operators / production observability. Pay-per-span, network egress. Reasonable default: curated, `coral_engine::datafusion=off`, sampled.
- Local UI filter → developers debugging locally. Disk only, bounded retention. Reasonable default: permissive, full DataFusion rule spans included.

#275 already half-acknowledges this with `HttpBodyFilteringSpanExporter` (HTTP bodies are fine locally but not in OTLP). Splitting the filter generalizes that pattern. Probably want to rename to `otlp_trace_filter` / `internal_trace_filter` (or similar) to avoid ambiguity, with a fallback for the old `trace_filter` name.

Implementation note: `tracing-subscriber` per-layer filters can only narrow, not widen, relative to a global filter. Each layer needs its own `Targets` filter via `.with_filter(...)` and no aggressive global filter above. Should be fine in the current setup but worth verifying when implementing.

**2. Optimizer-rule instrumentation as opt-in.** Now that #300 is merged and `RuleInstrumentationOptions::full()` is wired up, captured rule spans are the source of truth for planning. With separate filters, this is essentially driven by the local filter alone — flip `coral_engine::datafusion` to `trace` and the rule spans flow into the local store. No special toggle needed (or a thin `[otel].instrument_optimizer_rules` bool if we want it explicit).

**3. Trace UI = generic span renderer.** Rule spans show up alongside HTTP spans, query spans, etc. — no special-casing. `GetTraceResponse` slims down to `summary` + `spans`, `RecreatedQueryPlan` field goes away, `QueryManager` dependency drops out of `TraceService`. Big testability win.

**4. Explain plans as their own UI section + RPC.** `PlanService.PlanSql(workspace, sql)` is decoupled from traces, useful with or without a captured trace. Trace detail page can add a "view explain plan" button next to the captured SQL — clicking pre-fills workspace/SQL into the explain page. Page itself is upfront that it's a fresh plan against current state, so no drift trap (the captured rule spans, if present, remain the actual planning record from when the query ran).

**What this means for #275 in practice:**

- Drop \`RecreatedQueryPlan\` from \`traces.proto\`
- Drop \`recreate_query_plan\` / \`query_span_inputs\` from \`service.rs\`
- Drop \`QueryManager\` from \`TraceService\`
- Move \`plan_sql\` engine contract + \`PlanService.PlanSql\` RPC to a follow-up PR
- Filter split is its own follow-up (independent from this PR's scope)

Three (maybe four) PRs out of the current one — each independently reviewable and revertible. Happy to discuss any of these in more detail.

docs/getting-started/installation.mdx
```diff
@@ -101,6 +101,14 @@ Important files include:
 - `config.toml` for installed-source metadata and non-secret variables
 - imported source specs under `workspaces/<workspace>/sources/<source>/manifest.yaml`
 - source secrets stored separately within the same local trust boundary
+- JSONL trace files under `telemetry/traces/` when `[otel].enable_internal_tracing` is enabled
+
+To capture local traces without running an external OpenTelemetry store, add:
```
simonwhitaker: Couple of comments:

1. Imagine reading this as a new user of Coral. Should I even care about configuring this? (i.e. should the local web tracing UI be always-on?) If I do need to know about it, should I have to know or care that it's got anything to do with OpenTelemetry?
2. If the answer to those questions is yes, then it's probably a better fit for https://withcoral.com/docs/guides/observe-with-opentelemetry. I don't think it needs to be documented here under "Important files include...". Even users who care about the tracing probably don't care about how those trace data happen to be persisted to disk.

simonwhitaker (changes requested):

docs/getting-started/installation.mdx
sauldhernandez: Makes sense to me. I'm also wondering if we should just enable UI tracing by default to better showcase what coral can do.

simonwhitaker (approved):

jsummerfield (changes requested):
I like the direction here. The main thing I’d like us to tighten before this becomes the foundation is the boundary between external observability and product trace state. Right now those feel too coupled: local trace capture, OTLP export, filtering, body capture, and process-global tracing lifecycle are all flowing through the same config and initialisation path.

The product UI needs it's own Coral-owned trace lifecycle and policy boundary.

Concretely, I’d suggest we aim for:

- external OTLP config controls external export only
- A fixed local trace config controls UI trace retention, redaction, body previews, and storage
- `trace_filter` should not affect what the local product UI can inspect
- host-owned subscribers should not accidentally receive local-only body attributes
- `TraceService` should expose Coral-shaped execution evidence first, with raw spans available as diagnostic detail where useful

That keeps the best part of this PR, which is OTel-powered local trace capture, while avoiding a long-term coupling between observability plumbing and Coral’s product trace model.

docs/guides/observe-with-opentelemetry.mdx
```diff
@@ -18,17 +18,48 @@ Coral picks up the new settings on the next invocation, and spans, logs, and met
...
+| `service_name`                 | string  | `coral`                                                                | Value of the `service.name` resource attribute on every signal.                   |
+| `enable_internal_tracing`      | boolean | `true`                                                                 | Enables local trace inspection without requiring an OTLP collector.               |
+| `internal_trace_retention_days` | integer | `7`                                                                    | Number of days of local trace spans to retain.                                    |
+| `record_internal_http_bodies`  | boolean | `true`                                                                 | Records HTTP request and response bodies in local traces when internal tracing is enabled. |
+| `internal_http_body_max_bytes` | integer | `65536`                                                                | Maximum request or response body bytes recorded in a local trace.                 |
```
jsummerfield: This is a symptom of the coupling of product + o11y needs. `internal_*` should not live in the `otel` section.

sauldhernandez:
@cubic-dev-ai review

jsummerfield:
Thanks, the config split and removal of body capture address the big prior concerns. I think one lifecycle issue remains: `TraceService` is still registered from the current server config, but local trace export is process-global and first-init wins. That means repeated local server starts with different config dirs, or a host-owned subscriber, can expose a trace API reading a store that is not actually being written. Since this API powers product UI, registration should be tied to the actual local exporter/store installed for the process, or disabled/error on mismatch. Please add a regression for repeated starts with different config dirs and/or the host-subscriber path.

crates/coral-app/src/telemetry/config.rs
```diff
@@ -89,16 +141,47 @@ headers = "from=config"
 log_filter = "info"
 trace_filter = "coral_app=debug"
 service_name = "from-config"
+
+[local_traces]
```
jsummerfield: `local_traces` is an awkward name. This will age badly once we have a server.

`trace_history` ?

jsummerfield (approved):
Approving to unblock, but please address the lifecycle and trace_history comments before merging.

PR #280, opened 2026-05-06
fix: quote qualified table hints

mystic123:
## Summary

Fixes the quoted qualified-table mistake where SQL like `FROM "github.pulls"` is parsed as one identifier instead of `github`.`pulls`.

Changes:
- add an exact structured-error hint when a missing dotted table name exactly matches a visible `schema_name.table_name`
- keep MCP `name` compatibility and add `sql_reference` for SQL-safe `FROM` / `JOIN` usage
- update MCP guide wording to steer agents away from quoting the whole `schema.table` string
- add unit and integration coverage for engine and MCP surfaces

## Validation

- `cargo fmt`
- `cargo test -p coral-engine --lib table_not_found`
- `cargo test -p coral-engine --test engine quoted_fully_qualified`
- `cargo test -p coral-mcp list_tables`
- `cargo test -p coral-mcp`
- `make rust-checks`
- `git diff --check`


sauldhernandez (approved):

antonmry (approved):

Bradley-Butcher (approved):
Is there a plan for providing this same guidance via the CLI path?

mystic123:
> Is there a plan for providing this same guidance via the CLI path?

Yes. The CLI runtime path already gets the same guidance because `coral sql` renders the structured engine error hint, so a bad query like `FROM "github.pulls"` will surface the new `Hint:` from this PR without separate CLI product logic.

I added two prevention surfaces as follow-up:
  - This PR now also updates the MCP user docs to tell agents to use `sql_reference` and not quote the whole `schema.table` string.
  - I opened https://github.com/withcoral/skills/pull/14 to update the Coral CLI skill with the same component-wise table-reference guidance.

crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -351,6 +351,14 @@ fn table_not_found_hint(
         // schema for a close match — if `FROM account` has `stripe.accounts`
         // in the catalog, suggest the schema-qualified name instead of
         // sending the user to `coral.tables`.
+        if table.contains('.')
+            && let Some(info) = known_tables
+                .iter()
+                .find(|info| format!("{}.{}", info.schema_name, info.table_name) == table)
```
jsummerfield: Consider matching this case-insensitively when the match is unique. The rest of this error path already generally recovers case-preserving suggestions, but this exact byte-for-byte comparison means a visible `"GitHub"."Pulls"` table would not get the targeted whole-reference hint for `FROM "github.pulls"`. A small helper that checks exact first, then a unique `eq_ignore_ascii_case` fallback, would keep current behavior while covering mixed-case source/table names without making ambiguous case-only catalogs arbitrary.

jsummerfield (approved):

PR #281, opened 2026-05-06
feat(mcp): add regex table discovery

mystic123:
Draft stacked on #265.

## Summary
- Adds shared MCP discovery helpers for pagination, regex compilation, SQL string literals, and table-summary shaping.
- Adds `search_tables` over `coral.tables` with optional schema narrowing, regex matching, `matched_fields`, and pagination.
- Updates MCP guide/docs and MCP stdio test coverage.

## Validation
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p coral-mcp`
- `cargo build`
- Compiled local app smoke with connected sources: `github` had 369 visible tables, and MCP `search_tables {"pattern":"^pulls$","schema":"github"}` returned `total: 1`, `limit: 20`, `github.pulls`, `matched_fields: ["table_name"]`.

### DIFF
--- a/crates/coral-mcp/src/surface/discovery.rs
+++ b/crates/coral-mcp/src/surface/discovery.rs
@@ -1,0 +1,246 @@
+use coral_api::v1::TableSummary as ProtoTableSummary;
+use regex::{Regex, RegexBuilder};
+use rmcp::ErrorData;
+use serde_json::{Map, Value, json};
+
+use super::resources::format_schema_table_equivalent;
+
+const DEFAULT_LIMIT: u32 = 50;
+const MAX_LIMIT: u32 = 200;
+const MAX_METADATA_PATTERN_BYTES: usize = 256;
+const REGEX_SIZE_LIMIT_BYTES: usize = 1 << 20;
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) struct Pagination {
+    pub(crate) limit: u32,
+    pub(crate) offset: u32,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub(crate) struct Page<T> {
+    pub(crate) items: Vec<T>,
+    pub(crate) total: u32,
+    pub(crate) limit: u32,
+    pub(crate) offset: u32,
+    pub(crate) has_more: bool,
+    pub(crate) next_offset: Option<u32>,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub(crate) struct TableSummary {
+    pub(crate) schema_name: String,
+    pub(crate) table_name: String,
+    pub(crate) description: String,
+    pub(crate) guide: String,
+    pub(crate) required_filters: Vec<String>,
+}
+
+#[allow(
+    dead_code,
+    reason = "Column summaries are shared discovery scaffolding for the follow-up column tool."
+)]
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub(crate) struct ColumnSummary {
+    pub(crate) name: String,
+    pub(crate) data_type: String,
+    pub(crate) nullable: bool,
+    pub(crate) is_virtual: bool,
+    pub(crate) is_required_filter: bool,
+    pub(crate) description: String,
+    pub(crate) ordinal_position: u32,
+}
+
+impl TableSummary {
+    pub(crate) fn from_proto(table: &ProtoTableSummary) -> Self {
+        Self {
+            schema_name: table.schema_name.clone(),
+            table_name: table.name.clone(),
+            description: table.description.clone(),
+            guide: table.guide.clone(),
+            required_filters: table.required_filters.clone(),
+        }
+    }
+
+    pub(crate) fn matched_fields(&self, regex: &Regex) -> Vec<&'static str> {
+        let name = format!("{}.{}", self.schema_name, self.table_name);
+        let candidates = [
+            ("schema_name", self.schema_name.as_str()),
+            ("table_name", self.table_name.as_str()),
+            ("name", name.as_str()),
+            ("description", self.description.as_str()),
+            ("guide", self.guide.as_str()),
+        ];
+        let mut matches = candidates
+            .into_iter()
+            .filter_map(|(field, value)| regex.is_match(value).then_some(field))
+            .collect::<Vec<_>>();
+        if self
+            .required_filters
+            .iter()
+            .any(|filter| regex.is_match(filter))
+        {
+            matches.push("required_filters");
+        }
+        matches
+    }
+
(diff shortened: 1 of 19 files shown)


mystic123:
* **#283** <a href="https://app.graphite.com/github/pr/withcoral/coral/283?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#282** <a href="https://app.graphite.com/github/pr/withcoral/coral/282?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#281** <a href="https://app.graphite.com/github/pr/withcoral/coral/281?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/281?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#265** <a href="https://app.graphite.com/github/pr/withcoral/coral/265?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

jsummerfield:
Architecture note on the `search_tables` implementation:

I think the tool belongs in `coral-mcp`, but the current raw `coral.tables` SQL in the MCP handler feels like adapter creep for a surface we want to keep thin. The match arm ends up owning SQL construction, Arrow/JSON row decoding, row-shape assumptions, regex matching, pagination, and response shaping.

The cleaner narrow shape would be:

1. Add `guide` to `coral_engine::TableInfo`.
2. Populate it in `collect_tables` from `RegisteredTable.guide`.
3. Add `guide` as an additive field to the `coral-api` table metadata contract, at least `TableSummary` and probably `Table` for consistency.
4. Map it through `coral-app/src/transport.rs`.
5. Have `search_tables` call `ListTables { schema_name, pagination: limit 0, omit_columns: true }` and regex-filter typed summaries in MCP.

That keeps MCP responsible for MCP-specific behavior: tool arguments, regex semantics, `matched_fields`, pagination, and JSON result shape. It avoids making the adapter fabricate an internal SQL metadata query and decode generic JSON rows.

I would not add a dedicated `SearchTables` RPC yet. This looks like a typed metadata completeness issue, not a new backend operation.

mystic123:
Done: added `guide` to the typed table metadata path and changed `search_tables` to filter `ListTables` summaries instead of querying `coral.tables` from MCP.

simonwhitaker (approved):

jsummerfield (approved):
Looks great, thanks!

crates/coral-mcp/src/surface/discovery.rs
```diff
@@ -0,0 +1,197 @@
...
+            .into_iter()
+            .filter_map(|(field, value)| regex.is_match(value).then_some(field))
+            .collect()
+    }
+
+    pub(crate) fn search_result_value(&self, matched_fields: &[&'static str]) -> Value {
+        json!({
+            "schema_name": self.schema_name,
+            "table_name": self.table_name,
+            "name": format!("{}.{}", self.schema_name, self.table_name),
```
Bradley-Butcher, line 91: `search_result_value` returns `name: format!("{}.{}", schema_name, table_name)` but no `sql_reference`. `list_tables` exposes `sql_reference` via `format_schema_table_equivalent`, which quotes identifiers — see `sql_reference_quotes_each_identifier_independently` in `resources.rs:277`.

The naive concat breaks for case-sensitive or dot-containing identifiers. Example: a table `Pull.Requests` in schema `github`:
- `list_tables` → `sql_reference: github."Pull.Requests"`
- `search_tables` → `name: github.Pull.Requests`

An agent dropping the latter into `FROM github.Pull.Requests` gets a three-part identifier (schema `github`, schema `Pull`, table `Requests`) — parse error or silently wrong table.

Recommend including `sql_reference` on search results, formatted via the same helper.

Bradley-Butcher: Human comment: needing sql_reference feels more like a spec validation issue tbh

crates/coral-mcp/src/surface/discovery.rs
```diff
@@ -0,0 +1,197 @@
...
+            table_name: table.name.clone(),
+            description: table.description.clone(),
+            guide: table.guide.clone(),
+            required_filters: table.required_filters.clone(),
+        }
+    }
+
+    pub(crate) fn matched_fields(&self, regex: &Regex) -> Vec<&'static str> {
+        let name = format!("{}.{}", self.schema_name, self.table_name);
+        let required_filters = self.required_filters.join(",");
```
Bradley-Butcher: `matched_fields` joins `required_filters` with `,` and matches the regex against the joined string (line 64, line 71). This produces three surprises:

**1. Anchored exact match silently fails.** Agent looking for tables that require `repo`:
- Filters: `["owner", "repo"]` → haystack: `"owner,repo"`
- Pattern `^repo$` → no match (haystack starts with `owner`)
- The agent would have to write `^owner,repo$` to hit it — but they can't, because the join character isn't documented and they don't know the other filter names anyway.

**2. Cross-boundary false positives.** The comma is part of the haystack, so patterns straddle elements:
- Filters: `["abc", "def"]` → haystack: `"abc,def"`
- Pattern `c,d` matches, even though no filter contains `c,d`.
- Worse with `.`: filters `["owner", "repo"]`, pattern `r.r` matches (the `.` eats the comma).

**3. No way to express "filter exactly named X".** Substring `repo` matches `["repo"]` and `["repository_id"]` alike. Anchoring doesn't work (see #1). The only precise pattern requires knowing every other filter name.

**Suggested fix:** match per element instead of joining.

\`\`\`rust
let required_filters_match = self
    .required_filters
    .iter()
    .any(|f| regex.is_match(f));
\`\`\`

Then `^repo$` does what it looks like it does, and there's no cross-boundary haystack.

Bradley-Butcher (approved):
some comments, trust you to address/not appropriately

crates/coral-mcp/src/surface/discovery.rs
mystic123: Addressed: `search_tables` now includes `sql_reference` using the same identifier quoting helper as `list_tables`, with regression coverage for dotted/case-sensitive table names.

crates/coral-mcp/src/surface/discovery.rs
mystic123: Addressed: `required_filters` matching now evaluates each filter independently instead of matching a comma-joined string, with regression coverage for exact filter matches and cross-boundary false positives.

mystic123:
### Merge activity

* **May 12, 3:55 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/281).

PR #287, opened 2026-05-06
chore: update cloudwatch manifest hints

iaptsiauri:
Provides little bit more information about required AWS credentials


sources/cloudwatch_logs/manifest.yaml
```diff
@@ -19,11 +19,15 @@ inputs:
   AWS_ACCESS_KEY_ID:
     kind: secret
     hint: |
-      AWS access key ID with CloudWatch Logs read permissions.
+      AWS access key ID from an AWS access key pair with CloudWatch Logs
```
simonwhitaker: nit: it's not really a key pair. It's a secret key, and the secret key's ID. I'd suggest leaving this as-is.

jsummerfield (approved):

sources/cloudwatch_logs/manifest.yaml
```diff
@@ -19,11 +19,15 @@ inputs:
...
+      read permissions.
   AWS_SECRET_ACCESS_KEY:
     kind: secret
     hint: |
-      AWS secret access key for SigV4 signing.
+      AWS secret access key from the same access key pair as
```
simonwhitaker: Same comment here. I'd suggest rewriting the first line as simple "AWS secret access key", then keep the new bit about how Coral doesn't read your CLI profile, etc.

simonwhitaker (approved):

antonmry (approved):

PR #289, opened 2026-05-06
feat(engine): project column stats in catalog

mystic123:
## Ticket

- [SOURCE-470](https://linear.app/withcoral/issue/SOURCE-470/add-column-level-statistics-to-coralcolumns-nullability-rate)

## What This PR Does

Adds the engine-side contract that lets `coral.columns` project column statistics when the runtime is given a statistics profile.

Concretely, this PR:

- Defines the engine statistics profile types for source/table/column stats.
- Appends nullable statistics fields to `coral.columns`: `null_fraction`, `approx_distinct_count`, `stats_sample_count`, and `stats_precision`.
- Looks up stats by source, table, column, source version, and schema signature before projecting them.
- Keeps existing catalog behavior unchanged when no statistics profile is provided.
- Saturates oversized internal `sample_count` values to `i64::MAX` when projecting `stats_sample_count` as SQL `BIGINT`.

## What This PR Does Not Do

- Does not collect scan observations.
- Does not expose scan observations on `QueryExecution`.
- Does not persist statistics.
- Does not make `coral.columns` the storage model.
- Does not claim stats are exact or always present.

The important contract is: `coral.columns` is a nullable projection over runtime-provided stats, not the durable stats store. Scan observations are emitted later through telemetry in #290, not attached to final query results here.

## Stack Position

1. #289 engine projection contract for `coral.columns`
2. #290 telemetry scan observations
3. #291 app cache rebuild from local trace history
4. #292 Parquet scan observations
5. #293 docs and validation notes

## Review Follow-up

Addressed the latest #289 review notes:

- Removed the unpopulated `QueryExecution` observation channel from this bottom PR so there is no second, always-empty observation pipeline.
- Changed `ColumnStatistics::sample_count_i64()` to saturate at `i64::MAX` instead of projecting oversized `u64` sample counts as `NULL`.
- Added `sample_count_projection_saturates_bigint_overflow` to lock the overflow behavior.

## Validation

Rebased on fresh `origin/main` at `96532449`.

- `cargo fmt --all -- --check`
- `CARGO_TARGET_DIR=target/source-470-check cargo test --locked -p coral-engine`
- `CARGO_TARGET_DIR=target/source-470-check cargo clippy --locked -p coral-engine --all-targets --all-features -- -D warnings`
- Earlier stack validation also passed with `CARGO_TARGET_DIR=target/source-470-check make rust-checks` and compiled CLI local-source smoke with `target/source-470-check/debug/coral`:
[...]

PR #290, opened 2026-05-06
feat(engine): emit column-stat observations

mystic123:
## Ticket

## What This PR Does

Adds engine-side scan observation reporting through telemetry.

Concretely, this PR:

- Collects bounded column statistics for supported scalar scan output columns.
- Emits structured `column_statistics` observations as telemetry events on query/scan traces.
- Classifies observation scope from confirmed scan work rather than logical query shape where providers do not push filters/projections/limits.
- Emits conservative limited observations when a parent operator stops consuming file partitions before EOF.
- Leaves unsupported, awkward, JSON, nested, and unsafe cases nullable rather than inventing misleading stats.

## What This PR Does Not Do

- Does not persist statistics.
- Does not return statistics through `QueryExecution`.
- Does not update `coral.columns` directly.
- Does not promote non-global observations into table-global catalog stats.

The important contract is: observations are reported once through the telemetry pipeline; app code may later rebuild a cache from local trace history.

## Stack Position

Depends on #289. Feeds #291, which materializes a rebuildable app cache from local trace history.

## Validation

Rebased on fresh `origin/main` at `b4692fd5`.

- `CARGO_TARGET_DIR=target/source-470-check cargo test --locked -p coral-engine`
- `CARGO_TARGET_DIR=target/source-470-check cargo test --locked -p coral-app statistics`
- `CARGO_TARGET_DIR=target/source-470-check cargo clippy --locked -p coral-engine --all-targets --all-features -- -D warnings`

Fresh-main restack preserved main's chunked JSON execution path and file metadata providers while keeping SOURCE-470 observations on the telemetry pipeline.

### DIFF
--- a/crates/coral-engine/src/backends/shared/json_exec.rs
+++ b/crates/coral-engine/src/backends/shared/json_exec.rs
@@ -19,6 +19,10 @@ use datafusion::physical_plan::{
 use futures::{TryStreamExt, stream};
 use serde_json::Value;
 
+use crate::runtime::statistics::{
+    BatchStatisticsCollector, BatchStatisticsPlan, report_statistics_observation,
+};
+
 /// Fetches raw JSON rows for one logical table scan.
 #[async_trait]
 pub(crate) trait RowFetcher: fmt::Debug + Send + Sync {
@@ -42,6 +46,7 @@ pub(crate) struct JsonExec {
     fetcher: Fetcher,
     converter: Converter,
     projection: Option<Vec<usize>>,
+    statistics_plan: Option<BatchStatisticsPlan>,
 }
 
 impl fmt::Debug for JsonExec {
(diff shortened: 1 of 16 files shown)


mystic123:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/withcoral/coral/290?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#292** <a href="https://app.graphite.com/github/pr/withcoral/coral/292?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#291** <a href="https://app.graphite.com/github/pr/withcoral/coral/291?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#290** <a href="https://app.graphite.com/github/pr/withcoral/coral/290?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/290?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-engine/src/backends/shared/json_exec.rs
```diff
@@ -143,17 +209,28 @@ impl ExecutionPlan for JsonExec {
...
 
-            match &projection {
+            let batch = match &projection {
                 Some(indices) => batch.project(indices).map_err(|error| {
                     datafusion::error::DataFusionError::ArrowError(Box::new(error), None)
                 }),
                 None => Ok(batch),
+            }?;
+
+            if let (Some(plan), Some(sink)) = (statistics_plan, statistics_sink)
```
jsummerfield: This records the observation after `batch.project(indices)`, so the `TableGlobal` stats only include the query projection rather than the table. For example, `SELECT name ...` persists a global profile with only `name`, and `COUNT(*)` may produce no column stats at all.

Can we collect statistics from the full converted batch before applying the query projection, then return the projected batch to DataFusion? The tests should also assert `statistics_observations()` for projection and aggregate queries, not just returned rows.

crates/coral-engine/src/runtime/statistics.rs
```diff
@@ -0,0 +1,378 @@
...
+        DataType::Int64 => distinct_int64::<Int64Array>(batches, column_name),
+        DataType::Boolean => distinct_bool(batches, column_name),
+        DataType::Timestamp(TimeUnit::Microsecond, _) => {
+            distinct_int64::<TimestampMicrosecondArray>(batches, column_name)
+        }
+        _ => None,
+    }
+}
+
+fn distinct_utf8(batches: &[RecordBatch], column_name: &str) -> Option<u64> {
```
jsummerfield: This computes “approx_distinct_count” by storing every distinct value in an unbounded `HashSet`, so a normal scan over a high-cardinality string/id column can allocate proportional to the full result set. Since this runs as part of query execution, it can turn stats collection into an accidental memory sink.

Can we either cap/disable distinct collection for now, or switch this to a bounded estimator before persisting it as an approximate statistic?

mystic123: Addressed in `243e6cd5`: exact distinct collection now stops at a fixed cap and leaves `approx_distinct_count` NULL after the cap, with a high-cardinality regression test.

crates/coral-engine/src/backends/shared/json_exec.rs
mystic123: Addressed in `243e6cd5`: `JsonExec::execute` now collects observations from the full converted batch before applying projection, while still returning the projected batch. Added JSONL projection and COUNT aggregate assertions against `statistics_observations()`.

crates/coral-engine/src/runtime/statistics.rs
mystic123: Addressed in 243e6cd5: exact distinct collection now stops at a fixed cap and leaves approx_distinct_count NULL after the cap, with a high-cardinality regression test.

jsummerfield (approved):

PR #291, opened 2026-05-06
feat(app): cache column stats from traces

mystic123:
## Ticket

## What This PR Does

Adds the app-side cache that rebuilds workspace column statistics from local trace history after successful queries.

Concretely, this PR:

- Reads local trace-history telemetry observations after successful SQL execution.
- Rebuilds `$CORAL_CONFIG_DIR/workspaces/<workspace>/statistics/profile.json` as a materialized cache.
- Passes the retained profile into the engine runtime so `coral.columns` can project cached stats.
- Invalidates source stats on import/delete so stale stats are not projected.
- Keeps metadata listing and source validation independent of persisted stats.
- Preserves current credential-refresh behavior while adding stats profile loading.

## What This PR Does Not Do

- Does not make `profile.json` the durable observation source.
- Does not write observations directly from engine scan code.
- Does not materialize filtered/limited/projected observations as table-global stats.
- Does not require trace history to exist; without it, stats remain `NULL`.

The important contract is: local trace history is the durable local record of observations, and `profile.json` is only a rebuildable app-owned cache.

## Stack Position

Depends on #289 and #290. Enables later backend observation PRs, including #292, to become visible through `coral.columns` after app materialization.

## Validation

Rebased on fresh `origin/main` at `96532449`.

- `CARGO_TARGET_DIR=target/source-470-check make rust-checks`
- Compiled CLI local-source smoke with `target/source-470-check/debug/coral`:
  - JSONL: stats absent before first source query, present after query, survive runtime rebuild, and recover after corrupt `profile.json`.
  - HTTP: filtered query does not materialize global stats; later unfiltered query does.
  - Parquet: generated local fixture materializes stats through the same trace-history path.

### DIFF
--- a/crates/coral-app/src/query/manager.rs
+++ b/crates/coral-app/src/query/manager.rs
@@ -8,7 +8,8 @@ use std::time::Instant;
 use coral_engine::{
     CatalogInfo, CoralQuery, CoreError, DescribeTableInfo, QueryExecution,
     QueryExecutionProvenance, QueryPlan, QueryRuntimeConfig, QueryRuntimeContext, QuerySource,
-    RuntimeSourcePackage, SourceInputResolver, SourceValidationReport, StatusCode, TableInfo,
+    RuntimeSourcePackage, SourceInputResolver, SourceValidationReport, StatisticsProfile,
+    StatusCode, TableInfo,
 };
 use coral_spec::{ManifestInputKind, ManifestInputSpec};
 use opentelemetry::trace::Status as OtelStatus;
@@ -31,8 +32,8 @@ use crate::sources::materialization::{
 };
 use crate::sources::model::InstalledSource;
 use crate::sources::runtime_package::runtime_components_for_v4_source;
-use crate::state::{AppConfig, AppStateLayout, ConfigStore};
-use crate::telemetry::WORKSPACE_SPAN_ATTRIBUTE;
+use crate::state::{AppConfig, AppStateLayout, ConfigStore, StatisticsStore};
+use crate::telemetry::{InstalledLocalTraceStore, WORKSPACE_SPAN_ATTRIBUTE};
 use crate::workspaces::WorkspaceName;
 
 #[derive(Debug)]
@@ -63,6 +64,8 @@ struct LoadedQuerySource {
 pub(crate) struct QueryManager {
     config_store: ConfigStore,
     credential_manager: CredentialManager,
+    statistics_store: StatisticsStore,
+    local_trace_store: Option<InstalledLocalTraceStore>,
     runtime_context: QueryRuntimeContext,
     layout: AppStateLayout,
     engine_extensions_providers: Vec<Arc<dyn EngineExtensionsProvider>>,
@@ -79,10 +82,20 @@ impl QueryManager {
         Self {
             config_store,
             credential_manager,
+            statistics_store: StatisticsStore::new(layout.clone()),
+            local_trace_store: None,
             runtime_context,
             layout,
             engine_extensions_providers,
         }
+    }
+
+    pub(crate) fn with_local_trace_store(
+        mut self,
+        local_trace_store: Option<InstalledLocalTraceStore>,
+    ) -> Self {
+        self.local_trace_store = local_trace_store;
+        self
     }
 
     pub(crate) async fn list_tables(
@@ -108,6 +121,7 @@ impl QueryManager {
                         &loaded_sources,
                         &config,
                         CredentialResolutionMode::StoredOnly,
+                        StatisticsProfile::empty(),
                     )
                     .map_err(QueryManagerError::App)?;
                 let sources = query_sources_from_loaded(&loaded_sources);
@@ -143,6 +157,7 @@ impl QueryManager {
                         &loaded_sources,
                         &config,
                         CredentialResolutionMode::StoredOnly,
+                        StatisticsProfile::empty(),
                     )
                     .map_err(QueryManagerError::App)?;
                 let sources = query_sources_from_loaded(&loaded_sources);
@@ -203,7 +218,7 @@ impl QueryManager {
         sql: &str,
         attribution: &QueryAttribution,
     ) -> Result<QueryExecution, QueryManagerError> {
-        run_query_operation(
+        let result = run_query_operation(
             QueryOperation::ExecuteSql,
             workspace_name,
             sql,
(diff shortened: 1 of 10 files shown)


mystic123:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/withcoral/coral/291?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#292** <a href="https://app.graphite.com/github/pr/withcoral/coral/292?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#291** <a href="https://app.graphite.com/github/pr/withcoral/coral/291?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/291?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#290** <a href="https://app.graphite.com/github/pr/withcoral/coral/290?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

crates/coral-app/src/state/statistics.rs
```diff
@@ -0,0 +1,353 @@
...
+        };
+        table_stats
+            .columns
+            .entry(column.column_name.clone())
+            .and_modify(|existing| merge_column(existing, &observed))
+            .or_insert(observed);
+    }
+}
+
+fn merge_column(existing: &mut ColumnStatistics, observed: &ColumnStatistics) {
```
jsummerfield: Could we replace table-global stats instead of accumulating them here? A `TableGlobal` observation represents one full-table scan snapshot, so adding the previous persisted `sample_count`/`null_count` makes `coral.columns` depend on how often the same query ran. Rerunning the same full-table query would double the displayed counts without any data change.

crates/coral-app/src/state/statistics.rs
```diff
@@ -0,0 +1,353 @@
...
+        .or_insert_with(|| SourceStatistics {
+            schema_name: observation.schema_name.clone(),
+            source_version: observation.source_version.clone(),
+            tables: BTreeMap::default(),
+        });
+    source_stats
+        .source_version
+        .clone_from(&observation.source_version);
+
+    let replace_table = source_stats
```
jsummerfield: This replacement predicate should also account for `source_version` changes. With the current check, a same-schema source version bump merges the new observation into old stats and then saves the blended table under the new version, which bypasses the runtime version guard.

crates/coral-app/src/query/manager.rs
```diff
@@ -224,11 +246,37 @@ impl QueryManager {
...
 
-    fn runtime_config(&self, selected_sources: &[QuerySource]) -> QueryRuntimeConfig {
+    fn load_statistics_profile(
+        &self,
+        workspace_name: &WorkspaceName,
+        selected_sources: &[QuerySource],
+    ) -> StatisticsProfile {
+        match self.statistics_store.load_profile(workspace_name) {
+            Ok(mut profile) => {
+                profile.retain_sources(selected_sources.iter().map(QuerySource::source_name));
```
jsummerfield, line 477: Filtering by selected source name is necessary, but it is not enough to prove these stats belong to the current backing dataset. Source variables and secrets are resolved from app state, so a source can be reimported/rebound with the same name, version, and schema while pointing at different data. The narrow app-layer fix is to invalidate that source's persisted stats on source upsert/delete rather than pushing source identity into the engine.

crates/coral-app/src/query/manager.rs
```diff
@@ -68,7 +71,8 @@ impl QueryManager {
         let sources = self
             .load_query_sources(workspace_name)
             .map_err(QueryManagerError::App)?;
-        let runtime = self.runtime_config(&sources);
+        let statistics = self.load_statistics_profile(workspace_name, &sources);
```
jsummerfield: Does `list_tables` need the stats profile? The returned `TableInfo` does not expose statistics, so this adds locked profile-file I/O to listing calls without changing output. Same question for `list_catalog` below unless there is a planned contract change in this stack.

crates/coral-app/src/query/manager.rs
```diff
@@ -101,10 +106,23 @@ impl QueryManager {
...
                     .map_err(QueryManagerError::App)?;
-                let runtime = self.runtime_config(&sources);
-                CoralQuery::execute_sql(&sources, runtime, sql)
+                let statistics = self.load_statistics_profile(workspace_name, &sources);
+                let runtime = self.runtime_config(&sources, statistics);
+                let execution = CoralQuery::execute_sql(&sources, runtime, sql)
                     .await
-                    .map_err(QueryManagerError::Core)
+                    .map_err(QueryManagerError::Core)?;
+                if !execution.statistics_observations().is_empty()
```
jsummerfield: This makes successful SQL execution mutate workspace state and change future `coral.columns` output. Since that is CLI/MCP-visible behavior, this stack should include docs or MCP guidance explaining that `null_fraction`, `approx_distinct_count`, `stats_sample_count`, `stats_observed_at`, and `stats_precision` are observed best-effort persisted values, plus where the workspace profile is stored.

jsummerfield (commented):
A few architecture/correctness notes from tracing the stats persistence seam end to end.

crates/coral-app/src/query/manager.rs
mystic123: Addressed in stack docs commit `39c12532`: the MCP guide template and MCP user guide now describe these as best-effort persisted observations and include `CORAL_CONFIG_DIR/workspaces/<workspace>/statistics/profile.json`.

crates/coral-app/src/query/manager.rs
mystic123: Addressed in `495ec9b7`: `SourceManager` invalidates per-source stats on import/reimport and delete, with rollback on invalidation failure plus import/delete regression tests.

crates/coral-app/src/state/statistics.rs
```diff
@@ -0,0 +1,354 @@
...
+        self.save_unlocked(workspace_name, &profile)
+    }
+
+    pub(crate) fn invalidate_source(
+        &self,
+        workspace_name: &WorkspaceName,
+        source_name: &SourceName,
+    ) -> Result<(), AppError> {
+        let _lock = FileLock::exclusive(self.layout.state_lock())?;
+        let mut profile = self.load_unlocked(workspace_name)?;
```
jsummerfield: [P1] Could we make source invalidation tolerant of an unreadable stats profile? `load_unlocked()` returns JSON parse/read errors here, and the new source import/delete callers treat this as fatal and roll back. Since this file is derived, best-effort cache state, a corrupt `profile.json` should not be able to block source lifecycle operations; it should probably warn and behave like an empty/removed profile, matching the query path's fallback behavior.

crates/coral-app/src/sources/manager.rs
```diff
@@ -469,6 +478,13 @@ impl SourceManager {
             self.restore_source_rollback_state(workspace_name, &source_name, previous);
             return Err(error);
         }
+        if let Err(error) = self
```
jsummerfield: [P2] There is still a small stale-stats window here because the source config has already been committed before the old profile entry is removed. A concurrent query can load the new source config and the old stats profile between those two operations; if the source version/schema are unchanged but bindings now point at different data, `coral.columns` can briefly expose stale stats. Can this invalidation happen before publishing the new config, or under one state-store operation/lock so config and stats move together?

crates/coral-app/src/query/manager.rs
```diff
@@ -395,3 +441,205 @@ fn validate_required_variables(
...
+{"id":3,"category":"alpha","nullable_text":"three"}
+"#,
+        )
+        .expect("jsonl fixture");
+
+        let manifest = format!(
+            r#"name: local_stats_jsonl
+version: 0.1.0
+dsl_version: 3
+backend: jsonl
```
jsummerfield: You'll need to rebase since the `backend: file` PR went in.

crates/coral-app/src/sources/manager.rs
```diff
@@ -27,6 +27,7 @@ use tracing::warn;
 pub(crate) struct SourceManager {
     config_store: ConfigStore,
     credential_manager: CredentialManager,
+    statistics_store: StatisticsStore,
```
jsummerfield: [P2] I think this ownership seam would be cleaner if `SourceManager` did not coordinate `ConfigStore` and `StatisticsStore` directly. Both config and stats live under the app-state layout and share the same state lock, so source publish/remove plus stats invalidation should probably be one narrow state-layer operation, e.g. `upsert_source_and_invalidate_statistics` / `remove_source_and_invalidate_statistics`, implemented under one exclusive lock with an unlocked stats helper that can discard an unreadable profile. That would fix the stale-stats window and the corrupt-cache lifecycle failure without adding a broad repository abstraction or moving this concern into CLI/MCP/engine layers.

PR #292, opened 2026-05-06
feat(engine): observe parquet column stats

mystic123:
## Ticket

## What This PR Does

Adds Parquet scan observations to the telemetry-backed column-statistics pipeline.

Concretely, this PR:

- Emits telemetry observations for full, unfiltered, unprojected Parquet scans that can safely be treated as table-global.
- Downgrades projected, filtered, limited, and partition-fragment observations so they are not materialized as table-global catalog stats.
- Aggregates/collects incrementally with bounded memory behavior instead of retaining every scanned batch for stats.
- Applies the same distinct-count cap across string representations, including dictionary-backed strings.

## What This PR Does Not Do

- Does not extract exact Parquet footer statistics as the public stats source.
- Does not persist observations itself.
- Does not materialize projected or partition-fragment scans into `coral.columns` as table-global stats.
- Does not change the app cache model from #291.

The important contract is: Parquet participates in the same telemetry observation path as other backends; app materialization decides what is safe to project.

## Stack Position

Depends on #289, #290, and #291. #293 documents the public behavior and validation.

## Validation

Rebased on fresh `origin/main` at `96532449`.

- `CARGO_TARGET_DIR=target/source-470-check make rust-checks`
- Compiled CLI local-source smoke with `target/source-470-check/debug/coral`:
  - Parquet: generated deterministic `pyarrow` fixture with `Int64`, `Utf8`, nullable `Utf8`, and `Boolean` columns; queried through compiled CLI; verified `coral.columns` shows materialized stats.
  - JSONL and HTTP stack smokes also passed to verify cross-backend scope/materialization behavior still holds.

### DIFF
--- a/crates/coral-engine/src/backends/file/mod.rs
+++ b/crates/coral-engine/src/backends/file/mod.rs
@@ -216,7 +216,10 @@ impl CompiledBackendSource for FileCompiledSource {
         );
 
         for table in &self.manifest.tables {
-            let table_statistics = if matches!(table.format, FileFormat::Json | FileFormat::Jsonl) {
+            let table_statistics = if matches!(
+                table.format,
+                FileFormat::Json | FileFormat::Jsonl | FileFormat::Parquet
+            ) {
                 Some(FileStatisticsRegistration::new(
                     self.manifest.common.version.clone(),
                 ))
--- a/crates/coral-engine/src/runtime/statistics.rs
+++ b/crates/coral-engine/src/runtime/statistics.rs
@@ -728,8 +728,10 @@ fn record_bool_values(values: &mut HashSet<bool>, array: &dyn Array) -> Distinct
 mod tests {
     use std::sync::Arc;
 
-    use datafusion::arrow::array::{Float64Array, Int64Array, StringArray};
-    use datafusion::arrow::datatypes::{DataType, Field, Schema};
+    use datafusion::arrow::array::{
+        DictionaryArray, Float64Array, Int64Array, StringArray, StringViewArray, UInt16Array,
+    };
+    use datafusion::arrow::datatypes::{DataType, Field, Schema, UInt16Type};
 
     use super::{BatchStatisticsPlan, DISTINCT_COUNT_MAX_VALUES, collect_batch_statistics};
     use crate::contracts::{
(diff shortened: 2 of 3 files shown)


mystic123:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/withcoral/coral/292?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#292** <a href="https://app.graphite.com/github/pr/withcoral/coral/292?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/292?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#291** <a href="https://app.graphite.com/github/pr/withcoral/coral/291?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#290** <a href="https://app.graphite.com/github/pr/withcoral/coral/290?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

crates/coral-engine/src/backends/parquet/mod.rs
```diff
@@ -210,7 +237,285 @@ impl TableProvider for ParquetTableProvider {
...
+            return;
+        }
+        self.finished = true;
+        let observation = if failed {
+            self.statistics_accumulator.fail_partition()
+        } else {
+            self.statistics_accumulator.finish_partition()
+        };
+        if let Some(observation) = observation {
+            self.statistics_sink.observe(observation);
```
antonmry: **Question — why a bespoke sink/store instead of the existing telemetry pipeline?**

Inline here because this is the most concrete place to see the cost — but the question applies to the whole stack (#290 introduces the sink, #291 the store, #292 plugs Parquet in). I don't see this addressed in any of the descriptions or the ticket (apologies if I missed prior discussion).

We already have an in-process pipeline that does most of what these PRs build:

- `crates/coral-app/src/telemetry/local_store.rs` — `JsonlSpanExporter`, rolling local JSONL with retention + pruning. Local-by-default; no OTLP endpoint required.
- `crates/coral-api/proto/coral/v1/traces.proto` — `ListTraces` / `GetTrace` over gRPC, already consumed by the UI.
- `ui/src/views/traces/http-span-detail.tsx` — already renders structured payloads (HTTP request/response body previews) attached to scan spans.

#290's description notes that observations are emitted from `JsonExec` — which is already inside code that creates OTel spans for HTTP and JSONL scans. If observations were attached to the existing scan span as a structured event instead of pushed through `StatisticsObservationSink`:

- The producer surface collapses: no `StatisticsObservationSink`, no `RuntimeStatisticsContext` plumbing through `QueryRuntimeConfig` — just `span.add_event(...)` next to the body-preview attribute that's already being set.
- #291's `StatisticsStore` / `profile.json` becomes a derived materialized view over the trace store, rather than a parallel storage path with its own file lock, retention, and merge rules.
- "Null-rate of column X over the last 7 days" becomes a queryable trace-store query — the current `profile.json` design overwrites on each TableGlobal observation, so it literally can't answer that.
- The Traces UI gets stats visibility for free: click a slow query, see which observations it produced.

Worth flagging the direction-of-migration argument too: starting on the existing telemetry pattern is the cheaper default. We get storage, retention, UI consumption, and gRPC access for free, and if column statistics later outgrow that pipeline — plan-time read latency, retention semantics that need to diverge from traces, hosted-deployment fan-out — peeling them off into their own pipeline is a straightforward refactor we can do once we know the shape of the problem. The reverse move (parallel pipeline now, consolidate onto telemetry later) is more expensive once there are workspaces with persisted `profile.json` files in the wild — that means a migration path for the on-disk format and merge rules, not just a refactor.

mystic123: This is a fair direction to question, but I don't think we should rewrite this stack to make trace history the source of truth for `coral.columns`.

One bit of chronology/context: when the SOURCE-470 stack was designed/opened, base OTel/span instrumentation existed, but the local trace-store/inspection/UI pipeline cited here had not landed yet. The stack opened on May 6; the local trace storage + inspection API landed May 14, and the trace UI/body-viewer pieces landed after that. So the original design did not skip an already-merged app-level substrate.

Now that the trace pipeline exists, I agree there is value in emitting stats observations as trace events too. That would be useful for query debugging: click a trace, see which table/column observations were produced.

I don't think trace history should replace the workspace stats profile, though. `coral.columns` needs a low-latency materialized view during runtime/catalog construction, filtered to selected sources and invalidated with source lifecycle changes. Trace history has different semantics: retention-bound, user-configurable/disableable, optimized for list/get trace inspection, and historical by design. If catalog stats were derived directly from trace JSONL, runtime construction would either scan trace files or need a table-indexed materialized view. That materialized view is effectively the stats profile again, just downstream of telemetry.

antonmry: Good points — and the chronology context helps, thanks for laying it out.

@jsummerfield and I have been discussing this too. The principle we keep landing on: **decouple reporting from intelligence**. Report once via telemetry, then expose that data (directly or via materialized views like `profile.json`) to the UI, the catalog, or any future intelligence layer. That speaks to your "materialized view = stats profile again" point — same shape, but the source-of-truth shifts: telemetry becomes the durable record and `profile.json` becomes a rebuildable cache.

Telemetry's lifecycle (retention, disable toggle) could become a limiting factor later. But given how much engine complexity this removes, especially as we add more intelligence/UI features that all want the same underlying observations, it feels right to start that way and only add a dedicated pipeline once we've identified a concrete reason to. What do you think?

mystic123: Reworked the stack in this direction.

The engine now reports `StatisticsObservation` once as a structured telemetry event (`coral.statistics.observation`) on the active trace. The previous `StatisticsObservationSink` / `RuntimeStatisticsContext` / `QueryExecution` observation plumbing is gone, so JSONL, HTTP, and Parquet all use the same telemetry reporting path.

The app now treats local trace history as the durable observation log. After a successful query it force-flushes tracing, reads stats observation events from successful `coral.query` traces, and rebuilds `workspaces/<workspace>/statistics/profile.json` as a materialized cache filtered to the current workspace sources. Failed traces can still exist for diagnostics, but their observation events are not eligible for cache materialization. If trace history is disabled or unavailable, catalog stats stay `NULL` instead of using a parallel stats storage path.

I also updated the SOURCE-470 plan, Linear ticket, and PR descriptions to describe trace history as the observation source and `profile.json` as the rebuildable cache. Validation now includes `make rust-checks` and compiled CLI smokes against local JSONL, local HTTP, and generated local Parquet sources.

PR #301, opened 2026-05-07
build(cargo): standardize locked Cargo commands

jsummerfield:
## Intent

Cargo commands in our local workflow, CI, and docs were not consistently using the checked-in lockfile. That left some commands free to resolve a different semver-compatible dependency graph, which made rebuild behavior look arbitrary and could let local validation drift from committed state.

This change makes locked dependency resolution the default for repo-owned Cargo workflows and source-install instructions.

## Changes

- Add `--locked` to `clippy`, `doc`, and `xtask` docs-generation commands in the Makefile.
- Update contributor docs so the expanded `rust-checks` commands actually match the Makefile.
- Change source install docs to use `cargo install --path crates/coral-cli --locked`.
- Lock Cargo usage in build-timings CI and the `ryl` install path.
- Align the architecture docs' black-box CLI test command with the locked workflow.

## Verification

- `make rust-checks`
- `make docs-check`
- `make lint-sources`
- `cargo install --path crates/coral-cli --locked --root "$tmpdir" --debug`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,6 +74,13 @@ cargo test --workspace --all-targets --all-features --locked
 RUSTDOCFLAGS="-D warnings" cargo doc --workspace --all-features --no-deps --locked
 ```
 
+If a Cargo command fails because the lockfile needs to be updated but
+`--locked` was passed, refresh `Cargo.lock` with the narrowest practical
+command, such as `cargo update -p <crate>`, then commit the lockfile with the
+matching `Cargo.toml` change. For broader dependency changes, run one normal
+Cargo build without `--locked` to regenerate the lockfile before rerunning the
+locked checks.
+
 ### Source manifest linting
 
 Source manifests under `sources/*/manifest.yaml` are checked with
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -68,7 +68,6 @@ If you prefer to run steps individually, the equivalent commands are:
 
 ```bash
 cargo fmt --all -- --check
-cargo check --workspace --all-targets --all-features --locked
 cargo clippy --workspace --all-targets --all-features --locked -- -D warnings
 cargo test --workspace --all-targets --all-features --locked
 RUSTDOCFLAGS="-D warnings" cargo doc --workspace --all-features --no-deps --locked
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,9 +67,11 @@ update the relevant docs in the same pull request.
 If you prefer to run steps individually, the equivalent commands are:
 
 ```bash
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -1,9 +1,9 @@
 rust-checks:
 	cargo fmt --all -- --check
 	cargo check --workspace --all-targets --all-features --locked
-	cargo clippy --workspace --all-targets --all-features -- -D warnings
+	cargo clippy --workspace --all-targets --all-features --locked -- -D warnings
```
antonmry, line 3: This is going to fail if Cargo.lock wasn't updated.

`--locked` is usually used for reproducibility, not to improve compilation time.

Makefile
```diff
@@ -1,9 +1,9 @@
 rust-checks:
 	cargo fmt --all -- --check
 	cargo check --workspace --all-targets --all-features --locked
```
antonmry: Out of scope for this PR, but worth a follow-up: `cargo clippy` is a strict superset of `cargo check` (same compiler frontend, same `.rmeta` artifacts, plus lints). This line type-checks the workspace a second time and produces nothing the rest of the target reuses, so it's pure overhead on every `make rust-checks`. Dropping it would shave a full check pass with no loss in safety.

CONTRIBUTING.md
```diff
@@ -67,9 +67,11 @@ update the relevant docs in the same pull request.
...
 
 ```bash
-cargo fmt --all
-cargo clippy --workspace --all-targets -- -D warnings
-cargo test --workspace
+cargo fmt --all -- --check
+cargo check --workspace --all-targets --all-features --locked
+cargo clippy --workspace --all-targets --all-features --locked -- -D warnings
+cargo test --workspace --all-targets --all-features --locked
+RUSTDOCFLAGS="-D warnings" cargo doc --workspace --all-features --no-deps --locked
```
antonmry, line 73: Suggest adding a short note right after this block for new contributors: if any of these commands fail with `the lock file ... needs to be updated but --locked was passed`, run `cargo update -p <crate>` (or `cargo build` once without `--locked`) to refresh `Cargo.lock`, then commit it alongside the `Cargo.toml` change. `--locked` correctly catches forgotten lockfile updates, but the failure message is opaque the first time you hit it and the recovery isn't obvious.

antonmry (approved):

simonwhitaker (approved):

Makefile
jsummerfield: Ack, will follow up

jsummerfield:
### Merge activity

* **May 7, 3:17 PM UTC**: @jsummerfield merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/301).

PR #304, opened 2026-05-07
feat(engine): add final-result observer hook

Bradley-Butcher:
## Summary

Adds a generic `QueryResultObserver` engine extension that lets external extension crates inspect successful final SQL results after DataFusion collection and before `QueryExecution` is returned.

This wires observer lists through `EngineExtensions` and app extension-provider merging, preserving the existing layering for source decorators and request authenticators. Observer failures are mapped into structured `CoreError` diagnostics, and the new tests cover successful observation, error propagation, no-observer behavior, and filtered/projected final results.

## Intent

This is meant to be a small core hook, not a needle-specific feature.

The immediate use case is cloud/eval needle retrieval tracking: needles are injected inside Coral through an engine extension, and retrieval tracking should be able to observe the same typed final Arrow result regardless of whether the query was run through CLI, gRPC, MCP, tests, or a benchmark harness.

Keeping this in core as a generic observer avoids coupling tracking to one caller's output format, JSON rendering, or client-side decoding path. It also keeps needle-specific matching and logging out of `coral-engine`; those can live in the cloud/evals extension crate on top of this hook.

This does not attempt to provide row provenance. Observers see final materialized results only. They can answer questions like "does the final result contain the expected values?" but not "did this output physically derive from a specific source row through arbitrary SQL operators?"

## Checks

- `make rust-checks`

### DIFF
--- a/crates/coral-app/src/query/extensions.rs
+++ b/crates/coral-app/src/query/extensions.rs
@@ -49,6 +49,9 @@ pub(crate) fn engine_extensions_for_providers(
         let extra = provider.extensions_for(selected_sources);
         merged.source_decorators.extend(extra.source_decorators);
         merged
+            .query_result_observers
+            .extend(extra.query_result_observers);
+        merged
             .request_authenticators
             .extend(extra.request_authenticators);
     }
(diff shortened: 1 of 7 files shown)


sauldhernandez (commented):
Since observers run synchronously after result collection and before `QueryExecution` is returned, a slow observer will directly add latency to `execute_sql`. Is that coupling intentional for this extension point? If so, could we document that observers are expected to stay lightweight or hand off heavier work themselves, especially for observers that might persist results or call external services?

Bradley-Butcher:
@sauldhernandez thanks, yes, the in-band coupling is intentional: observers are a synchronous post-collection gate so observer failures can still fail the query before `QueryExecution` is returned.

I pushed `c520cbe` to make that contract explicit in the `QueryResultObserver` docs, including that observer work contributes directly to `execute_sql` latency and heavier persistence, network calls, or telemetry fanout should be handed off when they should not delay the query response.

On the default path I don't expect to ever see an observer. It should only be during benchmarking. Happy if you have a better suggestion, or are skewed towards making observation async.

sauldhernandez (approved):

PR #279, opened 2026-05-06
feat(engine): add DataFusion query tracing instrumentation

sauldhernandez:
feat(engine): add DataFusion query tracing instrumentation

Integrate `datafusion-tracing` to emit OpenTelemetry trace spans and
metrics for physical plan execution and optimizer rule phases.
- Instruments execution via `instrument_with_trace_spans!` macro
- Instruments optimizer rules via `instrument_rules_with_trace_spans!`

feat(ui): show local query traces

PR #305, opened 2026-05-07
feat(app): add SQL explain API

sauldhernandez:
## Summary

- add `QueryService.PlanSql(workspace, sql)` with transport-level plan messages
- move `CoralQuery::plan_sql` and the engine `QueryPlan` contract into this stacked PR
- add engine and gRPC coverage for logical and physical plan rendering

## Validation

- `RUSTC_WRAPPER= make rust-checks`

Note: `RUSTC_WRAPPER=` bypasses a local `sccache` dep-info write failure in this workspace.

### DIFF
--- a/crates/coral-api/proto/coral/v1/query.proto
+++ b/crates/coral-api/proto/coral/v1/query.proto
@@ -21,6 +21,31 @@ message ExecuteSqlResponse {
   int64 row_count = 2;
 }
 
+// Produces query-engine plan renderings for one SQL statement without executing
+// it.
+message PlanSqlRequest {
+  // Workspace that scopes the queryable source set.
+  Workspace workspace = 1;
+  // One read-only SQL statement.
+  string sql = 2;
+}
+
+// Query-engine plan renderings for one SQL statement.
+message QueryPlan {
+  // Parsed logical plan before logical optimization.
+  string unoptimized_logical_plan = 1;
+  // Logical plan after logical optimization.
+  string optimized_logical_plan = 2;
+  // Physical execution plan after physical optimization.
+  string physical_plan = 3;
+}
+
+// Query planning response.
+message PlanSqlResponse {
+  // Generated query plan.
+  QueryPlan plan = 1;
+}
+
 // Page request parameters for list-style APIs.
 message PaginationRequest {
   // Maximum number of items to return. Zero means unbounded for compatibility.
@@ -73,4 +98,6 @@ service QueryService {
   rpc ListTables(ListTablesRequest) returns (ListTablesResponse);
   // Executes one read-only SQL statement and returns the full result set.
   rpc ExecuteSql(ExecuteSqlRequest) returns (ExecuteSqlResponse);
+  // Plans one read-only SQL statement without executing it.
+  rpc PlanSql(PlanSqlRequest) returns (PlanSqlResponse);
 }
--- a/crates/coral-api/proto/coral/v1/query.proto
+++ b/crates/coral-api/proto/coral/v1/query.proto
@@ -21,9 +21,9 @@ message ExecuteSqlResponse {
   int64 row_count = 2;
 }
 
-// Produces query-engine plan renderings for one SQL statement without executing
-// it.
-message PlanSqlRequest {
+// Explains one SQL statement by producing query-engine plan renderings without
+// executing it.
+message ExplainSqlRequest {
   // Workspace that scopes the queryable source set.
   Workspace workspace = 1;
   // One read-only SQL statement.
@@ -40,8 +40,8 @@ message QueryPlan {
   string physical_plan = 3;
 }
 
-// Query planning response.
-message PlanSqlResponse {
+// SQL explanation response.
+message ExplainSqlResponse {
   // Generated query plan.
   QueryPlan plan = 1;
 }
@@ -98,6 +98,6 @@ service QueryService {
   rpc ListTables(ListTablesRequest) returns (ListTablesResponse);
   // Executes one read-only SQL statement and returns the full result set.
   rpc ExecuteSql(ExecuteSqlRequest) returns (ExecuteSqlResponse);
-  // Plans one read-only SQL statement without executing it.
-  rpc PlanSql(PlanSqlRequest) returns (PlanSqlResponse);
+  // Explains one read-only SQL statement without executing it.
+  rpc ExplainSql(ExplainSqlRequest) returns (ExplainSqlResponse);
 }
--- a/crates/coral-app/src/query/manager.rs
+++ b/crates/coral-app/src/query/manager.rs
@@ -97,13 +97,13 @@ impl QueryManager {
         .await
     }
 
-    pub(crate) async fn plan_sql(
+    pub(crate) async fn explain_sql(
         &self,
         workspace_name: &WorkspaceName,
         sql: &str,
     ) -> Result<QueryPlan, QueryManagerError> {
         run_query_operation(
-            QueryOperation::PlanSql,
+            QueryOperation::ExplainSql,
             workspace_name,
             sql,
             async {
@@ -111,7 +111,7 @@ impl QueryManager {
                     .load_query_sources(workspace_name)
                     .map_err(QueryManagerError::App)?;
                 let runtime = self.runtime_config(&sources);
-                CoralQuery::plan_sql(&sources, runtime, sql)
+                CoralQuery::explain_sql(&sources, runtime, sql)
                     .await
                     .map_err(QueryManagerError::Core)
             },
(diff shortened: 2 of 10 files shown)


sauldhernandez:
* **#305** <a href="https://app.graphite.com/github/pr/withcoral/coral/305?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/305?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#275** <a href="https://app.graphite.com/github/pr/withcoral/coral/275?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 3 other dependent PRs ([#274](https://github.com/withcoral/coral/pull/274) <a href="https://app.graphite.com/github/pr/withcoral/coral/274?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#279](https://github.com/withcoral/coral/pull/279) <a href="https://app.graphite.com/github/pr/withcoral/coral/279?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#376](https://github.com/withcoral/coral/pull/376) <a href="https://app.graphite.com/github/pr/withcoral/coral/376?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

crates/coral-engine/src/runtime/query.rs
```diff
@@ -133,6 +134,33 @@ impl QueryRuntimeAdapter {
...
+            .optimize(&logical_plan)
+            .map_err(|err| datafusion_to_core(&err, &self.tables))?;
+        let optimized_logical_plan_display =
+            optimized_logical_plan.display_indent_schema().to_string();
+        let physical_plan = session_state
+            .query_planner()
+            .create_physical_plan(&optimized_logical_plan, &session_state)
+            .await
+            .map_err(|err| datafusion_to_core(&err, &self.tables))?;
+        let physical_plan = displayable(physical_plan.as_ref()).indent(true).to_string();
```
antonmry: Really nice extraction — `plan_sql` ends up as just three clean DataFusion calls.

One small consistency observation: `display_indent_schema()` on the logical plans (lines 144 and 150) shows column names + types on each node, but `displayable(...).indent(true)` on the physical plan doesn't include schemas unless `.set_show_schema(true)` is also called. So in the response, logical nodes show schemas and physical nodes don't. For users tracing a column through the pipeline, that can be a bit jarring. Worth picking one (schemas everywhere or nowhere) and being consistent — totally a polish thing, not blocking.

crates/coral-api/proto/coral/v1/query.proto
```diff
@@ -21,6 +21,31 @@ message ExecuteSqlResponse {
...
+}
+
+// Query-engine plan renderings for one SQL statement.
+message QueryPlan {
+  // Parsed logical plan before logical optimization.
+  string unoptimized_logical_plan = 1;
+  // Logical plan after logical optimization.
+  string optimized_logical_plan = 2;
+  // Physical execution plan after physical optimization.
+  string physical_plan = 3;
```
antonmry, line 40: All three plans are returned as plain `string`. Totally reasonable for v1 — a UI renders these in a `<pre>` block and you're done. Just flagging that the proto shape forecloses structured rendering later: if the explain UI ever wants to highlight nodes, make them clickable, diff two plans side-by-side, or overlay metrics, the strings won't carry enough info. DataFusion has `display_pg_json()` for logical plans and `graphviz()` for both that produce structured / machine-readable output.

Two cheap ways to leave the door open without building anything now:
- Add an optional second representation field per plan (e.g., `string physical_plan_graphviz = 4`)
- Or use `oneof` so the format isn't baked in

Not blocking — just calling out that the proto is harder to change later than the renderer code is.

crates/coral-api/proto/coral/v1/query.proto
```diff
@@ -21,6 +21,31 @@ message ExecuteSqlResponse {
   int64 row_count = 2;
 }
 
+// Produces query-engine plan renderings for one SQL statement without executing
+// it.
+message PlanSqlRequest {
+  // Workspace that scopes the queryable source set.
+  Workspace workspace = 1;
+  // One read-only SQL statement.
+  string sql = 2;
```
antonmry, line 30: Related thought on the request side: `PlanSqlRequest` has no format selector, so clients always get whatever renderer the server picks. Fine for now — adding a `format` enum later is non-breaking (default = current behavior). Worth knowing that DataFusion ships a few alternatives that might be useful as the UI evolves:

- `DisplayableExecutionPlan::tree_render()` — newer, prettier physical-plan rendering (DF 44+, should be available given `datafusion-tracing` 53.x)
- `LogicalPlan::display_graphviz()` / `DisplayableExecutionPlan::graphviz()` — DOT format, useful if a visual rendering ever shows up

No action needed here — just flagging in case there's a planned UI shape that would benefit from one of those.

crates/coral-app/src/query/manager.rs
```diff
@@ -118,6 +118,20 @@ impl QueryManager {
         result
     }
 
+    pub(crate) async fn plan_sql(
```
antonmry: This bypasses the observability that `execute_sql` carefully sets up — no `Instant` timing, no `create_query_span`/`.instrument(...)`, no `metrics.count`/`metrics.duration`/`metrics.rows`, no `OtelStatus` recording. If the planning API is meant to back a UI (and the proto shape suggests it is), plan latency will be a primary user-perceived metric, and right now there's no signal on it.

Cheap to add by mirroring `execute_sql`'s instrumentation block — and much harder to retrofit once dashboards and alerts are built on the current (silent) shape. Not blocking, but worth doing in this PR while the structural symmetry is fresh.

antonmry (approved):

crates/coral-api/proto/coral/v1/query.proto
```diff
@@ -21,6 +21,31 @@ message ExecuteSqlResponse {
   int64 row_count = 2;
 }
 
+// Produces query-engine plan renderings for one SQL statement without executing
+// it.
+message PlanSqlRequest {
```
jsummerfield: This looks like product-facing EXPLAIN? I'd lean into that an name these ExplainSql etc.

crates/coral-engine/src/runtime/query.rs
```diff
@@ -171,6 +172,36 @@ impl QueryRuntimeAdapter {
         }
         Ok(())
     }
+
+    pub(crate) async fn plan_sql(&self, sql: &str) -> Result<QueryPlan, CoreError> {
+        let df = self
+            .ctx
+            .sql_with_options(sql, read_only_sql_options())
+            .await
+            .map_err(|err| datafusion_to_core(&err, &self.tables))?;
```
jsummerfield: `execute_sql` maps `sql_with_options` failures through `datafusion_to_core_with_sql(..., Some(sql))`, but `plan_sql` uses plain `datafusion_to_core` here. That drops the SQL text DataFusion diagnostic spans need to preserve quoted dotted identifiers, so cases like `SELECT * FROM "github.pulls"` lose the existing structured table-not-found hint covered by `quoted_fully_qualified_table_reference_reports_sql_reference_hint`. Please use `datafusion_to_core_with_sql(&err, &self.tables, Some(sql))` on this initial `sql_with_options` error path and add the same quoted fully-qualified table regression for `PlanSql`.

jsummerfield:
Anton's observability comment and the SQL-error comment point to the same design smell: `plan_sql` is becoming a parallel query lifecycle instead of an inspection mode of the existing query lifecycle.

`execute_sql` has accumulated real behavior: read-only SQL policy, SQL-aware error normalization, query spans and timing, metrics/status recording, source/runtime loading through `QueryManager`, and result observation. `plan_sql` currently redoes only the obvious DataFusion calls and misses the less obvious contract around them. That is exactly how sibling APIs drift.

I do not think that means we should delete `PlanSql`. I think it means Coral needs a shared query preparation/lifecycle path, with execution and explanation as two outcomes.

Smallest good shape:

- In `coral-engine`, factor the common `ctx.sql_with_options(sql, read_only_sql_options())` path into a helper that always uses the same SQL-aware error mapping.
- In `coral-app`, factor the `execute_sql` instrumentation wrapper into a query-operation lifecycle so `execute_sql` and `plan_sql` both get timing, span status, and metrics.
- Keep the final behaviors separate: `execute_sql` collects batches; `plan_sql` renders plan evidence.

I would avoid a grand abstraction here. The rule should be simple: if a user submits SQL to Coral, parsing, policy, structured errors, source loading, tracing, and metrics should be identical unless the divergence is intentional and named.

crates/coral-app/src/query/manager.rs
jsummerfield: It's interesting this is the 2nd place where we're drifting from `execute_sql`. I Left a top level PR comment with a suggestion.

jsummerfield (approved):

jsummerfield:
Nice!

PR #306, opened 2026-05-07
feat(engine): execute internal source UDTFs

Bradley-Butcher:
## Summary

Adds internal `DataFusion` UDTF execution for manifest-declared source table functions.

This is not the public `source.function(...)` SQL syntax. It is the engine execution substrate that turns an already-lowered internal UDTF call into an HTTP-backed table provider.

This PR and the next planner PR are intended to merge together. This branch is split out only to keep reviewable ownership boundaries: this PR handles internal execution; the next PR handles the public source-scoped SQL syntax.

## Why

The previous PR extracted the shared HTTP fetch target. This PR gives function specs a way to use that path.

Tables and functions bind SQL inputs differently:

- tables bind `WHERE` filters before scanning
- functions bind call arguments before scanning

After binding, both should use the same HTTP execution path: render request, paginate, decode response, and map rows into Arrow batches.

## What Changed

- HTTP manifests now register one hidden internal UDTF per function spec
- `HttpSourceTableFunction` validates positional call args and binds them into request values
- function scans use `http_json_exec` and `HttpFetchTarget::from_function(...)`
- function result `WHERE` predicates are not pushed down; `DataFusion` evaluates them locally
- SQL `LIMIT` still flows into HTTP fetching and pagination
- request rendering now supports function `arg`, `arg_int`, and `arg_bool` value sources
- non-HTTP backends return no table-function registrations
- internal UDTF names are hex-encoded to avoid collisions in `DataFusion`'s flat function registry

## Not in This PR

- public `github.find_issues(q => '...')` / `source.function(...)` planning; that is the next PR and should merge with this one
- bundled GitHub/Datadog function rollout
- MCP guide or docs changes
- provider-specific semantic argument design

## Validation

- `cargo test -p coral-engine --test engine table_function -- --nocapture`
- `make rust-checks`

### DIFF
--- a/crates/coral-engine/src/backends/http/mod.rs
+++ b/crates/coral-engine/src/backends/http/mod.rs
@@ -12,7 +12,7 @@ use datafusion::prelude::SessionContext;
 use crate::RequestAuthenticator;
 use crate::backends::{
     BackendCompileRequest, BackendRegistration, CompiledBackendSource, RegisteredSource,
-    RegisteredTable, build_registered_inputs, build_registered_table,
+    RegisteredTable, SourceTableFunctions, build_registered_inputs, build_registered_table,
     build_registered_table_function, internal_table_function_name, registered_columns_from_specs,
     required_filter_names,
 };
@@ -94,7 +94,7 @@ impl CompiledBackendSource for HttpCompiledSource {
             table_infos.push(registered_table(table));
         }
         let mut table_functions =
-            crate::SourceTableFunctions::with_capacity(self.manifest.functions.len());
+            SourceTableFunctions::with_capacity(self.manifest.functions.len());
         let mut table_function_infos = Vec::with_capacity(self.manifest.functions.len());
         for function in &self.manifest.functions {
             let internal_name =
--- a/crates/coral-engine/src/backends/http/mod.rs
+++ b/crates/coral-engine/src/backends/http/mod.rs
@@ -11,10 +11,10 @@ use datafusion::prelude::SessionContext;
 
 use crate::RequestAuthenticator;
 use crate::backends::{
-    BackendCompileRequest, BackendRegistration, BackendTableFunctionRegistration,
-    CompiledBackendSource, RegisteredSource, RegisteredTable, build_registered_inputs,
-    build_registered_table, build_registered_table_function, internal_table_function_name,
-    registered_columns_from_specs, required_filter_names,
+    BackendCompileRequest, BackendRegistration, CompiledBackendSource, RegisteredSource,
+    RegisteredTable, build_registered_inputs, build_registered_table,
+    build_registered_table_function, internal_table_function_name, registered_columns_from_specs,
+    required_filter_names,
 };
 use coral_spec::backends::http::{HttpSourceManifest, HttpTableSpec};
 pub(crate) mod auth;
@@ -93,20 +93,19 @@ impl CompiledBackendSource for HttpCompiledSource {
             tables.insert(table.name().to_string(), provider);
             table_infos.push(registered_table(table));
         }
-        let mut table_functions = Vec::with_capacity(self.manifest.functions.len());
+        let mut table_functions =
+            crate::SourceTableFunctions::with_capacity(self.manifest.functions.len());
         let mut table_function_infos = Vec::with_capacity(self.manifest.functions.len());
         for function in &self.manifest.functions {
             let internal_name =
                 internal_table_function_name(&self.manifest.common.name, &function.name);
-            let function_impl = function::HttpSourceTableFunction::new(
-                backend.clone(),
-                self.manifest.common.name.clone(),
-                function.clone(),
-            )?;
-            table_functions.push(BackendTableFunctionRegistration {
-                internal_name: internal_name.clone(),
-                function: Arc::new(function_impl),
-            });
+            let function_impl: Arc<dyn datafusion::catalog::TableFunctionImpl> =
+                Arc::new(function::HttpSourceTableFunction::new(
+                    backend.clone(),
+                    self.manifest.common.name.clone(),
+                    function.clone(),
+                )?);
+            table_functions.insert(internal_name, function_impl);
             table_function_infos.push(build_registered_table_function(
                 &self.manifest.common.name,
                 function,
--- a/crates/coral-engine/src/backends/http/mod.rs
+++ b/crates/coral-engine/src/backends/http/mod.rs
@@ -11,14 +11,16 @@ use datafusion::prelude::SessionContext;
 
 use crate::RequestAuthenticator;
 use crate::backends::{
-    BackendCompileRequest, BackendRegistration, CompiledBackendSource, RegisteredSource,
-    RegisteredTable, build_registered_inputs, build_registered_table,
-    build_registered_table_function, registered_columns_from_specs, required_filter_names,
+    BackendCompileRequest, BackendRegistration, BackendTableFunctionRegistration,
+    CompiledBackendSource, RegisteredSource, RegisteredTable, build_registered_inputs,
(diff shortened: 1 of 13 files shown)


Bradley-Butcher:
* **#306** <a href="https://app.graphite.com/github/pr/withcoral/coral/306?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/306?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-engine/src/backends/http/mod.rs
```diff
@@ -91,12 +93,25 @@ impl CompiledBackendSource for HttpCompiledSource {
...
-            .functions
-            .iter()
-            .map(|function| build_registered_table_function(&self.manifest.common.name, function))
-            .collect();
+        let mut table_functions = Vec::with_capacity(self.manifest.functions.len());
+        let mut table_function_infos = Vec::with_capacity(self.manifest.functions.len());
+        for function in &self.manifest.functions {
+            let internal_name =
+                internal_table_function_name(&self.manifest.common.name, &function.name);
+            let function_impl = function::HttpSourceTableFunction::new(
```
jsummerfield: This PR starts registering manifest-declared functions as executable UDTFs, but the existing HTTP source registration validation only walks table requests. A function request that references a missing `input` in its path/header/query/body can still be registered and advertised here, then fail late at execution time or silently omit `from: input` values. Please add the same input-dependency validation for `manifest.functions[*].request` before building/registering these UDTFs, with tests for missing function path/header/query/body inputs.

crates/coral-engine/src/backends/shared/template.rs
```diff
@@ -28,31 +28,18 @@ pub(crate) fn resolve_value_source(
             .get(key)
             .map(|v| Value::String(v.clone()))
             .or_else(|| default.clone())),
+        ValueSourceSpec::Arg { key, default } => Ok(filters
```
jsummerfield: This makes `arg` resolve from the same request-value map that table scans use for WHERE-derived filters. Function scans pass actual bound args through that map, but table scans pass filters, and source-scoped `request_headers` are shared by both paths. That means a malformed source-level header using `from: arg` / `{{arg.id}}` can now accidentally bind a table filter value instead of failing as “function argument outside a function request.” Please either reject `arg` in source-scoped rendering contexts, or split filter and function-arg values so the namespaces cannot alias each other.

crates/coral-engine/src/backends/http/mod.rs
jsummerfield: I think this points to a small cleanup rather than a broad new validation framework. The static shape checks in `coral-spec` and the resolved-input/auth checks in `coral-engine` are legitimately separate phases, but the engine-side pass currently has table-specific request traversal, so adding function requests created a second request-bearing surface that could drift. A targeted fix would be to make that pass iterate over all HTTP request sites — table default request, table route request, and function request — so future request surfaces get the same input-dependency checks by construction.

crates/coral-engine/tests/engine/http_tests.rs
```diff
@@ -42,6 +42,76 @@ fn base_http_manifest(name: &str, base_url: &str) -> Value {
...
+            },
+            "columns": [
+                { "name": "title", "type": "Utf8" },
+                { "name": "score", "type": "Float64" }
+            ]
+        }]
+    })
+}
+
+fn internal_table_function_name(schema: &str, function: &str) -> String {
```
jsummerfield, line 103: This helper is a sign that the internal UDTF naming seam is leaking. The engine deliberately hides source-scoped functions behind a flat DataFusion name, but the only way to execute the feature in tests is to duplicate that private encoding and call `__coral_udtf_*` directly. If table functions are meant to be externally invocable, the engine should own an invocation handle or rewrite from the public `(schema_name, function_name)` shape; if they are truly internal, the SQL generator should own the mapping through a shared helper instead of copying the encoding here.

crates/coral-engine/src/runtime/registry.rs
```diff
@@ -110,7 +113,10 @@ pub(crate) async fn register_sources(
                             compiled_source.schema_name(),
                             Arc::new(StaticSchemaProvider::new(decorated_tables)),
                         ) {
-                            Ok(_) => result.active_sources.push(registered_source),
+                            Ok(_) => {
+                                register_table_functions(ctx, table_functions);
```
jsummerfield, line 117: This introduces a second executable source surface after the table decorator seam has already run. Today `SourceDecorator::decorate_source` can wrap or replace table providers before registration, but table functions are registered separately and cannot be decorated, restricted, or observed through that same extension point. If decorators are meant to apply to source capabilities, include table functions in that seam; if they are intentionally table-only, rename/document the boundary so UDTFs do not accidentally bypass future policy or instrumentation decorators.

crates/coral-engine/src/backends/http/mod.rs
Bradley-Butcher: Addressed in the latest push. The engine-side resolved-input validation now iterates over HTTP request sites rather than only table requests: table default request, table route request, and function request. I also added coverage for missing function request inputs in path/query/header/body via `backend_client_rejects_unresolved_function_request_inputs`.

crates/coral-engine/src/backends/shared/template.rs
Bradley-Butcher: Addressed in the latest push. Request rendering now takes an explicit `RenderContext` with separate `filters`, `args`, `state`, and `resolved_inputs` maps. Tables pass filter values plus empty args; functions pass arg values plus empty filters. Added regression coverage for both directions: table filters no longer satisfy `arg.*`, and function args no longer satisfy `filter.*`.

crates/coral-engine/tests/engine/http_tests.rs
Bradley-Butcher: Left as an explicit stack seam for this PR. This branch only registers the flat internal DataFusion UDTF, so the internal execution test still needs a way to invoke that surface directly. The next PR in the stack adds the public source-scoped planner and tests the user-facing `source.function(...)` path, so callers do not need to know this encoding. I added a comment on the helper to make that boundary clear.

crates/coral-engine/src/runtime/registry.rs
Bradley-Butcher: Agreed this is a real capability/policy question, but I backed the attempted fix out of this PR. The latest push keeps `SourceDecorator` table-scoped and does not introduce the `SourceCapabilities` hook here. Table-function decoration/restriction/observation should land as a separate capability/decorator PR rather than being mixed into the HTTP UDTF execution change.

Bradley-Butcher:
Updated this PR to address the HTTP table/function drift comments.

What changed:
- Function requests now go through the same resolved-input validation path as table requests. The engine validates every HTTP request site: table default request, table route request, and function request.
- Request rendering now uses an explicit context with separate namespaces for table filters, function args, pagination state, and resolved source inputs. Tables pass filter values plus empty args; functions pass arg values plus empty filters.
- Added regression coverage so table filters cannot satisfy `arg.*`, function args cannot satisfy `filter.*`, and function request inputs fail at registration when missing.
- Backed out the broader decorator/capability hook change. Source decorators remain table-scoped in this PR; table-function decoration/policy should be a separate follow-up.

Also did a small naming cleanup in table scan code so table-derived values are called `filter_values` rather than generic `request_values`.

Validation:
- `cargo check -p coral-engine --tests --locked`
- `cargo test -p coral-engine --test engine request_headers_do_not_resolve --locked`
- `cargo test -p coral-engine backend_client_rejects_unresolved_function_request_inputs --locked`

jsummerfield (approved):

PR #308, opened 2026-05-07
ci(api): lint protobuf API with buf

jsummerfield:
## Summary

Add Buf as the protobuf contract lint gate for `coral-api` so API-shape and documentation drift are caught before merge. The new config enables Buf STANDARD plus explicit COMMENT_* rules, which forces RPC-specific response wrappers and documented enum values in the source service contract.

## Changes

- Add `crates/coral-api/buf.yaml` with the lint and breaking-change configuration.
- Update `SourceService` RPCs to return operation-specific response messages instead of shared domain messages or `google.protobuf.Empty`.
- Update Rust server, CLI, and test call sites for the generated response wrappers.
- Add `make lint-proto` and wire it into the Validate workflow as a path-filtered `proto-lint` job.

## Verification

- `buf lint`
- `make lint-proto`
- `make rust-checks`
- workflow YAML parse check
- `git diff --check`

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -25,6 +25,7 @@ jobs:
     outputs:
       github_actions: ${{ steps.filter.outputs.github_actions }}
       rust_checks: ${{ steps.filter.outputs.rust_checks }}
+      proto-lint: ${{ steps.filter.outputs.proto-lint }}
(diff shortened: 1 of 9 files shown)


antonmry (approved):

sauldhernandez (approved):
LGTM. Consider adding a CI check with `buf breaking` after proto changes stabilize.

PR #310, opened 2026-05-08
fix(spec): allow function-only HTTP manifests

jsummerfield:
## Why

Source-scoped HTTP table functions are meant to represent provider-native read operations that return rows without requiring a backing table. The new function spec and catalog support allowed that shape conceptually, but manifest parsing still required a `tables` block globally, so a function-only HTTP source failed schema validation before it could reach function validation or catalog discovery.

## What changed

- Allow HTTP manifests to define either `tables` or `functions`.
- Keep `tables` required for file-backed `jsonl` and `parquet` sources.
- Reject HTTP manifests that define neither tables nor functions.
- Add parser coverage for an HTTP manifest with functions and no tables.

## Verification

- `cargo test -p coral-spec parse_source_manifest_accepts_http_functions_without_tables --locked`
- `cargo test -p coral-spec --locked`
- `cargo test -p coral-spec validate_http_function --locked`
- `cargo test -p coral-engine --test engine catalog_tests::coral_table_functions_lists_source_functions --locked`
- `cargo test -p coral-engine --lib table_not_found --locked`
- `git diff --check`

### DIFF
--- a/crates/coral-spec/src/backends/http.rs
+++ b/crates/coral-spec/src/backends/http.rs
@@ -116,6 +116,7 @@ struct RawHttpSourceManifest {
     rate_limit: RateLimitSpec,
     #[serde(default)]
     inputs: Option<Value>,
+    #[serde(default)]
     tables: Vec<RawHttpTableSpec>,
     #[serde(default)]
     functions: Vec<SourceTableFunctionSpec>,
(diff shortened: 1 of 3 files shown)


Bradley-Butcher (approved):
Thanks had this initially but PRs were already bloated - was going to do it after but you've saved me the work!

simonwhitaker (approved):

PR #311, opened 2026-05-08
feat(spec): add search metadata hints

mystic123:
Add source-spec support for `kind: search`, `search_limits`, `detail_hints`, and filter metadata needed by retrieval-oriented sources.

Changes:
- Adds `search_limits` and `detail_hints` for HTTP tables/functions.
- Adds `kind: search` for source table functions and requires `search_limits` for search functions in Rust validation and JSON Schema.
- Validates detail hints against real target tables/filters, including result-column existence and compatible source/target data types.
- Aligns JSON Schema `filter.type` / `column.type` with Rust manifest data types.
- Marks `mode: search` as deprecated in the schema while continuing to accept it for compatibility tables.

Mode vs kind migration note:
- `mode: search` is the legacy filter-level marker for table-backed virtual filters. It binds SQL predicates into a table request. It remains accepted so existing manifests keep working, but new provider-native search surfaces should not use it.
- `kind: search` is the canonical surface-level marker for provider-native retrieval functions. It describes a callable operation that returns ranked candidate rows and must carry `search_limits`.
- Migration path: keep existing `mode: search` compatibility tables working, add `kind: search` functions for provider search endpoints, guide contributors toward functions, then remove/deprecate compatibility-table search filters after sources and docs have moved.

Stack dependencies:
- Follow-up validation hardening, including numeric caps and parquet filter parity, is split into #350 to keep this PR focused.
- Source-spec reference and MCP contributor guidance are intentionally split into #313. This PR should land as part of the stack, not independently for release, so the authoring docs move with the DSL additions.

Local validation:
- `cargo test -p coral-spec`
- `git diff --check`


mystic123:
* **#654** <a href="https://app.graphite.com/github/pr/withcoral/coral/654?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#610** <a href="https://app.graphite.com/github/pr/withcoral/coral/610?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#352** <a href="https://app.graphite.com/github/pr/withcoral/coral/352?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#313** <a href="https://app.graphite.com/github/pr/withcoral/coral/313?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#312** <a href="https://app.graphite.com/github/pr/withcoral/coral/312?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#350** <a href="https://app.graphite.com/github/pr/withcoral/coral/350?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#311** <a href="https://app.graphite.com/github/pr/withcoral/coral/311?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/311?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-spec/src/schema/source_manifest.schema.json
```diff
@@ -233,8 +293,14 @@
       "required": ["name", "request"],
       "properties": {
         "name": { "type": "string", "minLength": 1 },
+        "kind": { "type": "string", "enum": ["table", "search"] },
```
jsummerfield, line 319: [P2] This makes `kind: "search"` schema-valid without `search_limits`, but Rust validation rejects that shape in `validate_search_metadata` when `require_search_limits` is true. That leaves a schema/parser contract split where tooling can accept a manifest the parser refuses. Add an `if`/`then` schema rule requiring `search_limits` for `kind: "search"`, or relax the Rust requirement.

crates/coral-spec/src/schema/source_manifest.schema.json
```diff
@@ -223,8 +260,31 @@
       "required": ["name"],
       "properties": {
         "name": { "type": "string", "minLength": 1 },
+        "type": { "type": "string", "minLength": 1 },
```
jsummerfield: [P2] This accepts any non-empty filter type, while `FilterSpec::manifest_data_type()` only accepts the known manifest data types. The schema should enumerate those supported values so schema validation and Rust parsing agree; otherwise schema-valid manifests can still fail later, and file-backed filters get the same drift.

crates/coral-spec/src/validate.rs
```diff
@@ -186,6 +212,152 @@ pub(crate) fn validate_filters_and_column_exprs(
...
+                "{schema}.{} '{}' detail_hints",
+                source.surface_kind, source.surface_name
+            );
+            let Some(target) = resolve_detail_hint_target(schema, targets, &hint.table) else {
+                return Err(ManifestError::validation(format!(
+                    "{context} target table '{}' does not match any table in source '{schema}'",
+                    hint.table
+                )));
+            };
+            if !target
```
jsummerfield: [P2] This only verifies that the target filter exists. For `detail_hints` to be a reliable machine-readable path, validation also needs to compare the declared type of `search_result_column` against the declared type of the target `detail_filter` and reject incompatible pairs.

jsummerfield:
\>Legacy `mode: search` compatibility tables remain valid without forcing bundled provider migration metadata in this slice.  
  
It would be good to avoid acruing legacy cruft at this point. Can you remove it?

crates/coral-spec/src/schema/source_manifest.schema.json
mystic123: Fixed in this PR. The table-function schema now has an `if`/`then` rule that requires `search_limits` whenever `kind: "search"` is set, matching the Rust `validate_search_metadata(..., require_search_limits = true, ...)` path. I also added `validate_manifest_schema_rejects_search_function_without_search_limits` to keep the schema/Rust contract locked.

crates/coral-spec/src/schema/source_manifest.schema.json
mystic123: Fixed in this PR. `filter.type` now references the shared `manifest_data_type` enum instead of accepting any non-empty string, and the same enum is used for `column.type`. There is also a schema regression test for an unknown filter type. The file-backed Rust validation parity is further hardened in the follow-up PR #350.

crates/coral-spec/src/validate.rs
mystic123: Fixed in this PR. Detail-hint reference validation now resolves the declaring surface column and the target table filter, parses both declared manifest data types, and rejects incompatible pairs. Added `parse_http_manifest_rejects_detail_hint_type_mismatch` for that case.

mystic123:
Re: removing legacy `mode: search`: I did not remove it outright in this slice because it is the compatibility marker for existing table-backed virtual filters, while `kind: search` is the new canonical surface-level marker for provider-native retrieval functions.

What changed instead:
- `mode: search` is now explicitly deprecated in the JSON Schema and Rust docs/comments.
- New provider-native search should use `kind: search` table functions with `search_limits`.
- The stack now includes the real migration example in #352: Notion gets `notion.search_objects(query => ...)` as the canonical search function while `notion.search` remains only as the compatibility table.

So the migration path is: keep old manifests parsing, mark the old marker deprecated, add function-first provider search surfaces, then remove the compatibility search filters after sources have moved.

crates/coral-spec/src/schema/source_manifest.schema.json
```diff
@@ -159,6 +171,11 @@
           "items": { "$ref": "#/$defs/filter" }
         },
         "fetch_limit_default": { "type": "integer", "minimum": 1 },
+        "search_limits": { "$ref": "#/$defs/search_limits" },
+        "detail_hints": {
```
jsummerfield, line 175: [P2] This lets `detail_hints` be schema-valid while the parser can still reject them. `columns` remains optional, but Rust validation requires every `search_result_column` to exist in the declaring surface's columns. The new schema test currently accepts `search_result_column: id` without declaring an `id` column or a target filter, so schema-only tooling can approve a manifest that install-time parsing rejects. Either require/validate the referenced column shape in schema, or make the schema test go through `parse_source_manifest_yaml` so this drift is caught.

crates/coral-spec/src/schema/source_manifest.schema.json
```diff
@@ -217,24 +254,76 @@
...
+      "type": "string",
+      "enum": ["Utf8", "Int64", "Boolean", "Float64", "Timestamp", "Json"]
+    },
     "filter": {
       "type": "object",
       "additionalProperties": false,
       "required": ["name"],
       "properties": {
         "name": { "type": "string", "minLength": 1 },
+        "type": { "$ref": "#/$defs/manifest_data_type" },
```
jsummerfield, line 267: [P2] These are user-facing source-spec DSL additions, but the source-spec reference is not updated in this PR. The schema adds `filter.type`, `filter.description`, `kind`, `search_limits`, `detail_hints`, and deprecates `mode: search`; meanwhile `docs/reference/source-spec-reference.mdx` still documents the old table-function shape. If #313 is guaranteed to land with this stack before release, that is probably fine, but #311 should not merge independently with the authoring docs stale.

jsummerfield (commented):
Two follow-ups from a local review pass.

jsummerfield (approved):

crates/coral-spec/src/schema/source_manifest.schema.json
mystic123: Fixed in #311. The acceptance test now goes through `parse_source_manifest_yaml` instead of schema-only validation, and the fixture declares both the `id` result column and the target `id` filter referenced by `detail_hints`. That means this test now exercises the same schema + parser contract that install-time parsing uses.

crates/coral-spec/src/schema/source_manifest.schema.json
mystic123: Addressed in the PR metadata and stack shape. #311 now explicitly says source-spec reference and MCP contributor guidance are split into #313 and that #311 should land as part of the stack, not independently for release. #313 carries the source-spec/MCP docs updates for these DSL additions.

PR #322, opened 2026-05-08
fix(spec): validate inner expr for base64_decode

### LINKED ISSUE #318: bug(spec): base64_decode skips inner expr validation
While iterating on a custom HTTP source that used a derived `base64_decode` column, I hit what looks like a validation gap in `validate_expr`.

If the inner expression is a stale `from_filter` reference, `coral source lint` still reports the manifest as valid when that reference is nested under `base64_decode`.

Observed behavior:
- `coral source lint` accepts the manifest
- `coral source add --file ...` installs it
- querying the derived column returns blank/null at runtime

Expected behavior:
- validation should fail up front with the same kind of unknown-filter error that other wrapper expressions already surface

This seems to come from `ExprSpec::Base64Decode` not recursing into its inner `expr` in `crates/coral-spec/src/validate.rs`, unlike `format_timestamp`, `replace`, `coalesce`, and `template`.

I can open a small PR with a fix and regression test if that’s helpful.

siiddhantt:
Closes #318.

Validate the inner expression of base64_decode during source-spec validation so stale from_filter references fail at lint/install time instead of only surfacing later as blank derived values at query time.

Adds a regression test covering an unknown filter nested under base64_decode.

Validation:
- cargo fmt --all -- --check
- cargo clippy -p coral-spec --all-targets --all-features --locked -- -D warnings
- cargo test -p coral-spec --all-targets --all-features --locked

### DIFF
--- a/crates/coral-spec/src/validate.rs
+++ b/crates/coral-spec/src/validate.rs
@@ -487,7 +487,7 @@ fn validate_expr(expr: &ExprSpec, known_filters: &HashSet<String>, context: &str
                 "{context} references unknown filter '{filter_key}'"
             )));
         }
-        ExprSpec::FormatTimestamp { expr, .. } => {
+        ExprSpec::FormatTimestamp { expr, .. } | ExprSpec::Base64Decode { expr } => {
             validate_expr(expr, known_filters, context)?;
         }
         ExprSpec::Replace { expr, from, .. } => {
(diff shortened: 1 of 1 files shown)


simonwhitaker:
Thanks for the contribution! If you haven't done this already, please follow the instructions in [CONTRIBUTING.md](https://github.com/withcoral/coral/blob/main/CONTRIBUTING.md#licence) to agree to our contributor license agreement.

siiddhantt:
Thanks Simon, I’ve emailed `legal@withcoral.com` to confirm the individual CLA.

simonwhitaker (approved):
This is great @siiddhantt! Thanks for contributing to Coral, and fixing this gap.

PR #324, opened 2026-05-08
refactor(app): split source credential persistence

sauldhernandez:
## Intent

Move plaintext source credential material out of the general app state boundary and into a dedicated app-owned credentials boundary, while preserving the existing on-disk storage contract.

## What it enables

- Keeps source credential material owned by `crates/coral-app/src/credentials` instead of `state`.
- Introduces `CredentialManager` and `CredentialSetId` so source lifecycle and query loading depend on credential-set operations rather than source-secret store internals.
- Preserves the existing `workspaces/<workspace>/sources/<source>/secrets.env` path and `config.toml` `secrets` key behavior.
- Handles credential replacement without parsing malformed existing material, and keeps rollback/cleanup behavior explicit during source import and deletion.

## Usage examples

No user-facing command changes are required. Existing flows continue to use the same source lifecycle commands:

```bash
coral source add --file path/to/source.yaml
coral source test <source>
coral sql "select * from <source>.<table>"
```

Existing credential material remains under `workspaces/<workspace>/sources/<source>/secrets.env`.

## Validation

- `git diff --check $(git merge-base origin/main HEAD)`
- `cargo test -p coral-app`
- `make rust-checks`


sauldhernandez:
* **#324** <a href="https://app.graphite.com/github/pr/withcoral/coral/324?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/324?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jsummerfield:
I think this PR should either fix the lifecycle-level malformed credential case or narrow the claim about what has been fixed.

The store-level replacement path now avoids parsing existing material, which is good. But source replacement still calls `load_source_rollback_state()` before writing fresh material, and that snapshots credentials by parsing the existing `secrets.env`. So if an installed source has malformed credential material, `coral source add ...` still cannot repair it through the normal source lifecycle path. Delete has the same shape because it reads/parses existing material before removing the source.

This looks like preserved legacy behavior rather than a new regression, but because this PR is creating the credential boundary that OAuth will build on, I think the boundary should be honest here. A small way to do that would be to snapshot/restore credential material as opaque bytes for rollback instead of deserializing it into `BTreeMap<String, String>` before replacement/removal. Separately, `remove_material()` should probably take the same state lock as `replace_material()` and treat `NotFound` as success.

If fixing the lifecycle path is intentionally out of scope, the PR description/test coverage should be adjusted so it does not imply source replacement can recover from malformed existing credential material end-to-end.

jsummerfield (approved):

jsummerfield:
Thanks, the latest `fix(app): snapshot credential material for rollback` change addresses the lifecycle concern I raised.

The important bit is that rollback now snapshots/restores credential material as raw bytes instead of parsing it, so malformed existing `secrets.env` no longer blocks a normal source replacement or delete. `remove_material()` also now takes the state lock and treats missing files as success. The added tests for replacing and deleting sources with malformed credential material cover the case I was worried about.

No further concern from me on this PR.

crates/coral-app/src/sources/manager.rs
```diff
@@ -188,16 +190,23 @@ impl SourceManager {
...
                 )?),
             },
-            secrets: self
-                .secret_store
-                .read_source_secrets_for(workspace_name, source_name)?,
+            credential_material: self
+                .credential_manager
+                .snapshot_material(workspace_name, &credential_set_id)?,
         };
         if source_dir.exists()
```
jsummerfield: Small ownership cleanup: this deletes the source directory before routing credential deletion through `CredentialManager`, so the new credential-removal API is usually a no-op for source-backed material because `secrets.env` was already removed with the directory. I would flip the order and call `remove_material` before `remove_dir_all`; source lifecycle can still orchestrate the rollback, but the credential helper should own credential material deletion if we are introducing that boundary.

PR #343, opened 2026-05-11
feat(spec): model OAuth credential methods

sauldhernandez:
## Summary
- add manifest/protobuf models for secret credential retrieval methods
- expose OAuth method metadata through source discovery/info conversion
- document install-time credential methods separately from runtime auth

## Scope
This is the contract and metadata layer only. It does not run an OAuth flow or change CLI collection behavior yet; those are in the next PRs in the stack.

## Verification
- make rust-checks
- make docs-check


sauldhernandez:
* **#343** <a href="https://app.graphite.com/github/pr/withcoral/coral/343?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/343?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

simonwhitaker (approved):

jsummerfield:
Given the agreed phasing, I think this is the right scope for this PR.

I would not pull the future source -> surfaces -> operations/entities -> projections model into this change. Keeping OAuth as `credential.methods` metadata on `kind: secret` inputs is a reasonable phase-1 bridge as long as the contract remains explicit that this only controls install-time secret collection, not runtime request auth or the eventual connection model.

The validation/docs changes are doing the important work here: credentials only attach to secret inputs, `source_config` and `oauth` method shapes are rejected when mixed, OAuth flow/PKCE enums are explicit rather than defaulted, and the docs now separate collected inputs from runtime `auth`. That is enough for this layer to merge as current-manifest-compatible plumbing.

jsummerfield:
One correction to my earlier scope comment: the phase-1 manifest scope still looks right, but since the gRPC contract is internal and we are comfortable reshaping it, I think this PR should use `oneof` for the proto input shape instead of `kind` plus fields that are only valid for some kinds.

Right now `SourceInputSpec` can represent invalid states like a variable input with `credential`, or a secret input with `default_value`, and relies on Rust-side validation/decoding to keep those combinations out. A `oneof` would make the contract match the actual model:

```proto
message SourceInputSpec {
  string key = 1;
  bool required = 2;
  string hint = 3;

  oneof input {
    SourceVariableInput variable = 4;
    SourceSecretInput secret = 5;
  }
}

message SourceVariableInput {
  string default_value = 1;
}

message SourceSecretInput {
  SourceCredential credential = 1;
}
```

Same argument applies to credential methods: `oneof method { SourceConfigCredentialMethod source_config = ...; OAuthCredentialMethod oauth = ...; }` is cleaner than `type` plus an optional `oauth` payload. That is not pulling in the future source/surface/projection model; it is just making the current phase-1 contract harder to misuse.

crates/coral-api/proto/coral/v1/sources.proto
```diff
@@ -74,6 +74,145 @@ message SourceInputSpec {
   string default_value = 4;
   // Optional authored hint shown to the user while collecting this input.
   string hint = 5;
+  // Optional credential retrieval choices for secret inputs. If absent, clients
+  // collect the secret value as if a source_config method were declared.
+  SourceCredential credential = 6;
```
jsummerfield: [P1] Since this gRPC contract is internal and we are comfortable reshaping it, I think `SourceInputSpec` should model variable vs secret with a `oneof` instead of `kind` plus sibling fields. This shape can encode invalid states (`variable` + `credential`, `secret` + `default_value`, unspecified kind with populated fields) and then requires every decoder to remember the same cross-field rules. A stricter shape would be `oneof input { SourceVariableInput variable = ...; SourceSecretInput secret = ...; }`, with `default_value` only on the variable payload and `credential` only on the secret payload.

crates/coral-api/proto/coral/v1/sources.proto
```diff
@@ -74,6 +74,145 @@ message SourceInputSpec {
...
+// Credential retrieval choices for one source secret input.
+message SourceCredential {
+  // Authored retrieval methods in display order.
+  repeated SourceCredentialMethod methods = 1;
+}
+
+// One credential retrieval method.
+message SourceCredentialMethod {
+  // Method kind.
+  SourceCredentialMethodType type = 1;
```
jsummerfield: [P1] Same issue here: `type` plus optional variant payloads makes invalid states representable. This should be a `oneof method` with a `source_config` branch and an `oauth_authorization_code` branch. That keeps phase 1 scoped to current source inputs, but makes the transport contract match the actual mutually-exclusive method model.

crates/coral-api/proto/coral/v1/sources.proto
```diff
@@ -74,6 +74,145 @@ message SourceInputSpec {
...
+  // PKCE must be used for the OAuth flow.
+  O_AUTH_CREDENTIAL_PKCE_MODE_REQUIRED = 1;
+  // PKCE is not used for the OAuth flow.
+  O_AUTH_CREDENTIAL_PKCE_MODE_DISABLED = 2;
+}
+
+// OAuth flow settings.
+message OAuthCredentialFlow {
+  // Flow kind.
+  OAuthCredentialFlowType type = 1;
```
jsummerfield: [P2] This `flow.type` enum feels like premature abstraction. `OAuthCredentialMethod` is already authorization-code-specific and there is only one valid flow type, so the enum creates a second discriminant without adding real choice. I’d either rename the enclosing payload to `OAuthAuthorizationCodeCredentialMethod` and inline `pkce`, or make `flow` itself a `oneof` once another OAuth flow actually exists.

crates/coral-api/proto/coral/v1/sources.proto
```diff
@@ -74,6 +74,145 @@ message SourceInputSpec {
...
+  // Client ID resolution configuration.
+  OAuthCredentialClientId id = 1;
+  // Optional confidential-client secret configuration.
+  OAuthCredentialClientSecret secret = 2;
+}
+
+// OAuth client ID resolution configuration.
+message OAuthCredentialClientId {
+  // Optional manifest-authored default client ID.
+  string default = 1;
```
jsummerfield: [P2] I’d avoid `default` as a proto field name. It forces awkward generated names like Rust’s `r#default`, and it is vague at the call site. `default_value`, `default_client_id`, or `manifest_default_client_id` would be clearer.

crates/coral-api/proto/coral/v1/sources.proto
```diff
@@ -74,6 +74,145 @@ message SourceInputSpec {
...
+  OAuthCredentialClient client = 4;
+  // Optional scope parameter configuration.
+  OAuthCredentialScopes scopes = 5;
+}
+
+// Supported OAuth credential flow kinds.
+enum OAuthCredentialFlowType {
+  // Flow kind is unspecified. This is invalid for OAuth credential methods;
+  // clients must reject it instead of inferring a default flow.
+  O_AUTH_CREDENTIAL_FLOW_TYPE_UNSPECIFIED = 0;
```
jsummerfield: [P3] The `O_AUTH_...` enum prefix reads like the acronym got mechanically split. If these enums stay, prefer `OAUTH_CREDENTIAL_...` names; generated APIs and JSON/debug output will be much easier to read.

jsummerfield (commented):
Inline proto review.

jsummerfield (approved):

PR #347, opened 2026-05-11
feat: error message audit

iaptsiauri:
This is a proposal for error message audit it's aim is to provide user with the following 

- A clear, jargon-free message explaining what went wrong
- A hint field with an actionable next step
- Consistent formatting across CLI and server errors

What this PR does:
  - Adds a reusable `CliError::Diagnostic` path instead of many one-off CLI error variants.
  - Preserves existing semantic errors for `SourceNotInstalled` and `SourceNotFound`.
  - Improves source `install/test/lint` errors for missing inputs, invalid source names, manifest read/parse failures, server startup/connect failures, and non-query server statuses.
  - Reuses structured server ErrorInfo when available instead of falling back to unstructured anyhow output.
  - Quotes shell paths in interactive command hints so paths with spaces or shell meta characters are copyable.
  
  ```
   coral sql 'SELECT * FROM github.issuues LIMIT 1'
  └ Error: Table `github.issuues` not found
    Detail: No table `issues` exists in schema `github`.
    Hint: No source tables are currently queryable. Run `coral source discover`
    
    
    coral sql 'SELEC 1'
  └ Error: SQL query could not be parsed
    Detail: sql parser error: Expected: an SQL statement, found: SELEC at Line: 1, Column: 1
    Hint: Check the SQL syntax and retry. Use `SELECT * FROM coral.tables LIMIT 10` to inspect available tables.
```

- [ ] Run benchmarks, and check the impact of this change

### DIFF
--- a/crates/coral-app/src/query/manager.rs
+++ b/crates/coral-app/src/query/manager.rs
@@ -5,6 +5,7 @@ use std::future::Future;
 use std::sync::Arc;
 use std::time::Instant;
 
+use coral_api::{CORAL_ERROR_REASON_INVALID_INPUT, CORAL_ERROR_REASON_SOURCE_NOT_FOUND};
 use coral_engine::{
     CatalogInfo, CoralQuery, CoreError, QueryExecution, QueryPlan, QueryRuntimeConfig,
     QueryRuntimeContext, QuerySource, SourceValidationReport, StatusCode, TableInfo,
@@ -346,8 +347,8 @@ fn query_error_message(error: &QueryManagerError) -> String {
 
 fn app_error_type(error: &AppError) -> &'static str {
     match error {
-        AppError::SourceNotFound(_) => "SOURCE_NOT_FOUND",
-        AppError::InvalidInput(_) => "INVALID_INPUT",
+        AppError::SourceNotFound(_) => CORAL_ERROR_REASON_SOURCE_NOT_FOUND,
+        AppError::InvalidInput(_) => CORAL_ERROR_REASON_INVALID_INPUT,
         AppError::FailedPrecondition(_) => "FAILED_PRECONDITION",
         AppError::Io(_) => "IO",
         AppError::Yaml(_) => "YAML",
--- a/crates/coral-app/src/query/manager.rs
+++ b/crates/coral-app/src/query/manager.rs
@@ -167,9 +167,11 @@ impl QueryManager {
         workspace_name: &WorkspaceName,
     ) -> Result<Vec<QuerySource>, AppError> {
         let catalog = self.config_store.load_catalog()?;
+        let sources = catalog.workspace_sources(workspace_name);
         let mut query_sources = Vec::new();
-        for source in catalog.workspace_sources(workspace_name) {
-            match self.load_query_source(workspace_name, &source) {
+        let mut load_failures = Vec::new();
+        for source in &sources {
+            match self.load_query_source(workspace_name, source) {
                 Ok((query_source, _version)) => query_sources.push(query_source),
                 Err(error @ AppError::Credentials(CredentialsError::Unavailable(_))) => {
                     return Err(error);
@@ -180,8 +182,16 @@ impl QueryManager {
                         detail = %error,
                         "skipping source during query-source load"
                     );
+                    load_failures.push(format!("{}: {error}", source.name));
                 }
             }
+        }
+        if !sources.is_empty() && query_sources.is_empty() {
+            return Err(AppError::FailedPrecondition(format!(
+                "no installed sources could be loaded for workspace '{}': {}",
+                workspace_name.as_str(),
+                load_failures.join("; ")
+            )));
         }
         Ok(query_sources)
     }
--- a/crates/coral-cli/src/lib.rs
+++ b/crates/coral-cli/src/lib.rs
@@ -418,24 +418,60 @@ fn render_diagnostic(
 
 fn render_status_diagnostic(status: &tonic::Status) -> String {
     match decode_status_error(status) {
-        DecodedStatusError::Structured(error) => render_diagnostic(
-            error.summary,
-            if error.detail.is_empty() {
+        DecodedStatusError::Structured(error) => {
+            let coral_client::CoralQueryError {
+                reason,
+                summary,
+                detail,
+                hint: server_hint,
+                ..
+            } = *error;
+            let hint = cli_hint_for_app_reason(&reason).map_or_else(
+                || server_hint.unwrap_or_else(|| default_server_status_hint().to_string()),
+                str::to_string,
+            );
+            let detail = if detail.is_empty() {
                 status.message().to_string()
             } else {
-                error.detail
-            },
-            error.hint.unwrap_or_else(|| {
-                "Retry the command. If it keeps failing, check source setup and local Coral state."
-                    .to_string()
-            }),
-        ),
+                detail
+            };
+            render_diagnostic(summary, detail, hint)
+        }
         DecodedStatusError::Plain(message) => render_diagnostic(
             "Coral server request failed",
             message,
-            "Retry the command. If it keeps failing, check source setup and local Coral state.",
+            default_server_status_hint(),
         ),
     }
+}
+
+pub(crate) fn cli_hint_for_app_reason(reason: &str) -> Option<&'static str> {
+    match reason {
+        "SOURCE_NOT_FOUND" => Some(
+            "Run `coral source list` to see installed sources or `coral source discover` to see bundled sources available to install.",
+        ),
+        "INVALID_INPUT" => Some(
+            "Check the command input and retry. Run `coral --help` or the subcommand help for valid values.",
+        ),
+        "SETUP_REQUIRED" => Some(
+            "Run `coral source list` to inspect configured sources, then `coral source test <source>` for the source you are trying to use.",
+        ),
+        "INVALID_SECRETS_FILE" => Some(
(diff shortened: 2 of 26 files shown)

### CODE REVIEW

File: crates/coral-engine/src/contracts/query_error.rs [hunk 1]
```diff
@@ -130,6 +136,35 @@ impl StructuredQueryError {
...
+    }
+
+    /// Builds a structured `SQL_PARSE_ERROR` error.
+    pub(crate) fn sql_parse_error(detail: &str) -> Self {
+        Self::new(
+            SQL_PARSE_ERROR_REASON,
+            "SQL query could not be parsed",
+            detail,
+            Some(
+                "Check the SQL syntax and retry. Use `SELECT * FROM coral.tables LIMIT 10` to inspect available tables."
```
mystic123: I don't thinkg that this tip is useful here. we have a separate path/case when names are mispelled/missing and we display similar hint there.  
This error covers everything else, so we don't know much about the users' intention, right?

mystic123 (changes requested):

crates/coral-cli/src/lib.rs
```diff
@@ -204,37 +208,159 @@ pub enum CliError {
...
+                format!(
+                    "Run `coral source add {source_name}`, then retry `coral source test {source_name}`."
+                ),
             )),
-            Self::SourceNotFound { source_name } => Some(format!(
-                "source '{source_name}' was not found. Run `coral source list` to see installed sources or `coral source discover` to see bundled sources available to install.\n"
+            Self::SourceNotFound { source_name } => Some(render_diagnostic(
+                format!("Source `{source_name}` was not found"),
+                format!("No installed or bundled source named `{source_name}` exists."),
+                "Run `coral source list` to see installed sources or `coral source discover` to see bundled sources available to install.",
```
mystic123, line 385: `SourceNotFound` is also used by `source remove`, where a bundled-but-not-installed source is a normal case. In a fresh workspace `coral source remove github` now renders `Detail: No installed or bundled source named github￼exists` even though `source discover` lists github; use a remove-specific/not-installed diagnostic or make this generic detail not claim bundled nonexistence.

mystic123 (approved):

crates/coral-cli/src/source_ops.rs
```diff
@@ -139,25 +139,50 @@ async fn validate_source_request(
...
             workspace: Some(default_workspace()),
-            name: source_name_arg(Some(name))?,
+            name: normalized.clone(),
         }))
-        .await?
-        .into_inner();
+        .await
+    {
+        Ok(response) => response.into_inner(),
+        Err(status) if status.code() == tonic::Code::NotFound => {
```
sauldhernandez: Checking only `status.code() == NotFound` is too broad here. The app also maps local file failures with `io::ErrorKind::NotFound` to gRPC `NotFound` with reason `LOCAL_FILE_ERROR`, for example an installed imported source whose manifest file is missing. This branch would report the source as missing and drop the repair hint. Decode the structured status and only normalize `SOURCE_NOT_FOUND`; otherwise preserve `CliError::server_status(&status)`.

crates/coral-cli/src/source_ops.rs
```diff
@@ -211,42 +236,68 @@ fn print_source_info_response(source: &SourceInfo, verbose: bool) {
...
         .delete_source(Request::new(DeleteSourceRequest {
             workspace: Some(default_workspace()),
-            name: source_name_arg(Some(name))?,
+            name: normalized.clone(),
         }))
-        .await?;
+        .await
+    {
+        Ok(_) => Ok(()),
+        Err(status) if status.code() == tonic::Code::NotFound => {
```
sauldhernandez: Same issue here: `delete_source` can return `NotFound` for local file failures while removing an imported source, not just for an unknown source. Normalizing every `NotFound` hides the real `LOCAL_FILE_ERROR` diagnostic and can tell the user a listed source does not exist. Gate this on structured reason `SOURCE_NOT_FOUND`, and preserve other NotFound statuses with `server_status`.

crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -130,6 +136,32 @@ impl StructuredQueryError {
         }
     }
 
+    /// Builds a structured `EMPTY_SQL` error.
+    pub(crate) fn empty_sql() -> Self {
+        Self::new(
+            EMPTY_SQL_REASON,
+            "SQL query is empty",
+            "Coral cannot run an empty SQL string.",
+            Some("Try `coral sql \"SELECT * FROM coral.tables LIMIT 10\"`.".to_string()),
```
sauldhernandez: `StructuredQueryError` is an engine-level contract and this hint is forwarded unchanged to non-CLI consumers unless an adapter overrides it. Embedding `coral sql ...` here means MCP/SDK callers can receive a CLI command as remediation. Keep this hint transport-neutral, e.g. suggest the sample SQL statement only, and let CLI/MCP render surface-specific wording.

crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -345,6 +377,12 @@ fn table_not_found_hint(
     table: &str,
     known_tables: &[TableInfo],
 ) -> Option<String> {
+    if known_tables.is_empty() {
+        return Some(
+            "No source tables are currently queryable. Run `coral source discover`, then `coral source add <source>` to connect a source.".to_string(),
```
sauldhernandez: This no-tables hint is also engine-level structured output, so MCP/SDK callers will receive `coral source discover` / `coral source add` commands when no source tables are registered. Keep the engine hint surface-neutral, for example explain that no source tables are queryable and point callers at table/source discovery abstractly; adapters can add CLI-specific commands.

sauldhernandez (changes requested):
Reviewed with Review Buddy.

simonwhitaker (approved):

crates/coral-app/src/query/manager.rs
```diff
@@ -179,9 +182,17 @@ impl QueryManager {
                         detail = %error,
                         "skipping source during query-source load"
                     );
+                    load_failures.push(format!("{}: {error}", source.name));
                 }
             }
         }
+        if !sources.is_empty() && query_sources.is_empty() {
```
sauldhernandez, line 189: This breaks the current contract: if you attempt to execute a query that only uses introspection (e.g. coral catalog) and the workspace fails to load any real source, we will fail in this guard instead of actually executing a query we could've answered (since it didn't rely on any source)

crates/coral-cli/src/lib.rs
```diff
@@ -328,24 +334,55 @@ pub enum CliError {
         /// Normalized source name requested by the user.
         source_name: String,
     },
+    /// A non-query command failed with a rendered user-facing diagnostic.
+    #[error("command failed")]
+    Diagnostic {
```
sauldhernandez, line 339: The name `Diagnostic` does not seem to align with the other names in this enum which convey "types" of error, and instead seems to be more about how the error will be displayed/rendered. That smells like the wrong place for this.

If the intention is to shape out the rendering as the error bubbles up, consider something along the lines of `DiagnosedError` and keep the original error inside as part of the struct as well.

crates/coral-cli/src/lib.rs
```diff
@@ -355,13 +392,131 @@ impl CliError {
...
+        _ => None,
+    }
+}
+
+fn cli_auth_refresh_hint(metadata: &std::collections::HashMap<String, String>) -> String {
+    let source = metadata
+        .get("source")
+        .map_or_else(|| "<source>".to_string(), |source| shell_quote_arg(source));
+    format!(
+        "Run `coral source add {source} --interactive` to refresh saved credentials for bundled sources. For imported sources, run `coral source add --file <manifest-path> --interactive` with the manifest used to install the source."
```
sauldhernandez, line 501: This may be a bit vague now that we have OAuth with auto-refresh. Those creds should refresh with no user interaction, but manually inputed tokens will need updating as stated here.

crates/coral-cli/src/query_error.rs
```diff
@@ -42,19 +42,36 @@ pub(crate) fn telemetry_error_message(status: &tonic::Status) -> String {
...
     match code {
-        tonic::Code::InvalidArgument => "invalid argument",
-        tonic::Code::NotFound => "not found",
-        tonic::Code::FailedPrecondition => "failed precondition",
-        tonic::Code::Unavailable => "unavailable",
-        tonic::Code::Unimplemented => "unimplemented",
-        tonic::Code::Internal => "internal error",
-        _ => "error",
+        tonic::Code::InvalidArgument => (
+            "Query request is invalid",
```
sauldhernandez, line 52: I can see some of these been applicable to more surfaces than the CLI. Is the intention for each surface to fully control their error surface and not share common errors? Or is there room for sharing?

crates/coral-cli/src/lib.rs
```diff
@@ -328,24 +334,55 @@ pub enum CliError {
         /// Normalized source name requested by the user.
```
sauldhernandez, line 334: <p class='graphite__hidden'><i>[Re: line +333]</i></p>
Agent says this is dead code now. Can it be removed?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/withcoral/coral/347?utm_source=unchanged-line-comment">Graphite</a>.</p>

sauldhernandez (changes requested):

Bradley-Butcher:
Is this still alive? Are you still wanting a review from me?

PR #351, opened 2026-05-11
chore(cli): show git SHA in version output

simonwhitaker:
## Summary

Recreates the intent of #116 on current `main`.

- embeds the current short git SHA from the `coral-cli` build script
- includes that SHA in `coral --version` as SemVer build metadata (`+<sha>`)
- resolves git watch paths with `git rev-parse --git-path` so worktrees watch the common gitdir refs correctly
- keeps the existing embedded UI rebuild hints in the same build script

## Validation

- `make rust-checks`
- `cargo run -p coral-cli -- --version` -> `coral 0.2.0+96daa462`

### DIFF
--- a/crates/coral-cli/src/lib.rs
+++ b/crates/coral-cli/src/lib.rs
@@ -42,7 +42,11 @@ use tempfile as _;
 const DEFAULT_SERVER_PORT: u16 = 1457;
 
 #[derive(Debug, Parser)]
-#[command(name = "coral", version, arg_required_else_help = true)]
+#[command(
+    name = "coral",
+    version = concat!(env!("CARGO_PKG_VERSION"), "+", env!("CORAL_GIT_SHA")),
+    arg_required_else_help = true
+)]
 /// A local-first SQL interface for APIs, files, and other data sources.
 struct Cli {
     #[command(subcommand)]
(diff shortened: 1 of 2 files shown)


crates/coral-cli/src/lib.rs
```diff
@@ -42,7 +42,11 @@ use tempfile as _;
 const DEFAULT_SERVER_PORT: u16 = 1457;
 
 #[derive(Debug, Parser)]
-#[command(name = "coral", version, arg_required_else_help = true)]
+#[command(
+    name = "coral",
+    version = concat!(env!("CARGO_PKG_VERSION"), "-", env!("CORAL_GIT_SHA")),
```
sauldhernandez: Using `-<sha>` makes the reported version SemVer prerelease syntax. A release build for `0.2.0` will print something like `0.2.0-5914f96`, which SemVer parsers order below `0.2.0`; update checks or scripts parsing `coral --version` can treat the released binary as older than the release. Use SemVer build metadata (`0.2.0+<sha>`) or put the SHA outside the version token.

sauldhernandez (approved):
LGTM, just be mindful about SemVer parsing. See inline comment.

crates/coral-cli/src/lib.rs
simonwhitaker: Ah good catch, thanks Saúl. Fixing now.

PR #350, opened 2026-05-11
feat(spec): harden search metadata validation

mystic123:
Harden the search metadata spec validation added in #311.

Changes:
- Caps `search_limits` values in Rust and JSON Schema.
- Rejects aggregate candidate budgets above 10000 with checked multiplication.
- Keeps unsupported future backend table shapes from falling through schema table validation.
- Keeps backend-specific schemas aligned with backend parsers, including rejecting file-only `source` on HTTP tables.
- Validates parquet filter types with the same filter/type validator used elsewhere.
- Adds focused regression tests for the new rejection paths.

Local validation:
- `cargo fmt --all -- --check`
- `cargo check -p coral-spec`
- `cargo clippy -p coral-spec --all-targets -- -D warnings`
- `cargo test -p coral-spec`
- `git diff --check`

PR #313, opened 2026-05-08
docs(mcp): document search retrieval metadata

mystic123:
Document search/retrieval metadata in the source-spec reference and MCP guidance.

Changes:
- Documents `kind: search`, `search_limits`, filter metadata, and compatibility-table search metadata.
- Updates MCP guide/resource text to point agents at `coral.columns`, `coral.filters`, `coral.table_functions`, and `coral.inputs`.
- Keeps source-scoped search functions as the canonical retrieval surface and treats `mode: search` table filters as deprecated compatibility support.
- Leaves detail-hint catalog guidance out of the agent-facing docs for now; existing migrated sources do not require it.

The implementation plan file is intentionally not included in this PR.

Local validation:
- `cargo fmt --check`
- `cargo check -p coral-engine -p coral-mcp`
- `cargo test -p coral-mcp`
- `cargo run --locked -p xtask -- generate-docs --check`
- `git diff --check`

PR #312, opened 2026-05-08
feat(engine): expose search metadata catalog

mystic123:
Expose search/retrieval metadata through SQL catalog surfaces.

Why this metadata exists:
- The source spec now has provider-native search functions and bounded search behavior, but generic agents/tools cannot reliably use those semantics if they only exist in source YAML or prose docs.
- This metadata makes the retrieval contract queryable through the catalog: agents can ask what search functions exist, how to call them, what limits apply, which result columns come back, and which filters are available for follow-up table queries.
- This is the catalog layer needed by mechanical discovery flows such as `list_catalog` / `search_catalog`, without hardcoding provider-specific knowledge or relying on prose.
- Detail-hint catalog exposure is intentionally deferred: existing migrated sources are unambiguous enough through search result ids plus catalog-described tables, while the source-spec support can remain available for future providers with thinner or more ambiguous search results.

Changes:
- Adds `kind`, `arguments_json`, `result_columns_json`, and `search_limits_json` to `coral.table_functions`.
- Adds `search_limits_json` to `coral.tables`.
- Adds normalized `coral.filters` metadata, including filter type, mode, and descriptions.
- Exposes nullable `coral.columns.filter_mode` for columns that correspond to declared filters.

Local validation:
- `cargo fmt --check`
- `cargo check -p coral-engine -p coral-mcp`
- `cargo test -p coral-engine --test engine catalog`
- `cargo test -p coral-mcp`
- `cargo test -p coral-spec`
- `cargo run --locked -p xtask -- generate-docs --check`
- `git diff --check`

PR #352, opened 2026-05-11
feat(engine): execute Notion search as table function

mystic123:
## Summary
- Migrates Notion's real provider search surface to the executable source-scoped table function `notion.search_objects(query => ...)`.
- Uses Notion's API argument name `query` and binds it to the `/v1/search` request body.
- Adds `kind: search` and `search_limits` for the Notion provider search function.
- Keeps the legacy `notion.search` table as a deprecated compatibility surface and points its guide to the new function.
- Updates the HTTP engine fixture so function tests exercise `kind: search` plus `search_limits`.

## Syntax
```sql
SELECT object, id, url, public_url
FROM notion.search_objects(query => 'Coral')
LIMIT 10;
```

Use returned ids with the catalog-described Notion tables when full details are needed, for example:

```sql
SELECT *
FROM notion.pages
WHERE page_id = '<returned id>';
```

## Local validation
- `cargo fmt --check`
- `cargo check -p coral-engine -p coral-mcp`
- `cargo test -p coral-engine --test engine catalog`
- `cargo test -p coral-mcp`
- `cargo test -p coral-spec`
- `cargo run --locked -p xtask -- generate-docs --check`
- `git diff --check`

### DIFF
--- a/sources/core/notion/manifest.yaml
+++ b/sources/core/notion/manifest.yaml
@@ -34,13 +34,135 @@ request_headers:
     value: "2026-03-11"
 test_queries:
   - SELECT * FROM notion.search LIMIT 1
+  - SELECT * FROM notion.search_objects(query => 'Coral') LIMIT 1
+functions:
+  - name: search_objects
+    kind: search
+    description: Provider-ranked search results for pages and data sources shared with the integration
+    fetch_limit_default: 10
+    search_limits:
+      default_top_k: 10
+      max_top_k: 100
+      max_calls_per_query: 1
+    args:
+      - name: query
+        required: true
+        bind:
+          arg: query
+    request:
+      method: POST
+      path: /v1/search
+      body:
+        - path:
+            - query
+          from: arg
+          key: query
+    response:
+      rows_path:
+        - results
+    pagination:
+      mode: cursor_body
+      cursor_body_path:
+        - start_cursor
+      response_cursor_path:
+        - next_cursor
+      page_size:
+        default: 10
+        max: 100
+        body_path:
+          - page_size
+    columns:
+      - name: object
+        type: Utf8
+        nullable: false
+        description: Result object type (page or data_source)
+        expr:
+          kind: path
+          path:
+            - object
+      - name: id
+        type: Utf8
+        nullable: false
+        description: Notion object ID
+        expr:
+          kind: path
+          path:
+            - id
+      - name: created_time
+        type: Timestamp
+        nullable: true
+        description: Creation timestamp
+        expr:
+          kind: format_timestamp
+          input: iso8601
+          expr:
+            kind: path
+            path:
+              - created_time
+      - name: last_edited_time
+        type: Timestamp
+        nullable: true
+        description: Last edit timestamp
+        expr:
+          kind: format_timestamp
+          input: iso8601
+          expr:
+            kind: path
+            path:
+              - last_edited_time
+      - name: url
+        type: Utf8
+        nullable: true
+        description: Notion URL for page results
+        expr:
+          kind: path
+          path:
+            - url
+      - name: public_url
+        type: Utf8
+        nullable: true
+        description: Public URL for page results, when available
+        expr:
+          kind: path
+          path:
+            - public_url
+      - name: in_trash
+        type: Boolean
+        nullable: true
+        description: Whether the result is in trash
+        expr:
+          kind: path
+          path:
+            - in_trash
+      - name: parent
+        type: Json
+        nullable: true
+        description: Parent object
+        expr:
+          kind: path
+          path:
+            - parent
+      - name: properties
+        type: Json
+        nullable: true
+        description: Page or data source properties object
+        expr:
+          kind: path
+          path:
+            - properties
+      - name: raw
+        type: Json
+        nullable: false
+        description: Full raw Notion search result
+        expr:
+          kind: current_row
 tables:
   - name: search
     description: Pages and data sources shared with the integration
     guide: |
-      Start here to discover `page_id`, `database_id`, and `data_source_id`
-      values. Use `query` for title search and `object` with `page` or
-      `data_source` to narrow result types.
+      Compatibility table for discovering `page_id`, `database_id`, and
+      `data_source_id` values. Prefer `notion.search_objects(query => ...)`
+      for provider-ranked retrieval.
     filters:
       - name: query
         required: false
--- a/sources/core/notion/manifest.yaml
+++ b/sources/core/notion/manifest.yaml
@@ -49,6 +49,12 @@ functions:
         required: true
         bind:
           arg: query
+      - name: object
+        values:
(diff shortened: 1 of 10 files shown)


mystic123:
* **#654** <a href="https://app.graphite.com/github/pr/withcoral/coral/654?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#610** <a href="https://app.graphite.com/github/pr/withcoral/coral/610?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#352** <a href="https://app.graphite.com/github/pr/withcoral/coral/352?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/352?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#313** <a href="https://app.graphite.com/github/pr/withcoral/coral/313?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#312** <a href="https://app.graphite.com/github/pr/withcoral/coral/312?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#350** <a href="https://app.graphite.com/github/pr/withcoral/coral/350?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#311** <a href="https://app.graphite.com/github/pr/withcoral/coral/311?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

crates/coral-engine/src/backends/shared/template.rs
```diff
@@ -76,11 +80,36 @@ pub(crate) fn resolve_value_source(
             })?;
             Ok(Some(json!(parsed)))
         }
-        ValueSourceSpec::Arg { key, .. }
-        | ValueSourceSpec::ArgInt { key, .. }
-        | ValueSourceSpec::ArgBool { key, .. } => Err(DataFusionError::Execution(format!(
-            "function argument '{key}' cannot be resolved outside a function request"
-        ))),
+        ValueSourceSpec::Arg { key, default } => Ok(filters
```
jsummerfield: [P2] This makes `from: arg` value sources work by passing function arguments through the same map used for filters, but function request templates are still a hole. The spec validator accepts `{{arg.*}}` tokens for function requests, while `render_template` still only resolves input/filter/state tokens and rejects everything else. A function using `path: /search/{{arg.query}}` will validate but fail at execution time.

mystic123: Addressed in the current branch: request rendering now carries function args separately in `RenderContext`, `render_template` resolves `TemplateNamespace::Arg`, and the HTTP client passes `arg_values` when rendering base URLs, request paths, headers/query/body templates. The validator still rejects `{{arg.*}}` outside function requests.

sources/core/notion/manifest.yaml
```diff
@@ -34,13 +34,134 @@ request_headers:
     value: "2026-03-11"
 test_queries:
   - SELECT * FROM notion.search LIMIT 1
+functions:
+  - name: search_objects
+    kind: search
+    description: Provider-ranked search results for pages and data sources shared with the integration
+    fetch_limit_default: 10
+    search_limits:
```
jsummerfield, line 43: Could we either enforce these limits during function execution or avoid presenting them as hard bounds here? The HTTP function target currently carries `fetch_limit_default`, but not `search_limits`, and the fetch path uses `sql_limit.or(fetch_limit_default)`. That means an explicit `LIMIT 1000` can paginate beyond `max_top_k: 100` and `max_calls_per_query: 1`, even though the catalog metadata now advertises those values as the search budget.

sources/core/notion/manifest.yaml
```diff
@@ -34,13 +34,134 @@ request_headers:
     value: "2026-03-11"
 test_queries:
   - SELECT * FROM notion.search LIMIT 1
```
jsummerfield, line 36: Can we add a validation query for the new public function too? This PR makes `notion.search_objects(query => ...)` the preferred search surface, but `coral source test notion` still only exercises the legacy compatibility table. A small query like `SELECT * FROM notion.search_objects(query => 'Coral') LIMIT 1` would catch broken arg binding, request-body rendering, or cursor-body pagination for the new path.

jsummerfield (approved):

antonmry (approved):

sources/core/notion/manifest.yaml
mystic123: Addressed in 2c73bcc1: `HttpFetchTarget` now carries `search_limits`, and the shared HTTP fetch path derives the effective row cap, page-size cap, and provider-call cap from that metadata. An explicit `LIMIT` above the advertised budget is capped by `max_top_k * max_calls_per_query`, each provider page size is capped by `max_top_k`, and pagination stops after `max_calls_per_query` provider calls. Added `source_scoped_search_function_enforces_search_limits` to cover `LIMIT 250` being capped to one 100-row provider call for a search function.

sources/core/notion/manifest.yaml
mystic123: Addressed in 2c73bcc1: added `SELECT * FROM notion.search_objects(query => 'Coral') LIMIT 1` to `sources/core/notion/manifest.yaml` test_queries, so `coral source test notion` now exercises the new public table function path as well as the legacy compatibility table.

sources/core/notion/manifest.yaml
```diff
@@ -34,13 +34,135 @@ request_headers:
...
+functions:
+  - name: search_objects
+    kind: search
+    description: Provider-ranked search results for pages and data sources shared with the integration
+    fetch_limit_default: 10
+    search_limits:
+      default_top_k: 10
+      max_top_k: 100
+      max_calls_per_query: 1
+    args:
```
jsummerfield, line 47: Can we keep server-side object filtering on the preferred search surface? `search_objects` only accepts `query`, so `WHERE object = 'data_source'` is applied locally after fetching the bounded mixed search results. The legacy `notion.search` table still sends Notion’s `filter.property = object` / `filter.value = object`, which avoids missing data sources that were not in the first mixed provider-ranked page. Notion supports this filter on `/v1/search`, so this function should expose/push an object/result-type argument too.

jsummerfield: Defer to you on how valid this is.

mystic123: Implemented. `notion.search_objects` now accepts an optional `object` argument with `page` / `data_source` values and emits Notion `/v1/search` object filtering only when that argument is present. Added runtime support for conditional arg-backed body fields plus an HTTP regression test covering filtered and unfiltered requests.

mystic123:
### Merge activity

* **May 22, 10:15 AM UTC**: @mystic123 merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/352).

PR #353, opened 2026-05-11
feat(feedback): upload MCP feedback reports to Coral

jsummerfield:
## Summary

This adds the hosted feedback upload path behind the existing MCP feedback tool. The feedback tool writes the local JSONL report synchronously, then starts a best-effort hosted upload in the background so blocked-agent reports can be collected centrally without putting the remote ingest endpoint on the MCP tool critical path.

## Details

- Adds a hosted feedback publisher in `coral-app` with a fixed Coral ingest endpoint and timeout-bound upload behavior.
- Keeps local JSONL persistence as the only synchronous success condition for feedback submission.
- Starts hosted upload from `FeedbackManager` as a detached task after the local report is safely stored.
- Keeps the MCP/protobuf feedback response focused on the persisted report id and timestamp instead of exposing remote upload status to the LLM.
- Keeps `--enable-feedback` as the MCP discoverability toggle, while app bootstrap wires the hosted publisher for feedback submissions.
- Updates MCP tests and the MCP guide for the background hosted submission behavior.

## Verification

- `make rust-checks`
- Live smoke through `target/debug/coral mcp-stdio --enable-feedback` against `https://feedback.withcoral.com/ingest`; upload returned `accepted` and the local JSONL report was written.

### DIFF
--- a/crates/coral-api/proto/coral/v1/feedback.proto
+++ b/crates/coral-api/proto/coral/v1/feedback.proto
@@ -40,6 +40,6 @@ message SubmitFeedbackResponse {
 
 // Feedback submission APIs.
 service FeedbackService {
-  // Persists one blocked-agent feedback report locally.
+  // Persists one blocked-agent feedback report locally and may publish it remotely.
   rpc SubmitFeedback(SubmitFeedbackRequest) returns (SubmitFeedbackResponse);
 }
--- a/crates/coral-app/Cargo.toml
+++ b/crates/coral-app/Cargo.toml
@@ -19,6 +19,7 @@ opentelemetry.workspace = true
 opentelemetry-appender-tracing.workspace = true
 opentelemetry-otlp.workspace = true
 opentelemetry_sdk.workspace = true
+reqwest.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 serde_yaml.workspace = true
@@ -45,7 +46,6 @@ coral-spec.workspace = true
 coral-client.workspace = true
 prost.workspace = true
 opentelemetry_sdk = { workspace = true, features = ["testing"] }
-reqwest.workspace = true
 tempfile.workspace = true
 tokio.workspace = true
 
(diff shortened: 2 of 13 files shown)


sauldhernandez (commented):
A couple of concerns. First (minor), hosted upload is still on the feedback submission critical path: after the local JSONL write, SubmitFeedback awaits the remote POST and can hold the MCP tool/RPC response for up to the hosted upload timeout when the ingest endpoint is slow or unreachable. The tool call could return while the coral server is still trying the submission (if the submission result is not relevant to the LLM). Second (major), there is no real local-only opt-out. MCP users can omit --enable-feedback, but that disables the entire feedback tool; once enabled, they cannot keep local reports only, which can be useful in constrained network environments.

simonwhitaker (approved):

jsummerfield:
> A couple of concerns. First (minor), hosted upload is still on the feedback submission critical path: after the local JSONL write, SubmitFeedback awaits the remote POST and can hold the MCP tool/RPC response for up to the hosted upload timeout when the ingest endpoint is slow or unreachable. The tool call could return while the coral server is still trying the submission (if the submission result is not relevant to the LLM).

Good shout - have made it fire and forget.

> Second (major), there is no real local-only opt-out. MCP users can omit --enable-feedback, but that disables the entire feedback tool; once enabled, they cannot keep local reports only, which can be useful in constrained network environments.

That’s a fair callout, and I agree the behaviour is worth making explicit. The current design is intentional: the local filesystem path was mainly stepping stone, not the long-term destination for feedback. The goal of --enable-feedback is to make the feedback tool available for aggregated Coral feedback, with local persistence for individual review.

So I’d rather not add a local-only mode in this PR. It adds another operating mode we’d need to document and support, but it works against the purpose of the feature, which is to collect reports centrally so we can triage patterns across users. In constrained network environments, the hosted upload is best-effort and no longer blocks the tool response; the local report is still written first.

Bradley-Butcher (approved):
🔥

sauldhernandez (approved):
LGTM. Maybe worth being explicit on docs and MCP tool description that it **will** upload the data to our server so users can do an informed choice to use it. "Hosted submission" is a bit vague.

PR #360, opened 2026-05-12
feat(credentials): support OAuth device flow

sauldhernandez:
## Summary
- Add OAuth `device_code` as a first-class credential retrieval flow in the source spec, JSON schema, and gRPC contract.
- Implement app-owned device-code authorization: request a device code, poll the token endpoint, handle provider polling errors, and store returned token material.
- Update CLI OAuth retrieval to show the provider URL plus user code, while preserving authorization-code behavior.
- Document device-code vs authorization-code manifest fields.

## Stack
- Base: #343 (`feat(credentials): add OAuth credential retrieval`)
- This PR: generic device-flow support only
- Upstack: #361 adds GitHub source OAuth using this flow

## Validation
- `make rust-checks`
- `cargo test -p coral-app oauth --locked`
- `cargo test -p coral-spec oauth --locked`
- `cargo test -p coral-cli source_ops --locked`
- `make docs-check`
- `make lint-sources`
- `git diff --check`

PR #346, opened 2026-05-11
feat(sources/core/google_calendar): add Google Calendar source

sauldhernandez:
## Intent

Add Google Calendar as a bundled Coral source now that OAuth credential retrieval support is available, and place the manifest under `sources/core` so docs generation and runtime bundling both discover it.

## What it enables

Users can install `google_calendar` as a bundled source and query calendars, events, user settings, and color palettes through Coral SQL with a read-only Google Calendar OAuth access token.

## Usage examples

```sql
SELECT * FROM google_calendar.calendars LIMIT 10;
```

```sql
SELECT summary, start_date_time, end_date_time
FROM google_calendar.events
WHERE calendar_id = 'primary'
  AND single_events = true
  AND order_by = 'startTime'
LIMIT 20;
```

For secondary calendars, use `google_calendar.calendars.events_calendar_id` as the `events.calendar_id` value so calendar IDs containing `#` are path-safe.

PR #369, opened 2026-05-12
feat(sources/linear): add OAuth support

sauldhernandez:
## Intent

Linear auth should expose the credential choice users actually have without creating a second runtime credential slot. This PR keeps `LINEAR_API_KEY` as the single required Linear secret, lets guided OAuth or a pasted personal API key fill that secret, and adds the source-spec/runtime support needed to store and refresh OAuth tokens in the header shape Linear expects.

## What it enables

**Linear OAuth with one credential**

- `LINEAR_API_KEY` now offers two credential methods: `Connect with Linear OAuth` and `Paste a Linear API key`.
- Guided OAuth uses authorization-code + PKCE, the Linear `read` scope, default client ID `9d480c78cc389c7f7584afd60555aec1`, and fixed loopback redirect `http://127.0.0.1:53682/oauth/callback`.
- Linear auth reads one `Authorization` value from `LINEAR_API_KEY`; OAuth tokens are stored as `Bearer <token>` via `access_token_scheme: bearer`, while pasted API keys remain raw.
- Linear no longer needs a separate `LINEAR_OAUTH_ACCESS_TOKEN` input or a Linear-specific `from: one_of` auth header.

**Reusable OAuth and credential hardening**

- Adds OAuth `access_token_scheme` parsing/docs/schema support with `raw` as the default and `bearer` for providers that require `Authorization: Bearer <token>`.
- Persists the selected OAuth access-token scheme in Coral's internal OAuth metadata and reuses it during refresh, so refresh keeps the correct header format even if a manifest has multiple OAuth methods or method order changes.
- Treats empty secret strings as absent during import validation, query-source loading, and query-time input resolution. This prevents empty stored material from satisfying required-secret checks or being delegated to query execution, while preserving existing OAuth material when an explicitly empty binding means "do not replace this credential".
- Clears stale OAuth metadata when a user replaces a credential with a non-empty pasted value, so switching from guided OAuth to a manual key does not leave refresh metadata attached to the new secret.
- Updates generated Linear docs, source-spec docs, and maintained Coral source-authoring skill guidance to teach the single-secret OAuth/API-key pattern and reserve separate optional secrets plus `from: one_of` for genuinely separate credential slots.

## Usage examples

Connect Linear with guided OAuth:

```shell
coral source add linear --interactive
# Choose "Connect with Linear OAuth" and complete the browser flow.
```

Connect Linear with a pasted API key:

```shell
LINEAR_API_KEY=lin_api_... coral source add linear
```

Query after setup:

```sql
SELECT id, identifier, title
FROM linear.issues
LIMIT 10;
```

## Validation

- `cargo nextest run -p coral-app -E 'test(query_refresh_uses_stored_access_token_scheme_for_selected_oauth_method) or test(import_with_oauth_stores_bearer_prefixed_token)'`
- `make rust-checks`
- `make schema-check && make docs-check`
- `git diff --check`
- Structured autoreview on the combined branch and corrective patch: clean
- Structured autoreview on the final local patch before commit: clean


sauldhernandez:
* **#369** <a href="https://app.graphite.com/github/pr/withcoral/coral/369?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/369?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#370** <a href="https://app.graphite.com/github/pr/withcoral/coral/370?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#379** <a href="https://app.graphite.com/github/pr/withcoral/coral/379?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 6 other dependent PRs ([#346](https://github.com/withcoral/coral/pull/346) <a href="https://app.graphite.com/github/pr/withcoral/coral/346?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#360](https://github.com/withcoral/coral/pull/360) <a href="https://app.graphite.com/github/pr/withcoral/coral/360?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#374](https://github.com/withcoral/coral/pull/374) <a href="https://app.graphite.com/github/pr/withcoral/coral/374?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 3 others)
* `main`


<h2></h2>

sources/core/linear/manifest.yaml
```diff
@@ -8,18 +8,57 @@ description: >-
...
+    hint: |
+      Connect with Linear OAuth, or provide a raw OAuth access token.
+      Create a Linear OAuth application with the `read` scope and the
+      redirect callback URL `http://127.0.0.1:53682/oauth/callback`.
+      Coral includes a default Linear OAuth client ID; during interactive
+      setup you can press Enter to use it or provide your own client ID.
+      Coral stores OAuth refresh token metadata, but automatic refresh is
+      not implemented yet; rerun
+      `coral source add linear --interactive` if the access token expires.
+      See [Linear OAuth docs](https://linear.app/developers/oauth-2-0-authentication).
```
AlbertQM:
```suggestion
    hint: |
      You can connect with Linear OAuth, or provide a raw OAuth access token.
      Linear OAuth application needs the `read` scope and the
      redirect callback URL `http://127.0.0.1:53682/oauth/callback`.
      See [Linear OAuth docs](https://linear.app/developers/oauth-2-0-authentication).
```

AlbertQM (approved):
Wooo! I want to try this one (and the GitHub one) as soon as they're out 😄 

Just commenting on a tiny simplification of the `hint` for the new input

iaptsiauri (approved):

sources/core/linear/manifest.yaml
```diff
@@ -8,18 +8,56 @@ description: >-
...
 auth:
   type: HeaderAuth
   headers:
     - name: Authorization
-      from: template
-      template: "{{input.LINEAR_API_KEY}}"
+      from: one_of
+      values:
+        - from: input
+          key: LINEAR_API_KEY
```
jsummerfield: This ordering makes a switch-to-OAuth flow keep using a stale API key. `from: one_of` resolves the first non-empty value, and the app preserves existing expected secrets when a bundled source is re-added. If a user already has `LINEAR_API_KEY` stored and then re-runs setup with OAuth, both secrets can exist and this first branch still wins. The source should either clear the API-key credential when OAuth replaces auth, prefer OAuth when present, or otherwise make the migration path explicit.

sources/core/linear/manifest.yaml
```diff
@@ -8,18 +8,56 @@ description: >-
...
-      to specific teams.
+      You can paste a personal API key with the Read permission, which
+      gives read-only access to all tables. You can optionally scope the
+      key to specific teams.
       See [Linear API key docs](https://linear.app/docs/api-and-webhooks#api-keys).
+
+      Leave this empty during interactive setup to connect with Linear OAuth.
+  LINEAR_OAUTH_ACCESS_TOKEN:
+    kind: secret
+    required: false
```
jsummerfield: With both auth inputs optional, `coral source add linear` can persist a source with neither credential configured. The install path only enforces individually required inputs, and the CLI then treats validation failures as warnings, so users can end up with a source that was “Added” but cannot make any Linear request. Please add an install-time invariant that at least one auth branch is present, or make the source model enforce one required credential path.

sources/core/linear/manifest.yaml
```diff
@@ -8,18 +8,56 @@ description: >-
...
+    hint: |
+      You can connect with Linear OAuth, or provide a raw OAuth access token.
+      Linear OAuth application needs the `read` scope and the
+      redirect callback URL `http://127.0.0.1:53682/oauth/callback`.
+      See [Linear OAuth docs](https://linear.app/developers/oauth-2-0-authentication).
+    credential:
+      methods:
+        - type: oauth
```
jsummerfield, line 27: The hint above says users can provide a raw OAuth access token, but this credential list only exposes the browser OAuth method. In interactive setup the user gets OAuth or skip, not a paste-token path; pasting the OAuth token into `LINEAR_API_KEY` is not equivalent because that branch sends the value without `Bearer`. Please add a `type: source_config` method for `LINEAR_OAUTH_ACCESS_TOKEN`, or narrow the docs/hint to non-interactive env-var setup only.

crates/coral-app/src/sources/manager.rs
```diff
@@ -1039,6 +1050,20 @@ fn validate_bindings(
             _ => {}
         }
     }
+    for requirement in &candidate.auth_one_of_secret_requirements {
```
jsummerfield: This fixes the “at least one auth choice exists” case, but the replacement path still leaves stale siblings in the same `one_of` group. For Linear, if `LINEAR_OAUTH_ACCESS_TOKEN` is already stored and the user now supplies `LINEAR_API_KEY`, validation passes, but persist only clears metadata for the submitted key; the old OAuth token remains and still wins because the manifest tries OAuth first. When one key in an auth `one_of` group is explicitly replaced, the other keys in that group should be removed too.

sources/core/linear/manifest.yaml
```diff
@@ -8,18 +8,59 @@ description: >-
...
+      gives read-only access to all tables. You can optionally scope the
+      key to specific teams.
       See [Linear API key docs](https://linear.app/docs/api-and-webhooks#api-keys).
+
+      Leave this empty during interactive setup to connect with Linear OAuth.
+  LINEAR_OAUTH_ACCESS_TOKEN:
+    kind: secret
+    required: false
+    hint: |
+      Connect with Linear OAuth, or paste a raw OAuth access token.
```
jsummerfield: This pasted-token path needs to distinguish it from the guided OAuth flow. A raw pasted Linear OAuth token will not include Coral refresh metadata, and Linear OAuth access tokens are short-lived, so users who choose this path can hit an expiry cliff while guided OAuth can store/refresh with the returned refresh token. Please call out that pasted raw OAuth tokens must be rotated/reconnected when they expire, or steer users to guided OAuth for refresh-capable setup.

sources/core/linear/manifest.yaml
```diff
@@ -8,18 +8,63 @@ description: >-
...
+      Coral uses a built-in OAuth client by default; to use your own,
```
jsummerfield: This says users can “set `LINEAR_OAUTH_CLIENT_ID`”, but `prompt_oauth_credential_inputs` only calls `prompt_oauth_client_id` and never checks the environment for OAuth credential inputs. That makes the documented override a no-op unless the user retypes the client ID at the prompt. Can we either make `oauth.client.id.input` honor env vars before prompting, with a CLI test, or reword this to “enter `LINEAR_OAUTH_CLIENT_ID` when prompted”?

crates/coral-app/src/sources/model.rs
```diff
@@ -15,11 +15,78 @@ pub(crate) struct CandidateSource {
     pub(crate) description: String,
     pub(crate) version: String,
     pub(crate) inputs: Vec<ManifestInputSpec>,
+    pub(crate) auth_one_of_secret_requirements: Vec<AuthOneOfSecretRequirement>,
     pub(crate) installed: bool,
     pub(crate) origin: SourceOrigin,
     pub(crate) credential_storage: Option<CredentialStorageKind>,
 }
 
+/// Install-time invariant for an auth `from: one_of` whose branches are source
```
jsummerfield: We now enforce `auth_one_of_secret_requirements` server-side, but `candidate_source_to_proto` drops that model and `SourceInfo.inputs` can only say each Linear input is optional. That means clients cannot know “one of these two optional secrets is required” before submitting. Should we add a machine-readable credential group/one-of requirement to `SourceInfo` so setup UIs can render and validate this correctly?

jsummerfield (commented):
Two inline review comments from the Linear OAuth review.

jsummerfield (commented):
Four inline review comments from the refreshed Linear OAuth review.

crates/coral-app/src/sources/manager.rs
```diff
@@ -1016,10 +1018,24 @@ fn validate_bindings(
...
```
jsummerfield: This validation needs to consider the effective post-update material, not stored material unconditionally. If a request supplies an empty value for a `one_of` key while a stored sibling exists, validation passes here, but persistence treats that supplied key as a replacement, expands to all siblings, and clears the credential the validation relied on. Can we reject empty supplied secrets for these groups, or ignore them as not supplied before computing both validation and replacement sets?

crates/coral-api/proto/coral/v1/sources.proto
```diff
@@ -254,6 +261,8 @@ message SourceInfo {
   SourceOrigin origin = 6;
   // Where credential material is stored when this source is installed.
   SourceCredentialStorage credential_storage = 7;
+  // Input groups where at least one listed input must be configured.
+  repeated SourceInputOneOfRequirement one_of_input_requirements = 8;
```
jsummerfield: Now that `SourceInfo` carries `one_of_input_requirements`, `source info` should render those groups. Otherwise `coral source info linear` says both auth inputs as optional with no indication that at least one is required, which preserves the confusing UX this API field is meant to fix.

sources/core/linear/manifest.yaml
```diff
@@ -9,18 +9,64 @@ description: >-
...
-      Create a personal API key with the Read permission, which gives
-      read-only access to all tables. You can optionally scope the key
-      to specific teams.
+      You can paste a personal API key with the Read permission, which
+      gives read-only access to all tables. You can optionally scope the
+      key to specific teams.
       See [Linear API key docs](https://linear.app/docs/api-and-webhooks#api-keys).
+
+      Leave this empty during interactive setup to connect with Linear OAuth.
+  LINEAR_OAUTH_ACCESS_TOKEN:
```
jsummerfield: The source-local README is now stale: this manifest supports either `LINEAR_OAUTH_ACCESS_TOKEN` via OAuth or `LINEAR_API_KEY`, but `sources/core/linear/README.md` still says `LINEAR_API_KEY` is required. Please update the authentication section so the source-local docs match the new auth contract.

sources/core/linear/manifest.yaml
```diff
@@ -9,18 +9,64 @@ description: >-
 inputs:
   LINEAR_API_KEY:
     kind: secret
+    required: false
```
jsummerfield: This changes Linear's user-visible inputs and auth semantics but leaves the source version at `2.3.0`, the same as `origin/main`. Please bump the Linear source version so `source info/list` can distinguish the OAuth-capable manifest from the API-key-only one.

sauldhernandez:
@cubic-dev-ai review

PR #370, opened 2026-05-12
feat(spec): add typed auth fallback sources

sauldhernandez:
## Intent

Add first-class source-spec auth fallback so sources can support multiple credential shapes without embedding a mini expression language in templates, while keeping optional fallback credentials available at runtime.

## What it enables

- Adds `from: one_of` to resolve the first present, non-empty value source.
- Adds `from: bearer` to turn a raw secret input token into `Bearer <token>` and validates that bearer sources reference `kind: secret` inputs.
- Removes `||` and `concat(...)` expression templating from request/auth templates.
- Distinguishes required secrets from declared runtime secrets so optional OAuth fallback material is not dropped during query execution.
- Keeps HTTP, file, and MCP required-secret semantics consistent.
- Keeps input-reference validation, schema validation, runtime registration checks, source-spec docs, and maintained source-authoring skills aligned with the new auth model, including device-code OAuth guidance.

## Usage examples

A source can accept either a complete API-key authorization header or a raw OAuth access token:

```yaml
inputs:
  LINEAR_API_KEY:
    kind: secret
    required: false
  LINEAR_OAUTH_ACCESS_TOKEN:
    kind: secret
    required: false

auth:
  type: HeaderAuth
  headers:
    - name: Authorization
      from: one_of
      values:
        - from: input
          key: LINEAR_API_KEY
        - from: bearer
          key: LINEAR_OAUTH_ACCESS_TOKEN
```

If `LINEAR_API_KEY` is configured, Coral uses it as the header value. If only `LINEAR_OAUTH_ACCESS_TOKEN` is configured, Coral sends `Bearer <token>`.

## Validation

- `cargo test -p coral-spec bearer`
- `cargo test -p coral-spec parses_mcp_manifest_with_secret_input`
- `cargo test -p coral-app load_query_source_passes_present_optional_secrets_to_runtime`
- `env RUSTC_WRAPPER= make docs-check`
- `env RUSTC_WRAPPER= make rust-checks`

### DIFF
--- a/crates/coral-cli/src/onboard.rs
+++ b/crates/coral-cli/src/onboard.rs
@@ -174,9 +174,9 @@ async fn run_installed_source_menu(
                 .iter()
                 .map(manifest_input_from_proto)
                 .collect::<Result<Vec<_>, _>>()?;
-            let (variables, secrets) = source_ops::prompt_for_inputs(&inputs)?;
+            let inputs = source_ops::prompt_for_inputs_with_credential_methods(&inputs)?;
             let result =
-                source_ops::add_bundled_source(app, &source.name, variables, secrets).await?;
+                source_ops::add_bundled_source_with_credentials(app, &source.name, inputs).await?;
             println!("Reconfigured source {}", result.name);
             source_ops::validate_and_warn(
                 app,
@@ -197,8 +197,8 @@ async fn run_add_bundled_source(app: &AppClient, source: &SourceInfo) -> Result<
         .iter()
         .map(manifest_input_from_proto)
         .collect::<Result<Vec<_>, _>>()?;
-    let (variables, secrets) = source_ops::prompt_for_inputs(&inputs)?;
-    let result = source_ops::add_bundled_source(app, &source.name, variables, secrets).await?;
+    let inputs = source_ops::prompt_for_inputs_with_credential_methods(&inputs)?;
+    let result = source_ops::add_bundled_source_with_credentials(app, &source.name, inputs).await?;
     println!("Added source {}", result.name);
     source_ops::validate_and_warn(app, &result.name, source_ops::TableDisplayLimit::DEFAULT).await
 }
--- a/crates/coral-cli/src/onboard.rs
+++ b/crates/coral-cli/src/onboard.rs
@@ -174,7 +174,10 @@ async fn run_installed_source_menu(
                 .iter()
                 .map(manifest_input_from_proto)
                 .collect::<Result<Vec<_>, _>>()?;
-            let inputs = source_ops::prompt_for_inputs_with_credential_methods(&inputs)?;
+            let inputs = source_ops::prompt_for_inputs_with_credential_methods_in_mode(
+                &inputs,
+                source_ops::CredentialPromptMode::CredentialMethodFirst,
+            )?;
             let result =
                 source_ops::add_bundled_source_with_credentials(app, &source.name, inputs).await?;
             println!("Reconfigured source {}", result.name);
@@ -197,7 +200,10 @@ async fn run_add_bundled_source(app: &AppClient, source: &SourceInfo) -> Result<
         .iter()
         .map(manifest_input_from_proto)
         .collect::<Result<Vec<_>, _>>()?;
-    let inputs = source_ops::prompt_for_inputs_with_credential_methods(&inputs)?;
+    let inputs = source_ops::prompt_for_inputs_with_credential_methods_in_mode(
+        &inputs,
+        source_ops::CredentialPromptMode::CredentialMethodFirst,
+    )?;
     let result = source_ops::add_bundled_source_with_credentials(app, &source.name, inputs).await?;
     println!("Added source {}", result.name);
     source_ops::validate_and_warn(app, &result.name, source_ops::TableDisplayLimit::DEFAULT).await
--- a/crates/coral-spec/src/backends/http.rs
+++ b/crates/coral-spec/src/backends/http.rs
@@ -17,10 +17,13 @@ use serde_json::{Map, Value};
 
 use crate::{
     ColumnSpec, DetailHintDeclaringSurface, DetailHintSpec, DetailHintTargetTable, FilterSpec,
-    HeaderSpec, ManifestError, ManifestInputKind, ManifestInputSpec, PaginationSpec,
-    ParsedTemplate, RequestRouteSpec, RequestSpec, ResponseSpec, Result, SearchLimitsSpec,
-    SourceBackend, SourceManifestCommon, SourceTableFunctionSpec, TableCommon,
-    inputs::collect_source_inputs_value, validate::validate_template,
+    HeaderSpec, ManifestError, ManifestInputSpec, PaginationSpec, ParsedTemplate, RequestRouteSpec,
+    RequestSpec, ResponseSpec, Result, SearchLimitsSpec, SourceBackend, SourceManifestCommon,
+    SourceTableFunctionSpec, TableCommon,
+    inputs::{
+        collect_source_inputs_value, declared_secret_input_names, required_secret_input_names,
+    },
+    validate::validate_template,
     validate_detail_hint_references, validate_http_function, validate_http_function_names,
     validate_http_table, validate_table_names, validate_test_queries,
 };
@@ -212,16 +215,17 @@ impl HttpTableSpec {
 }
 
 impl HttpSourceManifest {
+    /// Returns all source secrets declared by this manifest.
+    pub fn declared_secret_names(&self) -> BTreeSet<String> {
+        declared_secret_input_names(&self.declared_inputs)
+    }
+
     /// Returns the source secrets required by this manifest.
     ///
     /// In the new input model, required declared inputs with `kind: secret`
     /// must be available before a source can compile or authenticate.
     pub fn required_secret_names(&self) -> BTreeSet<String> {
-        self.declared_inputs
-            .iter()
-            .filter(|input| input.kind == ManifestInputKind::Secret && input.required)
-            .map(|input| input.key.clone())
-            .collect()
+        required_secret_input_names(&self.declared_inputs)
     }
 }
 
--- a/crates/coral-spec/src/backends/http.rs
+++ b/crates/coral-spec/src/backends/http.rs
@@ -214,12 +214,12 @@ impl HttpTableSpec {
 impl HttpSourceManifest {
     /// Returns the source secrets required by this manifest.
     ///
-    /// In the new input model, every declared input with `kind: secret` is
-    /// required because secrets cannot carry defaults.
+    /// In the new input model, required declared inputs with `kind: secret`
+    /// must be available before a source can compile or authenticate.
     pub fn required_secret_names(&self) -> BTreeSet<String> {
         self.declared_inputs
             .iter()
-            .filter(|input| input.kind == ManifestInputKind::Secret)
+            .filter(|input| input.kind == ManifestInputKind::Secret && input.required)
             .map(|input| input.key.clone())
             .collect()
     }
--- a/crates/coral-spec/src/backends/mcp.rs
+++ b/crates/coral-spec/src/backends/mcp.rs
@@ -715,6 +715,17 @@ fn validate_table_tool_arg_value_source(
             }
             Ok(())
         }
+        ValueSourceSpec::OneOf { values } => {
+            if values.is_empty() {
+                return Err(ManifestError::validation(format!(
+                    "{context} one_of values must not be empty"
+                )));
+            }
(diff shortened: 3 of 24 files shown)


sauldhernandez:
* **#369** <a href="https://app.graphite.com/github/pr/withcoral/coral/369?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#370** <a href="https://app.graphite.com/github/pr/withcoral/coral/370?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/370?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#379** <a href="https://app.graphite.com/github/pr/withcoral/coral/379?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 6 other dependent PRs ([#346](https://github.com/withcoral/coral/pull/346) <a href="https://app.graphite.com/github/pr/withcoral/coral/346?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#360](https://github.com/withcoral/coral/pull/360) <a href="https://app.graphite.com/github/pr/withcoral/coral/360?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#374](https://github.com/withcoral/coral/pull/374) <a href="https://app.graphite.com/github/pr/withcoral/coral/374?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 3 others)
* `main`


<h2></h2>

docs/reference/source-spec-reference.mdx
```diff
@@ -946,6 +949,10 @@ Scopes are declared under `scopes.scope`; `delimiter: space` represents
 refresh tokens and OAuth metadata alongside the source secret, but automatic
 token refresh is not implemented yet.
 
+OAuth stores the token response `access_token` as the source secret value. If a
+runtime auth template needs a bearer header value, compose it in the template,
+for example `{{concat("Bearer ", input.OAUTH_TOKEN)}}`.
```
AlbertQM: This sounds like very useful information, but it's buried down this information-dense page. Can't think of a good place to surface this though, if you have ideas please shout! Can always update in the future if users have troubles with it.

AlbertQM (approved):

jsummerfield:
What would the alternative look like if we didn't add templating?

sauldhernandez:
> What would the alternative look like if we didn't add templating?

​We could enhance the auth block in the spec to support multiple options e.g.

```

auth:
  type: HeaderAuth
  headers:
  - name: Authorization
    from: one_of
    values:
    - from: input
      key: LINEAR_API_KEY
    - from: bearer
      key: LINEAR_OAUTH_ACCESS_TOKEN

```

Then coral can pick the first available, in order

crates/coral-spec/src/backends/http.rs
```diff
@@ -214,12 +214,12 @@ impl HttpTableSpec {
...
     ///
-    /// In the new input model, every declared input with `kind: secret` is
-    /// required because secrets cannot carry defaults.
+    /// In the new input model, required declared inputs with `kind: secret`
+    /// must be available before a source can compile or authenticate.
     pub fn required_secret_names(&self) -> BTreeSet<String> {
         self.declared_inputs
             .iter()
-            .filter(|input| input.kind == ManifestInputKind::Secret)
+            .filter(|input| input.kind == ManifestInputKind::Secret && input.required)
```
jsummerfield: Could we separate “must be configured” from “material to pass into runtime” here? `coral-app` currently reads and forwards only `source_spec.required_secret_names()` into `QuerySource` (`query/manager.rs` builds `resolved_secrets` from this set). With both fallback inputs marked `required: false`, a stored OAuth token is dropped before `resolve_inputs`, so the advertised `from: one_of` OAuth fallback resolves as missing. The runtime probably needs all configured declared secrets, while this set only answers the precondition check.

crates/coral-spec/src/inputs.rs
```diff
@@ -976,14 +987,15 @@ fn validate_value(value: &Value, is_root: bool, declared: &BTreeSet<String>) ->
 }
 
 fn validate_mapping(map: &Map<String, Value>, declared: &BTreeSet<String>) -> Result<()> {
-    if map.get("from").and_then(Value::as_str) != Some("input") {
+    let Some(source_kind @ ("input" | "bearer")) = map.get("from").and_then(Value::as_str) else {
```
jsummerfield, line 1023: `bearer` probably needs a stronger check than “declared”. Since it always serializes a credential-bearing header value, the referenced input should be `kind: secret`; otherwise a variable named something less obvious than `TOKEN` can be sent as `Bearer ...` and also leak through `coral.inputs`.

crates/coral-spec/src/backends/mcp.rs
```diff
@@ -770,8 +781,20 @@ fn validate_source_scoped_value_source(source: &ValueSourceSpec, context: &str)
...
+                )));
+            }
+            for value in values {
+                validate_source_scoped_value_source(value, context)?;
+            }
+            Ok(())
+        }
         ValueSourceSpec::Literal { .. }
         | ValueSourceSpec::Input { .. }
+        | ValueSourceSpec::Bearer { .. }
```
jsummerfield, line 801: Adding `bearer`/`one_of` here makes MCP participate in the new optional credential model too, but `McpSourceManifest::required_secret_names()` still returns every secret input. That means an MCP source with alternate optional credentials remains hard-required even though HTTP/file now honor `required: false`. Please update that path or this behavior becomes backend-dependent.

crates/coral-cli/src/onboard.rs
```diff
@@ -174,9 +174,9 @@ async fn run_installed_source_menu(
                 .iter()
                 .map(manifest_input_from_proto)
                 .collect::<Result<Vec<_>, _>>()?;
-            let (variables, secrets) = source_ops::prompt_for_inputs(&inputs)?;
+            let inputs = source_ops::prompt_for_inputs_with_credential_methods(&inputs)?;
```
jsummerfield: This preserves OAuth collection in onboard, but the helper consumes matching environment variables before it shows credential methods. In the guided “reconfigure” path, a stale `*_TOKEN` in the shell will silently bypass the OAuth/manual choice. It may be worth giving onboarding a mode that prompts for the credential method first, or at least asks before using env-provided secrets.

docs/reference/source-spec-reference.mdx
```diff
@@ -557,11 +557,31 @@ auth:
...
 ```
 
-`headers` entries share the same shape as request/table headers, but only `from: literal`, `from: template`, and `from: input` are meaningful here — filter and state tokens have no value at auth time.
+`headers` entries share the same value-source shape as request/table headers,
+but auth is source-scoped: `from: literal`, `from: template`, `from: input`,
+`from: bearer`, and `from: one_of` are meaningful here. Filter, argument, and
+state values have no value at auth time. Use `from: bearer` when Coral stores a
+raw token but the API expects an `Authorization: Bearer <token>` header.
+
+Use `from: one_of` when a source supports multiple credential shapes. Coral
```
jsummerfield, line 573: This needs to spell out the input-side `required: false` requirement. Secret inputs default to required, so authors following this `one_of` example without marking both alternatives optional will accidentally require both credentials and never get the intended fallback behavior.

plugins/coral/skills/coral-create-source-spec/references/http-source-checklist.md
```diff
@@ -28,6 +28,7 @@ Use:
 ## Authentication Setup
 
 - Keep credential collection separate from runtime auth. `inputs` stores values; `auth`, request headers, query params, or body fields decide where values are sent.
+- Use `from: bearer` for raw access tokens that must be sent as `Authorization: Bearer <token>`. Use `from: one_of` when runtime auth should choose the first present credential, such as a pasted API-key header before an OAuth bearer token.
```
jsummerfield, line 31: Since this same section now recommends OAuth fallback auth, can we also update the next OAuth bullets? The maintained skill still says OAuth means authorization-code only, but the parser/docs support `device_code` too. Leaving that stale will steer agents away from valid device-code sources.

jsummerfield (commented):
Re-review after the latest push. The onboarding env-first issue looks fixed; I found a few remaining contract/docs gaps.

crates/coral-spec/src/inputs.rs
```diff
@@ -975,19 +1009,28 @@ fn validate_value(value: &Value, is_root: bool, declared: &BTreeSet<String>) ->
...
+        ManifestError::validation(format!(
+            "manifest '{source_kind}' value source is missing key"
+        ))
+    })?;
+    let Some(kind) = declared.get(key) else {
         return Err(ManifestError::validation(format!(
             "manifest input '{key}' is referenced but not declared under top-level inputs"
         )));
+    };
+    if source_kind == "bearer" && *kind != ManifestInputKind::Secret {
```
jsummerfield, line 1037: Could we make the secret-kind check cover auth-scoped `from: input` too? The new `bearer` guard closes one path, but the fallback auth example still uses `from: input` for a complete `Authorization` value. I verified `coral source lint` accepts `HeaderAuth` with `Authorization` sourced from a `kind: variable` like `HEADER_VALUE`, which then exposes that credential through `coral.inputs`. Since the docs now say bearer/authorization values must be secrets, this likely needs context-aware validation for auth value sources.

docs/guides/write-a-custom-source.mdx
```diff
@@ -251,10 +251,14 @@ auth:
...
     - name: Authorization
-      from: template
-      template: Bearer {{input.DEMO_API_TOKEN}}
+      from: bearer
+      key: DEMO_API_TOKEN
 ```
 
+token, use `from: one_of` and put the pasted header first, then a `from: bearer`
+fallback for the raw OAuth token.
```
jsummerfield, line 261: Could this call out that the alternate credential inputs need `required: false`, or show the small `inputs:` snippet? Inputs without defaults are required by default, so an author following this fallback guidance will accidentally require both the pasted header and the OAuth token, which prevents `one_of` from behaving like a fallback.

plugins/coral/skills/coral-create-source-spec/SKILL.md
```diff
@@ -240,8 +240,23 @@ auth:
...
+  type: HeaderAuth
+  headers:
+    - name: Authorization
+      from: one_of
+      values:
+        - from: input
+          key: FOOBAR_API_KEY
+        - from: bearer
+          key: FOOBAR_OAUTH_ACCESS_TOKEN
 ```
```
jsummerfield, line 267: Same issue here for the maintained source-authoring skill: this `one_of` pattern needs the corresponding secret inputs marked `required: false`. Otherwise agents following the skill will generate specs that require both credential branches instead of allowing either branch to satisfy auth.

plugins/coral/skills/coral-create-source-spec/references/http-source-checklist.md
```diff
@@ -28,9 +28,10 @@ Use:
...
 - Use `type: source_config` when the user should provide a stored secret directly through an environment variable or prompt.
-- Use `type: oauth` when the provider should issue the source secret through an OAuth authorization-code flow during `coral source add --interactive`.
-- OAuth methods require `flow.type: authorization_code` and an explicit `flow.pkce` of `required` or `disabled`.
+- Use `type: oauth` when the provider should issue the source secret through an OAuth device-code or authorization-code flow during `coral source add --interactive`.
+- Device-code OAuth methods require `flow.type: device_code`, `endpoints.device_authorization_url`, `endpoints.token_url`, and a public client ID. Authorization-code OAuth methods require `flow.type: authorization_code` and an explicit `flow.pkce` of `required` or `disabled`.
```
jsummerfield, line 35: This device-code split is the right direction. Could we carry it through the top-level `coral-create-source-spec` skill text as well? That file still says OAuth credential methods currently mean authorization-code loopback setup, which contradicts this checklist and will steer source-authoring agents away from valid `device_code` sources.

docs/reference/source-spec-reference.mdx
```diff
@@ -690,6 +721,8 @@ Request query params, body fields, and headers support these `from` values:
 | `arg_split` | Split a source function argument string and serialize one part as a string |
 | `arg_split_int` | Split a source function argument string and serialize one part as a JSON integer |
 | `input` | Read a manifest-declared source input (variable/secret) by key |
+| `bearer` | Read a manifest-declared source input by key and serialize it as `Bearer <value>` |
```
jsummerfield: This should say `bearer` reads a manifest-declared **secret** input. The validator now rejects `from: bearer` when the referenced input is a variable, so the reference table should match that contract.

jsummerfield (approved):

PR #376, opened 2026-05-12
feat(app): keep local HTTP body previews off spans

sauldhernandez:
## Intent
Keep local HTTP request and response body previews available in Coral trace history without putting body contents on global tracing spans or OTLP exports.

## What it enables
- Local trace detail can show bounded HTTP body previews while host-owned subscribers and OTLP exporters only see body sizes and status metadata.
- Retry, rate-limit, and allowed-404 paths remain responsive when collecting best-effort body previews.
- Repeated local server starts use the same process-installed trace store for spans and body preview records.

## Usage examples
```toml
[trace_history]
record_http_bodies = false
```

```toml
[trace_history]
http_body_max_bytes = 65536
```

Agent/contributor behavior changed: no new workflow requirement in this PR; existing PR descriptions should continue to include intent, enabled capability, and usage examples.


sauldhernandez:
* **#376** <a href="https://app.graphite.com/github/pr/withcoral/coral/376?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/376?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#275** <a href="https://app.graphite.com/github/pr/withcoral/coral/275?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 3 other dependent PRs ([#274](https://github.com/withcoral/coral/pull/274) <a href="https://app.graphite.com/github/pr/withcoral/coral/274?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#279](https://github.com/withcoral/coral/pull/279) <a href="https://app.graphite.com/github/pr/withcoral/coral/279?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#305](https://github.com/withcoral/coral/pull/305) <a href="https://app.graphite.com/github/pr/withcoral/coral/305?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

crates/coral-engine/src/composition.rs
```diff
@@ -108,6 +111,15 @@ impl RequestAuthenticatorError {
     }
 }
 
+/// Request-time sink for local-only HTTP body records.
+pub trait HttpBodyRecorder: Send + Sync + std::fmt::Debug {
```
antonmry: Have you considered emitting each body capture as a child span with a dedicated target (e.g. `coral.http.body`) instead of routing it through a parallel sidecar store?

```rust
let _body = tracing::trace_span!(
    target: "coral.http.body",
    "http.request.body",
    direction = "request",
    body = body_preview,
    truncated = truncated,
).entered();
```

The existing `TargetFilteringSpanExporter` already filters by target — wrapping the OTLP exporter in one that drops `coral.http.body` would keep bodies on the local `JsonlSpanExporter` only, using the same mechanism that already separates the local and OTLP filters today. Operators would get a uniform knob:

```toml
[trace_history]
trace_filter = "coral_app=trace,coral.http.body=trace"

[otel]
trace_filter = "coral_app=info"   # body target implicitly off
```

If workable, this would let you drop: the `record_http_bodies` config flag, the `http-bodies-*.jsonl` format and the `file_prefix` generalization on `RollingJsonlWriter`, and the `apply_http_body_records` / `merge_http_body_record` / `prefixed_jsonl_file` read-time merge. Bodies would also become structural children of the request span rather than being joined by `(trace_id, span_id)`, so the orphan-record and span-export ordering edge cases would go away.

The main tradeoff is that `SpanData` is cloned to every `SpanProcessor` before per-processor filters run, so a body would briefly live in memory on the OTLP path before being dropped — worth weighing against the surface area of the parallel store.

crates/coral-app/src/telemetry/config.rs
```diff
@@ -60,13 +61,17 @@ impl Default for OtlpConfig {
...
+    pub(crate) record_http_bodies: bool,
+    pub(crate) http_body_max_bytes: usize,
 }
 
 impl Default for TraceHistoryConfig {
     fn default() -> Self {
         Self {
             enabled: true,
             retention_days: DEFAULT_TRACE_HISTORY_RETENTION_DAYS,
+            record_http_bodies: true,
```
antonmry: Suggest defaulting this to `false`. Recording bodies on disk by default — under the user's home directory, on a fresh install, with nothing in the UI pointing it out — is the kind of behavior a user is likely to discover only after the fact, which is a surprising default even for a local-only store.

Concretely:
- Flip the `Default` impl to `record_http_bodies: false` (and the matching test assertion at line 197).
- When the trace history UI renders a trace, show a hint when bodies are absent — something like *"HTTP bodies are not recorded. Set `[trace_history] record_http_bodies = true` in `config.toml` to capture request/response previews."* — so users discover the feature when they actually need it (debugging a failing source) rather than discovering retroactively that previews have been on disk all along.
- Optionally update `docs/guides/observe-with-opentelemetry.mdx` to lead with how to *enable* rather than how to *disable*.

jsummerfield:
I think this needs one more pass before merge.

The sidecar storage design is good, but `get_trace` currently loads every retained HTTP body record before filtering to the requested trace. `apply_http_body_records()` calls `read_http_body_records()`, and that builds a `Vec` from all `http-bodies-*` files for the full retention window before checking `record.trace_id`.

With body capture enabled by default and up to 64KiB per body, opening one trace detail can deserialize a large amount of unrelated body data into memory. It also means a malformed body record for a different trace can break trace detail for the trace the user actually asked for.

I’d stream-filter this path instead: while reading body JSONL files, keep only records matching the requested `trace_id`, and ideally also only records whose `span_id` is in the span set for that trace. That preserves the sidecar approach while keeping trace detail proportional to the selected trace.

AlbertQM (approved):
Docs LGTM

crates/coral-app/src/telemetry/config.rs
jsummerfield: Thinking about this more, I agree, false is the better default here, with good guidelines in the UI for how to enable.

crates/coral-engine/src/composition.rs
sauldhernandez: That was considered, and that's how the PR was initially shaped. But it's hard to guarantee that the span attribute will be dropped, because it depends on how the filtering is set up in the exporters. So we agreed it's safer to do it this way to avoid accidentally exports of body data via OLTP.

antonmry: Adding `coral.http.body=trace` to the otel trace filter by mistake doesn't seem a big concern vs. having to maintain the additional machinery (or having to remove it in the future if we want to support reporting data via OLTP, for example, to a local endpoint).

jsummerfield (approved):

lazybean:
need need need 

shipshipship

PR #377, opened 2026-05-12
feat(app): add OAuth authorization-code credential flow

sauldhernandez:
## Summary
- add the app-side OAuth credential manager for authorization-code sessions
- add start/complete source credential OAuth RPCs
- persist retrieved token material into source credential storage
- preserve gRPC tracing metadata for the new RPC handlers

## Scope
Builds on the credential-method contract from the previous PR. This adds backend/runtime behavior, but not the CLI interaction layer.

## Verification
- make rust-checks
- make docs-check


sauldhernandez:
* **#377** <a href="https://app.graphite.com/github/pr/withcoral/coral/377?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/377?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Bradley-Butcher (approved):
MCP tests lgtm

AlbertQM (approved):
Docs LGTM

jsummerfield:
Given the phasing we discussed, I think this PR should keep the existing import API additive rather than changing it in place.

The current diff changes `ImportSource` from unary to server-streaming and changes `ImportSourceResponse` from a direct `source` field into an `event` oneof. That is a generated-client break for every existing caller, including callers that never opt into OAuth. It gives us the double-break we are trying to avoid: one break now for OAuth, then another when the source/surface/projection model lands.

I’d keep the existing `ImportSource(ImportSourceRequest) returns (ImportSourceResponse)` shape unchanged for the non-OAuth path, and add a separate streaming RPC for the OAuth/interactive path, e.g. `ImportSourceWithCredentials(...) returns (stream ImportSourceWithCredentialsResponse)`. That lets phase 1 add OAuth support to the current manifest model without forcing all import callers to drain a stream or regenerate against a changed response shape.

While touching that RPC shape, I’d also avoid using `SourceVariable` for `credential_inputs`. Those values can include an OAuth client secret, and `SourceVariable` reads as non-secret/source-visible. A small method-scoped message like `CredentialInput { key, value }` with comments saying it may contain sensitive transient material would be less likely to leak through client logging or UI assumptions, without pulling in the future connection model now.

Everything else here can stay current-manifest-shaped for this phase. I would not block this stack on source -> surfaces -> operations/entities -> projections; that should happen once, later, with the planned source-model break.

jsummerfield:
Correction to my earlier comment after clarifying the intended contract boundary: if these gRPC APIs are considered internal and we are explicitly OK breaking/reforming them in this stack, then the unary-to-streaming `ImportSource` change is not a blocker by itself. The `oneof` event stream shape is a reasonable way to model progress plus final source completion.

I’d keep the remaining, narrower concern: `credential_inputs` should not reuse `SourceVariable` if it can carry OAuth client secrets. Even though the gRPC API is internal, the name and existing semantics of `SourceVariable` imply non-secret source config. A tiny OAuth/method-scoped input message would make the contract harder to misuse and easier to refactor later when credentials/connections become their own product concept.

crates/coral-app/src/sources/manager.rs
```diff
@@ -161,7 +204,42 @@ impl SourceManager {
...
+        command: ImportSourceWithCredentialsCommand,
+        events: mpsc::Sender<ImportSourceWithCredentialsEvent>,
+    ) -> Result<InstalledSource, AppError> {
+        let mut candidate =
+            describe_manifest(&command.manifest_yaml, SourceOrigin::Imported, false)?;
+        candidate.installed = self.source_exists(workspace_name, &candidate.name)?;
+        let stored_material = self
+            .read_source_material(workspace_name, &candidate.name)
+            .unwrap_or_default();
+        let oauth_material = self
```
jsummerfield, line 288: This does the browser OAuth flow before validating the ordinary import bindings. That creates a rough failure mode: a user can finish OAuth, then only afterward learn that a required non-OAuth variable is missing or a binding key is invalid. I would add a preflight step before `retrieve_oauth_material` that validates variables, duplicate/unknown bindings, selected OAuth inputs/methods, and source-config/OAuth conflicts using placeholders for the OAuth-filled secrets. The actual token retrieval can then be the first side-effect after the request is known to be structurally valid.

jsummerfield:
One docs follow-up before merging this PR: `docs/reference/source-spec-reference.mdx` still says Coral validates/exposes the OAuth redirect URI but “does not run the OAuth flow yet,” and later says OAuth token exchange / refresh-token storage are not implemented. That was accurate for the metadata PR, but this PR now runs authorization-code exchange and persists refresh tokens when providers return them. I would update this section so the docs distinguish implemented now (loopback auth-code exchange, access-token persistence, refresh-token storage) from still-not-implemented (automatic refresh).

crates/coral-api/proto/coral/v1/sources.proto
```diff
@@ -254,12 +254,66 @@ message ImportSourceRequest {
   repeated SourceVariable variables = 3;
   // Source-owned secret values.
   repeated SourceSecret secrets = 4;
+  // OAuth credentials the app should retrieve before validation and import.
+  repeated ImportSourceOAuthCredential oauth_credentials = 5;
```
jsummerfield: For a follow up (I can pick up) - I think we should split sources and connections at the API level.

jsummerfield (approved):

crates/coral-api/proto/coral/v1/sources.proto
jsummerfield: Apols for the back and forth - `OAuthCredentialRetrievalRequest` ? These aren't really credentials yet are they?

sauldhernandez: I agree the split should exist, let's discuss the shape of it.

sauldhernandez: true, I'll rename it

PR #381, opened 2026-05-12
chore: add skill for reviewing sources

simonwhitaker:


.agents/skills/review-coral-source/SKILL.md
```diff
@@ -0,0 +1,94 @@
...
+Open questions
+- Is endpoint X intentionally omitted from the first version?
+
+Review notes
+- I treated CI lint/schema checks as out of scope unless visible in the diff.
+```
+
+If no issues are found, say that directly and include any limits, such as not having live credentials or not inspecting CI logs.
+
+When highlighting a discrepancy between the source and the API documentation, always include links to the exact API documentation page.
```
jsummerfield, line 96: I might have missed it above, but I think this should be higher up. Most of the bugs codex caught have been a result of finding the authoritiative docs and cross-checking.

jsummerfield (approved):

.agents/skills/review-coral-source/SKILL.md
simonwhitaker: Finding the authoritative docs and cross-checking is implied further up, but I've added a line to make that explicit. This line was specifically around citing the source when reporting findings, rather than just asserting that something is wrong.

PR #383, opened 2026-05-12
feat(sources/community/cal): add Cal.com community source

### LINKED ISSUE #382: design: add Cal community source
## Summary

Add a new community `cal` source that lets Coral query scheduling, bookings, events, availability, and routing data from Cal.com through SQL. This enables querying meeting activity, scheduling analytics, host utilization, cancellations, and workflow automation data directly from Coral.

## Motivation

Cal.com is a widely used open-source scheduling and booking platform. Teams often need to answer questions like:

- “Which event types have the highest booking volume?”
- “What hosts are receiving the most meetings?”
- “How many meetings were cancelled or rescheduled?”
- “Which time slots have the highest utilization?”
- “How are routing forms and scheduling workflows performing?”

Without a bundled source, users need to manually integrate Cal.com APIs or export booking data into external systems. A first-party source makes these workflows accessible for SQL querying, dashboards, scheduling analytics, and operational workflows.

## Proposed design

Add a new source under `sources/community/cal`, implemented using Coral’s HTTP source-spec model.

Inputs:

- `CAL_API_KEY`
- optional `CAL_BASE_URL`

Auth:

- HTTP Bearer Token authentication

Supported environments:

- Cal.com Cloud
- self-hosted Cal.com instances

Default base URL:

- `https://api.cal.com`

Proposed tables:

- `users`
  - users and hosts accessible with the provided API key

- `bookings`
  - scheduled meetings, attendees, status, cancellations, and timestamps

- `event_types`
  - event configuration including duration, slug, scheduling settings, and visibility

- `availability`
  - host availability schedules and booking windows

- `routing_forms`
  - routing forms and lead qualification workflows

- `teams`
  - teams and organizational scheduling structures

Design details:

- use `Timestamp` for ISO-8601 date fields
- preserve nested payloads as `Json`
- support both cloud and self-hosted deployments
- keep v1 focused on scheduling analytics and operational workflows

Affected surface:

- community source only
- expected files:
  - `sources/community/cal/manifest.yaml`
  - `sources/community/cal/README.md`
  - generated source docs

No CLI, runtime, MCP, or config changes are proposed.

## Drawbacks and trade-offs

Large booking datasets may require pagination and filtering to avoid excessive API requests.

Some APIs and scheduling workflows may vary between self-hosted and cloud deployments depending on enabled modules and Cal.com version compatibility.

## Alternatives considered

- Exporting scheduling data into warehouses before querying adds operational overhead
- A bookings-only source would simplify implementation, but event types and availability data are important for operational analytics workflows

## Open questions

- Should workflows and webhook events be added in v2?
- Should payment and invoice integrations become separate tables later?
- Any concerns around pagination or rate limits for large scheduling datasets?

jishanahmed-shaikh:
Closes #382 

## What changed

Adds a new community source for Cal under `sources/community/cal/`.

**Tables (4):**
- `me` — profile of the authenticated user; confirms connectivity and surfaces username, time zone, and default schedule ID
- `event_types` — bookable meeting configurations with duration, slug, booking URL, schedule assignment, buffer times, and visibility settings
- `bookings` — scheduled meetings with status, attendee info, host details, cancellation and rescheduling metadata; cursor-paginated; optional filters for `status`, `attendee_email`, `attendee_name`, and `event_type_id`
- `schedules` — availability schedules with time zone, default flag, and availability windows

**Auth:** 
- Bearer token via `CAL_API_KEY` `(Settings → Security → API Keys)`
- Supports Cal.com Cloud and self-hosted instances via `CAL_BASE_URL` (default: `https://api.cal.com`). 
- Uses `cal-api-version: 2024-06-14`.

## Why it changed

Cal.com is a widely used open-source scheduling platform. Being able to query booking volume, cancellation rates, event type utilisation, and host activity via SQL is a natural fit for Coral — especially for teams running scheduling analytics or operational workflows.

## What was tested

- `coral source lint` — passes
- `uv` / `yamllint` / `make lint-sources` — passes across all sources
- `coral source test cal` — 2/2 test queries pass
- Live queries verified against a real Cal.com account:
  - `me` — returns authenticated user profile with correct columns
  - `event_types` — returns event types with correct columns
  - `schedules` — returns default schedule with time zone and availability
  - `bookings` — schema verified; cursor pagination confirmed working

## User-visible impact

New `cal` source installable via:

```bash
CAL_API_KEY=cal_live_... 
coral source add --file sources/community/cal/manifest.yaml
```

## Follow-on

- Routing forms table (`GET /v2/routing-forms`) could be added in a follow-up
- Teams table for team scheduling analytics

### DIFF
--- a/sources/community/cal/README.md
+++ b/sources/community/cal/README.md
@@ -1,0 +1,112 @@
+# Cal.com
+
+**Version:** 0.1.0
+**Backend:** HTTP
+**Tables:** 4
+**Base URL:** `https://api.cal.com` (override with `CAL_BASE_URL`)
+
+Query bookings, event types, schedules, and profile data from Cal.com
+(Cloud or self-hosted).
+
+## Authentication
+
+Requires a `CAL_API_KEY`. Find it in **Settings → Security → API Keys**.
+
+- Test mode keys have the prefix `cal_`
+- Live mode keys have the prefix `cal_live_`
+
+```bash
+CAL_API_KEY=cal_live_... coral source add --file sources/community/cal/manifest.yaml
+```
+
+Or interactively:
+
+```bash
+CAL_API_KEY=cal_live_... coral source add --file sources/community/cal/manifest.yaml --interactive
+```
+
+### Self-hosted
+
+Set `CAL_BASE_URL` to your instance URL:
+
+```bash
+CAL_API_KEY=cal_... CAL_BASE_URL=https://cal.example.com \
+  coral source add --file sources/community/cal/manifest.yaml
+```
+
+## Tables
+
+| Table | Description | Optional filters |
+|---|---|---|
+| `me` | Profile of the authenticated user | — |
+| `event_types` | Bookable meeting configurations | — |
+| `bookings` | Scheduled meetings and their status | `status`, `attendee_email`, `attendee_name`, `event_type_id` |
+| `schedules` | Availability schedules | — |
+
+## Quick start
+
+```bash
+# Confirm connectivity and see your profile
+coral sql "SELECT id, username, email, name, time_zone FROM cal.me"
+
+# List all event types
+coral sql "
+  SELECT id, title, slug, length_in_minutes, hidden, booking_url
+  FROM cal.event_types
+  ORDER BY length_in_minutes
+"
+
+# All bookings
+coral sql "
+  SELECT id, uid, title, status, start, duration, event_type_id
+  FROM cal.bookings
+  ORDER BY start DESC
+  LIMIT 20
+"
+
+# Filter bookings by status
+coral sql "
+  SELECT uid, title, start, host__name, host__email
+  FROM cal.bookings
+  WHERE status = 'cancelled'
+  ORDER BY start DESC
+  LIMIT 20
+"
+
+# Bookings for a specific event type
+coral sql "
+  SELECT uid, title, start, status
+  FROM cal.bookings
+  WHERE event_type_id = <your-event-type-id>
+  ORDER BY start DESC
+"
+
+# Booking volume by event type
+coral sql "
+  SELECT event_type__slug, status, COUNT(*) as count
+  FROM cal.bookings
+  GROUP BY event_type__slug, status
+  ORDER BY count DESC
+"
+
+# Availability schedules
+coral sql "
+  SELECT id, name, time_zone, is_default
+  FROM cal.schedules
+"
+```
+
+## Discovery order
+
+```text
+me
+  → default_schedule_id → schedules.id
+
+event_types
+  → id (event_type_id)
+    → bookings (WHERE event_type_id = '...')
+  → schedule_id → schedules.id
+
+bookings
+  → event_type_id → event_types.id
+```
--- a/sources/community/cal/README.md
+++ b/sources/community/cal/README.md
@@ -19,7 +19,7 @@ Requires a `CAL_API_KEY`. Find it in **Settings → Security → API Keys**.
 CAL_API_KEY=cal_live_... coral source add --file sources/community/cal/manifest.yaml
 ```
 
-Or interactively:
+Run from the repo root. Or interactively:
 
 ```bash
 CAL_API_KEY=cal_live_... coral source add --file sources/community/cal/manifest.yaml --interactive
--- a/sources/community/cal/README.md
+++ b/sources/community/cal/README.md
@@ -69,7 +69,7 @@ coral sql "
 
 # Recent bookings
 coral sql "
-  SELECT id, uid, status, start, duration, host_name, attendee_email
+  SELECT id, uid, status, start, end_time, duration, host_name, attendee_email
   FROM cal.bookings
   ORDER BY start DESC
   LIMIT 20
--- a/sources/community/cal/README.md
+++ b/sources/community/cal/README.md
@@ -12,7 +12,7 @@ Query bookings, event types, schedules, and profile data from Cal.com
 
 Requires a `CAL_API_KEY`. Find it in **Settings → Security → API Keys**.
 
-- Test mode keys have the prefix `cal_`
+- Test mode keys have the prefix `cal_test_`
 - Live mode keys have the prefix `cal_live_`
 
 ```bash
@@ -30,7 +30,7 @@ CAL_API_KEY=cal_live_... coral source add --file sources/community/cal/manifest.
 Set `CAL_BASE_URL` to your instance URL:
 
 ```bash
-CAL_API_KEY=cal_... CAL_BASE_URL=https://cal.example.com \
+CAL_API_KEY=cal_test_... CAL_BASE_URL=https://cal.example.com \
   coral source add --file sources/community/cal/manifest.yaml
 ```
 
@@ -40,8 +40,19 @@ CAL_API_KEY=cal_... CAL_BASE_URL=https://cal.example.com \
 |---|---|---|
 | `me` | Profile of the authenticated user | — |
 | `event_types` | Bookable meeting configurations | — |
-| `bookings` | Scheduled meetings and their status | `status`, `attendee_email`, `attendee_name`, `event_type_id` |
+| `bookings` | Scheduled meetings and their status | `status`, `attendee_email`, `attendee_name`, `event_type_id`, `after_start`, `before_end`, `after_created_at`, `before_created_at` |
 | `schedules` | Availability schedules | — |
+
+### Bookings status filter note
+
+The `status` filter and the `status` response column use different vocabularies:
+
+| | Values |
+|---|---|
+| Filter (`WHERE status = '...'`) | `upcoming`, `recurring`, `past`, `cancelled`, `unconfirmed` |
+| Response column | `accepted`, `cancelled`, `pending`, `rejected` |
+
+The filter accepts **at most one value** per query — passing multiple values returns 400.
 
 ## Quick start
 
@@ -56,17 +67,26 @@ coral sql "
   ORDER BY length_in_minutes
 "
 
-# All bookings
+# Recent bookings
 coral sql "
-  SELECT id, uid, title, status, start, duration, event_type_id
+  SELECT id, uid, status, start, duration, host_name, attendee_email
   FROM cal.bookings
   ORDER BY start DESC
   LIMIT 20
 "
 
-# Filter bookings by status
+# Bookings in the last 7 days
 coral sql "
-  SELECT uid, title, start, host__name, host__email
+  SELECT uid, title, status, start, host_name, attendee_email
+  FROM cal.bookings
+  WHERE after_start = '2026-05-06T00:00:00Z'
+    AND before_end = '2026-05-13T00:00:00Z'
+  ORDER BY start DESC
+"
+
+# Cancelled bookings
+coral sql "
+  SELECT uid, title, start, cancellation_reason, cancelled_by_email
   FROM cal.bookings
   WHERE status = 'cancelled'
   ORDER BY start DESC
@@ -75,25 +95,22 @@ coral sql "
 
 # Bookings for a specific event type
 coral sql "
-  SELECT uid, title, start, status
+  SELECT uid, title, start, status, attendee_email
   FROM cal.bookings
   WHERE event_type_id = <your-event-type-id>
   ORDER BY start DESC
 "
 
-# Booking volume by event type
+# Booking volume by event type slug
 coral sql "
-  SELECT event_type__slug, status, COUNT(*) as count
+  SELECT event_type_slug, status, COUNT(*) as count
   FROM cal.bookings
-  GROUP BY event_type__slug, status
+  GROUP BY event_type_slug, status
   ORDER BY count DESC
 "
 
 # Availability schedules
-coral sql "
-  SELECT id, name, time_zone, is_default
-  FROM cal.schedules
-"
+coral sql "SELECT id, name, time_zone, is_default FROM cal.schedules"
 ```
 
 ## Discovery order
--- a/sources/community/cal/README.md
+++ b/sources/community/cal/README.md
@@ -97,7 +97,7 @@ coral sql "
 coral sql "
   SELECT uid, title, start, status, attendee_email
   FROM cal.bookings
-  WHERE event_type_id = <your-event-type-id>
+  WHERE event_type_id = 12345
   ORDER BY start DESC
 "
 
--- a/sources/community/cal/manifest.yaml
+++ b/sources/community/cal/manifest.yaml
@@ -315,7 +315,7 @@ tables:
     description: Scheduled meetings and their status
     guide: |
       Each row is one booking. `event_type_id` joins to `cal.event_types`.
-      `start` and `end` are Timestamps — use them for time-based filtering
+      `start` and `end_time` are Timestamps — use them for time-based filtering
       (e.g. WHERE start > NOW() - INTERVAL '7 days').
 
       The `status` filter and the `status` response column use different
--- a/sources/community/cal/manifest.yaml
+++ b/sources/community/cal/manifest.yaml
@@ -17,8 +17,9 @@ inputs:
     kind: variable
     default: https://api.cal.com
     hint: |
-      Base URL for the Cal.com API. Use the default for Cal.com Cloud.
-      For self-hosted Cal.com instances use your own host URL.
+      Base URL for the Cal.com API. Use the default (https://api.cal.com)
+      for Cal.com Cloud. For self-hosted Cal.com instances use your own
+      host URL.
 base_url: "{{input.CAL_BASE_URL}}"
 auth:
   type: HeaderAuth
@@ -159,7 +160,8 @@ tables:
       meeting duration. `hidden = true` means the event type is not
       publicly visible. Results are scoped to the API key owner —
       individual keys return only their own event types; org-scoped keys
-      return all event types in the organization.
+      return all event types in the organization. Org-scoped keys may
+      return large result sets as this table is not paginated.
     request:
       method: GET
       path: /v2/event-types
@@ -326,7 +328,10 @@ tables:
       and merge results.
 
       Use `after_start` / `before_end` for time-window queries instead of
-      fetching all bookings and filtering client-side.
+      fetching all bookings and filtering client-side. Note that
+      `after_start`, `before_end`, `after_created_at`, and
+      `before_created_at` are filter-only parameters and do not appear
+      as selectable columns in the result.
     filters:
       - name: status
         required: false
--- a/sources/community/cal/manifest.yaml
+++ b/sources/community/cal/manifest.yaml
@@ -445,7 +445,7 @@ tables:
             kind: path
             path:
               - start
-      - name: end
+      - name: end_time
         type: Timestamp
         nullable: true
         description: Booking end time
--- a/sources/community/cal/manifest.yaml
+++ b/sources/community/cal/manifest.yaml
@@ -10,8 +10,8 @@ inputs:
     kind: secret
     hint: |
       Your Cal.com API key. Find it in Settings > Security > API Keys.
-      Test mode keys have the prefix cal_ and live mode keys have the
-      prefix cal_live_.
+      Test mode keys have the prefix cal_test_ and live mode keys have
+      the prefix cal_live_.
       See [Cal.com API docs](https://cal.com/docs/api-reference/v2/introduction).
   CAL_BASE_URL:
     kind: variable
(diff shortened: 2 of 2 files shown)


jishanahmed-shaikh:
@cubic-dev-ai Review

antonmry:
Hi @jishanahmed-shaikh , thanks for the contribution. Have you already completed the CLA process? https://github.com/withcoral/coral/blob/main/CONTRIBUTING.md#licence

jishanahmed-shaikh:
> Hi @jishanahmed-shaikh , thanks for the contribution. Have you already completed the CLA process? https://github.com/withcoral/coral/blob/main/CONTRIBUTING.md#licence

Yes I have already completed the Process

sources/community/cal/manifest.yaml
```diff
@@ -0,0 +1,696 @@
...
+    pagination:
+      mode: cursor_query
+      cursor_param: cursor
+      response_cursor_path:
+        - pagination
+        - nextCursor
+      page_size:
+        default: 100
+        max: 100
+        query_param: take
```
antonmry: Hi! Thanks for putting this together. While reading the [Cal v2 _Get all bookings_ docs](https://cal.com/docs/api-reference/v2/bookings/get-all-bookings) I noticed the page-size query parameter there is documented as `limit` (default 50), not `take`. Could you double-check whether `take` is actually being honoured by Cal, or whether the request silently falls back to Cal's default 50? It might also be worth re-confirming the `max: 100` value against the docs — I couldn't find a documented upper bound. WDYT?

sources/community/cal/manifest.yaml
```diff
@@ -0,0 +1,696 @@
...
+      For self-hosted Cal.com instances use your own host URL.
+base_url: "{{input.CAL_BASE_URL}}"
+auth:
+  type: HeaderAuth
+  headers:
+    - name: Authorization
+      from: template
+      template: Bearer {{input.CAL_API_KEY}}
+request_headers:
+  - name: cal-api-version
```
antonmry: One thing I'd love your read on: Cal v2 seems to use per-endpoint `cal-api-version` headers — e.g., the docs I read mention `2024-06-14` for event-types, `2024-06-11` for [schedules](https://cal.com/docs/api-reference/v2/schedules/get-all-schedules), and `2026-05-01` for [bookings](https://cal.com/docs/api-reference/v2/bookings/get-all-bookings). Pinning a single global version here works today (likely via a server-side fallback), but do you think it's worth moving the header into each table's `request` so we're explicit about which schema we expect? Curious what you've seen in live testing.

sources/community/cal/manifest.yaml
```diff
@@ -0,0 +1,696 @@
...
+      person's meetings. Results are cursor-paginated newest-first.
+    filters:
+      - name: status
+        required: false
+      - name: attendee_email
+        required: false
+      - name: attendee_name
+        required: false
+      - name: event_type_id
+        required: false
```
antonmry, line 343: Curious whether you considered adding time-range filters here? The Cal v2 [bookings endpoint](https://cal.com/docs/api-reference/v2/bookings/get-all-bookings) documents `afterStart`, `beforeEnd`, `afterCreatedAt`, `beforeCreatedAt`, `afterUpdatedAt` / `beforeUpdatedAt`, plus a multi-value `eventTypeIds`. Without those, common analytical queries ("cancellations this week", "bookings in last 30 days") have to fetch every booking and filter client-side. Would you be open to wiring at least the date-range ones through?

sources/community/cal/manifest.yaml
```diff
@@ -0,0 +1,696 @@
...
+          path:
+            - disableGuests
+  - name: bookings
+    description: Scheduled meetings and their status
+    guide: |
+      Each row is one booking. Use `status` to filter by accepted,
+      cancelled, or pending. `event_type_id` joins to `cal.event_types`.
+      `start` and `end` are ISO 8601 timestamps. Filter by `status` to
+      analyse cancellations or by `attendee_email` to look up a specific
+      person's meetings. Results are cursor-paginated newest-first.
```
antonmry: Reading the Cal docs, it looks like the `status` query filter uses one vocabulary (`upcoming` / `recurring` / `past` / `cancelled` / `unconfirmed`, single-value-only — multiple values returns 400) while the response `status` field uses another (`ACCEPTED` / `CANCELLED` / `PENDING` / `REJECTED`). The current column description and this guide refer to the response vocabulary, but the same name `status` is also a filter. Do you think it's worth calling that distinction out in the guide so users don't accidentally write `WHERE status IN ('pending', 'cancelled')` and hit a 400?

sources/community/cal/manifest.yaml
```diff
@@ -0,0 +1,696 @@
...
+          kind: from_filter
+          key: status
+      - name: filter_attendee_email
+        type: Utf8
+        nullable: true
+        virtual: true
+        description: Echoes the attendee_email filter applied to this query
+        expr:
+          kind: from_filter
+          key: attendee_email
```
antonmry: Could you walk me through the intent of the `filter_status` / `filter_attendee_email` virtual columns? Looking at how `from_filter` is used in `sources/core/linear/manifest.yaml` and `sources/core/notion/manifest.yaml`, the existing pattern seems to be one column per filter (named the same as the filter), without a parallel "real" column from the response. Here we end up with both a `status` response column and a `filter_status` virtual column — and `attendee_name` / `event_type_id` filters don't have echo columns at all. Would the source be easier to use if we picked one convention?

sources/community/cal/manifest.yaml
```diff
@@ -0,0 +1,696 @@
...
+      - name: meeting_url
+        type: Utf8
+        nullable: true
+        virtual: false
+        description: Video meeting URL
+        expr:
+          kind: path
+          path:
+            - meetingUrl
+      - name: host__id
```
antonmry: Small consistency check — the nested-field flattening here uses double underscores (`host__id`, `host__name`, `event_type__slug`), but existing manifests like `sources/core/linear/manifest.yaml` flatten with single underscores (e.g., `team_id`, `team_name`). Is there a reason to keep the `__`, or would you be open to renaming to `host_id` / `event_type_slug` to match the rest of the repo?

sources/community/cal/manifest.yaml
```diff
@@ -0,0 +1,696 @@
...
+      - name: host__time_zone
+        type: Utf8
+        nullable: true
+        virtual: false
+        description: Host time zone (first host)
+        expr:
+          kind: first_array_item_path
+          path:
+            - hosts
+          item_path:
```
antonmry: The `bookings` table surfaces the first host (id/name/email/time_zone) but I couldn't find any attendee columns, even though the table accepts `attendee_email` / `attendee_name` as filters. If a user filters by an attendee they won't be able to see who the attendee actually is on the result row. Would it make sense to expose attendee fields the same way you've done for hosts?

sources/community/cal/manifest.yaml
```diff
@@ -0,0 +1,696 @@
...
+        - data
+      allow_404_empty: false
+      row_strategy: direct
+    pagination:
+      mode: none
+    columns:
+      - name: id
+        type: Int64
+        nullable: false
+        virtual: false
```
antonmry: Style nit: existing manifests in `sources/core/` (intercom, notion, linear) only set `virtual: true` and omit the field otherwise. Would you mind dropping `virtual: false` throughout to match the house style? It would also trim ~50 lines from the file.

sources/community/cal/manifest.yaml
```diff
@@ -0,0 +1,696 @@
...
+      - name: end
+        type: Utf8
+        nullable: true
+        virtual: false
+        description: Booking end time (ISO 8601)
+        expr:
+          kind: path
+          path:
+            - end
+      - name: duration
```
antonmry, line 464: Cal documents `start`, `end`, `createdAt`, `updatedAt` as ISO 8601 strings. The notion source models the equivalent fields as `Timestamp` rather than `Utf8`, which lets users do things like `WHERE start > NOW() - INTERVAL '7 days'`. Given bookings is the table users will most often time-filter, do you think `Timestamp` would be a friendlier choice here? (Happy to ignore if there's a reason the manifest is staying string-typed.)

sources/community/cal/README.md
```diff
@@ -0,0 +1,112 @@
...
+**Base URL:** `https://api.cal.com` (override with `CAL_BASE_URL`)
+
+Query bookings, event types, schedules, and profile data from Cal.com
+(Cloud or self-hosted).
+
+## Authentication
+
+Requires a `CAL_API_KEY`. Find it in **Settings → Security → API Keys**.
+
+- Test mode keys have the prefix `cal_`
```
antonmry: Quick nit — I think Cal's test-mode keys actually carry the prefix `cal_test_` (paired with `cal_live_`), not bare `cal_`. Could you double-check and tighten the wording so users don't expect any `cal_…` key without `_live_` to automatically be a test key?

sources/community/cal/manifest.yaml
```diff
@@ -0,0 +1,696 @@
...
+        virtual: false
+        description: Organization ID if the user belongs to an organization
+        expr:
+          kind: path
+          path:
+            - organizationId
+  - name: event_types
+    description: Event types configured by the authenticated user
+    guide: |
+      Each row is one event type — a bookable meeting configuration. Use
```
antonmry, line 157: The Cal v2 [event-types endpoint](https://cal.com/docs/api-reference/v2/event-types/get-all-event-types) supports `username`, `usernames`, `eventSlug`, `orgSlug`, `orgId` as filters. For single-user API keys none of this matters, but org-scoped keys will return a lot. Could the guide at least note that scope depends on the key's owner? Or, if you'd rather, expose `username` as a filter — interested in your take.

sources/community/cal/manifest.yaml
```diff
@@ -0,0 +1,696 @@
...
+          path:
+            - bio
+      - name: organization_id
+        type: Int64
+        nullable: true
+        virtual: false
+        description: Organization ID if the user belongs to an organization
+        expr:
+          kind: path
+          path:
```
antonmry, line 143: Minor: you surface `organization_id` here, but the `/v2/me` response also returns a nested `organization` object (with `isPlatform`). Worth either exposing both fields or mentioning in the guide that org details require a separate lookup — what do you think?

sources/community/cal/manifest.yaml
jishanahmed-shaikh: You're right, the current Cal v2 bookings docs document limit (default 50) as the page-size parameter, not take. I'll switch to limit and drop the max: 100 since there's no documented upper bound in the spec. Will refactor

sources/community/cal/manifest.yaml
jishanahmed-shaikh: Good catch. The bookings endpoint actually requires cal-api-version: 2026-05-01 per the current docs, that's the version that exposes the full filter set including afterStart, beforeEnd, etc. I'll move the version header into each table's request block so we're explicit about the schema each table expects, rather than relying on server-side fallback

sources/community/cal/manifest.yaml
jishanahmed-shaikh: Completely agree, these are the most useful filters for scheduling analytics. I'll wire through after_start, before_end, after_created_at, and before_created_at at minimum. I already has this in mind, working on this.

sources/community/cal/manifest.yaml
jishanahmed-shaikh: You're right, This might become an issue. The filter status and the response status column use completely different vocabularies. I'll update the guide to explicitly document both sets of values and warn that the filter accepts only one value at a time.

sources/community/cal/manifest.yaml
jishanahmed-shaikh: Fair point. The issue is that status is both a real response field and a filter, so I added a filter_status prefix to avoid collision. Looking at the existing pattern more carefully, the right approach is to drop the echo columns entirely for filters that have a real response column with the same name, the value is already visible in the row. I'll remove filter_status and filter_attendee_email and keep only the real columns.

sources/community/cal/manifest.yaml
jishanahmed-shaikh: I'll add attendee__name, attendee__email, and attendee__time_zone columns mapped from the first item in the attendees array, matching the same pattern used for host fields.

sources/community/cal/manifest.yaml
jishanahmed-shaikh: Will remove all virtual: false declarations to match the house style.

sources/community/cal/manifest.yaml
jishanahmed-shaikh: Good suggestion. I'll try and test to work around switching start, end, created_at, and updated_at on bookings to Timestamp with format_timestamp: iso8601 so time-based filtering works natively.

sources/community/cal/manifest.yaml
jishanahmed-shaikh: Will add a note to the guide that results are scoped to the key owner, individual keys return only their own event types, org-scoped keys return all org event types.

sources/community/cal/README.md
jishanahmed-shaikh: You're right, I'll correct the README to say test keys have the prefix cal_test_ and live keys have cal_live_.

sources/community/cal/manifest.yaml
jishanahmed-shaikh: Will add organization__is_platform column to surface the nested field.

jishanahmed-shaikh:
Feedback from @antonmry is acknowledged and worked upon, @cubic-dev-ai Review the changes

sources/community/cal/manifest.yaml
jishanahmed-shaikh: Acknowledged, will work on the consistency with single _

jishanahmed-shaikh:
@cubic-dev-ai Review

sources/community/cal/manifest.yaml
antonmry: Thanks! Tiny cross-thread nudge — to stay consistent with what you agreed in https://github.com/withcoral/coral/pull/383#discussion_r3229797737, would the new columns land as `attendee_name` / `attendee_email` / `attendee_time_zone` (single underscore) rather than `attendee__*`? Same question applies to the `organization__is_platform` over in https://github.com/withcoral/coral/pull/383#discussion_r3229800400 — likely `organization_is_platform`. Happy to defer if you'd rather keep `__` everywhere, just flagging so the renames and the new columns land on the same convention.

sources/community/cal/manifest.yaml
antonmry: Sounds good — one open question before you cut the echo columns: do you know whether Coral's planner pushes `WHERE status = 'cancelled'` against the *real* response column down to the `status` query param, or does the filter only fire when the predicate targets a `from_filter` virtual column? Looking at `sources/core/linear/manifest.yaml` and `sources/core/notion/manifest.yaml`, the existing pattern only uses virtuals for filterable fields, which makes me suspect pushdown needs the virtual. Worth a quick local test (or a check with a spec author) — otherwise dropping the echoes could lose API-side filtering entirely and silently turn `WHERE status = …` into a client-side filter over the full booking history.

jishanahmed-shaikh: So I dug into the Coral engine source to confirm. As per my understanding Filter pushdown is driven entirely by the `filters:` block in the manifest (`extract_filter_values` in `filter_expr.rs` matches SQL predicates against `defined_filters` by name). The `from_filter` virtual column is only for echoing the filter value back in the result row, it plays no role in whether the API-side filter fires. So `WHERE status = 'cancelled'` pushes down to the Cal API as long as `status` is in the `filters:` list, regardless of whether a virtual echo column exists. Dropping `filter_status` and `filter_attendee_email` is safe, API-side filtering is preserved.

Also confirmed with a local test and ran the `coral-review-source-spec`, it suggested a few "Low" level guide wording tweaks, making those changes and updating.

sources/community/cal/manifest.yaml
jishanahmed-shaikh: Confirmed the consistency, the latest commit uses single underscore throughout: `attendee_name`, `attendee_email`, `attendee_time_zone`, and `organization_is_platform`. No __ anywhere in the new columns.

jishanahmed-shaikh:
Implemented Minor Changes based off suggestion from running `coral-review-source-spec`, Review the Changes @cubic-dev-ai

sources/community/cal/manifest.yaml
```diff
@@ -0,0 +1,723 @@
...
+        required: false
+      - name: before_created_at
+        required: false
+    request:
+      method: GET
+      path: /v2/bookings
+      headers:
+        - name: cal-api-version
+          from: literal
+          value: "2026-05-01"
```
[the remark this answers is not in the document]
jishanahmed-shaikh: @cubic-dev-ai

The 2026-05-01 version is correct, the Cal v2 bookings docs explicitly require it and it was verified in live testing.

Reference Docs : https://cal.com/docs/api-reference/v2/bookings/get-all-bookings

sources/community/cal/README.md
```diff
@@ -0,0 +1,129 @@
...
+  WHERE status = 'cancelled'
+  ORDER BY start DESC
+  LIMIT 20
+"
+
+# Bookings for a specific event type
+coral sql "
+  SELECT uid, title, start, status, attendee_email
+  FROM cal.bookings
+  WHERE event_type_id = <your-event-type-id>
```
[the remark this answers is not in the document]
jishanahmed-shaikh: @cubic-dev-ai

Fixed the SQL placeholder issue. Replaced `<your-event-type-id>` with 12345, a valid SQL literal users can substitute

Commit Reference: https://github.com/withcoral/coral/pull/383/commits/6b0e8de7050f0bcc223cd6a3cb5c0d143266f24b

jishanahmed-shaikh:
@cubic-dev-ai, as for the `"[sources/community/cal/manifest.yaml]"`

```yaml
headers:
  - name: cal-api-version
    from: literal
    value: "2026-05-01"
```

`2026-05-01` is the correct and documented version for the Cal.com v2 bookings endpoint. The Cal docs explicitly state it must be set to `2026-05-01` for that endpoint, and also verified it works in live testing.

Reference Docs : "https://cal.com/docs/api-reference/v2/bookings/get-all-bookings"

Review @cubic-dev-ai

jishanahmed-shaikh:
> > @cubic-dev-ai, as for the `"[sources/community/cal/manifest.yaml]"`
> > ```yaml
> > headers:
> >   - name: cal-api-version
> > ...
> > ```
> 
> @jishanahmed-shaikh I checked the Cal.com API reference at the URL you referenced, and there's a discrepancy worth flagging.
> 
> The [Cal.com `GET /v2/bookings` docs](https://cal.com/docs/api-reference/v2/bookings/get-all-bookings) currently show `2026-02-25` as the required version:
> 
> ```yaml
> operationId: BookingsController_2026_02_25_getBookings
> parameters:
>   - name: cal-api-version
>     in: header
>     description: Must be set to 2026-02-25.
>     required: true
>     schema:
>       type: string
>       default: '2026-02-25'
> ```
> 
> Your manifest uses `2026-05-01`, which doesn't match what's documented. However, a few possibilities:
> 
> 1. **The docs may have been updated since you tested** — if Cal.com pushed a new version (`2026-05-01`) and the API reference on the doc site hasn't been updated yet, your live-tested value could be correct and the docs stale.
> 2. **The API may accept both** — Cal.com might apply a "latest available" fallback if a version isn't strictly matched, which would explain why `2026-05-01` works in testing even if it's not the documented version.
> 3. **The docs may be accurate and `2026-02-25` is the correct value** — in which case `2026-05-01` might be silently falling back to the default/older version, potentially missing the features you need (like the full filter set).
> 
> **My recommendation:** If you can confirm that `2026-02-25` also passes the filter tests, I'd align it with the documented value to avoid surprises when Cal.com eventually enforces strict version matching. If `2026-02-25` doesn't support the full filter set and `2026-05-01` does (and you've verified it), then keep `2026-05-01` but consider adding a brief inline comment noting that this was verified against live testing and may differ from the published docs.

`2026-05-01` is the correct version for `GET /v2/bookings`. The rendered docs at (https://cal.com/docs/api-reference/v2/bookings/get-all-bookings) explicitly state "Must be set to 2026-05-01" and the curl example on that page uses it. Each cal.com endpoint has its own version, 2026-02-25 appears on other endpoints like the recordings endpoint, but that is unrelated. Verified in live testing as well. 

@cubic-dev-ai

jishanahmed-shaikh:
@cubic-dev-ai Review

antonmry (approved):
Thanks for the contribution and for addressing the feedback

PR #396, opened 2026-05-13
feat(slack): expose rich message payload columns

lazybean:
## Summary

Implements Linear [AOL-1](https://linear.app/withcoral/issue/AOL-1/expose-slack-message-attachments-and-files-in-coral) as a **manifest-only** update to the bundled Slack source.

The change exposes rich Slack message payload data so users can inspect files, images, attachments, blocks, and link-preview metadata from SQL instead of seeing those messages as blank or text-only rows.

## What changed

Only `sources/core/slack/manifest.yaml` is changed.

- Adds raw JSON columns to both `slack.messages` and `slack.thread_replies`:
  - `files`
  - `attachments`
  - `blocks`
- Adds flattened convenience columns for common file/image metadata, including first-file fields and joined fields for IDs, names, URLs, thumbnails, attachment image URLs, block types, and block image URLs.
- Adds manifest `test_queries` showing how to query rich payload columns for known Slack messages/threads.
- Updates the Slack source description/help text to mention rich payloads and the optional `files:read` scope for private file URL access.
- Preserves existing columns and behavior.

## Validation

- `RUSTC_WRAPPER= cargo test -p coral-spec --locked`
- `RUSTC_WRAPPER= cargo test -p coral-app --locked sources`
  - catalog/unit source tests passed
  - sandbox blocked unrelated gRPC port-binding tests in `coral-app --test grpc`

### DIFF
--- a/crates/coral-app/src/sources/catalog.rs
+++ b/crates/coral-app/src/sources/catalog.rs
@@ -163,8 +163,8 @@ mod tests {
             "Slack manifest hint should mention {FILES_SCOPE}"
         );
         assert!(
-            manifest.manifest_yaml.contains(ENCODED_FILES_SCOPE),
-            "prefilled Slack app link should request {FILES_SCOPE}"
+            !manifest.manifest_yaml.contains(ENCODED_FILES_SCOPE),
+            "prefilled Slack app link should stay on the baseline scopes"
         );
     }
 
--- a/crates/coral-app/src/sources/catalog.rs
+++ b/crates/coral-app/src/sources/catalog.rs
@@ -152,16 +152,19 @@ mod tests {
 
     #[test]
     fn slack_manifest_documents_required_files_scope() {
+        const FILES_SCOPE: &str = "files:read";
+        const ENCODED_FILES_SCOPE: &str = "files%3Aread";
+
         let slack = SourceName::parse("slack").expect("source");
         let manifest = load_bundled_source(&slack).expect("slack bundled source");
 
         assert!(
-            manifest.manifest_yaml.contains("files:read"),
-            "Slack manifest hint should mention files:read"
+            manifest.manifest_yaml.contains(FILES_SCOPE),
+            "Slack manifest hint should mention {FILES_SCOPE}"
         );
         assert!(
-            manifest.manifest_yaml.contains("files%3Aread"),
-            "prefilled Slack app link should request files:read"
+            manifest.manifest_yaml.contains(ENCODED_FILES_SCOPE),
+            "prefilled Slack app link should request {FILES_SCOPE}"
         );
     }
 
--- a/crates/coral-app/src/sources/catalog.rs
+++ b/crates/coral-app/src/sources/catalog.rs
@@ -151,6 +151,21 @@ mod tests {
     }
 
     #[test]
+    fn slack_manifest_documents_required_files_scope() {
+        let slack = SourceName::parse("slack").expect("source");
+        let manifest = load_bundled_source(&slack).expect("slack bundled source");
+
+        assert!(
+            manifest.manifest_yaml.contains("files:read"),
+            "Slack manifest hint should mention files:read"
+        );
+        assert!(
+            manifest.manifest_yaml.contains("files%3Aread"),
+            "prefilled Slack app link should request files:read"
+        );
+    }
+
+    #[test]
     fn describe_manifest_extracts_declared_inputs() {
         let source = describe_manifest(
             r#"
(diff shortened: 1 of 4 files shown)


lazybean (commented):
I should only change `sources/core/slack/manifest.yaml`
Other files should remain unchanged

lazybean:
Addressed the review comment: PR #396 is now rewritten to a single manifest-only commit changing only `sources/core/slack/manifest.yaml`.

sources/core/slack/manifest.yaml
```diff
@@ -29,6 +32,28 @@ auth:
       template: Bearer {{input.SLACK_TOKEN}}
 test_queries:
   - SELECT * FROM slack.channels LIMIT 1
+  - >
+    SELECT ts, thread_ts, text, files, attachments, blocks, file_id,
+    file_name, file_title, file_mimetype, file_filetype, file_url_private,
+    file_url_private_download, file_thumb_360, file_thumb_720, file_ids,
+    attachment_image_urls, block_types
+    FROM slack.messages
+    WHERE channel = 'C0AFMHJTMK8'
```
sauldhernandez: This makes the bundled Slack validation workspace-specific. `coral source test slack` runs these queries for every installed Slack source, so most users will get Slack `channel_not_found`/access failures even with a valid token. For `slack.messages`, only `channel`/`oldest`/`latest` are pushed to `conversations.history`, so the `thread_ts` and `ts` predicates are applied after fetching and can crawl a large channel. Keep `test_queries` workspace-agnostic, and cover rich payloads with mocked fixture tests instead.

sources/core/slack/manifest.yaml
```diff
@@ -182,6 +232,274 @@ tables:
           kind: path
           path:
             - text
+      - name: files
```
sauldhernandez, line 281: Inserting the new columns before the existing `ts`/`permalink` fields changes the ordinal schema for `SELECT *`, Arrow IPC, and CSV consumers; `schema_from_columns` preserves manifest order. Make this additive by appending the rich-payload columns after the existing columns in both `messages` and `thread_replies`.

sauldhernandez (changes requested):
Reviewed with Review Buddy.

lazybean:
Verification note from #ludo-sandbox (C0B3UF0CYH1): this branch exposes the rich payload columns as expected, and the channel currently demonstrates non-null `files` and `blocks` payloads but no non-null `attachments` payloads. One follow-up question: should Coral provide access to the actual Slack file contents, not just the message-embedded file metadata/preview? For example, `plugin-explainer.html` is exposed in `files` with metadata and a truncated Slack preview (737 lines total, preview shows only the first few lines; `preview_is_truncated=true`). If we want full contents, I think we would need a deliberate design for how to fetch `url_private` / `url_private_download` with the Slack token (likely requiring `files:read`), and decide whether that belongs as a separate table/function/virtual column rather than inflating `slack.messages` rows. Can we clarify whether full file body retrieval is in scope for this PR/ticket, and if so what shape we want?

sources/core/slack/manifest.yaml
```diff
@@ -1,9 +1,10 @@
 name: slack
-version: 2.1.0
+version: 2.1.1
 dsl_version: 3
 backend: http
 description: >-
-  Query channels, messages, thread replies, and users from your Slack workspace.
+  Query channels, messages, thread replies, users, and rich message payloads
```
AlbertQM, line 6: I think we can keep the old description; as a user, I'd expect "messages" to encompass everything about a message (ie its "rich payload")

```suggestion
  Query channels, messages, thread replies, and users from your Slack workspace.
```

sources/core/slack/manifest.yaml
```diff
@@ -14,7 +15,9 @@ inputs:
       If you already have a Slack app with the required scopes, you can
       find your Bot User OAuth Token on its OAuth & Permissions page. The
       required scopes are: channels:read, channels:history, groups:read,
-      groups:history, users:read, users:read.email.
+      groups:history, users:read, users:read.email. Add `files:read` if you
+      plan to inspect or download the private file URLs exposed in message
+      payloads.
```
AlbertQM, line 20: Should we just update the URL below to also include this new scope? Otherwise we'll get runtime errors when Coral tries to fetch files?

AlbertQM (commented):
Thank you! Two tiny nits to consider 😄

crates/coral-engine/tests/engine/http_tests.rs
lazybean, line 1: (A)I should not change that.

crates/coral-app/src/sources/catalog.rs
lazybean, line 1: A)I should not change that.

PR #398, opened 2026-05-13
feat(cli,mcp): embed build identity in `--version` and `coral://build`

enigma:
## Why

When testing Coral — verifying a freshly built binary, checking which worktree is
running, confirming a server under test is the version you expect — there was no
reliable way to tell from the binary itself. `coral --version` only showed the semver.
This matters at the CLI, and doubly so when Coral is running as an MCP server: agents
and test harnesses have no shell access to run external probes.

## What changed

- `coral --version` now embeds the git SHA as semver build metadata in all builds:
  `0.2.0+abc1234`. In debug builds it also includes the working-tree hash and source
  path: `0.2.0+abc1234 (debug, tree: d917cd2, src: /path/to/worktree)`. A dirty
  working tree gets a `-wip` suffix on the tree hash.
- The MCP `initialize` response now includes the same full version string in
  `serverInfo.version`. Clients receive it at connect time with no extra tool calls.

## Test plan

**CLI — release build:**
```bash
cargo build --release -p coral-cli && ./target/release/coral --version
# coral 0.2.0+<sha>
```

**CLI — debug build:**
```bash
cargo build -p coral-cli && ./target/debug/coral --version
# coral 0.2.0+<sha> (debug, tree: <tree>, src: /path/to/worktree)
```

**MCP serverInfo version matches `--version`:**
```bash
echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"test","version":"0"}}}' \
  | ./target/debug/coral mcp-stdio 2>/dev/null | head -1 | jq '.result.serverInfo'
# { "name": "coral", "version": "0.2.0+<sha> (debug, tree: <tree>, src: ...)" }
```

**Resource list no longer includes `coral://build`:**
```bash
printf '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"test","version":"0"}}}\n{"jsonrpc":"2.0","method":"notifications/initialized","params":{}}\n{"jsonrpc":"2.0","id":2,"method":"resources/list","params":{}}\n' \
  | ./target/debug/coral mcp-stdio 2>/dev/null | jq -r 'select(.id==2) | [.result.resources[].uri]'
# ["coral://guide","coral://tables"]
```


enigma:
* **#398** <a href="https://app.graphite.com/github/pr/withcoral/coral/398?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/398?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mystic123 (approved):

crates/coral-cli/build.rs
```diff
@@ -9,16 +9,26 @@
...
+
+    if profile == "debug" {
+        let source_path = repo_root().unwrap_or_else(|| "unknown".to_owned());
+        let (tree, is_wip) = wip_tree().unwrap_or_else(|| ("unknown".to_owned(), false));
+        let version_tree = if is_wip {
+            format!("{tree}-wip")
+        } else {
+            tree.clone()
+        };
+        long_version = format!("{long_version} (debug, tree: {version_tree}, src: {source_path})");
```
sauldhernandez, line 26: This embeds a working-tree identity, but the build script only reruns on the git ref paths below. Because the script emits `rerun-if-changed`, Cargo will not rerun it for ordinary repo source changes, staged index changes, or changes in sibling crates, so a debug rebuild can keep a stale `CORAL_LONG_VERSION`/`CORAL_WIP_TREE`. Add rerun coverage for the files/index that feed the tree hash, or compute this identity somewhere that is refreshed for each debug run.

crates/coral-mcp/src/surface/resources.rs
```diff
@@ -26,6 +28,17 @@ pub(crate) fn tables_resource(visible_table_count: usize) -> Resource {
...
 
+pub(crate) fn build_resource() -> Resource {
+    RawResource::new("coral://build", "build")
+        .with_description("Coral MCP runtime build identity: version, git SHA, working-tree hash, source path. Use to confirm the running binary matches the worktree being inspected.")
+        .with_mime_type("text/plain")
+        .no_annotation()
+}
+
+pub(crate) fn build_resource_content(build_identity: BuildIdentity) -> String {
+    build_identity.long_version.to_owned()
```
sauldhernandez: This returns the raw `CORAL_LONG_VERSION`, but Clap renders `coral --version` as `coral {CORAL_LONG_VERSION}`. That breaks the advertised contract that `coral://build` returns the same plain-text identity as `coral --version`, so direct comparisons will false-negative. Either return the actual CLI display string here, or rename/update the contract and tests to say this is the raw build version string.

sauldhernandez (approved):
LGTM in general. Two small inline comments

jsummerfield (changes requested):
Can you update the description with the use case? Without the why behind this/what problem you're trying to solve, it's hard to give meaningful feedback.

My main concern: We should not advertise `coral://build`  unconditionally in the MCP surface. This feels like a relatively niche debugging affordance, and every always-on MCP capability adds a bit more discovery/context surface for clients and agents to carry around.

crates/coral-cli/build.rs
enigma: Good catch, removed, thank you!

crates/coral-mcp/src/surface/resources.rs
enigma: Both fixed and then some — after discussion we dropped `coral://build` entirely and moved the version string into `serverInfo.version` in the MCP `initialize` response instead. Clients get it at connect time with no extra resource, and the surface stays clean. The `coral ` prefix issue is moot now.

enigma (commented):
Updated the PR description with the why — the short version is: `coral --version` only showed semver before, making it hard to tell binaries apart when testing, and doubly so when Coral is running as an MCP server where you have no shell access.

On `coral://build`: good point. I dropped the resource entirely. The version string now lives in `serverInfo.version` in the `initialize` response — standard MCP, zero extra surface, clients already have it in context from the handshake.

jsummerfield:
This looks like it can still report the wrong build identity after normal Git ref movement.

`build.rs` now embeds Git-derived data into `CORAL_LONG_VERSION`, but the old `cargo:rerun-if-changed` tracking for `.git/HEAD`, the checked-out ref, and `packed-refs` was removed. Cargo can therefore reuse the previous build-script output when only Git state moves.

I reproduced this with adjacent PR commits:

```bash
git checkout 26330b0e
CARGO_TARGET_DIR=/tmp/coral-pr398-target cargo build -p coral-cli
/tmp/coral-pr398-target/debug/coral --version
# coral 0.2.1+26330b0e ...

git checkout 3c4d0bac
CARGO_TARGET_DIR=/tmp/coral-pr398-target cargo build -p coral-cli
/tmp/coral-pr398-target/debug/coral --version
# still coral 0.2.1+26330b0e ...
```

That makes the feature unreliable in exactly the scenario it is meant to solve: verifying which binary/server is running. The MCP `serverInfo.version` path inherits the same stale `CORAL_LONG_VERSION`, so clients can also receive the wrong identity at initialize time.

I think this needs to restore explicit Git ref rebuild triggers, or otherwise force the build script to rerun when the Git identity inputs change. It would also be worth adding a targeted test or xtask-style check for the stale-ref case, since ordinary unit/integration tests do not exercise Cargo's build-script cache behavior.

crates/coral-cli/src/lib.rs
```diff
@@ -446,6 +446,7 @@ async fn run_app_command(
                 coral_mcp::McpOptions {
                     feedback_enabled: args.enable_feedback,
                     trace_parent: ctx.and_then(|ctx| ctx.trace_parent.clone()),
+                    long_version: env!("CORAL_LONG_VERSION"),
```
jsummerfield, line 449: Since this now changes the MCP initialize surface by putting the binary identity into `serverInfo.version`, can we add a direct assertion for that in the raw stdio initialize test? Right now `mcp_stdio_raw_tools_list_advertises_client_compatible_schemas` only checks that `protocolVersion` exists, so this wiring could regress without a test failure.

jsummerfield (commented):
One additional inline test-coverage note.

jsummerfield:
My May 15 comment still stands: the current implementation can still report stale build identity after normal Git ref movement unless the build script restores tracking for the Git identity inputs or uses another explicit freshness mechanism.

PR #399, opened 2026-05-13
feat: predicate push down implementation

iaptsiauri:
# Dependent Predicate Pushdown Stack

This PR is the first PR in the dependent predicate pushdown stack. The stack should be reviewed in order because each branch builds on the previous one.

## Review Order

1. #399 - `feat: predicate push down spec`
   - Branch: `iaptsiauri/predicate-push-down-spec`
   - Base: `main`
   - Introduces the source-spec capability surface for dependent predicate pushdown.
   - Adds bindable filter metadata, V1 wire type metadata, search-index safety validation, JSON schema/docs updates, and engine registration guards for unsupported bindable backends.

2. #409 - `feat: predicate push down planner`
   - Branch: `iaptsiauri/predicate-push-down-planner`
   - Base: `iaptsiauri/predicate-push-down-spec`
   - Adds dependent join candidate analysis, the DataFusion logical extension node, optimizer rewrite, safe scan peeling, and planner tests.
   - This PR detects supported joins but does not yet perform dependent HTTP execution.

3. #605 - `feat(engine): execute dependent joins`
   - Branch: `iaptsiauri/predicate-push-down-exec-core`
   - Base: `iaptsiauri/predicate-push-down-planner`
   - Adds physical dependent join execution: binding projection, dependent HTTP fetch dispatch, local joined output construction, extension planner lowering, and core execution tests.

4. #606 - `fix(engine): harden dependent join execution`
   - Branch: `iaptsiauri/predicate-push-down-exec-hardening`
   - Base: `iaptsiauri/predicate-push-down-exec-core`
   - Hardens HTTP completeness, literal filter preservation, cap/error behavior, resolver-overflow fallback, trace propagation, EXPLAIN metrics, and related tests.
   - Also removes manifest-level dependent-join caps from the final API shape. Row, binding, and fanout limits are engine runtime policy, not source manifest metadata.

5. #607 - `feat(sources): enable dependent pushdown pilots`
   - Branch: `iaptsiauri/predicate-push-down-pilots`
   - Base: `iaptsiauri/predicate-push-down-exec-hardening`
   - Enables pilot bindable filters in the real source manifests for GitHub pull requests, Linear issues, and Slack messages.

## Final Feature Shape

The complete stack implements V1 dependent predicate pushdown for inner equi-joins where one side is an HTTP/API-backed dependent table with bindable filters. The intended execution model is:

1. Plan a normal SQL inner equi-join.
2. Detect when one HTTP table can be safely bound by join keys.
3. Execute the resolver side.
4. Derive distinct non-null binding tuples.
5. Fetch dependent HTTP rows using the existing HTTP client path.
6. Complete the SQL join locally for correctness.

Important final constraints:

- Only inner equi-joins are in scope.
- Search-index-backed tables cannot expose bindable filters.
- Bindable filters must be equality filters.
- V1 wire serialization is string-only.
- Required dependent filters must be satisfied by join bindings or literals.
- Binding fanout and per-binding dependent row overflow are hard errors.
- Resolver-row overflow can retry the original query without the dependent join rewrite.
- Dependent fetches must preserve HTTP pagination completeness and must not interpret cache or page hints as SQL completeness limits.

## Validation

Validation run on the top of the reconstructed stack:

```sh
cargo fmt --check
cargo test -p coral-spec bindable --no-fail-fast
cargo test -p coral-engine dependent_join --no-fail-fast
```

Results:

- `cargo fmt --check` passed
- `cargo test -p coral-spec bindable --no-fail-fast` passed
- `cargo test -p coral-engine dependent_join --no-fail-fast` passed

## Reviewer Notes

For this PR specifically, focus on the source-spec contract and validation shape. The planner, execution, fallback, and pilot-source details are intentionally handled in later PRs in the stack.


crates/coral-engine/src/backends/common.rs
```diff
@@ -111,6 +111,10 @@ pub(crate) trait CompiledBackendSource: Send + Sync {
 
     fn source_name(&self) -> &str;
 
+    fn backend_kind(&self) -> &'static str;
```
antonmry: Consider returning a `BackendKind` enum here instead of `&'static str`. Then `validate_bindable_backend_support` can use an exhaustive `match` (or `matches!` with a named variant), and adding a new backend forces an explicit decision in the V1 allow-list rather than silently passing the `!= "http"` check at runtime. Also removes the risk of typos in the string comparison.

antonmry (approved):

docs/reference/source-spec-reference.mdx
```diff
@@ -507,6 +508,62 @@ rate_limit:
...
+  max_concurrency: 8
+```
+
+### HTTP dependent-join capability metadata
+
+Coral can mark HTTP table filters as safe for future dependent predicate
+pushdown. This metadata is table-local: it says a filter may be supplied from a
+runtime join key, not that the table depends on a specific source pair.
+
+V1 capability metadata is deliberately narrow:
```
AlbertQM: Don't think we need to say this out loud 😄 

```suggestion

```

AlbertQM (approved):

crates/coral-spec/src/backends/http.rs
```diff
@@ -80,6 +80,8 @@ pub struct RateLimitSpec {
     pub remaining_header: Option<String>,
     #[serde(default)]
     pub reset_header: Option<String>,
+    #[serde(default)]
+    pub max_concurrency: Option<usize>,
```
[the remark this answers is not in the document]
iaptsiauri: Ignore this, this is moved to config and is not part of the manifest anymore

jsummerfield:
I think #399 should narrow the source-spec surface. The feature we need is not a general public API for dependent joins; it is a small contract that says when an existing filter/input is safe for Coral to bind from another relation as an exact runtime lookup key.

The durable contract should be:

> This equality filter is semantically safe for exact dependent lookup.

Everything else should live somewhere else:

- request wiring belongs to request metadata / generated OpenAPI projection metadata
- serialization belongs to generated operation metadata
- caps, fanout, concurrency, timeout, and fallback behavior belong to engine/runtime policy
- search/retrieval classification should use existing `kind: search` / `mode: search` concepts, not a second `search_index` flag

## Why

Fields like `dependent_join`, `wire_type`, `search_index`, and broad `bindable` metadata mix separate layers into the human-authored manifest:

- provider/API wiring: where the upstream parameter goes: path, query, body
- Coral projection choices: table vs table function vs search surface vs lookup surface
- semantic/planner meaning: exact, complete, equality-like, safe to bind from another relation
- runtime policy: caps, fanout, concurrency, timeout, error/fallback behavior

That layering matters more now because of #934. Once we transpile from OpenAPI, the mechanical parts should be generated from provider evidence where possible. OpenAPI can usually tell us parameter location, wire names, scalar-ish types, request-body fields, and requiredness. It cannot reliably prove that an endpoint is an exhaustive exact lookup, or that using a parameter for a dependent join preserves SQL correctness. That semantic bit needs either conservative inference or a small reviewed overlay.

## Before / After Examples

### 1. Path Parameter Lookup

Before:

```yaml
filters:
  - name: owner
    bindable: true
    wire_type: string
  - name: repo
    bindable: true
    wire_type: string

request:
  path: /repos/{{filter.owner}}/{{filter.repo}}/pulls

dependent_join:
  max_bindings: 500
  max_resolver_rows: 10000
```

After:

```yaml
filters:
  - name: owner
    lookup_key: true
  - name: repo
    lookup_key: true

request:
  path: /repos/{{filter.owner}}/{{filter.repo}}/pulls
```

Meaning: `owner` and `repo` are semantically safe equality lookup keys. The fact that they appear in the path is already request wiring. Runtime caps are engine policy.

### 2. Query Parameter Lookup

Before:

```yaml
filters:
  - name: channel
    bindable: true
    wire_type: string

request:
  path: /api/conversations.history
  query:
    - name: channel
      from: filter
      key: channel
```

After:

```yaml
filters:
  - name: channel
    lookup_key: true

request:
  path: /api/conversations.history
  query:
    - name: channel
      from: filter
      key: channel
```

Meaning: `channel` being sent as a query param is not the important semantic fact. The important fact is that filtering by channel is exact enough for dependent lookup.

### 3. Body/GraphQL Lookup

Before:

```yaml
filters:
  - name: team_id
    bindable: true
    wire_type: string

request:
  method: POST
  path: /graphql
  body:
    - path: [variables, filter, team, id, eq]
      from: filter
      key: team_id
```

After:

```yaml
filters:
  - name: team_id
    lookup_key: true

request:
  method: POST
  path: /graphql
  body:
    - path: [variables, filter, team, id, eq]
      from: filter
      key: team_id
```

This is the example that proves “path template filters only” is too narrow. The lookup key can be in the body, not only the URL path. But the source author still should not have to describe executor behavior.

### 4. Search Surface

Before:

```yaml
tables:
  - name: search_issues
    search_index: true
    filters:
      - name: q
        mode: search
```

After:

```yaml
functions:
  - name: search_issues
    kind: search
    args:
      - name: q
        required: true
        bind:
          arg: q
```

Or, for legacy table filters:

```yaml
filters:
  - name: q
    mode: search
```

Search should not imply lookup eligibility. Search returns ranked/capped/fuzzy candidate sets, not exhaustive equality-filtered rows. So we should not need a second `search_index` field just to prevent dependent lookup; lookup eligibility should be opt-in.

### 5. Runtime Policy

Before:

```yaml
dependent_join:
  max_bindings: 500
  max_resolver_rows: 10000
  max_rows_per_binding: 50000

rate_limit:
  max_concurrency: 8
```

After:

```yaml
filters:
  - name: repo
    lookup_key: true
```

Runtime policy should be owned by the engine:

```text
dependent_join.max_bindings
dependent_join.max_rows_per_binding
dependent_join.timeout
dependent_join.max_concurrency
dependent_join.fallback_behavior
```

Those are not properties of the provider API. They are Coral execution choices and may change by runtime, workspace, source size, plan shape, or operator defaults.

## Minimum Ask For #399

For this PR, I’d make the public spec change much smaller:

1. Remove `dependent_join` caps from the public manifest surface.
2. Do not add `wire_type` to the public source-spec API.
3. Do not add `search_index` as a second search classification.
4. Keep one explicit semantic hook on equality filters/inputs.
5. Prefer naming it `lookup_key`; if we keep `bindable`, document it narrowly as “safe for Coral to bind from another relation as an exact equality lookup.”
6. Document this as a v3/manual-source bridge. In the v4/OpenAPI path, mechanical operation metadata should be generated, with reviewed overlays only for semantic facts OpenAPI cannot prove.

## Short Version

The stable source-spec contract should be exact lookup eligibility: which equality filters are safe for Coral to bind from another relation without changing SQL semantics. Route/body/query placement, wire serialization, and execution limits are not that contract. They should live in generated operation metadata or engine policy, especially with #934 moving us toward OpenAPI-derived manifests. This keeps #399 useful for the planner/executor stack without freezing engine internals into the human-facing source-spec API.

crates/coral-engine/src/contracts/query.rs
```diff
@@ -219,6 +235,124 @@ impl QueryRuntimeConfig {
...
+    /// Maximum key-supplying rows allowed for one join-key combination.
+    pub max_resolver_rows_per_binding: usize,
+    /// Maximum concurrent upstream requests issued by one dependent join.
+    pub max_concurrency: usize,
+}
+
+impl Default for DependentJoinConfig {
+    fn default() -> Self {
+        Self {
+            enabled: true,
```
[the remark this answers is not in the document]
iaptsiauri: This is intentional. Dependent predicate pushdown is meant to be enabled by default for new runtimes

crates/coral-engine/tests/engine/dependent_join_execution_tests.rs
```diff
@@ -0,0 +1,2995 @@
...
+        .and(path_regex(r"^/repos/withcoral/coral/pulls/[1-5]$"))
+        .respond_with({
+            let active = Arc::clone(&active);
+            let max_active = Arc::clone(&max_active);
+            move |request: &Request| {
+                let in_flight = active.fetch_add(1, Ordering::AcqRel) + 1;
+                max_active.fetch_max(in_flight, Ordering::AcqRel);
+
+                let active = Arc::clone(&active);
+                tokio::spawn(async move {
```
sauldhernandez: This test's concurrency oracle is racy. The mock increments `active` when the request handler runs, but the decrement is a separate timer with the same delay as `ResponseTemplate::set_delay(delay)`. If the response delay wakes first, the dependent-join driver can correctly start the next fetch before this timer decrements `active`, so `max_active` can transiently exceed `max_concurrency` even when the implementation respects the limit. Make the oracle deterministic, for example by tying the decrement to actual response completion or by using a semaphore/barrier around request admission rather than a parallel sleep.

crates/coral-spec/src/validate.rs
```diff
@@ -450,6 +462,16 @@ fn validate_detail_hints(
     Ok(())
 }
 
+fn validate_filter_capabilities(filter: &FilterSpec) -> Result<()> {
+    if filter.bindable && filter.mode != FilterMode::Equality {
```
sauldhernandez: This only validates `bindable => mode=equality`; it doesn't reject bindable filters on HTTP tables that declare `search_limits`. The docs added in this PR say capped/ranked/incomplete search-result filters must not be bindable because dependent joins need complete filtered result sets, but `fetch_complete` still caps search targets at `max_top_k * max_calls_per_query`. A source can therefore declare an equality bindable filter on a capped search table and the optimizer can build a dependent join that silently drops matches beyond the cap. Please make that invariant enforceable in spec validation, or otherwise prove the target is complete, instead of relying on author guidance.

jsummerfield (commented):
I left a few focused inline comments from the local join validation and trace review. I skipped the search-limits/bindable point here because that thread already exists on the PR.

crates/coral-engine/src/runtime/dependent_join/optimizer.rs
```diff
@@ -0,0 +1,836 @@
...
+
+        if !dependent_has_column(dependent, dependent_column) {
+            return DependentJoinAnalysis::Fallback(DependentJoinFallbackReason::NonEqui);
+        }
+
+        let Some(filter) = dependent
+            .table
+            .filters()
+            .iter()
+            .find(|filter| filter.name == dependent_column.name)
```
jsummerfield, line 254: Can we avoid making the SQL output column name be the bindable filter name here? This works for the simple fixtures, but it leaks planner constraints into source shape: GitHub has to expose a duplicate `pull_number` column even though the natural API/output field is `number`. An explicit filter-to-column binding in the manifest would let the optimizer bind `filter: pull_number` from `column: number` without forcing source authors to add duplicate columns.

crates/coral-engine/src/runtime/query.rs
```diff
@@ -211,16 +321,71 @@ impl QueryRuntimeAdapter {
...
+                    cap = cap_error.cap,
+                    disposition = "fallback",
+                    "dependent join resolver row cap exceeded",
+                );
+
+                let fallback = fallback_runtime
+                    .get_or_build_without_dependent_join()
+                    .await?;
+
+                self.execute_sql_once(&fallback.ctx, sql)
```
jsummerfield: This retry is not semantically equivalent for required-filter dependent joins. A query can satisfy required filters only through join bindings, but this fallback disables the rewrite and reruns the original SQL as a normal scan; in that case the user sees `MISSING_REQUIRED_FILTER` instead of the resolver-row cap error. I reproduced this with `max_resolver_rows = 1`: the trace showed `DependentJoinExec` started, no dependent fetches ran, and the final `coral.query` error was `MISSING_REQUIRED_FILTER`. Can we preserve the original cap error when the fallback plan cannot satisfy the required filters?

crates/coral-engine/src/backends/http/provider.rs
```diff
@@ -103,32 +137,61 @@ pub(crate) fn http_json_exec(request: HttpJsonExecRequest<'_>) -> Result<Arc<dyn
...
+                if let Some(limit) = post_filter_limit {
+                    filtered_items.truncate(limit);
+                }
+                &filtered_items
+            };
             convert_items(
                 target.columns(),
                 schema.clone(),
-                &filter_values,
+                &request_filter_values,
```
jsummerfield: This conversion only receives request-consumed filters, but local filtering above evaluated rows with `active_filter_values`. Exact local `from_filter`/template columns can therefore be retained during local filtering and then materialize as null/missing in the converted batch, where DataFusion's residual predicate re-checks them. Can we pass the active filter values into conversion, or add a focused test for an exact local filter backed by a `from_filter`/template column?

crates/coral-engine/src/backends/http/provider.rs
```diff
@@ -103,32 +137,61 @@ pub(crate) fn http_json_exec(request: HttpJsonExecRequest<'_>) -> Result<Arc<dyn
...
+                items
+            } else {
+                filtered_items = filter_items_by_column_values(
+                    target.columns(),
+                    &local_filter_values,
+                    &active_filter_values,
+                    &arg_values,
+                    items,
+                );
+                if let Some(limit) = post_filter_limit {
```
jsummerfield, line 192: This truncates after the exact local filters but before DataFusion has applied any residual predicates. In a mixed case with one exact local filter plus another residual predicate, an early row that later fails the residual can consume the limit and drop a later valid row. The new LIKE/no-local-filter coverage does not hit this path; can we either delay this cap until after residuals or add coverage for exact-local + residual + LIMIT?

crates/coral-spec/src/validate.rs
iaptsiauri: added helper `validate_bindable_filters_compatible_with_search_limits` which rejects manifest with both bindable and search_limits

crates/coral-engine/src/runtime/dependent_join/optimizer.rs
jsummerfield: Clarifying this after re-checking the intended v3 contract: if `pull_number` was just a bad manifest projection, I don't think this PR needs new filter-to-column metadata. The narrower fix is to treat bindable filter names as Coral's projected SQL names, rename this GitHub filter to `number`, wire the request through `{{filter.number}}`, and remove the duplicate `pull_number` SQL surface. I'd keep the current optimizer same-name contract for v3 once the manifest/test prove `pr.number = ...` takes the dependent path.

sources/core/github/manifest.yaml
```diff
@@ -89797,6 +89799,7 @@ tables:
         required: false
       - name: pull_number
         required: false
+        bindable: true
```
jsummerfield: Can we rename this bindable filter to `number` and wire the per-PR route as `{{filter.number}}` instead of carrying the provider parameter name `pull_number` into the SQL surface? That keeps the v3 contract simple: bindable filter names are Coral's projected SQL names. Please also add or adjust a regression that joins on `github.pulls.number` and proves the dependent path calls the per-PR endpoint.

jsummerfield (commented):
One follow-up clarification from the bindable-column discussion.

jsummerfield (approved):

PR #402, opened 2026-05-13
feat(mcp): add MCP-backed source support

antonmry:
## Summary

Adds a new `backend: mcp` source type so Coral can expose MCP tools as SQL
tables and source-scoped table functions.

This PR includes:

- `coral-spec` manifest parsing, JSON schema support, and validation for MCP
  sources.
- `coral-engine` runtime support for stdio MCP servers through `rmcp`.
- Tool-backed SQL tables with filter pushdown into MCP tool arguments.
- Source-scoped table functions with named SQL arguments.
- Shared JSON response row extraction for HTTP and MCP backends.
- Structured MCP query errors for missing filters/args, server startup,
  initialization, tool call, tool-returned error, and decode failures.
- A community ClickHouse Cloud MCP source under
  `sources/community/clickhouse_mcp`.

No public source-spec docs are included yet.

## Manifest Shape

MCP sources declare a stdio server command plus tables and/or functions.

```yaml
dsl_version: 3
name: clickhouse_mcp
version: 0.1.0
backend: mcp

server:
  transport: stdio
  command: npx
  args:
    - -y
    - mcp-remote
    - https://mcp.clickhouse.cloud/mcp
```

### Tables

Tables are for stable collection-shaped tool calls.

```yaml
tables:
  - name: services
    tool: get_services_list
    filters:
      - name: organization_id
        required: true
        tool_arg: organizationId
    response:
      rows_path: [result]
    columns:
      - name: organization_id
        type: Utf8
        virtual: true
        expr:
          kind: from_filter
          key: organization_id
      - name: id
        type: Utf8
        nullable: false
      - name: name
        type: Utf8
        nullable: false
```

Query:

```sql
SELECT id, name, state
FROM clickhouse_mcp.services
WHERE organization_id = 'org_123';
```

The table provider extracts supported equality filters from DataFusion,
serializes them into MCP tool arguments, calls the tool, extracts rows with
`ResponseSpec`, and converts JSON rows to Arrow through `JsonExec`.

Tables can also opt into pushing SQL `LIMIT` into a tool argument:

```yaml
limit_binding:
  tool_arg: page_size
  max: 200
```

The engine still applies the row cap after extraction, so `limit_binding`
improves upstream efficiency without changing SQL semantics.

### Table Functions

Functions are for operation-shaped calls where explicit invocation arguments
are clearer than table filters.

```yaml
functions:
  - name: run_select_query
    tool: run_select_query
    args:
      - name: query
        required: true
        bind:
          arg: query
      - name: service_id
        required: true
        bind:
          arg: serviceId
    response:
      rows_path: [result, data]
    columns:
      - name: row
        type: Json
        expr:
          kind: current_row
```

Query:

```sql
SELECT *
FROM clickhouse_mcp.run_select_query(
  query => 'SELECT database, name FROM system.tables LIMIT 10',
  service_id => 'svc_123'
);
```

Function arguments must be named. The engine validates required arguments
before planning the scan.

### MCP-specific manifest names, shared binding semantics

MCP table manifests use MCP-native names like `tool_args`,
`filters[].tool_arg`, and `limit_binding.tool_arg`. They do not expose HTTP
request vocabulary such as query params, body fields, headers, or routes.

Internally, MCP reuses the same lower-level concepts where they are generic:

- source input/value rendering through `resolve_value_source`
- SQL filter extraction through the shared filter helper
- JSON row mapping through `convert_items`
- Arrow execution through `JsonExec`

### Tables and functions both matter

Tables make collection-shaped tools easy to query and join:

```sql
SELECT *
FROM clickhouse_mcp.service_backups
WHERE organization_id = 'org_123' AND service_id = 'svc_123';
```

Functions make operation-shaped tools explicit:

```sql
SELECT *
FROM clickhouse_mcp.get_service_details(
  organization_id => 'org_123',
  service_id => 'svc_123'
);
```

The backend supports both instead of forcing every MCP tool into one SQL shape.

### Avoid backend duplication

HTTP and MCP both produce JSON payloads. This PR moves row extraction into
`backends/shared/response_rows.rs` so `Direct`, `DictEntries`, and
`SeriesPointList` behavior stays consistent across JSON-producing backends.

The intended boundary is:

1. Backend-specific code fetches one JSON payload.
2. Shared `ResponseSpec` logic extracts rows.
3. Shared mapping/execution code produces Arrow batches.

MCP does not reuse HTTP request structs because those carry HTTP-only concepts.

### Static manifest contract

The engine does not introspect MCP tools at query time. The manifest remains
the contract for SQL names, filters, arguments, response shape, and columns.

Tool/schema introspection is left for a future scaffolding workflow.

## Error Handling

MCP errors now flow through structured query errors where possible:

- `MISSING_REQUIRED_FILTER`
- `MISSING_REQUIRED_FUNCTION_ARG`
- `MCP_SERVER_START_FAILED`
- `MCP_INITIALIZE_FAILED`
- `MCP_TOOL_CALL_FAILED`
- `MCP_TOOL_RETURNED_ERROR`
- `MCP_RESULT_DECODE_FAILED`

This gives CLI/MCP consumers stable error codes and metadata instead of only
plain DataFusion error strings.

## ClickHouse MCP Source

Adds `sources/community/clickhouse_mcp` as the first real MCP source example.
It uses `mcp-remote` to bridge Coral's stdio-only MCP runtime to the remote
ClickHouse Cloud MCP endpoint.

The source exposes:

- Cloud management tables such as organizations, services, ClickPipes,
  backups, costs, databases, and tables.
- Functions for service details, private endpoint config, SQL execution, and
  ClickPipe operations.

It lives under `sources/community` rather than `sources/core`, so it is a
repo-distributed source and not bundled into the binary.

## Pending Work

- ~~Richer `LIMIT`/pagination pushdown beyond the initial `limit_binding`
  support~~ 14954692836cf17fcd460661730d632d98149338
- ~~Better typed function argument serialization if MCP tools require non-string
  arguments.~~ c0c06af3df5c2f5d3a7223ea53662198b8188db0
- Add telemetry
- Public docs for `backend: mcp` once the manifest shape is stable enough.
- Real stdio MCP fixture/integration test that starts a child MCP server.
- Support HTTP streaming
- Session- or source-scoped MCP process pooling if process startup becomes
  material for repeated scans.
- MCP resource and resource-template backed tables.
- Optional scaffolding from `tools/list`, `inputSchema`, and `outputSchema`.

## Validation

Recommended validation before merging:

```sh
cargo fmt --all -- --check
cargo clippy --workspace --all-targets --all-features --locked -- -D warnings
cargo test --workspace --all-targets --all-features --locked
RUSTDOCFLAGS="-D warnings" cargo doc --workspace --all-features --no-deps --locked
```

Targeted areas covered by tests in this branch:

- MCP manifest parsing and validation.
- MCP tables and functions in the engine.
- Required table filters and required function args.
- Filter pushdown into MCP tool args.
- Projection and limit behavior.
- `limit_binding` behavior.
- MCP catalog metadata.
- Shared response row extraction.
- Structured MCP error mapping.


antonmry:
@cubic-dev-ai re-run review

jsummerfield (commented):
Source review finding:

- Blocking: this PR does not currently have the `cla signed` label. The repo contribution rules require the CLA before a community source can be approved. Please complete the CLA flow and get the label applied, then this can receive a source-spec content review.

jsummerfield (commented):
Findings:

- High: this branch currently deletes 21 unrelated files under `sources/community` when compared with current `main`. A community-source PR should add or update its own source directory without removing existing sources. Please merge/rebase current `main` and restore these unrelated files before this can be approved. Examples from the diff:

- `sources/community/cal/README.md`
- `sources/community/cal/manifest.yaml`
- `sources/community/clickhouse/README.md`
- `sources/community/clickhouse/manifest.yaml`
- `sources/community/cloudflare/README.md`
- `sources/community/cloudflare/manifest.yaml`

jsummerfield (commented):
Findings:

- High: `sources/community/clickhouse_mcp/manifest.yaml:3` sets `backend: mcp`, but the current source manifest schema only accepts `http`, `parquet`, or `jsonl`; the schema validator also rejects the top-level `server` field and each function `tool` field. This source will not parse as a valid Coral source manifest today. Please either move this behind the PR that adds MCP-backed source-spec support or update it to the source-spec shape that is actually implemented in this repository.

Review note: this PR still also needs the `cla signed` label before it can be approved.

crates/coral-engine/src/backends/mcp/fetch.rs
```diff
@@ -0,0 +1,99 @@
...
+                        tool: self.tool_name.clone(),
+                        detail: format!("exceeded pagination max_pages={max_pages}"),
+                    },
+                )));
+            }
+
+            let arguments = self.arguments_for_cursor(next_cursor.as_deref());
+            let payload = self
+                .backend
+                .call_tool(&self.relation, &self.tool_name, arguments)
```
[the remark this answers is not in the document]
antonmry: Real, and acknowledged — but it's the same process-per-query limitation Coral has had since the MVP, not pagination-specific. Every scan (paginated or not) spawns a fresh stdio MCP server; pagination just makes the per-page cost more visible. Tracked under "Pending Work" in the PR description ("Session- or source-scoped MCP process pooling if process startup becomes material for repeated scans") and in PLAN_mcp.md "Open Decisions". No community source in-tree currently exposes a paginated MCP tool, so this isn't a regression on the existing surface. Deferring to the pooling work.

crates/coral-engine/src/backends/mod.rs
```diff
@@ -1,4 +1,71 @@
...
+//! - `client.rs` = configured stateful wrapper + transport interface.
+//! - `transport.rs` = the actual round-trip / process spawn / store handle.
+//! - `response.rs` = decode one response.
+//! - `fetch.rs` = multi-request orchestration / pagination loop.
+//!
+//! If a concern doesn't apply, omit the file — do not repurpose the name.
+//!
+//! ## Reference backends
+//!
+//! - [`http`] — fullest surface: separate `client`, `transport`, `response`,
```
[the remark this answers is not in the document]
antonmry: that's solved in a different PR

docs/reference/community-sources.mdx
```diff
@@ -50,6 +50,7 @@ After importing, the source behaves like any other installed source. It appears
 | [clear_ml](https://github.com/withcoral/coral/tree/main/sources/community/clear_ml) | Query ClearML projects, tasks, models, pipelines, workers, queues, artifact metadata, and current user context. |
 | [clerk](https://github.com/withcoral/coral/tree/main/sources/community/clerk) | Query users, organizations, memberships, invitations, and sessions from Clerk. |
 | [clickhouse](https://github.com/withcoral/coral/tree/main/sources/community/clickhouse) | Query ClickHouse databases, tables, columns, query history, running processes, server metrics, storage parts, and replication health via the ClickHouse HTTP interface. |
+| [clickhouse_mcp](https://github.com/withcoral/coral/tree/main/sources/community/clickhouse_mcp) | ClickHouse Cloud remote MCP server (mcp.clickhouse.cloud). |
```
simonwhitaker, line 53: This is the only change I see under docs. It looks like we're missing a few other places, e.g. https://withcoral.com/docs/reference/source-spec-reference.

simonwhitaker (approved):

docs/reference/community-sources.mdx
AlbertQM: Hmm yeah the docs should be auto-generated, and CI should fail if e.g. bits are missing or are out of sync -- will have a look at what's wrong
EDIT: For community sources, this is all we auto-generate :smile:

Agreed that a whole new backend needs a bit more details in the docs

antonmry: it's mentioned in the PR description. I would prefer to add docs in a separate PR once this was tested by the team and the spec is stable. `docs/reference/community-sources.mdx` needs to be updated to avoid the CI Docs freshness check, though

crates/coral-engine/src/backends/mcp/provider.rs
```diff
@@ -0,0 +1,247 @@
...
+                            table: self.table.name().to_string(),
+                            column: filter.name.clone(),
+                        },
+                    )));
+                }
+                None => {}
+            }
+        }
+
+        let effective_limit = effective_limit(
```
jsummerfield: I think `limit_binding.max` is getting mixed into the final row limit here. That makes `max` behave like “never return more than N rows”, not “never ask the MCP server for more than N rows per call”. For a paginated table, `limit_binding.max = 100` would make row 101 unreachable even when the server returns a next cursor. Could we keep two separate values: one pushed into the MCP tool arg/page size, and one final result limit from SQL `LIMIT` / `fetch_limit_default`?

sources/community/clickhouse_mcp/manifest.yaml
```diff
@@ -0,0 +1,561 @@
...
+      - name: query
+        required: true
+        bind:
+          arg: query
+      - name: service_id
+        required: true
+        bind:
+          arg: serviceId
+    response:
+      rows_path: [result, data]
```
jsummerfield, line 568: I tested `run_select_query` with invalid ClickHouse SQL. The CLI exited successfully and returned zero rows because this `rows_path` skips the tool's error branch. Telemetry also recorded `status=ok` with `row_count=0`, so an upstream query failure is indistinguishable from a legitimate empty result. Could we either route the error branch into an engine error or expose enough raw error state for this source to fail visibly?

crates/coral-engine/src/backends/mcp/provider.rs
```diff
@@ -0,0 +1,247 @@
...
+                            .tool_arg_for_filter(&filter.name)
+                            .ok_or_else(|| {
+                                DataFusionError::Plan(format!(
+                                    "{}.{} filter '{}' is missing its MCP tool binding",
+                                    self.source_schema,
+                                    self.table.name(),
+                                    filter.name
+                                ))
+                            })?;
+                    arguments.insert(tool_arg.to_string(), Value::String(value.clone()));
```
jsummerfield: This currently serializes every pushed table filter as a JSON string. The manifest supports typed MCP filters (`Int64`, `Boolean`, `Float64`, etc.), so a tool that expects JSON `10` or `true` will instead receive `"10"` or `"true"`. Function args already preserve scalar JSON types; could table filters use the declared filter type to preserve JSON scalar values too?

sources/community/clickhouse_mcp/README.md
```diff
@@ -0,0 +1,259 @@
...
+
+If the refresh token ever expires, re-run the command above to re-auth.
+
+### Register the source
+
+```bash
+coral source add --file sources/community/clickhouse_mcp/manifest.yaml
+```
+
+This spawns the MCP child process to confirm connectivity, then prints the
```
jsummerfield: This overstates what `source add` verifies. I tested this path with an isolated Coral config: `source add`/`source test` registered and reported the source, but the MCP child process and OAuth/connectivity path did not run until the first real `coral sql` query. Since this manifest has no `test_queries`, could we reword this to say the source is registered/cataloged here and connectivity is verified by the follow-up query?

jsummerfield (approved):

sources/community/clickhouse_mcp/manifest.yaml
antonmry: Done in bd700ba0a2d16e30c1bde5dbf901f2a5fd592813

crates/coral-engine/src/backends/mcp/provider.rs
antonmry: Done with 44d1c457765917cc43fe63d6aabbaf41ba8aae9f

crates/coral-engine/src/backends/mcp/provider.rs
antonmry: Done with e7841306870c1630ab2dd862e3484785de4d4379

sources/community/clickhouse_mcp/README.md
antonmry: solved with 233dda11e6dea98370df013b0a524e56d1f88f03

crates/coral-engine/src/backends/mcp/fetch.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+    let value = get_path_value(payload, &response.error_path)?;
+    if value.is_null() {
+        return None;
+    }
+    Some(match value {
+        Value::String(text) => text.clone(),
+        other => other.to_string(),
+    })
+}
```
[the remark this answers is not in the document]
antonmry: @cubic-dev-ai recheck

PR #440, opened 2026-05-14
feat(sources/community/postman): add Postman API source

### LINKED ISSUE #439: design: add Postman community source
## Summary

Add a new `postman` community source that lets Coral query Postman workspaces, API collections, request metadata, environments, and monitors through the Postman API. This makes it easy for users and agents to discover accessible Postman workspaces, inspect API collections and request structure, manage test environments, and validate monitor health for API inventory analysis and governance workflows.

## Motivation

Postman is widely used for API development, testing, and documentation. Teams often need a quick read-only inventory of Postman collections, request endpoints, test environments, and API monitors across workspaces, especially for API discovery, endpoint analytics, governance checks, and agent workflows.

Without a community source, users must hand-roll Postman API calls and pagination logic for common metadata checks like endpoint discovery, request method distribution, and monitor health status.

## Proposed design

Add a new community source under `sources/community/postman`, implemented as a manifest-only HTTP source.

Inputs:

- `POSTMAN_API_KEY` required secret
  - Postman API key from account settings
  - recommended scopes: workspaces.read, collections.read, environments.read, monitors.read

Auth:

- header auth
- `X-API-Key: {{input.POSTMAN_API_KEY}}`

Proposed v1 tables:

- `postman.me`
  - `GET /me`
  - validates API key and returns current user metadata
- `postman.workspaces`
  - `GET /workspaces`
  - lists Postman workspaces accessible to the caller
- `postman.collections`
  - `GET /collections`
  - lists API collections with request counts and organization metadata
- `postman.collection_requests`
  - `GET /collections/{collection_uid}`
  - requires `collection_uid`
  - exposes request-level endpoint metadata: method, URL, description, auth, scripts
- `postman.environments`
  - `GET /environments`
  - lists test/deployment environments and variable groups
- `postman.monitors`
  - `GET /monitors`
  - lists API health monitors with failure counts and schedule config

Design details:

- keep v1 read-only
- use Postman API v1 only
- document API key setup and scopes in README
- include offset pagination for list endpoints
- expose high-value flattened fields (owner_id, workspace_id) plus raw JSON for complex structures (auth, headers, scripts, variables)
- require `collection_uid` filter on collection_requests to prevent unbounded scans

Affected surface:

- community source only
- expected files:
  - `sources/community/postman/manifest.yaml`
  - `sources/community/postman/README.md`

No CLI, config, MCP, runtime, or bundled source changes are proposed.

## Drawbacks and trade-offs

This source requires a Postman API key rather than OAuth. Users must create an API key through their Postman account settings.

Mock servers, flows, specifications, and pull requests are intentionally out of scope for v1 because they add complexity without immediate value for core API discovery and monitoring workflows.

## Alternatives considered

- Including mock server metadata: rejected for v1 because mock servers add operational complexity without immediate query value.
- Including API governance features (APIs, comments, forks): rejected for v1 because these surfaces are less stable and add breadth without depth.
- Querying Postman Collection Format v2.1 directly: rejected because Postman's REST API provides structured, normalized metadata that's easier to query.

## Open questions

- Should mock servers become part of a future v1.1 release?
- Should workspace team/permission metadata be exposed in future versions?
- Should enterprise Postman API features be queried when available?

siddhiguptas:
## Summary

Fixes #439 

Add a new `postman` community source for querying Postman workspace inventory, API collections, request metadata, test environments, and API monitors using the Postman Management API.

This change is manifest-only and uses Coral's existing HTTP source-spec model. It adds read-only Postman metadata querying without changing CLI, MCP, config, runtime, or bundled-source contracts.

## What changed

Added:

- `sources/community/postman/manifest.yaml`
- `sources/community/postman/README.md`

## Source scope

Inputs:

- `POSTMAN_API_KEY`

Tables:

- `postman.me`
  - `GET /me`
  - validates API key and returns user metadata
- `postman.workspaces`
  - `GET /workspaces`
  - lists accessible workspaces with pagination
- `postman.collections`
  - `GET /collections`
  - lists collections with owner, workspace, and metadata
- `postman.collection_requests`
  - `GET /collections/{collection_uid}`
  - requires `collection_uid` (required filter)
  - exposes request-level endpoint data: method, URL, auth, scripts, headers
- `postman.environments`
  - `GET /environments`
  - lists environments with variables and owner metadata
- `postman.monitors`
  - `GET /monitors`
  - lists API health monitors with status and failure counts

## Implementation notes

- uses Postman API key header auth (`X-API-Key`)
- uses offset pagination (limit/offset) for list endpoints
- supports 25-item default page size on paginated endpoints
- requires `collection_uid` filter on collection_requests to prevent unbounded scans
- exposes high-value flattened fields (owner_id, workspace_id) plus raw JSON for complex structures (auth, headers, scripts, variables)
- treats missing data (e.g., unlinked monitors) as empty results
- README documents API key setup, scopes, and example queries

## Validation

Live API testing completed:

- All 6 endpoints tested with real Postman API credentials
- All response structures verified against actual API responses
- Field mappings validated with real data (9 workspaces, 9 collections, 4 environments tested)
- Pagination validated on multi-page endpoints
- Critical field mapping issues identified and fixed during testing

Local validation:

- Manifest YAML structure verified
- All Arrow types correctly declared
- All columns have nullable declarations
- All columns have expr path mappings
- Response extraction paths all validated

## Testing notes

Testing was performed with actual Postman API using a real access token. All 6 endpoints returned 200 status and real data structures. Field mappings were corrected based on actual API response structure discovered during testing:

- `/me` returns user object directly: `{user: {...}}`
- `/collections` returns owner as numeric ID, not nested object
- `/environments` returns values array for variables
- `/collection_requests` handles URL as both string and object

All issues discovered during testing were fixed before this PR.

## Out of scope

Not included in v1:

- Mock servers
- Flows and specifications
- Forks and pull requests
- Firestore/Realtime Database querying
- write operations such as creating collections or modifying monitors

### DIFF
--- a/sources/community/postman/manifest.yaml
+++ b/sources/community/postman/manifest.yaml
@@ -447,7 +447,7 @@ tables:
       - name: variables
         type: Json
         nullable: true
-        description: Environment variables (JSON structure). Secret values are masked in v1.
+        description: "Environment variables (JSON structure). Note: the Postman API returns secret values as plaintext."
         expr:
           kind: path
           path: [values]
--- a/sources/community/postman/manifest.yaml
+++ b/sources/community/postman/manifest.yaml
@@ -38,7 +38,7 @@ tables:
       method: GET
       path: /me
     response:
-      rows_path: []
+      rows_path: [user]
     columns:
       - name: user_id
         type: Int64
@@ -203,34 +203,27 @@ tables:
         expr:
           kind: path
           path: [uid]
-      - name: name
-        type: Utf8
-        nullable: false
-        description: Collection name
-        expr:
-          kind: path
-          path: [name]
-      - name: owner_id
-        type: Int64
-        nullable: true
-        description: Owner user ID (flattened from owner object)
-        expr:
-          kind: path
-          path: [owner, id]
-      - name: owner_name
-        type: Utf8
-        nullable: true
-        description: Owner username (flattened from owner object)
-        expr:
-          kind: path
-          path: [owner, username]
-      - name: workspace_id
-        type: Utf8
-        nullable: true
-        description: Parent workspace ID (flattened from workspace)
-        expr:
-          kind: path
-          path: [workspace, id]
+       - name: name
+         type: Utf8
+         nullable: false
+         description: Collection name
+         expr:
+           kind: path
+           path: [name]
+       - name: owner_id
+         type: Int64
+         nullable: true
+         description: Owner user ID
+         expr:
+           kind: path
+           path: [owner]
+       - name: workspace_id
+         type: Utf8
+         nullable: true
+         description: Parent workspace ID
+         expr:
+           kind: path
+           path: [workspace]
       - name: created_at
         type: Timestamp
         nullable: true
@@ -332,13 +325,13 @@ tables:
         expr:
           kind: path
           path: [request, method]
-      - name: url
-        type: Utf8
-        nullable: true
-        description: Request URL (flattened from request object)
-        expr:
-          kind: path
-          path: [request, url, raw]
+       - name: url
+         type: Utf8
+         nullable: true
+         description: Request URL string or raw URL
+         expr:
+           kind: path
+           path: [request, url]
       - name: description
         type: Utf8
         nullable: true
@@ -431,55 +424,48 @@ tables:
         expr:
           kind: path
           path: [uid]
-      - name: name
-        type: Utf8
-        nullable: false
-        description: Environment name
-        expr:
-          kind: path
-          path: [name]
-      - name: owner_id
-        type: Int64
-        nullable: true
-        description: Owner user ID (flattened from owner object)
-        expr:
-          kind: path
-          path: [owner, id]
-      - name: owner_name
-        type: Utf8
-        nullable: true
-        description: Owner username (flattened from owner object)
-        expr:
-          kind: path
-          path: [owner, username]
-      - name: workspace_id
-        type: Utf8
-        nullable: true
-        description: Parent workspace ID (flattened from workspace)
-        expr:
-          kind: path
-          path: [workspace, id]
-      - name: created_at
-        type: Timestamp
-        nullable: true
-        description: Environment creation timestamp (ISO 8601)
-        expr:
-          kind: path
-          path: [createdAt]
-      - name: updated_at
-        type: Timestamp
-        nullable: true
-        description: Environment last updated timestamp (ISO 8601)
-        expr:
-          kind: path
-          path: [updatedAt]
-      - name: variables
-        type: Struct
-        nullable: true
-        description: Environment variables (JSON structure). Secret values are masked in v1.
-        expr:
-          kind: path
-          path: [value]
+       - name: name
+         type: Utf8
+         nullable: false
+         description: Environment name
+         expr:
+           kind: path
+           path: [name]
+       - name: owner_id
+         type: Int64
+         nullable: true
+         description: Owner user ID
+         expr:
+           kind: path
+           path: [owner]
+       - name: workspace_id
+         type: Utf8
+         nullable: true
+         description: Parent workspace ID
+         expr:
+           kind: path
+           path: [workspace]
+       - name: created_at
+         type: Timestamp
+         nullable: true
+         description: Environment creation timestamp (ISO 8601)
+         expr:
+           kind: path
+           path: [createdAt]
+       - name: updated_at
+         type: Timestamp
+         nullable: true
+         description: Environment last updated timestamp (ISO 8601)
+         expr:
+           kind: path
+           path: [updatedAt]
+       - name: variables
+         type: Struct
+         nullable: true
+         description: Environment variables (JSON structure). Secret values are masked in v1.
+         expr:
+           kind: path
+           path: [values]
 
   - name: monitors
     description: API monitors configured for continuous testing and health checking. Enables API reliability analytics and incident summarization. Important for API governance workflows.
--- a/sources/community/postman/manifest.yaml
+++ b/sources/community/postman/manifest.yaml
@@ -102,12 +102,13 @@ tables:
     response:
       rows_path: [workspaces]
     pagination:
-      mode: offset
-      offset_param: offset
+      mode: cursor_query
       page_size:
         default: 25
         max: 100
         query_param: limit
+      cursor_param: cursor
+      response_cursor_path: [meta, nextCursor]
     columns:
       - name: id
         type: Utf8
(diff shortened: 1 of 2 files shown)


sources/community/postman/manifest.yaml
```diff
@@ -0,0 +1,567 @@
...
+          path: [isPublic]
+
+  - name: workspaces
+    description: Postman workspaces accessible to the authenticated user. Workspaces are top-level organizational entities containing collections, environments, and monitors.
+    fetch_limit_default: 25
+    request:
+      method: GET
+      path: /workspaces
+      query:
+        - name: limit
```
antonmry: Thanks for the detailed source and README. Could you run `coral source lint` against the manifest and check the DSL fields here?

On the current head I get schema validation failures because `from: limit` / `from: offset` are not supported request value sources; Coral adds pagination params from `pagination.page_size.query_param` and `pagination.offset_param` automatically. The same lint run also flags the later `Struct` column types, and this table uses `path_params` / `kind: context`, which do not appear to be supported by the current DSL. As written, the source will not install yet.

sources/community/postman/manifest.yaml
```diff
@@ -0,0 +1,567 @@
...
+          kind: path
+          path: [visibility]
+      - name: created_by
+        type: Int64
+        nullable: true
+        description: User ID who created the workspace
+        expr:
+          kind: path
+          path: [createdBy]
+      - name: created_at
```
antonmry, line 148: Could you double-check the timestamp mappings before this lands?

Postman returns these as ISO 8601 strings, but Coral `Timestamp` columns need the value converted with `kind: format_timestamp` and `input: iso8601`; using a plain `path` expression here is likely to produce NULL timestamps at runtime. This applies to the other `created_at` / `updated_at` columns too.

sources/community/postman/manifest.yaml
```diff
@@ -0,0 +1,567 @@
...
+      path: /workspaces
+      query:
+        - name: limit
+          from: limit
+        - name: offset
+          from: offset
+    response:
+      rows_path: [workspaces]
+    pagination:
+      mode: offset
```
antonmry: Could you verify the pagination mode for `/workspaces` against the current Postman API docs?

The docs I found describe `limit` plus `cursor` / `nextCursor` for this endpoint, not offset pagination. If that is still accurate, using `mode: offset` here could repeat or miss rows beyond the first page. It may need `cursor_query` with `cursor_param: cursor` and the response cursor path, or no pagination if the endpoint shape does not expose one in the tested account.

Docs: https://www.postman.com/postman/postman-public-workspace/request/mes2rzs/get-all-workspaces

sources/community/postman/manifest.yaml
```diff
@@ -0,0 +1,567 @@
...
+          kind: path
+          path: [createdAt]
+      - name: updated_at
+        type: Timestamp
+        nullable: true
+        description: Environment last updated timestamp (ISO 8601)
+        expr:
+          kind: path
+          path: [updatedAt]
+      - name: variables
```
antonmry: Could you check whether this field is present on the list endpoint response?

Postman's documented `GET /environments` response includes environment metadata (`id`, `name`, `createdAt`, `updatedAt`, `owner`, `uid`, etc.) but not the `values` array; variables appear to require fetching an individual environment. If that matches the live API, this column will always be NULL in this table and should either be removed, documented as not available from the list endpoint, or backed by a different request shape.

Docs: https://www.postman.com/postman/postman-public-workspace/request/bqbz8iw/get-all-environments

sources/community/postman/manifest.yaml
```diff
@@ -0,0 +1,567 @@
...
+        required: true
+    request:
+      method: GET
+      path: /collections/{collection_uid}
+      path_params:
+        collection_uid:
+          from: filter
+          key: collection_uid
+    response:
+      rows_path: [item]
```
antonmry: Could you confirm the intended behavior for folders and nested requests?

`rows_path: [item]` only selects the top-level collection items. In Postman collections, folders can contain their own `item` arrays, so this table may omit nested requests and may also emit folder rows with NULL request fields. If Coral cannot recursively flatten this response in the manifest today, it would be helpful to either document that v1 only exposes top-level requests or expose the raw collection/folder JSON instead of presenting this as request-level coverage.

Docs: https://learning.postman.com/docs/collections/use-collections/manage-collections

sources/community/postman/manifest.yaml
```diff
@@ -0,0 +1,556 @@
...
+        type: Timestamp
+        nullable: true
+        description: Environment last updated timestamp (ISO 8601)
+        expr:
+          kind: format_timestamp
+          input: iso8601
+          expr:
+            kind: path
+            path: [updatedAt]
+      - name: variables
```
antonmry: This contradicts the README. The column maps to `[values]` here, but the README "environments" section says: "The list endpoint does not return environment variable values. Variable details require fetching individual environments, which is not supported in v1."

If the README note is correct, this column will always be NULL and should be dropped for v1 (or backed by a per-environment fetch). If your live testing actually got `values` back from `/environments`, the README note is wrong instead. One of the two needs to change so the manifest and docs agree.

sources/community/postman/manifest.yaml
```diff
@@ -0,0 +1,556 @@
...
+      top-level organizational entities containing collections, environments,
+      and monitors.
+    fetch_limit_default: 25
+    request:
+      method: GET
+      path: /workspaces
+    response:
+      rows_path: [workspaces]
+    pagination:
+      mode: offset
```
antonmry: The README "Limits" section says "List endpoints return all items without server-side pagination," but this table (and `collections`, `environments`, `monitors`) is still configured with `mode: offset` + `limit`/`offset` query params.

If the Postman API ignores `?offset=` on these endpoints, the runtime will keep re-requesting the same first page and emit duplicate rows in a loop until the page comes back short. Could you confirm — on a workspace with more than 25 items — that incrementing `offset` actually returns a different slice? If it doesn't, switch these tables to `mode: none` (matching the README claim). If it does, update the README so it doesn't contradict the manifest.

sources/community/postman/manifest.yaml
```diff
@@ -0,0 +1,556 @@
...
+        type: Utf8
+        nullable: true
+        description: Request URL string or raw URL
+        expr:
+          kind: path
+          path: [request, url, raw]
+      - name: description
+        type: Utf8
+        nullable: true
+        description: Request description
```
antonmry, line 322: Minor: Postman request `description` can come back as either a string or an object of the form `{content, type}` (e.g. when the description is Markdown). With a plain `kind: path` and `Utf8` column type, the object form will surface as NULL or a stringified blob.

Either extract `request.description.content` (with the string form as a fallback), or document that rows with structured descriptions render as NULL.

antonmry (approved):

PR #453, opened 2026-05-14
feat(file)!: add unified file backend

jsummerfield:
## Intent
Unify Coral's file-source model around DataFusion's native file abstractions so file format and object transport are independent, while keeping Coral-specific code limited to behavior DataFusion does not provide.

## Summary
- Replace the legacy `parquet` and `jsonl` source backends with unified `backend: file` support.
- Add table-level `format`/`source` handling for Parquet, JSONL, JSON, and CSV.
- Route native file formats through DataFusion listing-table primitives, with custom JSON handling only where Coral needs JSON text normalization or segment partitions.
- Add strict path partition extraction/pruning and typed S3 object-store policy support.

## Validation
- `make rust-checks`


jsummerfield:
* **#458** <a href="https://app.graphite.com/github/pr/withcoral/coral/458?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#457** <a href="https://app.graphite.com/github/pr/withcoral/coral/457?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#456** <a href="https://app.graphite.com/github/pr/withcoral/coral/456?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#455** <a href="https://app.graphite.com/github/pr/withcoral/coral/455?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#454** <a href="https://app.graphite.com/github/pr/withcoral/coral/454?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#453** <a href="https://app.graphite.com/github/pr/withcoral/coral/453?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/453?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

AlbertQM (approved):
Docs LGTM

crates/coral-engine/src/backends/file/mod.rs
```diff
@@ -0,0 +1,159 @@
+//! Native file table provider backed by local files or object-store URLs.
+
+mod json;
+mod listing;
+mod parquet_schema;
+mod partitions;
+mod provider;
```
antonmry, line 8: Consider adding a `mod error;` placeholder here, alongside the others. HTTP (#651) and MCP (#402) both ship an `error.rs` with structured error variants (`MISSING_REQUIRED_FILTER`, `MCP_TOOL_RETURNED_ERROR`, etc.) so the CLI / MCP consumer layer can distinguish failure modes by stable code rather than parsing strings. This backend surfaces `DataFusionError` directly.

Fine for V1 — file-IO errors are mostly self-explanatory — but worth putting the file in place now, even if it starts with just one or two variants (file-not-found, schema-mismatch, unsupported-format). Benefits:

- Matches the module layout of the other two backends, so newcomers don't have to wonder why one is missing.
- Gives an obvious home for the inevitable case where a consumer wants to discriminate "file not found" from "permission denied" or "schema mismatch."
- Costs almost nothing today; avoids a small refactor later.

crates/coral-engine/src/backends/file/listing.rs
```diff
@@ -0,0 +1,328 @@
+//! Shared file listing, URL, format, and object-store setup helpers.
```
antonmry: Nit: worth expanding this header to call out *why* there's no `transport.rs` / `client.rs` / `response.rs` / `fetch.rs` in this backend. Readers coming from `http/` or `mcp/` (both of which have those files) will look for them and be briefly confused.

Suggested wording:

```rust
//! File listing, URL, format, and object-store setup helpers.
//!
//! The file backend deliberately omits the `client` / `transport` /
//! `response` / `fetch` files used by the HTTP and MCP backends: DataFusion's
//! `ListingTable` + `ObjectStore` registry own per-request transport,
//! response decoding, and parallel-scan orchestration internally. This
//! module is the setup-and-hand-off layer — it builds an `ObjectStore`
//! handle and a `ListingTableConfig`, then hands them to DataFusion.
```

(For the bigger backend-layout convention itself, there's a candidate write-up in #402 that lands a `//!` block on `backends/mod.rs` describing the standard layout, so this header would just point at "this backend is the documented exception" rather than re-explain the whole thing.)

antonmry (approved):

mystic123 (approved):

crates/coral-engine/src/backends/file/listing.rs
```diff
@@ -0,0 +1,486 @@
...
+}
+
+fn build_object_store(
+    source_schema: &str,
+    table_path: &ListingTableUrl,
+    source: &FileSourceSpec,
+    resolved_inputs: &BTreeMap<String, String>,
+) -> Result<Arc<dyn ObjectStore>> {
+    match table_path.scheme() {
+        "file" => Ok(Arc::new(LocalFileSystem::new())),
```
sauldhernandez, line 260: This scan now follow symlinked directories, which is a contract change from the previous JSONL walker that skipped symlinks. Is that change intentional here?

crates/coral-engine/src/backends/file/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    fn source_name(&self) -> &str {
+        &self.manifest.common.name
+    }
+
+    async fn register(&self, ctx: &SessionContext) -> Result<BackendRegistration> {
+        let mut tables: HashMap<String, Arc<dyn TableProvider>> = HashMap::new();
+        let mut table_infos = Vec::with_capacity(self.manifest.tables.len());
+        let resolved_inputs = coral_spec::resolve_inputs(
```
sauldhernandez, line 80: nothing prevents a spec from defining the s3 credentials as literals or non-secret inputs. Not a deal-breaker, but maybe consider steering the users into having to use secrets for this information.

sauldhernandez (approved):

crates/coral-engine/src/backends/file/listing.rs
jsummerfield: Intentional. This backend now delegates local listing to DataFusion/object_store rather than Coral’s old JSONL-specific walker, and `object_store::local::LocalFileSystem` follows symlinks by design, including symlinked directories.

Given the direction of this PR is to use DataFusion’s file primitives as the contract, I’d rather keep that behaviour than reintroduce a custom walker just to preserve the old JSONL edge case. We should document this explicitly as part of the file source contract.

crates/coral-engine/src/backends/file/mod.rs
jsummerfield: Good point. I agree. I've seen this crop up in community sources too by accident.

I’m going to handle that in a follow-up PR in `coral-spec`, so the rule is enforced at manifest validation time rather than inside the engine path.

jsummerfield:
### Merge activity

* **May 26, 7:48 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/453).
* **May 26, 7:49 AM UTC**: @jsummerfield merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/453).

PR #493, opened 2026-05-15
feat(sources/community/buildkite): add Buildkite source spec

sahil9001:
## Summary
- add a community Buildkite HTTP source spec for current user, organizations, pipelines, organization builds, pipeline builds, and agents
- document setup, table columns, example SQL, validation, and v1 limitations
- checked existing repo sources and open PRs before implementation; Buildkite was not present and no open Buildkite source PR was found

## Validation
- Ruby YAML parse: passed
- coral source lint sources/community/buildkite/manifest.yaml: passed
- git diff --check: passed
- make rust-checks: not run because cargo is not installed in this environment

## Contributor behavior
- No agent/contributor operating guidance changed.


sahil9001:
@antonmry Can you please help with the review? Thanks!

sources/community/buildkite/manifest.yaml
```diff
@@ -0,0 +1,589 @@
...
+      - name: org_slug
+        required: true
+      - name: pipeline_slug
+        required: true
+      - name: state
+      - name: branch
+      - name: commit
+      - name: creator
+    request:
+      method: GET
```
antonmry, line 401: This table is missing several columns that `organization_builds` (L229) exposes for the same Build resource. The two endpoints return identical payload shapes, so I'd expect the column sets to match. Notable absences here:

- `tag` (git tag)
- `source` (build source — `webhook`, `api`, `ui`, etc.)
- `creator__id` (only `creator__email` is exposed; users joining to `current_user.id` will need both)
- `scheduled_at` — particularly useful for "queue time" analysis (`started_at - scheduled_at`)
- `url` (API URL for the build)
- `graphql_id`

The asymmetry is the kind that bites users who write a query against `organization_builds` and copy-paste it to `pipeline_builds`. Easy fix — just lift the same column blocks across.

sources/community/buildkite/manifest.yaml
```diff
@@ -0,0 +1,589 @@
...
+          input: iso8601
+          expr: {kind: path, path: [created_at]}
+      - name: archived_at
+        type: Timestamp
+        nullable: true
+        description: Pipeline archive time.
+        expr:
+          kind: format_timestamp
+          input: iso8601
+          expr: {kind: path, path: [archived_at]}
```
antonmry, line 211: Tiny suggestion — Buildkite's pipeline payload returns both `archived_at` (timestamp) and `archived` (boolean) directly. The manifest only exposes `archived_at`. `WHERE archived = false` is the natural filter to "show me active pipelines"; doing `WHERE archived_at IS NULL` works too but is less obvious.

Adding alongside the existing column:

```yaml
- name: archived
  type: Boolean
  nullable: true
  description: Whether the pipeline is archived.
  expr: {kind: path, path: [archived]}
```

Optional — feel free to skip if you'd rather keep the column count tight.

sources/community/buildkite/manifest.yaml
```diff
@@ -0,0 +1,589 @@
...
+      template: Bearer {{input.BUILDKITE_API_TOKEN}}
+
+request_headers:
+  - name: Accept
+    from: literal
+    value: application/json
+
+test_queries:
+  - SELECT id, email, name FROM buildkite.current_user LIMIT 1
+
```
antonmry, line 32: Suggestion — please ignore if you'd rather not bloat the test list. Adding one query against a paginated table would exercise the `link_header` pagination path on small data:

```yaml
test_queries:
  - SELECT id, email, name FROM buildkite.current_user LIMIT 1
  - SELECT name, slug FROM buildkite.organizations LIMIT 5
```

`current_user` doesn't paginate (single-row response), so the `link_header` config on the other 5 tables isn't actually exercised by `coral source test` today. The `organizations` query above is already used as an example in the README, so it's a natural fit.

sources/community/buildkite/manifest.yaml
```diff
@@ -0,0 +1,589 @@
+name: buildkite
```
antonmry, line 1: Could you confirm whether this manifest has been run end-to-end against a live Buildkite token? The PR description says `coral source lint` passed but explicitly notes `make rust-checks: not run because cargo is not installed in this environment`, and there's no live query output in the PR.

The other recent community-source PRs (#525 shopify, #549 trino, #533 sendgrid) all include actual query output from a real workspace. Buildkite's REST API is well-documented and the manifest looks correct on paper, but Link-header pagination has some quirks (relative vs absolute URLs, `rel="next"` ordering, `per_page` enforcement) that are best caught with a single live run.

The README's "Validation" section lists the right commands — just confirming they were actually executed before this lands. No harm if they were, just want to be sure.

sources/community/buildkite/manifest.yaml
```diff
@@ -0,0 +1,589 @@
...
+        - name: creator
+          from: filter
+          key: creator
+    response:
+      row_strategy: direct
+    pagination:
+      mode: link_header
+      page_size:
+        default: 30
+        max: 100
```
antonmry, line 266: Optional follow-on (already mentioned in the README "Limitations" section, just calling out the priority): exposing `jobs` as a separate `buildkite.jobs` table would be the single highest-value v2 addition. Common ops queries that need it:

- Job-level success rate per pipeline (`SELECT pipeline__slug, state, COUNT(*) FROM jobs GROUP BY ...`)
- Retry analysis (jobs with `retry_count > 0` or `retried_in_job_id IS NOT NULL`)
- Agent placement / queue analysis (which agent ran which jobs)
- Step-level timing breakdowns

Today these have to be done with JSON unnesting on `organization_builds.jobs` / `pipeline_builds.jobs`, which is workable but awkward. Not a v1 blocker — flagging it as the obvious next step.

sources/community/buildkite/manifest.yaml
```diff
@@ -0,0 +1,589 @@
...
+          expr: {kind: path, path: [finished_at]}
+      - name: url
+        type: Utf8
+        nullable: true
+        description: API URL for the build.
+        expr: {kind: path, path: [url]}
+      - name: web_url
+        type: Utf8
+        nullable: true
+        description: Browser URL for the build.
```
antonmry, line 379: Tiny consistency nit, please ignore: `organization_builds.pipeline__name` doubles up with what users can already get from `buildkite.pipelines.name` once they have the slug. `pipeline_builds` skips it (the slug is in the filter, so users already know which pipeline). Either:

1. Drop `pipeline__name` from `organization_builds` for symmetry, or
2. Add `pipeline__name` to `pipeline_builds` (less useful since the filter pins the pipeline).

Option 1 feels cleaner. Up to you — entirely cosmetic.

sources/community/buildkite/manifest.yaml
sahil9001: I have not run this end-to-end with a live Buildkite token locally; I do not have one available in this environment. I did run `coral source lint sources/community/buildkite/manifest.yaml` successfully after the changes. `make rust-checks` still cannot run here because `cargo` is not installed.

sources/community/buildkite/manifest.yaml
sahil9001: Addressed in 38cea6d. `pipeline_builds` now exposes the matching build-resource columns from `organization_builds`: `graphql_id`, `tag`, `source`, `creator__id`, `pipeline__slug`, `scheduled_at`, and `url`.

sources/community/buildkite/manifest.yaml
sahil9001: Addressed in 38cea6d. Added `SELECT name, slug FROM buildkite.organizations LIMIT 5` to `test_queries` and updated the README verification text to mention the paginated check.

sources/community/buildkite/manifest.yaml
sahil9001: Addressed in 38cea6d. Added the `archived` boolean alongside `archived_at` and updated the README table docs.

sources/community/buildkite/manifest.yaml
sahil9001: Agreed. I left this as the documented v2 follow-up for now; the v1 manifest still exposes jobs as JSON on the build rows.

sources/community/buildkite/manifest.yaml
sahil9001: Addressed in 38cea6d using option 1. Removed `pipeline__name` from `organization_builds` and updated the README so the build tables stay consistent.

jsummerfield (approved):
Reviewed as a community source. CLA label is present, the README gives setup/schema/examples, and I did not find blocking source-spec or provider-surface issues in this pass. I did not run live API calls because I do not have provider credentials for this source.

jsummerfield:

- Make Buildkite API token setup and required organization/pipeline access explicit.
- Keep the CI operations workflows strong: organizations, pipelines, builds, jobs, annotations, and build states should be queryable with clear examples.
- Clarify pagination, branch/created-at filters, and rate-limit behavior for large CI histories.
- Include a first-success query that proves the token and organization slug are configured correctly.

jsummerfield:
Closing due to inactivity

PR #506, opened 2026-05-16
feat(sources/community/x): add x community source

### LINKED ISSUE #501: design: add X community source
## Summary

Add a new community `x` source that lets Coral query X/Twitter public and account-authorized data through SQL. For users, this would make posts, user profiles, recent search results, timelines, mentions, followers, and following lists queryable for social monitoring, support triage, developer-relations workflows, content audits, and lightweight research, while staying within Coral's current HTTP source-spec model and keeping the first version intentionally read-only.

## Motivation

X is widely used for product announcements, customer support, developer relations, incident communication, open-source project discussion, and public market/research signals. Teams often need to answer questions like:

- "What recent posts match this product, incident, or brand query?"
- "What did a specific account post recently?"
- "Who mentioned our account over the last few days?"
- "What public profile metadata is available for this handle?"
- "Which accounts follow, or are followed by, a specific user?"
- "What engagement counts are visible for recent posts?"
- "Which posts should be joined with GitHub issues, incidents, launches, or
  changelog entries?"

Without a Coral source, users who want X-aware agent workflows must either call the X API manually, write custom scripts, use SDKs directly, or copy data from dashboards. A Coral source would make common read-only social monitoring and investigation workflows available through SQL, and would complement existing Coral sources for GitHub, Slack, incidents, analytics, and observability.

## Proposed design

This would be a new community source under `sources/community/x`, implemented as a manifest-only HTTP source.

**Inputs:**

- `X_BEARER_TOKEN`
  - required secret
  - X API bearer token from an approved Developer Platform app, or an OAuth
    user access token when user-context endpoints are included
  - used in the `Authorization: Bearer ...` header
  - minimum scopes for OAuth 2.0 user tokens: `tweet.read` and `users.read`;some optional follow/list/bookmark/space tables may also need  additional  scopes if they are added later
  - app-only bearer tokens can read many public lookup/search/timeline
    endpoints, subject to the account's X API access tier

**Auth:**

- `HeaderAuth`
- header: `Authorization: Bearer {{input.X_BEARER_TOKEN}}`

**Base URL:**

- `https://api.x.com`

**Proposed v1 tables/functions:**

- **`x.users_by_id`**
  - `GET /2/users/{id}`
  - row key: `data`
  - required filter: `id`
  - key columns: `id`, `name`, `username`, `created_at`, `description`,
    `location`, `url`, `protected`, `verified`, `profile_image_url`,
    `pinned_tweet_id`, `public_metrics` (Json), `verified_type`,
    `affiliation` (Json)

- **`x.users_by_username`**
  - `GET /2/users/by/username/{username}`
  - row key: `data`
  - required filter: `username`
  - key columns: same user-profile columns as `x.users_by_id`
  - useful as the first discovery step before user-scoped tables

- **`x.recent_search`**
  - `GET /2/tweets/search/recent`
  - row key: `data`
  - required filter: `query`
  - optional filters: `start_time`, `end_time`, `since_id`, `until_id`,
    `sort_order`
  - cursor pagination with `pagination_token` and `meta.next_token`
  - page size via `max_results`; X's provider default is 10 and the documented
    range is 10-100, so Coral can use default 100 and max 100
  - key columns: `id`, `text`, `author_id`, `conversation_id`,
    `created_at`, `lang`, `possibly_sensitive`, `reply_settings`,
    `source`, `in_reply_to_user_id`, `public_metrics` (Json),
    `referenced_tweets` (Json), `entities` (Json), `attachments` (Json),
    `context_annotations` (Json), `edit_history_tweet_ids` (Json),
    `withheld` (Json)

- **`x.user_posts`**
  - `GET /2/users/{id}/tweets`
  - row key: `data`
  - required filter: `user_id`
  - optional filters: `start_time`, `end_time`, `since_id`, `until_id`,
    `exclude`
  - cursor pagination with `pagination_token` and `meta.next_token`
  - page size via `max_results`; documented range is 5-100
  - same post columns as `x.recent_search`

- **`x.user_mentions`**
  - `GET /2/users/{id}/mentions`
  - row key: `data`
  - required filter: `user_id`
  - optional filters: `start_time`, `end_time`, `since_id`, `until_id`
  - cursor pagination with `pagination_token` and `meta.next_token`
  - page size via `max_results`; documented range is 5-100
  - same post columns as `x.recent_search`

- **`x.followers`**
  - `GET /2/users/{id}/followers`
  - row key: `data`
  - required filter: `user_id`
  - cursor pagination with `pagination_token` and `meta.next_token`
  - page size via `max_results`; documented range is 1-1000
  - key columns: same user-profile columns as `x.users_by_id`

- **`x.following`**
  - `GET /2/users/{id}/following`
  - row key: `data`
  - required filter: `user_id`
  - cursor pagination with `pagination_token` and `meta.next_token`
  - page size via `max_results`; documented range is 1-1000
  - key columns: same user-profile columns as `x.users_by_id`

- **Optional follow-up: `x.me`**
  - `GET /2/users/me`
  - row key: `data`
  - no required filters
  - requires an OAuth user access token; it is not suitable as a generic
    app-only bearer-token smoke test
  - key columns: same user-profile columns as `x.users_by_id`
  - useful for validating user-context credentials and discovering the
    authenticated user's ID if the source chooses OAuth user-token auth

**Design details:**

- use X API v2 at `https://api.x.com/2`
- all v1 tables are read-only `GET` endpoints
- use bearer-token `HeaderAuth` because the read endpoints accept
  `Authorization: Bearer <token>`
- use cursor query pagination where supported:
  - request parameter: `pagination_token`
  - response cursor path: `meta.next_token`
  - page-size parameter: `max_results`, with endpoint-specific ranges
- set `rows_path` to `data`; X list endpoints return rows in `data` and pagination metadata in `meta`. Single-object lookup endpoints return one
  object in `data`, so the manifest should use Coral's direct row strategy for object responses.
- expose X IDs as `Utf8`; X documents IDs as strings to avoid large-integer issues in clients
- request common field expansions through query parameters where useful:
  - `tweet.fields=attachments,author_id,context_annotations,conversation_id,created_at,entities,geo,id,in_reply_to_user_id,lang,possibly_sensitive,public_metrics,referenced_tweets,reply_settings,source,text,withheld`
  - `user.fields=created_at,description,entities,id,location,name,pinned_tweet_id,profile_image_url,protected,public_metrics,url,username,verified,verified_type,withheld`
- preserve nested objects such as `public_metrics`, `entities`, `referenced_tweets`, `attachments`, `context_annotations`, `withheld`, and  `affiliation` as `Json`
- expose post `created_at` fields as `Utf8` in v1 unless live validation confirms Coral can parse the returned values safely. X docs show user
  `created_at` as ISO 8601, but some post examples are RFC 2822-style strings, so post timestamps should not be assumed to parse as `Timestamp`.
- require filters for search and user-scoped tables so Coral does not make broad or expensive API calls by default
- include cheap smoke `test_queries`, likely:
  - `SELECT id, name, username FROM x.users_by_username WHERE username = 'XDevelopers' LIMIT 1`
  - `SELECT id, name, username FROM x.users_by_id WHERE id = '2244994945' LIMIT 1`
- keep the source read-only; no posting, deleting, liking, reposting,  following, unfollowing, bookmarking, list mutation, direct messageswebhooks, or stream connections in v1

**Endpoint verification:**

Checked against X's current docs on 2026-05-16:

Result: the proposed v1 endpoints are publicly documented X API v2 endpoints. They require valid X Developer Platform credentials and are subject to the authenticated app/account's access tier, but they are not private, partner-only, or undocumented APIs.

- API overview: https://docs.x.com/overview
- Authentication mapping:
  https://docs.x.com/fundamentals/authentication/guides/v2-authentication-mapping
- Recent search:
  https://docs.x.com/x-api/posts/search-recent-posts
- User lookup:
  https://docs.x.com/x-api/users/lookup/introduction
- Timelines:
  https://docs.x.com/x-api/posts/timelines/introduction
- User posts:
  https://docs.x.com/x-api/posts/user-posts-timeline-by-user-id
- User mentions:
  https://docs.x.com/x-api/posts/user-mention-timeline-by-user-id
- Follows:
  https://docs.x.com/x-api/users/follows/introduction

**Existing-source check:**

Checked locally and through public GitHub search on 2026-05-16:

- no local `sources/core/x`
- no local `sources/core/twitter`
- no local `sources/community/x`
- no local `sources/community/twitter`
- no open or closed Coral GitHub issue/PR found for `twitter`,
  `api.x.com`, `developer.x.com`, `tweets/search/recent`, `twitter-api`,
  `sources/community/x`, or `sources/community/twitter`

**Affected surface:**

- community source only
- expected files:
  - `sources/community/x/manifest.yaml`
  - `sources/community/x/README.md`

No CLI, config, MCP, or runtime surface changes are proposed.

## Drawbacks and trade-offs

X API access is tiered and can be restrictive. Some users may not have access to recent search, timeline, follows, or higher-volume read operations depending on their Developer Platform plan and app settings. The README should make the credential and tier requirements explicit, and validation should distinguish source-spec errors from upstream access-tier errors. If `x.me` is added, it should clearly require an OAuth user access token rather than an app-only bearer token.

Rate limits are likely to matter. The source should keep v1 conservative by requiring filters for broad endpoints, using provider page-size limits, and documenting that large searches or high-cardinality follower/following scans may hit X rate limits quickly.

X's branding and API terminology are in transition: docs use X, while many endpoints still use `tweets` in paths and field names. The proposed source name is `x` to match current branding, but table and column descriptions should mention X/Twitter so users can find it.

Search is not a normal unfiltered table. `x.recent_search` should require a `query` filter, or possibly be implemented as a table function, so users do not accidentally run broad requests.

Some useful data appears in nested response objects or optional includes. The first version should expose stable scalar identifiers and common scalar fields, while preserving richer nested structures as `Json` instead of over-flattening every possible expansion.

Timestamps need live validation. X API v2 normally exposes `created_at` values as strings, and the docs show ISO 8601 user timestamps but RFC 2822-style post examples. The safest v1 is to expose post dates as `Utf8` unless live testing proves the returned values can be parsed consistently by Coral.

Public X data can still be sensitive or operationally important. The README should tell users to respect X Developer Terms, avoid over-collection, and be careful with personal data, account metadata, and joined data from other sources.

## Alternatives considered

A smaller source could start with only `users_by_username`, `users_by_id`, and `user_posts`. That would be easier to validate, but would miss the main social monitoring workflow: recent search and mentions.

A broader source could include likes, bookmarks, lists, spaces, communities, post counts, trends, news, usage, compliance, streams, webhooks, and direct messages. That expands scope too quickly for v1 and mixes public lookup workflows with user-private or streaming surfaces.

Write operations such as posting, deleting posts, liking, reposting, following, and list mutations were considered out of scope. Coral sources should start as read-only query surfaces, and write operations would add permission, safety, and product-surface concerns that do not belong in the first version.

Using unofficial scraping, browser automation, or third-party X data APIs was rejected. The initial source should use the official documented X API v2 surface.

Keeping the source name as `twitter` was considered for familiarity and search discoverability. The proposed name is `x` to match the current provider brand, with X/Twitter wording in the README and descriptions.

## Open questions

- Should the source schema be named `x`, `twitter`, or `x_twitter`?
- Should search-style endpoints such as recent search be modeled as required   filter tables or as source table functions?
- Should `x.recent_search` be included in v1 despite access-tier variation, or  should v1 start with user lookup and user timeline endpoints only?
- Should `followers` and `following` be included in v1, or deferred because  they can be high-cardinality and rate-limit heavy?
- Should `x.me` be deferred unless the source explicitly requires OAuth  user-token authentication?
- Should user `created_at` columns be `Timestamp` after live validation while  post `created_at` columns stay `Utf8`, or should all date fields stay `Utf8`  for maximum safety?
- Which optional scopes beyond `tweet.read` and `users.read` should be  documented for the first version?

Happy to adjust based on feedback.

---

Yashagarwal9798:
## Summary

Adds a new community source for X/Twitter under `sources/community/x/`.

This source enables querying X API v2 data (user lookup, recent search, user timelines, mentions, followers, and following lists) through SQL. The source is manifest-only, read-only, and uses Coral's existing HTTP source-spec model with bearer-token auth and cursor pagination. No CLI, MCP, config, or runtime changes.

Closes #501 

## What changed

**Added:**

- `sources/community/x/manifest.yaml`
- `sources/community/x/README.md`

### Tables

| Table | Endpoint | Pagination |
|-------|----------|------------|
| `users_by_id` | `GET /2/users/{id}` | none |
| `users_by_username` | `GET /2/users/by/username/{username}` | none |
| `recent_search` | `GET /2/tweets/search/recent` | cursor |
| `user_posts` | `GET /2/users/{id}/tweets` | cursor |
| `user_mentions` | `GET /2/users/{id}/mentions` | cursor |
| `followers` | `GET /2/users/{id}/followers` | cursor |
| `following` | `GET /2/users/{id}/following` | cursor |

### Auth

Bearer token auth via `X_BEARER_TOKEN`.

**Inputs:**

- `X_BEARER_TOKEN` - required secret. X API bearer token from an approved Developer Platform app, or an OAuth token with access to the selected X API v2 read endpoints. For OAuth 2.0 user-context tokens, the minimum read scopes are typically `tweet.read` and `users.read`. Endpoint availability is still   subject to the app/account's X API access tier.

## Implementation notes

- **X API v2 over HTTP** - all endpoints use `method: GET` against `https://api.x.com/2/...`.
- **HeaderAuth** - the manifest sends  `Authorization: Bearer {{input.X_BEARER_TOKEN}}`.
- **No `/2/users/me` table in v1** - that endpoint requires OAuth user-context authentication, so v1 avoids using it as a generic app-bearer-token smoke test.
- **Explicit `row_strategy: direct`** - `users_by_id` and `users_by_username` set `row_strategy: direct` explicitly for clarity, since they return a single object in `data` rather than an array.
- **Cursor pagination** - list tables use `mode: cursor_query` with `pagination_token` and `meta.next_token`. Page size is sent through `max_results`.
- **Endpoint-specific page sizes** - `recent_search` uses max 100, `user_posts` and `user_mentions` use max 100, and `followers` /  `following` use max 1000.
- **`recent_search` as a search function** - `recent_search` is modeled under `functions:` with `kind: search` and `search_limits`, using named arguments (`query =>`, `start_time =>`, etc.) instead of `WHERE` filters. SQL syntax is `SELECT ... FROM x.recent_search(query => '...')`.
- **Required filters for scoped endpoints** - user-scoped tables require `user_id`; user lookup tables require `id` or `username`.
- **Filter echo columns** - required filters are exposed as columns where users need to write `WHERE` predicates. `followers` and `following` expose `user_id` as a `from_filter` column so `WHERE user_id = '...'` satisfies the  required filter and remains visible in results.
- **IDs as Utf8** - X user and post IDs are exposed as strings to avoid large-integer precision issues.
- **Timestamps as Timestamp** - `created_at` values are parsed from ISO 8601 and exposed as `Timestamp` columns. This enables natural time predicates such as `WHERE created_at > NOW() - INTERVAL '24 hours'` without requiring `CAST`.
- **Nested objects as Json** - complex structures such as `public_metrics`, `entities`, `attachments`, `referenced_tweets`, `context_annotations`, `note_tweet`, `affiliation`, and `withheld` are preserved as `type: Json`.
- **Long-form post support** - `note_tweet` is requested via `tweet.fields` and exposed as a `Json` column on `recent_search`, `user_posts`, and `user_mentions`. Without it, X silently truncates posts over 280 characters in the `text` field.
- **Common field expansions** - requests include `tweet.fields` and `user.fields` values for useful scalar and nested metadata.

## Validation

- `cargo run -p coral-cli -- source lint sources/community/x/manifest.yaml` - **passes** (`Manifest is valid`)
- Live validation was run with a real `X_BEARER_TOKEN`.
- Live validation caught and fixed a required-filter mapping bug in `followers` and `following`: the required filter was `user_id`, but the echo column was initially named `source_user_id`. Both tables now expose the filter column as `user_id`, matching `user_posts` and `user_mentions`.
- All 7 tables verified against official X API documentation:

| Check | Result |
|-------|--------|
| Endpoint URLs match X API v2 docs | all 7 |
| Auth header matches X bearer-token docs | pass |
| Required filters match path/query needs | pass |
| `rows_path: [data]` matches X response shape | pass |
| Cursor config uses `pagination_token` and `meta.next_token` | pass |
| `max_results` limits match endpoint docs | pass |
| README examples match manifest table/filter names | pass |

## User-visible impact

New `x` source installable via:

```sh
X_BEARER_TOKEN=... \
coral source add --file sources/community/x/manifest.yaml

## Example queries:

-- Find a user ID from a handle
SELECT id, name, username, public_metrics
  FROM x.users_by_username
 WHERE username = 'XDevelopers';

-- Fetch a user profile by ID
SELECT id, name, username, created_at, verified
  FROM x.users_by_id
 WHERE id = '2244994945';

-- Search recent public posts
SELECT id, author_id, text, created_at, public_metrics
  FROM x.recent_search
 WHERE query = 'from:XDevelopers'
 LIMIT 25;

-- Recent posts from a user timeline
SELECT id, text, created_at, public_metrics
  FROM x.user_posts
 WHERE user_id = '2244994945'
 LIMIT 25;

-- Recent posts mentioning a user
SELECT id, author_id, text, created_at
  FROM x.user_mentions
 WHERE user_id = '2244994945'
 LIMIT 25;

-- Followers for a user
SELECT user_id, id, username, name, public_metrics
  FROM x.followers
 WHERE user_id = '2244994945'
 LIMIT 50;

-- Accounts followed by a user
SELECT user_id, id, username, name, public_metrics
  FROM x.following
 WHERE user_id = '2244994945'
 LIMIT 50;

```

## Known limitations

- **X API access tiers** - a valid token can still receive provider errors if the app/account's X API plan does not include a requested endpoint.
- **No `/2/users/me` in v1** - current-user lookup is deferred because it requires OAuth user-context authentication and is not suitable for a generic app bearer token source.
- **Recent search availability** - `recent_search` may not be available on all X API plans.
- **Rate limits** - followers/following can be high-cardinality for popular accounts. Use `LIMIT` and targeted workflows to avoid rate-limit pressure.
- **Timestamps as Timestamp** - `created_at` columns are parsed from ISO 8601 and exposed as `Timestamp`. Natural time-range SQL predicates work without `CAST`.
- **Nested data as Json** - nested structures are preserved as Json columns and should be queried with Coral JSON functions when needed.
- **Public data sensitivity** - public X data can still contain personal or operationally sensitive information. Users should follow X Developer Terms and be careful when joining this data with other sources.

## Out of scope

Not included in v1:

- Current authenticated user lookup (`GET /2/users/me`)
- Posting, deleting, liking, reposting, following, or unfollowing
- Direct messages
- Bookmarks
- Lists
- Spaces
- Communities
- Streaming APIs
- Webhooks/event subscriptions
- Media download
- Write operations of any kind

## Follow-on

- **X API `expansions` parameter** — adding `expansions=author_id,referenced_tweets.id,attachments.media_keys` as a literal query param would cause X to return related users/tweets/media in a top-level `includes` block. This would let users get author names alongside search results without a separate `users_by_id` lookup. Option 1: expose `includes` as a top-level `Json` column. Option 2: add a virtual table function that joins includes back into the rows (better UX, but may need DSL/runtime support). Applies to `recent_search`, `user_posts`, `user_mentions`, `followers`, and `following`.


sources/community/x/manifest.yaml
```diff
@@ -0,0 +1,1105 @@
...
+        expr:
+          kind: path
+          path: [username]
+      - name: created_at
+        type: Utf8
+        nullable: true
+        description: User creation time as returned by X.
+        expr:
+          kind: path
+          path: [created_at]
```
antonmry: Suggestion for v1 — please ignore if you'd really rather wait, but I think this is safe to ship today.

The X API v2 docs document `created_at` (and the other tweet/user timestamps in this manifest) as ISO 8601 / RFC 3339 — that's been the documented format since v2 launched. Exposing them as `Utf8` means the very natural query `WHERE created_at > NOW() - INTERVAL '24 hours'` doesn't work without `CAST(created_at AS TIMESTAMP)` on every reference.

Converting is mechanical:

```yaml
- name: created_at
  type: Timestamp
  nullable: true
  description: User creation time.
  expr:
    kind: format_timestamp
    input: iso8601
    expr:
      kind: path
      path: [created_at]
```

Same pattern would apply to `user_posts.created_at`, `user_mentions.created_at`, `recent_search.created_at`, `followers.created_at`, `following.created_at` — six call sites. The "live validation before converting" hedge in the PR description is fair caution, but since X's format is well-documented and stable, this should be a free win.

sources/community/x/manifest.yaml
```diff
@@ -0,0 +1,1105 @@
...
+  - name: recent_search
+    description: >-
+      Search recent public X/Twitter posts matching an X API query. Requires a
+      query filter and is useful for brand, incident, launch, and topic
+      monitoring.
+    fetch_limit_default: 100
+    filters:
+      - name: query
+        required: true
+      - name: start_time
```
antonmry: Tiny docs nit worth doing before this lands: X's `/2/tweets/search/recent` only returns posts from the last 7 days — anything older is silently absent. Right now the `description` and the `guide` (and the README) all describe this as "search recent public posts" without calling out the 7-day cap.

A user reading the column list and writing:

```sql
SELECT id, text FROM x.recent_search
WHERE query = 'from:XDevelopers' AND created_at > '2026-01-01'
```

gets 0 rows back for old date ranges and won't know whether their query is wrong or the data isn't there. Worth a one-line addition to the table `guide`:

> Only returns posts from the last 7 days. For older posts, use the X full-archive search (`/2/tweets/search/all`) which requires a higher API tier.

sources/community/x/manifest.yaml
```diff
@@ -0,0 +1,1105 @@
...
+        - name: until_id
+          from: filter
+          key: until_id
+        - name: sort_order
+          from: filter
+          key: sort_order
+        - name: tweet.fields
+          from: literal
+          value: attachments,author_id,context_annotations,conversation_id,created_at,edit_history_tweet_ids,entities,geo,id,in_reply_to_user_id,lang,possibly_sensitive,public_metrics,referenced_tweets,reply_settings,source,text,withheld
+    response:
```
antonmry: Follow-on idea, not a blocker — the killer ergonomics feature in X API v2 is the `expansions` parameter. Adding `expansions=author_id,referenced_tweets.id,attachments.media_keys` causes X to return related users/tweets/media inline under a top-level `includes` block, so users wouldn't need a separate `users_by_id` lookup just to display author names alongside search results.

Wiring it up would mean:

1. Add `expansions=...` as a literal query param.
2. Either expose `includes` as a top-level Json column on `recent_search`, or add a virtual table function that joins includes back into the rows.

Probably a v2 thing — option 2 is the better UX but requires DSL/runtime support that may not exist yet. Worth adding to the "Follow-on" section in the PR description so it doesn't get lost.

(Same applies to `user_posts`, `user_mentions`, `followers`, `following`.)

sources/community/x/manifest.yaml
```diff
@@ -0,0 +1,1105 @@
...
+    fetch_limit_default: 100
+    filters:
+      - name: query
+        required: true
+      - name: start_time
+      - name: end_time
+      - name: since_id
+      - name: until_id
+      - name: sort_order
+    request:
```
antonmry: Small docs improvement — please ignore if you'd rather keep the guide tight. X's search query DSL has a fair number of operators that users won't discover from this manifest alone:

```
from:XDevelopers          (author)
to:@user                  (replied to)
@username                 (mentioned)
lang:en                   (language)
is:reply / is:retweet     (type)
has:media / has:links     (content)
url:example.com           (URL filter)
-keyword                  (negation)
```

Adding 2-3 example operators in the table `guide` plus a link to `https://developer.x.com/en/docs/x-api/tweets/search/integrate/build-a-query/standard-operators` would meaningfully reduce the new-user friction. The README has `from:XDevelopers` as an example but the surface is much larger.

sources/community/x/manifest.yaml
```diff
@@ -0,0 +1,1105 @@
...
+          key: until_id
+        - name: exclude
+          from: filter
+          key: exclude
+        - name: tweet.fields
+          from: literal
+          value: attachments,author_id,context_annotations,conversation_id,created_at,edit_history_tweet_ids,entities,geo,id,in_reply_to_user_id,lang,possibly_sensitive,public_metrics,referenced_tweets,reply_settings,source,text,withheld
+    response:
+      rows_path: [data]
+    pagination:
```
antonmry, line 548: Tiny addition that probably belongs in v1 — X Premium users can post longer than 280 chars, and those posts only return the truncated `text` unless you request `note_tweet`. To get the full body, add `note_tweet` to the `tweet.fields` literal:

```
tweet.fields: ...,note_tweet,...
```

Then expose `note_tweet.text` as a column (or just keep the whole `note_tweet` object as Json). Otherwise `SELECT text FROM x.user_posts` will quietly truncate long-form posts at 280 chars, which is a surprising failure mode.

Same applies to `recent_search` and `user_mentions` `tweet.fields` literals above.

sources/community/x/README.md
```diff
@@ -0,0 +1,233 @@
...
+X_BEARER_TOKEN="..." coral source add x
+```
+
+## Tables
+
+### `users_by_username`
+
+Fetch a user profile by username. Use this first when you have a handle and
+need the stable user ID for timeline, mention, follower, or following queries.
+
```
antonmry, line 66: Worth adding to the README's Tables section — X's API tiers gate endpoints differently:

| Endpoint | Free | Basic | Pro | Enterprise |
|---|---|---|---|---|
| `users_by_id` / `users_by_username` | ✓ | ✓ | ✓ | ✓ |
| `recent_search` | ❌ (read-only Free is very limited) | ✓ | ✓ | ✓ |
| `user_posts` / `user_mentions` | very low limit | ✓ | ✓ | ✓ |
| `followers` / `following` | very low limit | ✓ | ✓ | ✓ |

Exact rate limits change frequently — a sentence pointing to `https://developer.x.com/en/docs/x-api/getting-started/about-x-api` plus a note that "users on the Free tier will hit endpoint-not-allowed errors on `recent_search`" would save users a confused first-run.

sources/community/x/manifest.yaml
```diff
@@ -0,0 +1,1105 @@
...
+      method: GET
+      path: /2/users/{{filter.id}}
+      query:
+        - name: user.fields
+          from: literal
+          value: affiliation,created_at,description,entities,id,location,name,pinned_tweet_id,profile_image_url,protected,public_metrics,url,username,verified,verified_type,withheld
+    response:
+      rows_path: [data]
+    pagination:
+      mode: none
```
antonmry: Tiny consistency nit, please ignore if you'd rather keep the manifest tight. `users_by_id` and `users_by_username` both return `{data: {single object}}`, and the manifest relies on the default `row_strategy` (which is `Direct`) to wrap that single object as one row. That works correctly, but for consistency with `sources/community/cal/manifest.yaml` — which sets `row_strategy: direct` explicitly on every table including singletons — it'd be marginally clearer to declare it here too:

```yaml
response:
  rows_path: [data]
  row_strategy: direct
```

Pure cosmetic — same behavior either way.

jsummerfield (approved):
Automated review as a community source. CLA label is present, the README gives setup/schema/examples, and I did not find blocking source-spec or provider-surface issues in this pass. I did not run live API calls because I do not have provider credentials for this source.

Please address Anton's comments above before merging.

jsummerfield (approved):
Reviewed against the source checklist and current X API v2 docs. The required filters, endpoint paths, bearer auth, cursor pagination (`pagination_token` / `meta.next_token`), and endpoint-specific page sizes line up with the provider docs; IDs are safely modeled as strings, high-cardinality follower/following tables have required filters, and `coral source lint` passes. I did not run it with my own X token.

sources/community/x/manifest.yaml
Yashagarwal9798: Done converted all seven created_at columns to Timestamp with format_timestamp / iso8601, updated the README and PR description to match.

sources/community/x/manifest.yaml
Yashagarwal9798: Added  both the manifest description and the README now call out the 7-day window and point to /2/tweets/search/all for older posts.

sources/community/x/manifest.yaml
Yashagarwal9798: added to a Follow-on section in the PR description so it's tracked for v2. Agreed option 2 is the better UX once DSL/runtime support is there.

sources/community/x/manifest.yaml
Yashagarwal9798: Added  the README now lists common query operators and links to the X query building guide.

sources/community/x/manifest.yaml
Yashagarwal9798: Done added note_tweet to tweet.fields on all three post tables and exposed it as a Json column.

sources/community/x/README.md
Yashagarwal9798: Added the README Tables section now has a tier availability table and a link to the X API access levels page.

sources/community/x/manifest.yaml
Yashagarwal9798: Added row_strategy: direct to both users_by_id and users_by_username for explicit clarity.

jsummerfield:
Thanks for adding this source. I found a few issues I think should be fixed before merge:

1. The README documents `X_BEARER_TOKEN="..." coral source add x` under “For bundled installs”, but this PR only adds `sources/community/x`. Current Coral rejects that command with `unknown bundled source 'x'`, so users following the README cannot install it that way. Please remove that block or replace it with the file-based community-source command.

2. `recent_search` should be modeled as a Coral search function, not as a normal table with a required `query` filter. The X endpoint is provider-native search: it takes a query string and returns provider-ranked/recent matching posts. Coral’s source-spec guidance says those surfaces should live under `functions` with `kind: search` and `search_limits`, which also means the README SQL example should use named function arguments rather than `WHERE query = ...`.

3. The README access-tier table appears stale. The linked X access-level page now redirects to current X docs describing Pay-per-use and Enterprise pricing, not Free/Basic/Pro/Enterprise tiers. Since access and cost are a major footgun for this source, I’d replace the table with a shorter note that endpoint availability and pricing are controlled by X’s current pay-per-use/enterprise model and link to the current pricing/docs page.

Yashagarwal9798:
@jsummerfield done all three addressed:
1. Removed the bundled `coral source add x` command  only `--file` remains.
2. Moved `recent_search` to `functions:` with `kind: search` and `search_limits`. SQL is now `SELECT ... FROM x.recent_search(query => '...')`.
3. Replaced the stale tier table with a shorter note linking to X's current access level docs.

jsummerfield (changes requested):
Findings:

- Medium: the OAuth 2.0 user-token guidance understates the scopes needed for follows tables. `sources/community/x/README.md:17` and `sources/community/x/manifest.yaml:14` mention `tweet.read users.read offline.access`, but this source exposes `followers` and `following`, and X’s v2 auth mapping requires `follows.read` for those user-context endpoints. App-only bearer tokens are fine for public reads, but a user token minted exactly as documented can fail on these tables. Docs: https://docs.x.com/fundamentals/authentication/guides/v2-authentication-mapping and https://docs.x.com/fundamentals/authentication/oauth-2-0/authorization-code

- Low: the `recent_search` access wording is stale. `sources/community/x/README.md:57` says it is not available on the free tier and will return endpoint-not-allowed, but current X docs describe recent search as available to all developers under the current pricing model. Please replace that tier language with current pricing/access guidance so users do not misdiagnose billing or credit problems. Docs: https://docs.x.com/x-api/posts/search/introduction and https://docs.x.com/x-api/getting-started/pricing

- Low: the README still does not show a reproducible live validation path (`sources/community/x/README.md:23`). Since the built-in test query hits billable `recent_search`, please state that `coral source test x` needs usable X API credits plus lookup/search/follows-capable auth, or include sanitized successful output from the current head.

The previous install-path, search-function, and stale tier-table issues are addressed, and `coral source lint` passes. This should be close after the auth/live-validation docs are corrected.

jsummerfield (changes requested):
Findings:

- Medium: `sources/community/x/README.md:57` still links the old developer.x.com access-level page and says `recent_search` is not available on the free tier. The current X docs have moved to docs.x.com and describe recent search as available to all developers, while full-archive search is Pay-per-use/Enterprise. Please update this section to match the current X pricing/access model and link the current docs. Docs: https://docs.x.com/x-api/posts/search/introduction and https://docs.x.com/x-api/getting-started/pricing

- Medium: `sources/community/x/manifest.yaml:14` and `sources/community/x/manifest.yaml:37` still use stale “API tier” / “higher API tier” wording in user-facing descriptions. Please use the current pay-per-use/Enterprise framing consistently so users know what token/access they need.

The earlier install-command and `recent_search` modeling issues are fixed on this head. Thanks for tightening those; the remaining issue is just the access/pricing guidance, which matters a lot for first success with X.

jsummerfield:

jsummerfield (changes requested):
Thanks for the updates here. The manifest still lints cleanly, and the install-command/search-function items are addressed. I found two remaining blockers on the current head:

- `sources/community/x/README.md` still points to the old developer.x.com access page and says `recent_search` is not available on the free tier, while `sources/community/x/manifest.yaml` still uses “API tier” / “higher API tier” wording. Please update this to the current X docs framing: recent search is documented under the current X API search surface, while full-archive search is the pay-per-use/Enterprise path. Current docs: https://docs.x.com/x-api/posts/search/introduction and https://docs.x.com/x-api/getting-started/pricing

- The PR body says live validation was run with a real token, but I still do not see reproducible sanitized output from the current head. Please include the exact `coral source add` / representative query output, with the token and any sensitive values redacted, so we can verify the source actually works against X with usable API access.

Everything else I checked in this pass is either fixed or non-blocking.

jsummerfield:

jsummerfield:

PR #525, opened 2026-05-16
feat(sources/community/shopify): add shopify community source

### LINKED ISSUE #522: design: add shopify community source
## Summary

Add a new community `shopify` source that lets Coral query Shopify Admin GraphQL data through SQL. For users, this would make store metadata, products, product variants, collections, orders, customers, locations, and inventory items queryable for commerce analytics, catalog audits, fulfillment triage, inventory checks, support workflows, and cross-source joins with tools like Stripe, Slack, GitHub, incidents, and observability sources.

## Motivation

Shopify is a common system of record for ecommerce businesses. Teams often need
to answer questions like:

- "Which products or variants changed recently?"
- "Which SKUs are active, draft, archived, or low inventory?"
- "Which orders are paid but not fulfilled?"
- "Which customers have the most orders or highest lifetime spend?"
- "Which locations can fulfill online orders?"
- "Which inventory items are tracked and tied to specific SKUs?"
- "How do orders and customers relate to support tickets, incidents, payments, or warehouse workflows?"
Without a Coral source, users need custom scripts, Shopify SDKs, exports, or manual admin work to inspect this data. A read-only Coral source would expose common operational Shopify data through SQL while keeping the first version bounded to documented Admin GraphQL queries.

## Proposed design

This would be a new community source under `sources/community/shopify`, implemented as a manifest-only HTTP source.

**Inputs:**

- `SHOPIFY_SHOP_DOMAIN`
  - required variable
  - Shopify shop domain, for example `example.myshopify.com`
  - used to build the Admin GraphQL endpoint
- `SHOPIFY_API_VERSION`
  - optional variable
  - defaults to `2026-04`, which Shopify docs list as the latest stable Admin
    GraphQL API version checked on May 16, 2026
- `SHOPIFY_ADMIN_ACCESS_TOKEN`
  - required secret
  - Admin API access token for a custom or public app
  - sent as the `X-Shopify-Access-Token` header
  - recommended read scopes depend on the tables used:
    - `read_products` for products, product variants, and collections
    - `read_orders` for orders
    - `read_customers` for customers
    - `read_locations` for locations
    - `read_inventory` for inventory items
    - `read_all_orders` only when the app needs approved access beyond the  default recent-orders window

**Auth:**

- `HeaderAuth`
- header: `X-Shopify-Access-Token: {{input.SHOPIFY_ADMIN_ACCESS_TOKEN}}`
- request headers: `Accept: application/json` and `Content-Type: application/json`

**Base URL:**

- `https://{{input.SHOPIFY_SHOP_DOMAIN}}/admin/api/{{input.SHOPIFY_API_VERSION}}`

**Proposed v1 tables:**

- **`shopify.shop`**
  - GraphQL query: `shop`
  - no pagination
  - useful as the first smoke test and account metadata table
  - key columns: `id`, `name`, `myshopify_domain`, `email`, `contact_email`, `currency_code`, `iana_timezone`, `created_at`, `updated_at`, `primary_domain__url`, `primary_domain__host`

- **`shopify.products`**
  - GraphQL query: `products`
  - cursor pagination with `first`, `after`, and `pageInfo.endCursor`
  - optional `query` filter using Shopify API search syntax
  - key columns: `id`, `title`, `handle`, `status`, `vendor`, `product_type`, `tags`, `total_inventory`, `has_only_default_variant`, `created_at`, `updated_at`, `published_at`, price range columns

- **`shopify.product_variants`**
  - GraphQL query: `productVariants`
  - cursor pagination with `first`, `after`, and `pageInfo.endCursor`
  - optional `query` filter using Shopify API search syntax, such as SKU, barcode, product ID, collection ID, or inventory filters
  - key columns: `id`, `product_id`, `product_title`, `title`, `display_name`, `sku`, `barcode`, `price`, `compare_at_price`, `inventory_item_id`, `inventory_quantity`, `inventory_policy`, `selected_options`, `created_at`, `updated_at`

- **`shopify.collections`**
  - GraphQL query: `collections`
  - cursor pagination with `first`, `after`, and `pageInfo.endCursor`
  - optional `query` filter using Shopify API search syntax
  - key columns: `id`, `title`, `handle`, `description`, `sort_order`, `products_count`, `rule_set`, `seo__title`, `seo__description`, `updated_at`

- **`shopify.orders`**
  - GraphQL query: `orders`
  - cursor pagination with `first`, `after`, and `pageInfo.endCursor`
  - optional `query` filter using Shopify API search syntax, such as financial status, fulfillment status, customer ID, SKU, or date filters
  - key columns: `id`, `name`, `created_at`, `updated_at`, `processed_at`, `cancelled_at`, `display_financial_status`, `display_fulfillment_status`, `currency_code`, total/subtotal/tax amount columns, `customer_id`, `customer_display_name`, `tags`, `test`

- **`shopify.customers`**
  - GraphQL query: `customers`
  - cursor pagination with `first`, `after`, and `pageInfo.endCursor`
  - optional `query` filter using Shopify API search syntax
  - key columns: `id`, `display_name`, `first_name`, `last_name`, `default_email`, `default_phone`, `state`, `number_of_orders`, `amount_spent`, `amount_spent_currency_code`, `verified_email`,  `tax_exempt`, `tags`, `created_at`, `updated_at`

- **`shopify.locations`**
  - GraphQL query: `locations`
  - cursor pagination with `first`, `after`, and `pageInfo.endCursor`
  - optional filters: `query`, `include_inactive`, `include_legacy`
  - key columns: `id`, `name`, `is_active`, `fulfills_online_orders`, `has_active_inventory`, `ships_inventory`, address fields, and `fulfillment_service__handle`

- **`shopify.inventory_items`**
  - GraphQL query: `inventoryItems`
  - cursor pagination with `first`, `after`, and `pageInfo.endCursor`
  - optional `query` filter using Shopify API search syntax, such as SKU, ID, created date, or updated date
  - key columns: `id`, `sku`, `tracked`, `requires_shipping`,  `duplicate_sku_count`, `country_code_of_origin`,  `province_code_of_origin`, `harmonized_system_code`, `inventory_history_url`, `created_at`, `updated_at`

**Design details:**

- use Shopify Admin GraphQL API rather than REST because Shopify marks REST Admin API as legacy and requires new public apps to use GraphQL Admin API
- all v1 tables are read-only GraphQL queries sent as HTTP `POST`
- use `X-Shopify-Access-Token` header auth
- use `cursor_body` pagination for connection queries:
  - request cursor body path: `variables.after`
  - response cursor path: `data.<connection>.pageInfo.endCursor`
  - page size body path: `variables.first`
- keep provider IDs as `Utf8` because Shopify GraphQL IDs are global ID strings such as `gid://shopify/Product/...`
- expose Shopify GraphQL `DateTime` values as `Utf8` in v1 to keep live validation simple and avoid timestamp conversion assumptions
- expose nested arrays and objects such as `tags`, `selected_options`, `rule_set`, and money/SEO structures as scalar columns where practical and
  `Json` where they are richer structures
- keep the source read-only; no product, inventory, customer, order, or  fulfillment mutations in v1
- include cheap smoke `test_queries`, likely:
  - `SELECT id, name, myshopify_domain FROM shopify.shop LIMIT 1`
  - `SELECT id, title, status FROM shopify.products LIMIT 1`

**Endpoint verification:**

Checked against Shopify's current public docs on May 16, 2026:

Result: the proposed v1 tables map to publicly documented Shopify Admin GraphQL API queries. They require a valid Admin API access token and the appropriate read scopes, but the APIs are documented and not private, partner-only, or unsupported surfaces.

- Admin GraphQL overview, auth, and endpoint: https://shopify.dev/docs/api/admin-graphql/latest
- Access scopes:
  https://shopify.dev/docs/api/usage/access-scopes
- GraphQL pagination:
  https://shopify.dev/docs/api/usage/pagination-graphql
- REST legacy note:
  https://shopify.dev/docs/api/admin-rest/latest
- `shop` query:
  https://shopify.dev/docs/api/admin-graphql/latest/queries/shop
- `products` query:
  https://shopify.dev/docs/api/admin-graphql/latest/queries/products
- `productVariants` query:
  https://shopify.dev/docs/api/admin-graphql/latest/queries/productVariants
- `collections` query:
  https://shopify.dev/docs/api/admin-graphql/latest/queries/collections
- `orders` query:
  https://shopify.dev/docs/api/admin-graphql/latest/queries/orders
- `customers` query:
  https://shopify.dev/docs/api/admin-graphql/latest/queries/customers
- `locations` query:
  https://shopify.dev/docs/api/admin-graphql/latest/queries/locations
- `inventoryItems` query:
  https://shopify.dev/docs/api/admin-graphql/latest/queries/inventoryItems

**Existing-source check:**

Checked locally before authoring this branch and through public GitHub search on May 16, 2026:

- no local `sources/core/shopify`
- no pre-existing local `sources/community/shopify`
- no open or closed Coral GitHub issue/PR found for `shopify`
- no Coral GitHub issue/PR found for `sources/community/shopify`
- upstream `sources/core` and `sources/community` directory listings do not
  include Shopify

**Affected surface:**

- community source only
- expected files:
  - `sources/community/shopify/manifest.yaml`
  - `sources/community/shopify/README.md`

No CLI, config, MCP, runtime, or source-spec changes are proposed.

## Drawbacks and trade-offs

Shopify Admin API access depends on app scopes. Users with a token that only has `read_products` can query catalog tables, but order, customer, location, and inventory tables require their own scopes. The README should clearly map tables to scopes so provider permission errors are not mistaken for source-spec bugs.

Orders have extra constraints. Shopify's docs say apps can access only the default recent-orders window unless they request and receive approval for `read_all_orders` in addition to `read_orders`. The first source should document that limitation and avoid assuming full historical order access.

Customer and order data can include protected customer data. The source should stay read-only, preserve least-privilege guidance, and warn users that Shopify may restrict protected customer data unless the app meets Shopify's protected customer data requirements.

GraphQL Admin API is cost-rate-limited. The first version should keep page sizes conservative, expose query filters, and encourage users to use `LIMIT` and Shopify search syntax rather than broad scans against large stores.

Shopify's API surface is large. The first source should avoid over-scoping into fulfillment orders, discounts, metafields, blogs/pages, themes, markets, returns, analytics, or write operations. Those can be follow-ups once the core commerce tables are validated.

## Alternatives considered

Using REST Admin API was rejected because Shopify marks REST Admin API as legacy and requires new public apps to use GraphQL Admin API. GraphQL also matches Coral's existing POST + `cursor_body` manifest pattern used by other sources.

A smaller source could start with only `shop`, `products`, `product_variants`, and `collections`. That would be easier to validate but would miss the main operational workflows around orders, customers, locations, and inventory.

A broader source could include fulfillment orders, inventory levels, metafields, discounts, pages, blogs, articles, returns, markets, and analytics. That expands scope too quickly for v1 and mixes core commerce data with specialized surfaces that have different permission and data-volume risks.

Write operations such as product updates, inventory adjustments, fulfillment actions, customer edits, refunds, or order mutations were considered out of scope. Coral sources should start as read-only query surfaces.

Using the Storefront API was rejected for this contribution because the target is admin/operations data, not buyer-facing storefront data.

## Open questions

- Should v1 include all eight proposed tables, or start with catalog and shop metadata first?
- Should `orders` and `customers` be included in v1 despite protected customer data and scope requirements, or deferred to reduce review risk?
- Should date/time columns stay as `Utf8` for v1, or should live validation convert Shopify `DateTime` fields to Coral `Timestamp`?
- Should `inventory_levels` be included in v1, or deferred because it requires nested quantity handling by location?
- Should the source expose `metafields` as follow-up table functions rather than nested columns on each resource?

---

Yashagarwal9798:
## Summary

Adds a new community source for Shopify under `sources/community/shopify/`.

This source enables querying Shopify Admin GraphQL data through SQL. It covers
shop metadata, products, product variants, collections, orders, customers,
locations, and inventory items. The source is manifest-only, read-only, and
uses Coral's existing HTTP source-spec model with `POST` GraphQL requests,
`X-Shopify-Access-Token` auth, and cursor pagination. No CLI, MCP, config,
runtime, or source-spec changes are included.

Closes #522 
## What changed

**Added:**

- `sources/community/shopify/manifest.yaml`
- `sources/community/shopify/README.md`

### Tables

| Table | GraphQL root field | Pagination |
|---|---|---|
| `shop` | `shop` | none |
| `products` | `products` | cursor |
| `product_variants` | `productVariants` | cursor |
| `collections` | `collections` | cursor |
| `orders` | `orders` | cursor |
| `customers` | `customers` | cursor |
| `locations` | `locations` | cursor |
| `inventory_items` | `inventoryItems` | cursor |

### Auth

Admin API token auth via `SHOPIFY_ADMIN_ACCESS_TOKEN`.

**Inputs:**

- `SHOPIFY_SHOP_DOMAIN` - required variable. Shopify shop domain such as
  `example.myshopify.com`.
- `SHOPIFY_API_VERSION` - optional variable. Defaults to `2026-04`.
- `SHOPIFY_ADMIN_ACCESS_TOKEN` - required secret. Shopify Admin API access
  token from a custom or public app. Coral sends it as
  `X-Shopify-Access-Token`.

## Implementation notes

- **Admin GraphQL over HTTP** - all tables use `method: POST` against
  `/admin/api/{{input.SHOPIFY_API_VERSION}}/graphql.json`.
- **API version maintenance** - the source defaults to Shopify Admin GraphQL
  `2026-04`. Future bumps should check Shopify's API versioning release
  schedule (`https://shopify.dev/docs/api/usage/versioning#release-schedule`)
  because Shopify releases stable versions quarterly, supports each stable
  version for at least 12 months, and version bumps can require GraphQL query
  updates for renamed, deprecated, or removed fields and enum values.
- **HeaderAuth** - the manifest sends
  `X-Shopify-Access-Token: {{input.SHOPIFY_ADMIN_ACCESS_TOKEN}}`.
- **JSON request headers** - the source sends `Accept: application/json` and
  `Content-Type: application/json`.
- **Cursor body pagination** - connection tables use `mode: cursor_body` with
  `variables.after` and `data.<connection>.pageInfo.endCursor`.
- **Body page size** - connection tables send page size through
  `variables.first`, defaulting to 100 with max 250.
- **Optional Shopify search filters** - connection tables expose a `query`
  filter using Shopify Admin API search syntax. `locations` also exposes
  `include_inactive` and `include_legacy`.
- **IDs as Utf8** - Shopify GraphQL global IDs are exposed as strings.
- **Money amounts as raw and ergonomic columns** - Shopify money amount fields
  are preserved as exact decimal strings in `*_decimal` columns and also
  exposed through natural `Float64` amount columns for arithmetic.
- **Timestamps as Timestamp** - Shopify `DateTime` fields are parsed from ISO
  8601 and exposed as `Timestamp` columns for date-range filtering and
  ordering.
- **Nested data as Json** - arrays and richer structures such as product tags,
  selected options, order tags, customer tags, and smart collection rule sets
  are preserved as `Json`.
- **Read-only v1** - no product, inventory, customer, order, fulfillment, or
  other mutations are included.

## Endpoint verification

Checked against Shopify's public documentation on May 16, 2026:

| Check | Result |
|---|---|
| Admin GraphQL endpoint format is documented | pass |
| `X-Shopify-Access-Token` auth header is documented | pass |
| `2026-04` is the latest Admin GraphQL version shown by Shopify docs | pass |
| REST Admin API is marked legacy for new public apps | pass |
| `shop` query is public/documented | pass |
| `products` query is public/documented | pass |
| `productVariants` query is public/documented | pass |
| `collections` query is public/documented | pass |
| `orders` query is public/documented | pass |
| `customers` query is public/documented | pass |
| `locations` query is public/documented | pass |
| `inventoryItems` query is public/documented | pass |
| GraphQL cursor pagination with `first`, `after`, and `pageInfo.endCursor` is documented | pass |
| README table list, filters, and scopes match the manifest | pass |

Primary docs checked:

- https://shopify.dev/docs/api/admin-graphql/latest
- https://shopify.dev/docs/api/usage/access-scopes
- https://shopify.dev/docs/api/usage/pagination-graphql
- https://shopify.dev/docs/api/admin-rest/latest
- https://shopify.dev/docs/api/admin-graphql/latest/queries/shop
- https://shopify.dev/docs/api/admin-graphql/latest/queries/products
- https://shopify.dev/docs/api/admin-graphql/latest/queries/productVariants
- https://shopify.dev/docs/api/admin-graphql/latest/queries/collections
- https://shopify.dev/docs/api/admin-graphql/latest/queries/orders
- https://shopify.dev/docs/api/admin-graphql/latest/queries/customers
- https://shopify.dev/docs/api/admin-graphql/latest/queries/locations
- https://shopify.dev/docs/api/admin-graphql/latest/queries/inventoryItems

## Validation

Lint completed separately. Live validation was rerun on May 18, 2026 against a
Shopify development store (`coral-dev-testing.myshopify.com`).

### Lint

```
coral source lint sources/community/shopify/manifest.yaml
Manifest is valid
```

### Source add and test queries

```
coral source add --file sources/community/shopify/manifest.yaml
Added source shopify

  shopify connected successfully

    shopify (8 tables)
    - collections
    - customers
    - inventory_items
    - locations
    - orders
    - product_variants
    - products
    - shop
    Query tests
    2 declared, 2 passed, 0 failed

    PASS SELECT id, name, myshopify_domain FROM shopify.shop LIMIT 1
      1 row

    PASS SELECT id, title, status FROM shopify.products LIMIT 1
      0 rows
```

### Live SQL queries - all 8 tables

| Table | Query | Result |
|---|---|---|
| `shop` | `SELECT id, name, myshopify_domain FROM shopify.shop LIMIT 1` | 1 row: `gid://shopify/Shop/76681248933`, `Coral Dev Testing`, `coral-dev-testing.myshopify.com` |
| `products` | `SELECT id, title, status, min_variant_price_amount, min_variant_price_amount_decimal, created_at FROM shopify.products LIMIT 5` | 0 rows (empty dev store) - query ran without error |
| `product_variants` | `SELECT id, product_title, sku, price, price_decimal, created_at FROM shopify.product_variants LIMIT 5` | 0 rows - query ran without error |
| `collections` | `SELECT id, title, updated_at FROM shopify.collections LIMIT 5` | 1 row: `gid://shopify/Collection/335299084453`, `Home page`, `2026-05-16T14:23:34Z` |
| `orders` | `SELECT id, name, current_total_price_amount, current_total_price_amount_decimal, created_at FROM shopify.orders LIMIT 5` | 0 rows - query ran without error |
| `customers` | `SELECT id, display_name, amount_spent, amount_spent_decimal, created_at FROM shopify.customers LIMIT 5` | 0 rows - query ran without error |
| `locations` | `SELECT id, name, is_active, created_at FROM shopify.locations LIMIT 5` | 1 row: `gid://shopify/Location/86482845861`, `Shop location`, `true`, `2026-05-16T14:23:31Z` |
| `inventory_items` | `SELECT id, sku, tracked, created_at FROM shopify.inventory_items LIMIT 5` | 0 rows - query ran without error |

### Review-change validation

| Check | Query | Result |
|---|---|---|
| Float64 money aggregation | `SELECT SUM(current_total_price_amount) AS total_order_value FROM shopify.orders` | Query ran without error; result was `NULL` because the dev store has no orders |
| Timestamp date-range predicate | `SELECT id, title, updated_at FROM shopify.collections WHERE updated_at > NOW() - INTERVAL '30 days' LIMIT 5` | 1 row: `gid://shopify/Collection/335299084453`, `Home page`, `2026-05-16T14:23:34Z` |

All 8 tables responded successfully. Tables with 0 rows are expected on a
fresh development store with no sample products, orders, or customers
populated.

## User-visible impact

New `shopify` source installable via:

```sh
SHOPIFY_SHOP_DOMAIN=example.myshopify.com \
SHOPIFY_ADMIN_ACCESS_TOKEN=shpat_... \
coral source add --file sources/community/shopify/manifest.yaml
```

Example queries:

```sql
-- Confirm the authenticated shop
SELECT id, name, myshopify_domain
FROM shopify.shop
LIMIT 1;

-- Active products
SELECT id, title, status, vendor, total_inventory
FROM shopify.products
WHERE query = 'status:active'
LIMIT 25;

-- Product variants by SKU
SELECT id, product_title, sku, price, inventory_quantity
FROM shopify.product_variants
WHERE query = 'sku:ABC*'
LIMIT 25;

-- Paid unfulfilled orders
SELECT id, name, display_financial_status, display_fulfillment_status
FROM shopify.orders
WHERE query = 'financial_status:paid fulfillment_status:unfulfilled'
LIMIT 25;

-- Aggregate paid unfulfilled order value
SELECT SUM(current_total_price_amount) AS total_unfulfilled_value
FROM shopify.orders
WHERE query = 'financial_status:paid fulfillment_status:unfulfilled';

-- Inventory items by SKU
SELECT id, sku, tracked, locations_count
FROM shopify.inventory_items
WHERE query = 'sku:ABC*'
LIMIT 25;
```

## Known limitations

- Shopify Admin API access depends on app scopes. A token with only
  `read_products` can query catalog tables but not orders, customers,
  locations, or inventory tables that require other scopes.
- Historical orders require approved `read_all_orders` in addition to
  `read_orders`.
- Order and customer data can include protected customer data and may require
  Shopify protected customer data approval.
- Shopify GraphQL Admin API is cost-rate-limited. Users should prefer targeted
  Shopify search filters and SQL `LIMIT`.
- Shopify money amounts are exact decimal strings in the API. This source keeps
  exact raw values in `*_decimal` columns and exposes `Float64` amount columns
  for analytics.
- Shopify `DateTime` columns are exposed as `Timestamp` values parsed from ISO
  8601 strings.
- Nested structures are preserved as `Json` columns.

## Out of scope

Not included in v1:

- Product, variant, collection, order, customer, inventory, or fulfillment
  mutations
- Fulfillment orders
- Inventory levels by location
- Metafields as standalone tables or functions
- Discounts
- Markets
- Returns
- Analytics
- Pages, blogs, articles, and themes
- Bulk operations
- Storefront API data
- Webhooks or event subscriptions
- Write operations of any kind

### DIFF
--- a/sources/community/shopify/manifest.yaml
+++ b/sources/community/shopify/manifest.yaml
@@ -1,0 +1,1693 @@
+name: shopify
+version: 0.1.0
+dsl_version: 3
+backend: http
+description: >-
+  Query Shopify shop metadata, products, product variants, collections,
+  orders, customers, locations, and inventory items.
+inputs:
+  SHOPIFY_SHOP_DOMAIN:
+    kind: variable
+    required: true
+    hint: |
+      Shopify shop domain for the Admin GraphQL API, for example
+      `example.myshopify.com`. Do not include `https://` or a trailing slash.
+  SHOPIFY_API_VERSION:
+    kind: variable
+    default: "2026-04"
+    required: false
+    hint: |
+      Shopify Admin GraphQL API version. Defaults to `2026-04`, the latest
+      stable version checked for this source. Override only when you have
+      validated the source against another supported Admin API version.
+  SHOPIFY_ADMIN_ACCESS_TOKEN:
+    kind: secret
+    required: true
+    hint: |
+      Shopify Admin API access token from a custom or public app. The token is
+      sent as `X-Shopify-Access-Token`. Grant only the read scopes needed for
+      the tables you use: `read_products`, `read_orders`, `read_customers`,
+      `read_locations`, and `read_inventory`. Historical orders require
+      approved `read_all_orders` in addition to `read_orders`.
+base_url: https://{{input.SHOPIFY_SHOP_DOMAIN}}/admin/api/{{input.SHOPIFY_API_VERSION}}
+auth:
+  type: HeaderAuth
+  headers:
+    - name: X-Shopify-Access-Token
+      from: template
+      template: "{{input.SHOPIFY_ADMIN_ACCESS_TOKEN}}"
+request_headers:
+  - name: Accept
+    from: literal
+    value: application/json
+  - name: Content-Type
+    from: literal
+    value: application/json
+test_queries:
+  - SELECT id, name, myshopify_domain FROM shopify.shop LIMIT 1
+  - SELECT id, title, status FROM shopify.products LIMIT 1
+tables:
+  - name: shop
+    description: Shopify shop metadata for the authenticated Admin API token.
+    guide: |
+      Start here to validate credentials and confirm which shop a token
+      belongs to. This table returns one row from the Admin GraphQL `shop`
+      query.
+    fetch_limit_default: 1
+    request:
+      method: POST
+      path: /graphql.json
+      body:
+        - path: [query]
+          from: literal
+          value: |
+            query ShopifyShop {
+              shop {
+                id
+                name
+                myshopifyDomain
+                email
+                contactEmail
+                currencyCode
+                ianaTimezone
+                timezoneAbbreviation
+                taxesIncluded
+                taxShipping
+                createdAt
+                updatedAt
+                url
+                weightUnit
+                primaryDomain {
+                  id
+                  host
+                  url
+                  sslEnabled
+                }
+                plan {
+                  displayName
+                  partnerDevelopment
+                  shopifyPlus
+                }
+              }
+            }
+    response:
+      rows_path: [data, shop]
+      row_strategy: direct
+    pagination:
+      mode: none
+    columns:
+      - name: id
+        type: Utf8
+        nullable: false
+        description: Globally unique Shopify shop ID.
+        expr:
+          kind: path
+          path: [id]
+      - name: name
+        type: Utf8
+        nullable: false
+        description: Shop name.
+        expr:
+          kind: path
+          path: [name]
+      - name: myshopify_domain
+        type: Utf8
+        nullable: false
+        description: Canonical myshopify.com domain.
+        expr:
+          kind: path
+          path: [myshopifyDomain]
+      - name: email
+        type: Utf8
+        nullable: true
+        description: Shop owner email address.
+        expr:
+          kind: path
+          path: [email]
+      - name: contact_email
+        type: Utf8
+        nullable: true
+        description: Public contact email address for the shop.
+        expr:
+          kind: path
+          path: [contactEmail]
+      - name: currency_code
+        type: Utf8
+        nullable: true
+        description: Shop currency code.
+        expr:
+          kind: path
+          path: [currencyCode]
+      - name: iana_timezone
+        type: Utf8
+        nullable: true
+        description: Shop IANA timezone.
+        expr:
+          kind: path
+          path: [ianaTimezone]
+      - name: timezone_abbreviation
+        type: Utf8
+        nullable: true
+        description: Shop timezone abbreviation.
+        expr:
+          kind: path
+          path: [timezoneAbbreviation]
+      - name: taxes_included
+        type: Boolean
+        nullable: true
+        description: Whether product prices include taxes.
+        expr:
+          kind: path
+          path: [taxesIncluded]
+      - name: tax_shipping
+        type: Boolean
+        nullable: true
+        description: Whether the shop charges taxes on shipping.
+        expr:
+          kind: path
+          path: [taxShipping]
+      - name: created_at
+        type: Utf8
+        nullable: true
+        description: Shop creation timestamp from Shopify.
+        expr:
+          kind: path
+          path: [createdAt]
+      - name: updated_at
+        type: Utf8
+        nullable: true
+        description: Shop last updated timestamp from Shopify.
+        expr:
+          kind: path
+          path: [updatedAt]
+      - name: url
+        type: Utf8
+        nullable: true
+        description: Shop online store URL.
+        expr:
+          kind: path
+          path: [url]
+      - name: weight_unit
+        type: Utf8
+        nullable: true
+        description: Shop primary weight unit.
+        expr:
+          kind: path
+          path: [weightUnit]
+      - name: primary_domain__id
+        type: Utf8
+        nullable: true
+        description: Primary domain ID.
+        expr:
+          kind: path
+          path: [primaryDomain, id]
+      - name: primary_domain__host
+        type: Utf8
+        nullable: true
+        description: Primary domain host.
+        expr:
+          kind: path
+          path: [primaryDomain, host]
+      - name: primary_domain__url
+        type: Utf8
+        nullable: true
+        description: Primary domain URL.
+        expr:
+          kind: path
+          path: [primaryDomain, url]
+      - name: primary_domain__ssl_enabled
+        type: Boolean
+        nullable: true
+        description: Whether SSL is enabled on the primary domain.
+        expr:
+          kind: path
+          path: [primaryDomain, sslEnabled]
+      - name: plan__display_name
+        type: Utf8
+        nullable: true
+        description: Shopify plan display name.
+        expr:
+          kind: path
+          path: [plan, displayName]
+      - name: plan__partner_development
+        type: Boolean
+        nullable: true
+        description: Whether this is a partner development shop.
+        expr:
+          kind: path
+          path: [plan, partnerDevelopment]
+      - name: plan__shopify_plus
+        type: Boolean
+        nullable: true
+        description: Whether the shop is on Shopify Plus.
+        expr:
+          kind: path
+          path: [plan, shopifyPlus]
+
+  - name: products
+    description: Products in the Shopify catalog.
+    guide: |
+      Requires `read_products`. Use the optional `query` filter with Shopify
+      API search syntax, for example `status:active`, `vendor:Snowdevil`, or
+      `updated_at:>2026-01-01`. Join `id` to `product_variants.product_id`.
+    fetch_limit_default: 100
+    filters:
+      - name: query
+        required: false
+        mode: search
+    request:
+      method: POST
+      path: /graphql.json
+      body:
+        - path: [query]
+          from: literal
+          value: |
+            query ShopifyProducts($first: Int!, $after: String, $query: String) {
+              products(first: $first, after: $after, query: $query) {
+                nodes {
+                  id
+                  title
+                  handle
+                  status
+                  vendor
+                  productType
+                  tags
+                  totalInventory
+                  hasOnlyDefaultVariant
+                  createdAt
+                  updatedAt
+                  publishedAt
+                  priceRangeV2 {
+                    minVariantPrice {
+                      amount
+                      currencyCode
+                    }
+                    maxVariantPrice {
+                      amount
+                      currencyCode
+                    }
+                  }
+                }
+                pageInfo {
+                  endCursor
+                  hasNextPage
+                }
+              }
+            }
+        - path: [variables, first]
+          from: literal
+          value: 100
+        - path: [variables, query]
+          from: filter
+          key: query
+    response:
+      rows_path: [data, products, nodes]
+      row_strategy: direct
+    pagination:
+      mode: cursor_body
+      cursor_body_path: [variables, after]
+      response_cursor_path: [data, products, pageInfo, endCursor]
+      page_size:
+        default: 100
+        max: 250
+        body_path: [variables, first]
+    columns:
+      - name: query
+        type: Utf8
+        nullable: true
+        virtual: true
+        description: Shopify API search query filter.
+        expr:
+          kind: from_filter
+          key: query
+      - name: id
+        type: Utf8
+        nullable: false
+        description: Globally unique Shopify product ID.
+        expr:
+          kind: path
+          path: [id]
+      - name: title
+        type: Utf8
+        nullable: false
+        description: Product title.
+        expr:
+          kind: path
+          path: [title]
+      - name: handle
+        type: Utf8
+        nullable: true
+        description: Product handle.
+        expr:
+          kind: path
+          path: [handle]
+      - name: status
+        type: Utf8
+        nullable: true
+        description: Product status such as ACTIVE, DRAFT, or ARCHIVED.
+        expr:
+          kind: path
+          path: [status]
+      - name: vendor
+        type: Utf8
+        nullable: true
+        description: Product vendor.
+        expr:
+          kind: path
+          path: [vendor]
+      - name: product_type
+        type: Utf8
+        nullable: true
+        description: Merchant-defined product type.
+        expr:
+          kind: path
+          path: [productType]
+      - name: tags
+        type: Json
+        nullable: true
+        description: Product tags.
+        expr:
+          kind: path
+          path: [tags]
+      - name: total_inventory
+        type: Int64
+        nullable: true
+        description: Total inventory across variants.
+        expr:
+          kind: path
+          path: [totalInventory]
+      - name: has_only_default_variant
+        type: Boolean
+        nullable: true
+        description: Whether the product has only the default variant.
+        expr:
+          kind: path
+          path: [hasOnlyDefaultVariant]
+      - name: created_at
+        type: Utf8
+        nullable: true
+        description: Product creation timestamp from Shopify.
+        expr:
+          kind: path
+          path: [createdAt]
+      - name: updated_at
+        type: Utf8
+        nullable: true
+        description: Product last updated timestamp from Shopify.
+        expr:
+          kind: path
+          path: [updatedAt]
+      - name: published_at
+        type: Utf8
+        nullable: true
+        description: Product publication timestamp, when present.
+        expr:
+          kind: path
+          path: [publishedAt]
+      - name: min_variant_price_amount
+        type: Utf8
+        nullable: true
+        description: Minimum variant price amount in the product price range.
+        expr:
+          kind: path
+          path: [priceRangeV2, minVariantPrice, amount]
+      - name: min_variant_price_currency_code
+        type: Utf8
+        nullable: true
+        description: Currency code for the minimum variant price.
+        expr:
+          kind: path
+          path: [priceRangeV2, minVariantPrice, currencyCode]
+      - name: max_variant_price_amount
+        type: Utf8
+        nullable: true
+        description: Maximum variant price amount in the product price range.
+        expr:
+          kind: path
+          path: [priceRangeV2, maxVariantPrice, amount]
+      - name: max_variant_price_currency_code
+        type: Utf8
+        nullable: true
+        description: Currency code for the maximum variant price.
+        expr:
+          kind: path
+          path: [priceRangeV2, maxVariantPrice, currencyCode]
+
+  - name: product_variants
+    description: Product variants and their SKU, price, and inventory item metadata.
+    guide: |
+      Requires `read_products`. Use the optional `query` filter with Shopify
+      API search syntax, for example `sku:ABC-123`, `barcode:ABC-abc-1234`,
+      `product_id:123`, or `collection_id:123`.
+    fetch_limit_default: 100
+    filters:
+      - name: query
+        required: false
+        mode: search
+    request:
+      method: POST
+      path: /graphql.json
+      body:
+        - path: [query]
+          from: literal
+          value: |
+            query ShopifyProductVariants($first: Int!, $after: String, $query: String) {
+              productVariants(first: $first, after: $after, query: $query) {
+                nodes {
+                  id
+                  title
+                  displayName
+                  sku
+                  barcode
+                  price
+                  compareAtPrice
+                  availableForSale
+                  inventoryQuantity
+                  inventoryPolicy
+                  selectedOptions {
+                    name
+                    value
+                  }
+                  createdAt
+                  updatedAt
+                  product {
+                    id
+                    title
+                    handle
+                  }
+                  inventoryItem {
+                    id
+                    sku
+                    tracked
+                    requiresShipping
+                  }
+                }
+                pageInfo {
+                  endCursor
+                  hasNextPage
+                }
+              }
+            }
+        - path: [variables, first]
+          from: literal
+          value: 100
+        - path: [variables, query]
+          from: filter
+          key: query
+    response:
+      rows_path: [data, productVariants, nodes]
+      row_strategy: direct
+    pagination:
+      mode: cursor_body
+      cursor_body_path: [variables, after]
+      response_cursor_path: [data, productVariants, pageInfo, endCursor]
+      page_size:
+        default: 100
+        max: 250
+        body_path: [variables, first]
+    columns:
+      - name: query
+        type: Utf8
+        nullable: true
+        virtual: true
+        description: Shopify API search query filter.
+        expr:
+          kind: from_filter
+          key: query
+      - name: id
+        type: Utf8
+        nullable: false
+        description: Globally unique Shopify product variant ID.
+        expr:
+          kind: path
+          path: [id]
+      - name: product_id
+        type: Utf8
+        nullable: true
+        description: Parent product ID.
+        expr:
+          kind: path
+          path: [product, id]
+      - name: product_title
+        type: Utf8
+        nullable: true
+        description: Parent product title.
+        expr:
+          kind: path
+          path: [product, title]
+      - name: product_handle
+        type: Utf8
+        nullable: true
+        description: Parent product handle.
+        expr:
+          kind: path
+          path: [product, handle]
+      - name: title
+        type: Utf8
+        nullable: false
+        description: Variant title.
+        expr:
+          kind: path
+          path: [title]
+      - name: display_name
+        type: Utf8
+        nullable: true
+        description: Display name combining product and variant context.
+        expr:
+          kind: path
+          path: [displayName]
+      - name: sku
+        type: Utf8
+        nullable: true
+        description: Variant SKU.
+        expr:
+          kind: path
+          path: [sku]
+      - name: barcode
+        type: Utf8
+        nullable: true
+        description: Variant barcode.
+        expr:
+          kind: path
+          path: [barcode]
+      - name: price
+        type: Utf8
+        nullable: true
+        description: Variant price in the shop currency.
+        expr:
+          kind: path
+          path: [price]
+      - name: compare_at_price
+        type: Utf8
+        nullable: true
+        description: Variant compare-at price, when present.
+        expr:
+          kind: path
+          path: [compareAtPrice]
+      - name: available_for_sale
+        type: Boolean
+        nullable: true
+        description: Whether the variant is available for sale.
+        expr:
+          kind: path
+          path: [availableForSale]
+      - name: inventory_quantity
+        type: Int64
+        nullable: true
+        description: Total sellable quantity of the variant.
+        expr:
+          kind: path
+          path: [inventoryQuantity]
+      - name: inventory_policy
+        type: Utf8
+        nullable: true
+        description: Inventory policy for out-of-stock ordering.
+        expr:
+          kind: path
+          path: [inventoryPolicy]
+      - name: selected_options
+        type: Json
+        nullable: true
+        description: Selected option name/value pairs for the variant.
+        expr:
+          kind: path
+          path: [selectedOptions]
+      - name: inventory_item_id
+        type: Utf8
+        nullable: true
+        description: Inventory item ID connected to this variant.
+        expr:
+          kind: path
+          path: [inventoryItem, id]
+      - name: inventory_item_sku
+        type: Utf8
+        nullable: true
+        description: Inventory item SKU.
+        expr:
+          kind: path
+          path: [inventoryItem, sku]
+      - name: inventory_item_tracked
+        type: Boolean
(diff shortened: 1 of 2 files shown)


sources/community/shopify/manifest.yaml
```diff
@@ -0,0 +1,1693 @@
...
+          from: filter
+          key: query
+    response:
+      rows_path: [data, products, nodes]
+      row_strategy: direct
+    pagination:
+      mode: cursor_body
+      cursor_body_path: [variables, after]
+      response_cursor_path: [data, products, pageInfo, endCursor]
+      page_size:
```
antonmry, line 312: Just flagging for awareness, not a blocker: this `cursor_body` setup will issue one extra request at the end of each pagination sweep.

Relay-style GraphQL connections (which Shopify uses) set `pageInfo.endCursor` on every page that has at least one edge, regardless of `pageInfo.hasNextPage`. Coral's runtime (`crates/coral-engine/src/backends/http/client.rs:336–348`) stops only when `response_cursor_path` resolves to null or an empty string. So on the last non-empty page, the runtime will:

1. Read the (non-null) `endCursor`.
2. Issue one more `after: <endCursor>` request.
3. Get back an empty `nodes` array (no edges → no `endCursor`).
4. Stop.

Same behavior on `product_variants`, `collections`, `orders`, `customers`, `locations`, and `inventory_items` since they all use the same shape.

Not a correctness bug — just one wasted round-trip per pagination traversal, which on Shopify's cost-rate-limited API isn't free. The cleanest fix would be a DSL feature to stop when a separate "has_more" path is false (`hasNextPage` in this case), but that's a Coral-side change rather than something to fix here. No action needed for this PR.

sources/community/shopify/manifest.yaml
```diff
@@ -0,0 +1,1693 @@
...
+          kind: path
+          path: [currentTotalPriceSet, shopMoney, amount]
+      - name: current_total_price_currency_code
+        type: Utf8
+        nullable: true
+        description: Current total order price currency code.
+        expr:
+          kind: path
+          path: [currentTotalPriceSet, shopMoney, currencyCode]
+      - name: current_subtotal_price_amount
```
antonmry: Tiny ergonomics note (and the same applies to `current_subtotal_price_amount`, `current_total_tax_amount`, `customers.amount_spent`, and the `min/max_variant_price` columns on `products`): Shopify returns money `amount` fields as decimal **strings** (e.g. `"12.34"`) for precision reasons, so keeping them as `Utf8` faithfully reflects the API.

But the most natural query shape on order/customer money columns is aggregation — `SELECT SUM(current_total_price_amount) FROM orders WHERE ...` — which requires a `CAST(... AS DOUBLE)` on every column today. Two options worth considering:

1. Add a brief mention in the column descriptions (e.g. "decimal string — CAST to DOUBLE for arithmetic") so users aren't surprised.
2. Expose them as `Float64` directly via path expression. You lose the decimal-precision guarantee, but for analytics queries that's usually the right tradeoff.

Hacker News does something similar with `created_at_i` (Int64 epoch) alongside `created_at` (Timestamp). The dual-column pattern works well — both raw + ergonomic forms exposed.

sources/community/shopify/manifest.yaml
```diff
@@ -0,0 +1,1693 @@
...
+        expr:
+          kind: path
+          path: [taxShipping]
+      - name: created_at
+        type: Utf8
+        nullable: true
+        description: Shop creation timestamp from Shopify.
+        expr:
+          kind: path
+          path: [createdAt]
```
antonmry: Worth considering for v1 if you have the bandwidth — date columns are likely the single biggest ergonomics win available here.

Shopify's GraphQL `DateTime` scalar is documented as ISO 8601. The manifest currently exposes `created_at`, `updated_at`, `processed_at`, `cancelled_at`, and `publishedAt` across all tables as `Utf8`. That means a natural query like:

```sql
SELECT id, name FROM shopify.orders
WHERE created_at > NOW() - INTERVAL '7 days'
```

won't work without `CAST(created_at AS TIMESTAMP)` or `TO_TIMESTAMP(...)` on every reference. Converting them to real `Timestamp` columns is mechanical:

```yaml
- name: created_at
  type: Timestamp
  nullable: true
  description: Order creation timestamp.
  expr:
    kind: format_timestamp
    input: iso8601
    expr:
      kind: path
      path: [createdAt]
```

Date-range filtering on orders/customers is probably the single most common query shape against this source, so the ergonomics impact is high. The PR description already lists "Date/time columns are exposed as Utf8 in v1" as a known limitation, so this is consistent with the author's plan — just flagging in case it's an easy v1 lift.

sources/community/shopify/manifest.yaml
```diff
@@ -0,0 +1,1693 @@
...
+    required: false
+    hint: |
+      Shopify Admin GraphQL API version. Defaults to `2026-04`, the latest
+      stable version checked for this source. Override only when you have
+      validated the source against another supported Admin API version.
+  SHOPIFY_ADMIN_ACCESS_TOKEN:
+    kind: secret
+    required: true
+    hint: |
+      Shopify Admin API access token from a custom or public app. The token is
```
antonmry, line 24: Tiny FYI, not a blocker — `required: true` and `required: false` on input declarations are silently ignored by the DSL today. Per `crates/coral-spec/src/inputs.rs:126–131`, the `required` flag is **derived** from whether a `default:` is set, not read directly from the manifest:

```rust
ordered.push(ManifestInputSpec {
    key: key.clone(),
    kind,
    required: default_value.is_none(),
    ...
});
```

So `SHOPIFY_SHOP_DOMAIN` (no default) is already required, `SHOPIFY_API_VERSION` (with default `2026-04`) is already optional, and the explicit `required:` lines are no-ops. Doesn't hurt anything, but you can remove them for clarity if you'd like. Same goes for the `required: false` on the filter declarations further down — `false` is the default.

sources/community/shopify/manifest.yaml
```diff
@@ -0,0 +1,1693 @@
...
+  - name: products
+    description: Products in the Shopify catalog.
+    guide: |
+      Requires `read_products`. Use the optional `query` filter with Shopify
+      API search syntax, for example `status:active`, `vendor:Snowdevil`, or
+      `updated_at:>2026-01-01`. Join `id` to `product_variants.product_id`.
+    fetch_limit_default: 100
+    filters:
+      - name: query
+        required: false
```
antonmry: Trivial nit — same applies to the `query` and `include_inactive`/`include_legacy` filters on `product_variants`, `collections`, `orders`, `customers`, `locations`, and `inventory_items`: `required: false` is the default for `FilterSpec`, so the explicit lines are redundant. Looking at `crates/coral-spec/src/common.rs:140–147`:

```rust
pub struct FilterSpec {
    pub name: String,
    #[serde(default)]
    pub required: bool,  // defaults to false
    #[serde(default)]
    pub mode: FilterMode,
}
```

Feel free to remove them across all the tables to keep things tight — but it's purely cosmetic.

sources/community/shopify/manifest.yaml
```diff
@@ -0,0 +1,1693 @@
...
+      - name: plan__shopify_plus
+        type: Boolean
+        nullable: true
+        description: Whether the shop is on Shopify Plus.
+        expr:
+          kind: path
+          path: [plan, shopifyPlus]
+
+  - name: products
+    description: Products in the Shopify catalog.
```
antonmry, line 251: Style nit, please ignore: this manifest uses the legacy array form for request bodies:

```yaml
body:
  - path: [query]
    from: literal
    value: |
      query ShopifyProducts(...) { ... }
```

The DSL also supports a newer tagged form (`body: {format: json, fields: [...]}`) — see `sources/community/milvus/manifest.yaml` for examples. Both parse equivalently to a JSON body, so there's no functional difference. Just flagging since reviewers comparing the two recent manifests will notice the inconsistency. Up to you whether to switch.

sources/community/shopify/manifest.yaml
```diff
@@ -0,0 +1,1693 @@
...
+    hint: |
+      Shopify shop domain for the Admin GraphQL API, for example
+      `example.myshopify.com`. Do not include `https://` or a trailing slash.
+  SHOPIFY_API_VERSION:
+    kind: variable
+    default: "2026-04"
+    required: false
+    hint: |
+      Shopify Admin GraphQL API version. Defaults to `2026-04`, the latest
+      stable version checked for this source. Override only when you have
```
antonmry, line 19: Minor maintenance note: Shopify deprecates Admin GraphQL versions on roughly a yearly cadence (each version is supported for ~12 months). Pinning to `2026-04` is the right v1 call — but it would help future maintainers if the README (or a comment here) called out:

1. How often this needs to be bumped (~1× per year),
2. Where to check current support windows (`https://shopify.dev/docs/api/usage/versioning#release-schedule`),
3. What user-visible churn to expect when bumping (mostly enum/field rename diffs in the GraphQL queries).

Not a blocker — `SHOPIFY_API_VERSION` already lets users override on their own clusters. Just thinking about the next contributor who picks this up.

sources/community/shopify/manifest.yaml
```diff
@@ -0,0 +1,1693 @@
...
+        description: Whether the shop is on Shopify Plus.
+        expr:
+          kind: path
+          path: [plan, shopifyPlus]
+
+  - name: products
+    description: Products in the Shopify catalog.
+    guide: |
+      Requires `read_products`. Use the optional `query` filter with Shopify
+      API search syntax, for example `status:active`, `vendor:Snowdevil`, or
```
antonmry, line 254: v2 enhancement idea, please ignore for this PR: Shopify's Admin GraphQL also exposes singleton lookups `product(id: ID!)`, `customer(id: ID!)`, `order(id: ID!)` etc. that fetch a single resource by global ID without going through the paginated connection. Today, "look up order #123" requires `WHERE query = 'name:#123'` via the search filter, which works but is awkward when you already know the global ID.

A v2 enhancement could add companion tables (`shopify.product`, `shopify.customer`, `shopify.order`) keyed on a required `id` filter, each issuing a single non-paginated GraphQL query. Useful for joins from external sources (e.g. an order ID coming out of Stripe). Not a blocker.

sources/community/shopify/manifest.yaml
Yashagarwal9798: Thanks for the detailed explanation. That makes sense, and I agree this looks like a Coral-side pagination feature rather than something to work around in the Shopify manifest.

Would it be useful if I opened a follow-up issue for adding an optional pagination.has_more_path field, so cursor_query / cursor_body sources can stop when a boolean path like pageInfo.hasNextPage is false? I can take a look at implementing it separately after this source PR, with docs and tests, if that direction sounds right.

sources/community/shopify/manifest.yaml
Yashagarwal9798: Thanks, updated this to expose both forms: exact Shopify decimal strings are now available as *_decimal columns, while the natural money amount columns are Float64 for aggregation/query ergonomics.

sources/community/shopify/manifest.yaml
Yashagarwal9798: Thanks, cleaned this up by removing the redundant required fields from inputs and filters. The required/optional behavior now relies on the DSL defaults/default values.

sources/community/shopify/manifest.yaml
Yashagarwal9798: Thanks, I’ll leave this one as-is for now since both body forms are equivalent, but I’ll use the newer tagged body form for future source manifests.

sources/community/shopify/manifest.yaml
Yashagarwal9798: Thanks, added a README note covering the Shopify API version support window, where to check the release schedule, and the kind of GraphQL query churn to expect when bumping the default version.

sources/community/shopify/manifest.yaml
Yashagarwal9798: Thanks, agreed. I’ll leave singleton ID lookup tables out of v1 and treat them as a follow-up enhancement for direct joins from external sources.

sources/community/shopify/manifest.yaml
Yashagarwal9798: Thanks, updated the Shopify DateTime fields to real Timestamp columns using format_timestamp with iso8601 input. I also live-validated a date-range predicate against shopify.collections successfully.

sources/community/shopify/manifest.yaml
```diff
@@ -0,0 +1,1791 @@
...
+        type: Utf8
+        nullable: true
+        description: Variant price as Shopify's decimal string.
+        expr:
+          kind: path
+          path: [price]
+      - name: price
+        type: Float64
+        nullable: true
+        description: Variant price as Float64 for arithmetic.
```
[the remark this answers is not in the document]
Yashagarwal9798: This is intentional for query ergonomics: the exact Shopify decimal strings are still preserved in the matching `*_decimal` columns, while the `Float64` columns are companion analytics columns for aggregation. Users who need exact financial/accounting precision can use the `*_decimal` fields.

antonmry (approved):
Thanks for the thorough turnaround — every comment got a concrete response, and the dual-column money pattern (`*_decimal` Utf8 + bare Float64) is genuinely elegant. Same shape as `sources/community/hn`'s `created_at_i` / `created_at` pair, just applied to money instead of time. Worth promoting that pattern across other community sources where vendors return precision-sensitive scalars in unergonomic formats.

Verified at HEAD (`26b9d32c`):

- **DateTime → Timestamp** (`L166–182`, `L388–414`, `L681–697`, `L833+`): all switched to `format_timestamp input: iso8601`, with live date-range validation against `shopify.collections`. ✅
- **Money dual-column** (`L418–445` and parallels on `customers.amount_spent`, `orders.current_*_amount`, `product_variants.price`/`compare_at_price`): `*_decimal` preserves Shopify's exact decimal string, bare column exposes `Float64` for analytics. ✅
- **`required:` cleanup**: removed from both inputs and filters since the DSL derives required-ness from `default` presence. ✅
- **API version maintenance note**: added to the README. ✅

Items 1 (Relay pagination overshoot), 6 (legacy body form), and 8 (singleton-lookup tables) are all reasonably deferred — they're DSL gaps or v2 enhancements, not v1 blockers.

Approving. I'll file the Coral-side issue for the `hasNextPage` / `has_more` cursor stop signal so this pattern gets tracked beyond the Shopify manifest.

PR #533, opened 2026-05-16
feat(sources/community/sendgrid): add SendGrid community source

### LINKED ISSUE #532: design: add SendGrid community source
## Summary

Add a community source for [SendGrid](https://sendgrid.com) (Twilio) — an email delivery and marketing platform. This source enables Coral users to query SendGrid account data (API keys, teammates, sender identities, templates, suppression lists, marketing lists, and single sends) through SQL using the [SendGrid Web API v3](https://docs.sendgrid.com/api-reference).

## Motivation

SendGrid is widely used for transactional and marketing email. Teams managing SendGrid accounts currently need to navigate the dashboard or write custom API scripts to audit key management, review suppression lists, and manage sender configurations. A Coral source would let users:

- **Audit API keys** — list all keys and their names for rotation/cleanup
- **Review team access** — audit teammates, admin privileges, and user types
- **Manage sender identities** — verified senders, marketing senders, from/reply-to addresses, verification status
- **Inspect templates** — list transactional templates (legacy and dynamic), generation type, versions
- **Monitor deliverability** — query bounces, blocks, spam reports, and invalid emails with reasons and timestamps
- **Audit suppressions** — review suppression (unsubscribe) groups, default status, and unsubscribe counts
- **Review marketing** — list contact lists with counts, single sends with status and scheduling

### Inputs

| Input | Kind | Description |
|---|---|---|
| `SENDGRID_API_KEY` | secret | SendGrid API key (starts with `SG.`) |

### Auth

- `HeaderAuth` with `Authorization: Bearer {{input.SENDGRID_API_KEY}}`

### Proposed v1 tables

| Table | Endpoint | Pagination | Columns | Notes |
|---|---|---|---|---|
| `api_keys` | `GET /api_keys` | none | 2 | Key ID and name |
| `teammates` | `GET /teammates` | offset (max 500) | 7 | Username, email, admin, scopes |
| `verified_senders` | `GET /verified_senders` | none | 11 | Email, name, address, verified, locked |
| `templates` | `GET /templates` | none | 5 | Both legacy+dynamic, versions as JSON |
| `suppression_groups` | `GET /asm/groups` | none | 5 | Unsubscribe groups with counts |
| `bounces` | `GET /suppression/bounces` | offset (max 500) | 4 | Email, reason, SMTP status, timestamp |
| `blocks` | `GET /suppression/blocks` | offset (max 500) | 4 | Email, reason, status, timestamp |
| `spam_reports` | `GET /suppression/spam_reports` | offset (max 500) | 3 | Email, IP, timestamp |
| `invalid_emails` | `GET /suppression/invalid_emails` | offset (max 500) | 3 | Email, reason, timestamp |
| `marketing_lists` | `GET /marketing/lists` | none | 3 | List name and contact count |
| `marketing_senders` | `GET /marketing/senders` | none | 13 | From/reply-to, address, verified, locked |
| `singlesends` | `GET /marketing/singlesends` | none | 8 | Name, status, send_at, categories, A/B test |

### Design details

- **Mixed response shapes** — SendGrid endpoints return data in different wrappers: `result` array (api_keys, teammates, templates, lists, singlesends), `results` array (verified_senders, marketing_senders), or top-level JSON array (suppression groups, bounces, blocks, spam_reports, invalid_emails)
- **Offset pagination** — suppression endpoints and teammates use `limit` + `offset` query params
- **Page-token pagination not followed** — templates, marketing_lists, and singlesends support page_token pagination that Coral can't fully follow yet; the source requests the largest documented first page
- **Unix timestamps** — suppression tables use Unix epoch integers for `created`, not ISO 8601
- **Nested from/reply_to** — marketing_senders flattens nested `from.email`, `from.name`, `reply_to.email`, `reply_to.name` into top-level columns

### Affected surface

- Community source only
- Expected files: `sources/community/sendgrid/manifest.yaml`, `sources/community/sendgrid/README.md`
- No CLI, MCP, config, or runtime changes

## Drawbacks and trade-offs

- **Read-only** — no create, update, or delete operations
- **No global stats** — `GET /stats` requires date range parameters and returns time-series data that doesn't fit well into a simple table
- **No contacts export** — contact data requires async export jobs, not simple GET
- **Page-token tables may miss data** — templates, marketing_lists, and singlesends return only the first page; accounts with >200 templates or >1000 lists won't see all data
- **EU base URL** — EU-hosted accounts need to manually change the base URL to `https://api.eu.sendgrid.com/v3`

## Open questions

- Should page-token pagination support be added to Coral's engine to fully paginate templates/lists/singlesends?
- Should stats tables be added with required date-range filters?

---

Arnav1709:
## Summary

Adds a new community source for Twilio SendGrid under `sources/community/sendgrid/`.

This source enables querying SendGrid account data (API keys, teammates, sender identities, templates, suppression lists, marketing resources) through SQL using the SendGrid Web API v3. The source is manifest-only, read-only, and uses Coral's existing HTTP source-spec model with `HeaderAuth`. No CLI, MCP, config, or runtime changes.

Closes #532 

## What changed

Added:

- `sources/community/sendgrid/manifest.yaml` (859 lines)
- `sources/community/sendgrid/README.md` (291 lines)

## Tables

| Table | Endpoint | Pagination | Columns |
|---|---|---|---|
| `api_keys` | `GET /api_keys` | none | 2 |
| `teammates` | `GET /teammates` | offset (max 500) | 7 |
| `verified_senders` | `GET /verified_senders` | none | 11 |
| `templates` | `GET /templates` | none | 5 |
| `suppression_groups` | `GET /asm/groups` | none | 5 |
| `bounces` | `GET /suppression/bounces` | offset (max 500) | 4 |
| `blocks` | `GET /suppression/blocks` | offset (max 500) | 4 |
| `spam_reports` | `GET /suppression/spam_reports` | offset (max 500) | 3 |
| `invalid_emails` | `GET /suppression/invalid_emails` | offset (max 500) | 3 |
| `marketing_lists` | `GET /marketing/lists` | none | 3 |
| `marketing_senders` | `GET /marketing/senders` | none | 13 |
| `singlesends` | `GET /marketing/singlesends` | none | 8 |

## Auth

Bearer token authentication with `Authorization: Bearer <SG.key>`.

Input:

- `SENDGRID_API_KEY` — required secret. SendGrid API key from Settings > API Keys.

## Implementation notes

- **Mixed response shapes** — endpoints return data as `result` array (api_keys, teammates, templates, lists, singlesends), `results` array (verified_senders, marketing_senders), or top-level JSON array via `row_strategy: direct` (suppression groups, bounces, blocks, spam_reports, invalid_emails)
- **Offset pagination** — teammates and suppression endpoints use `limit` + `offset`; other endpoints return full results in a single request
- **Page-token limitation** — templates, marketing_lists, and singlesends support page_token-based pagination that Coral can't follow; source requests largest documented first page where possible
- **Static query param** — templates sends `generations=legacy,dynamic` to return both template types
- **Marketing senders** — uses `/marketing/senders` (not `/senders`), flattens nested `from.email`/`from.name` and `reply_to.email`/`reply_to.name` into separate columns
- **Unix timestamps** — suppression tables use integer epoch timestamps for `created`; marketing_senders uses integer `created_at`/`updated_at`
- **EU support** — EU-hosted accounts need to change base_url to `https://api.eu.sendgrid.com/v3`

## Validation

- `coral source lint sources/community/sendgrid/manifest.yaml` — passes
- `coral source add` with real key — 12 tables exposed, test query passes
- **Live query validation — all 12 tables queried successfully using real API key:**
- **Mock server validation** — docs-shaped responses verified for all 12 tables including nested from/reply_to flattening and template versions JSON

## User-visible impact

New `sendgrid` source installable via:

```bash
export SENDGRID_API_KEY="SG.your_key_here"
coral source add --file sources/community/sendgrid/manifest.yaml
```

Example queries:

```sql
-- Audit API keys
SELECT api_key_id, name FROM sendgrid.api_keys;

-- Review team access
SELECT username, email, is_admin, user_type
FROM sendgrid.teammates;

-- Check bounces by status code
SELECT email, reason, status
FROM sendgrid.bounces
WHERE status LIKE '5.%'
LIMIT 20;

-- Review suppression group stats
SELECT name, description, unsubscribes, is_default
FROM sendgrid.suppression_groups;

-- List marketing campaigns
SELECT name, status, send_at, is_abtest
FROM sendgrid.singlesends;

-- Audit sender verification
SELECT nickname, from_email, from_name, verified
FROM sendgrid.marketing_senders;
```

## Known limitations

- **Read-only** — no create, update, or delete operations
- **No global stats** — would require date-range filters and time-series modeling
- **No contacts** — contact export requires async jobs
- **Page-token tables** — templates (>200), marketing_lists (>1000), singlesends may not return all data on large accounts
- **EU accounts** — must manually update base_url in manifest

### DIFF
--- a/sources/community/sendgrid/manifest.yaml
+++ b/sources/community/sendgrid/manifest.yaml
@@ -7,14 +7,21 @@ description: >-
   reports, invalid emails), marketing lists, single sends,
   suppression groups, sender identities, verified senders,
   API keys, and teammates from Twilio SendGrid.
-base_url: https://api.sendgrid.com/v3
 
 inputs:
+  SENDGRID_BASE_URL:
+    kind: variable
+    default: https://api.sendgrid.com/v3
+    hint: >-
+      Base URL for the SendGrid API. The default works for US-hosted
+      accounts. For EU-hosted accounts use https://api.eu.sendgrid.com/v3.
   SENDGRID_API_KEY:
     kind: secret
     hint: >-
       SendGrid API key. Create one at https://app.sendgrid.com/settings/api_keys
       with at least read-only access to the resources you want to query.
+
+base_url: "{{input.SENDGRID_BASE_URL}}"
 
 auth:
   type: HeaderAuth
@@ -25,6 +32,8 @@ auth:
 
 test_queries:
   - SELECT api_key_id, name FROM sendgrid.api_keys LIMIT 1
+  - SELECT id, nickname FROM sendgrid.verified_senders LIMIT 1
+  - SELECT email FROM sendgrid.bounces LIMIT 1
 
 tables:
   # ──────────────────────────────────────────────────────────────────────
@@ -36,6 +45,8 @@ tables:
       and creation/last-used metadata.
     guide: >
       `SELECT api_key_id, name FROM sendgrid.api_keys`.
+      SendGrid documents `limit` for this endpoint, but no offset or cursor;
+      this table requests the first 100 API keys.
     request:
       method: GET
       path: /api_keys
@@ -266,6 +277,9 @@ tables:
     guide: >
       `SELECT id, name, generation, updated_at
       FROM sendgrid.templates LIMIT 20`.
+      This endpoint returns `_metadata.next` as a full URL, which Coral
+      cannot follow as a cursor yet, so the table requests the largest
+      documented first page.
     request:
       method: GET
       path: /templates
@@ -309,7 +323,7 @@ tables:
       - name: updated_at
         type: Utf8
         nullable: true
-        description: Last update timestamp.
+        description: Last update timestamp in `YYYY-MM-DD HH:MM:SS` format.
         expr:
           kind: path
           path:
@@ -392,9 +406,27 @@ tables:
     guide: >
       `SELECT email, reason, status, created
       FROM sendgrid.bounces LIMIT 50`.
+      Optional filters: `start_time`, `end_time`, and `email`.
+    filters:
+      - name: start_time
+        required: false
+      - name: end_time
+        required: false
+      - name: email
+        required: false
     request:
       method: GET
       path: /suppression/bounces
+      query:
+        - name: start_time
+          from: filter
+          key: start_time
+        - name: end_time
+          from: filter
+          key: end_time
+        - name: email
+          from: filter
+          key: email
     response:
       row_strategy: direct
     pagination:
@@ -406,6 +438,22 @@ tables:
         max: 500
         query_param: limit
     columns:
+      - name: start_time
+        type: Int64
+        nullable: true
+        virtual: true
+        description: Optional start time filter as Unix epoch seconds.
+        expr:
+          kind: from_filter
+          key: start_time
+      - name: end_time
+        type: Int64
+        nullable: true
+        virtual: true
+        description: Optional end time filter as Unix epoch seconds.
+        expr:
+          kind: from_filter
+          key: end_time
       - name: email
         type: Utf8
         nullable: false
@@ -430,7 +478,7 @@ tables:
           kind: path
           path:
             - status
-      - name: created
+      - name: created_i
         type: Int64
         nullable: true
         description: Unix timestamp when the bounce was recorded.
@@ -438,6 +486,17 @@ tables:
           kind: path
           path:
             - created
+      - name: created
+        type: Timestamp
+        nullable: true
+        description: When the bounce was recorded.
+        expr:
+          kind: format_timestamp
+          input: seconds
+          expr:
+            kind: path
+            path:
+              - created
 
   # ──────────────────────────────────────────────────────────────────────
   # blocks
@@ -449,9 +508,27 @@ tables:
     guide: >
       `SELECT email, reason, status, created
       FROM sendgrid.blocks LIMIT 50`.
+      Optional filters: `start_time`, `end_time`, and `email`.
+    filters:
+      - name: start_time
+        required: false
+      - name: end_time
+        required: false
+      - name: email
+        required: false
     request:
       method: GET
       path: /suppression/blocks
+      query:
+        - name: start_time
+          from: filter
+          key: start_time
+        - name: end_time
+          from: filter
+          key: end_time
+        - name: email
+          from: filter
+          key: email
     response:
       row_strategy: direct
     pagination:
@@ -463,6 +540,22 @@ tables:
         max: 500
         query_param: limit
     columns:
+      - name: start_time
+        type: Int64
+        nullable: true
+        virtual: true
+        description: Optional start time filter as Unix epoch seconds.
+        expr:
+          kind: from_filter
+          key: start_time
+      - name: end_time
+        type: Int64
+        nullable: true
+        virtual: true
+        description: Optional end time filter as Unix epoch seconds.
+        expr:
+          kind: from_filter
+          key: end_time
       - name: email
         type: Utf8
         nullable: false
@@ -487,7 +580,7 @@ tables:
           kind: path
           path:
             - status
-      - name: created
+      - name: created_i
         type: Int64
         nullable: true
         description: Unix timestamp when the block was recorded.
@@ -495,6 +588,17 @@ tables:
           kind: path
           path:
             - created
+      - name: created
+        type: Timestamp
+        nullable: true
+        description: When the block was recorded.
+        expr:
+          kind: format_timestamp
+          input: seconds
+          expr:
+            kind: path
+            path:
+              - created
 
   # ──────────────────────────────────────────────────────────────────────
   # spam_reports
@@ -506,9 +610,27 @@ tables:
     guide: >
       `SELECT email, ip, created
       FROM sendgrid.spam_reports LIMIT 50`.
+      Optional filters: `start_time`, `end_time`, and `email`.
+    filters:
+      - name: start_time
+        required: false
+      - name: end_time
+        required: false
+      - name: email
+        required: false
     request:
       method: GET
       path: /suppression/spam_reports
+      query:
+        - name: start_time
+          from: filter
+          key: start_time
+        - name: end_time
+          from: filter
+          key: end_time
+        - name: email
+          from: filter
+          key: email
     response:
       row_strategy: direct
     pagination:
@@ -520,6 +642,22 @@ tables:
         max: 500
         query_param: limit
     columns:
+      - name: start_time
+        type: Int64
+        nullable: true
+        virtual: true
+        description: Optional start time filter as Unix epoch seconds.
+        expr:
+          kind: from_filter
+          key: start_time
+      - name: end_time
+        type: Int64
+        nullable: true
+        virtual: true
+        description: Optional end time filter as Unix epoch seconds.
+        expr:
+          kind: from_filter
+          key: end_time
       - name: email
         type: Utf8
         nullable: false
@@ -536,7 +674,7 @@ tables:
           kind: path
           path:
             - ip
-      - name: created
+      - name: created_i
         type: Int64
         nullable: true
         description: Unix timestamp when the spam report was recorded.
@@ -544,6 +682,17 @@ tables:
           kind: path
           path:
             - created
+      - name: created
+        type: Timestamp
+        nullable: true
+        description: When the spam report was recorded.
+        expr:
+          kind: format_timestamp
+          input: seconds
+          expr:
+            kind: path
+            path:
+              - created
 
   # ──────────────────────────────────────────────────────────────────────
   # invalid_emails
@@ -555,9 +704,27 @@ tables:
     guide: >
       `SELECT email, reason, created
       FROM sendgrid.invalid_emails LIMIT 50`.
+      Optional filters: `start_time`, `end_time`, and `email`.
+    filters:
+      - name: start_time
+        required: false
+      - name: end_time
+        required: false
+      - name: email
+        required: false
     request:
       method: GET
       path: /suppression/invalid_emails
+      query:
+        - name: start_time
+          from: filter
+          key: start_time
+        - name: end_time
+          from: filter
+          key: end_time
+        - name: email
+          from: filter
+          key: email
     response:
       row_strategy: direct
     pagination:
@@ -569,6 +736,22 @@ tables:
         max: 500
         query_param: limit
     columns:
+      - name: start_time
+        type: Int64
+        nullable: true
+        virtual: true
+        description: Optional start time filter as Unix epoch seconds.
+        expr:
+          kind: from_filter
+          key: start_time
+      - name: end_time
+        type: Int64
+        nullable: true
+        virtual: true
+        description: Optional end time filter as Unix epoch seconds.
+        expr:
+          kind: from_filter
+          key: end_time
       - name: email
         type: Utf8
         nullable: false
@@ -585,7 +768,7 @@ tables:
           kind: path
           path:
             - reason
-      - name: created
+      - name: created_i
         type: Int64
         nullable: true
         description: Unix timestamp when recorded.
@@ -593,6 +776,17 @@ tables:
           kind: path
           path:
             - created
+      - name: created
+        type: Timestamp
+        nullable: true
+        description: When the invalid email was recorded.
+        expr:
+          kind: format_timestamp
+          input: seconds
+          expr:
+            kind: path
+            path:
+              - created
 
   # ──────────────────────────────────────────────────────────────────────
   # marketing_lists
@@ -604,6 +798,9 @@ tables:
     guide: >
       `SELECT id, name, contact_count
       FROM sendgrid.marketing_lists`.
+      This endpoint returns `_metadata.next` as a full URL, which Coral
+      cannot follow as a cursor yet, so the table requests the largest
+      documented first page.
     request:
       method: GET
       path: /marketing/lists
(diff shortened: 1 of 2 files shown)


sources/community/sendgrid/manifest.yaml
```diff
@@ -0,0 +1,859 @@
...
+    guide: >
+      `SELECT username, email, is_admin, user_type
+      FROM sendgrid.teammates`.
+    request:
+      method: GET
+      path: /teammates
+    response:
+      rows_path:
+        - result
+    pagination:
```
antonmry, line 96: Sorry if I've got this wrong — could you double-check this against a live run? The SendGrid teammates docs at `https://www.twilio.com/docs/sendgrid/api-reference/teammates/retrieve-all-teammates` describe the response wrapper as `results` (plural), not `result` (singular), which would make this `rows_path: [result]` extract `null` and return 0 rows on every query.

The other tables in this manifest use both forms based on the actual endpoint shape — `api_keys` and `templates` correctly use `result`, while `verified_senders` and `marketing_senders` correctly use `results`. Teammates seems to fall in the `results` bucket per the docs.

The PR description says "Live query validation — all 12 tables queried successfully" — if `teammates` really did return rows, either the docs are inaccurate or the endpoint accepts both forms, in which case ignore this comment. Otherwise the fix would just be:

```yaml
response:
  rows_path:
    - results
```

sources/community/sendgrid/manifest.yaml
```diff
@@ -0,0 +1,859 @@
...
+      page_size:
+        default: 100
+        max: 500
+        query_param: limit
+    columns:
+      - name: email
+        type: Utf8
+        nullable: false
+        description: Bounced email address.
+        expr:
```
antonmry, line 461: Optional but probably worth doing now — SendGrid's `/suppression/bounces` accepts `start_time`, `end_time`, and `email` query params for server-side filtering (per `https://www.twilio.com/docs/sendgrid/api-reference/bounces-api/retrieve-all-bounces`). The PR's headline example query —

```sql
SELECT email, reason, status FROM sendgrid.bounces WHERE status LIKE '5.%'
```

— currently downloads the full bounce list to filter `status` client-side. Adding `start_time`/`end_time` filters would make "bounces from the last 7 days" a single short request instead of a full scan, which I'd guess is the most common shape of the query.

The same three filters (`start_time`, `end_time`, `email`) apply to `/suppression/blocks`, `/suppression/spam_reports`, and `/suppression/invalid_emails`, so if you decide to add them, they go in all four tables. Up to you whether to land it now or as a follow-on.

sources/community/sendgrid/manifest.yaml
```diff
@@ -0,0 +1,859 @@
...
+        nullable: true
+        description: Unix timestamp when the bounce was recorded.
+        expr:
+          kind: path
+          path:
+            - created
+
+  # ──────────────────────────────────────────────────────────────────────
+  # blocks
+  # ──────────────────────────────────────────────────────────────────────
```
antonmry, line 503: Tiny ergonomics suggestion, please ignore if this was deliberate: `created` here (and the equivalent columns on `blocks`, `spam_reports`, `invalid_emails`, plus `marketing_senders.created_at`/`updated_at`) are Unix epoch seconds. They could be exposed as actual `Timestamp` columns using:

```yaml
- name: created
  type: Timestamp
  nullable: true
  description: When the bounce was recorded.
  expr:
    kind: format_timestamp
    input: seconds
    expr:
      kind: path
      path:
        - created
```

That would let users write `WHERE created > NOW() - INTERVAL '7 days'` naturally instead of doing epoch arithmetic. If you'd rather keep the raw int, the `sources/community/hn` source has the nice pattern of exposing both: a `created_at_i` Int64 alongside a `created_at` Timestamp derived from it. Up to you.

sources/community/sendgrid/manifest.yaml
```diff
@@ -0,0 +1,859 @@
...
+          kind: path
+          path:
+            - generation
+      - name: updated_at
+        type: Utf8
+        nullable: true
+        description: Last update timestamp.
+        expr:
+          kind: path
+          path:
```
antonmry, line 329: Minor docs nit — per the SendGrid templates docs, `updated_at` uses a custom format (`"YYYY-MM-DD HH:MM:SS"`, e.g. `"2020-11-12 12:00:09"`), not ISO 8601 or Unix epoch. Keeping `Utf8` here is reasonable since `format_timestamp` doesn't have a native pattern for this shape — but if you could call out the format in the column description, users won't assume they can compare it to ISO timestamps from other sources.

Same caveat applies to `singlesends.send_at` / `created_at` / `updated_at` if those are also in this custom format (their column descriptions currently say "ISO 8601" — worth confirming).

sources/community/sendgrid/manifest.yaml
```diff
@@ -0,0 +1,859 @@
...
+        - name: page_size
+          from: literal
+          value: 200
+    response:
+      rows_path:
+        - result
+    pagination:
+      mode: none
+    columns:
+      - name: id
```
antonmry, line 299: Not a blocker — just acknowledging the documented limitation here. SendGrid's `/templates` returns `_metadata.next` as a full URL (per the docs), which Coral's `cursor_query` mode can't consume because it would pass the entire URL verbatim as the next request's cursor value. So hardcoding `page_size=200` (the documented max) is the right call for now.

Same pagination gap applies to `marketing_lists` and `singlesends` further down in this manifest. The README's "Pagination" section calls all three out — could you mirror that note here in the table `guide` block (or a top-level comment) so contributors editing this file don't accidentally try to add `mode: cursor_query` and assume it works?

(Same DSL gap shows up in the Snyk and Beehiiv PRs — probably worth a small DSL feature down the line, but nothing to do about it on a per-source PR.)

sources/community/sendgrid/manifest.yaml
```diff
@@ -0,0 +1,859 @@
+name: sendgrid
+version: 0.1.0
+dsl_version: 3
+backend: http
+description: >-
+  Query email templates, suppression lists (bounces, blocks, spam
+  reports, invalid emails), marketing lists, single sends,
+  suppression groups, sender identities, verified senders,
+  API keys, and teammates from Twilio SendGrid.
+base_url: https://api.sendgrid.com/v3
```
antonmry: Tiny ergonomics suggestion: EU-hosted SendGrid accounts need `https://api.eu.sendgrid.com/v3`, and the PR notes those users have to "manually update base_url in manifest" — which means forking this file. Following the cal.com pattern (`sources/community/cal/manifest.yaml:16–23`), you could add a `SENDGRID_BASE_URL` variable input with the US URL as default:

```yaml
inputs:
  SENDGRID_BASE_URL:
    kind: variable
    default: https://api.sendgrid.com/v3
    hint: |
      Base URL for the SendGrid API. The default works for US-hosted
      accounts. For EU-hosted accounts use
      https://api.eu.sendgrid.com/v3.
  SENDGRID_API_KEY:
    ...
base_url: "{{input.SENDGRID_BASE_URL}}"
```

That way EU users can just set an env var instead of editing the manifest. Not a blocker.

sources/community/sendgrid/manifest.yaml
```diff
@@ -0,0 +1,859 @@
...
+    guide: >
+      `SELECT api_key_id, name FROM sendgrid.api_keys`.
+    request:
+      method: GET
+      path: /api_keys
+      query:
+        - name: limit
+          from: literal
+          value: 100
+    response:
```
antonmry, line 57: Minor — please ignore if `/api_keys` really doesn't paginate. The endpoint accepts a `limit` query param but I couldn't tell from the docs whether it's an actual page-size cap or just a result limit (with no way to page past it). With `limit=100` hardcoded as a literal, an account with >100 API keys would be silently truncated.

Most accounts have well under 100 keys so this is mostly theoretical, but if `limit` does cap the response, it might be cleaner to either:

1. Document the 100-row cap in the table description, or
2. Move it to a `page_size`/`offset` pagination block if SendGrid actually supports `offset` on this endpoint (the docs I found didn't say).

sources/community/sendgrid/manifest.yaml
```diff
@@ -0,0 +1,859 @@
...
+    - name: Authorization
+      from: template
+      template: "Bearer {{input.SENDGRID_API_KEY}}"
+
+test_queries:
+  - SELECT api_key_id, name FROM sendgrid.api_keys LIMIT 1
+
+tables:
+  # ──────────────────────────────────────────────────────────────────────
+  # api_keys
```
antonmry, line 40: Tiny suggestion, easy to skip: this source has three distinct response shapes (`result` array, `results` array, top-level array via `row_strategy: direct`), and the test queries currently only exercise the first one. Adding one query per shape would make CI catch any future `rows_path` regression on the others. Something like:

```yaml
test_queries:
  - SELECT api_key_id, name FROM sendgrid.api_keys LIMIT 1
  - SELECT id, nickname FROM sendgrid.verified_senders LIMIT 1
  - SELECT email FROM sendgrid.bounces LIMIT 1
```

(Bounces returning 0 rows on a clean account is fine — the query itself still validates the parsing path.)

sources/community/sendgrid/manifest.yaml
Arnav1709: I double-checked against a live SendGrid run.

For `GET /v3/teammates?limit=1&offset=0`, the live API response used `result` singular, not `results` plural:

- top-level keys: `["result"]`
- `result` type: array
- `results` type: null
- `result_count`: 1
- `results_count`: 0

I also installed the source locally and queried through Coral with the current manifest. `SELECT COUNT(*) FROM sendgrid.teammates` returned 1 row, and a follow-up query against `user_type` / `is_admin` returned valid data.

So I think the manifest is correct here with:

response:
  rows_path:
    - result

Looks like the docs may be inconsistent with the live endpoint behavior for this one.


image
image

sources/community/sendgrid/manifest.yaml
Arnav1709: Added `start_time`, `end_time`, and `email` filters to `bounces`, `blocks`, `spam_reports`, and `invalid_emails`, and updated the README examples to show server-side filtering.

sources/community/sendgrid/manifest.yaml
Arnav1709: Done. I kept the raw epoch values as `created_i` / `created_at_i` / `updated_at_i` and added derived `Timestamp` columns for `created`, `created_at`, and `updated_at`.

sources/community/sendgrid/manifest.yaml
Arnav1709: Updated. `templates.updated_at` now explicitly documents the `YYYY-MM-DD HH:MM:SS` format. I also checked `singlesends`; those fields are documented as date-time, so I left their ISO wording unchanged.

sources/community/sendgrid/manifest.yaml
Arnav1709: Done. Added manifest guide notes for `templates`, `marketing_lists`, and `singlesends` explaining that `_metadata.next` is a full URL and Coral cannot follow it as a cursor yet.

sources/community/sendgrid/manifest.yaml
Arnav1709: Done. Added `SENDGRID_BASE_URL` with `https://api.sendgrid.com/v3` as the default and documented `https://api.eu.sendgrid.com/v3` for EU-hosted accounts.

sources/community/sendgrid/manifest.yaml
Arnav1709: Done. I left the query as a fixed `limit=100` because the docs do not show offset/cursor support, and documented that this table requests the first 100 API keys.

sources/community/sendgrid/manifest.yaml
Arnav1709: Done. Added test queries for all three response shapes: `result`, `results`, and top-level array via `row_strategy: direct`.

antonmry (approved):
Thanks for the very thorough response — and particularly for re-running against the live API on the `teammates` finding. You're right and the docs are wrong: SendGrid's live response uses `result` (singular) for teammates, not `results` (plural). Glad we caught that with empirical evidence rather than trusting the doc. Apologies for the false alarm — that's exactly the kind of verification I was hoping for.

Verified the other fixes at HEAD (`6ec4f81a`):

- **Suppression filters** (`L411–415` bounces, `L513–519` blocks, `L545–551` spam_reports, `L617–619` invalid_emails): `start_time`, `end_time`, and `email` all wired up as server-side query params. ✅
- **Dual timestamp columns** (`L481–499` bounces, plus equivalents on blocks/spam_reports/invalid_emails/marketing_senders): `created_i` (Int64 epoch) + `created` (Timestamp) — exactly the `sources/community/hn` pattern. Best of both worlds. ✅
- **`SENDGRID_BASE_URL` input** (`L12`, with `base_url: "{{input.SENDGRID_BASE_URL}}"` at L24): EU users now just set an env var. ✅
- **`templates.updated_at` format**: documented as `YYYY-MM-DD HH:MM:SS`. ✅
- **`_metadata.next` pagination limitation**: documented in the table guides for `templates`, `marketing_lists`, `singlesends`. ✅
- **`api_keys` 100-row cap**: documented. ✅
- **Test queries**: now cover all three response shapes (`result`, `results`, top-level array). ✅

Approving. Really nice work — the dual `_i` / `Timestamp` column pattern in particular is something I'd love to see other community sources adopt.

PR #549, opened 2026-05-17
feat(sources/community/trino): add Trino community source

### LINKED ISSUE #519: design: add trino community source
Add a read-only community source for Trino.

It would expose coordinator and query state over the Trino cluster REST
API so agents can inspect a cluster with SQL.

Proposed tables:
- info     => coordinator version, environment, uptime
- nodes    => worker nodes, heartbeat, recent failure ratio
- queries  => active and recent queries: state, user, error, CPU,
             input size, timings

Auth: X-Trino-User header (Trino requires a user on every request).
Works with self-managed Trino coordinators.

Scope is observability only, it does not run SQL via the
/v1/statement protocol, so it adds no query-execution load.

Worth adding

hassaansaleem28:
Adds a read-only Trino community source. Closes #519.

**Tables**:
- info     — coordinator version, environment, state, uptime
- queries  — active and recent queries: state, user, query type, error,
             timings, driver counts, memory usage

**Auth**: 

X-Trino-User header (Trino requires a user on /v1/query). Scope is
observability only — it does not run SQL via /v1/statement, so it adds no
query-execution load.

Note: the `nodes` table from the issue was dropped on purpose. /v1/node
returns 404 in modern Trino (verified on 481), so a worker table would be
broken. Worker info is available via system.runtime.nodes with a regular
Trino SQL client.

**Verified locally** against Trino 481:

$ coral source test trino
  ✓ trino connected successfully
    trino (2 tables)

$ coral sql "SELECT version, environment, state FROM trino.info"
| 481 | docker | ACTIVE |

$ coral sql "SELECT user, state, error_type, error_code_name FROM trino.queries WHERE state='FAILED'"
| bob | FAILED | USER_ERROR | MISSING_SCHEMA_NAME |

**Repro**:

docker run -d --name trino-coral -p 8080:8080 trinodb/trino:latest
TRINO_URL=http://localhost:8080 TRINO_USER=coral \
coral source add --file sources/community/trino/manifest.yaml
coral source test trino

**Demo video below.**
[Screencast from 2026-05-17 09-40-02.webm](https://github.com/user-attachments/assets/45594470-b62c-46b7-b578-761b1500047f)

### DIFF
--- a/sources/community/trino/README.md
+++ b/sources/community/trino/README.md
@@ -1,0 +1,126 @@
+# Trino
+
+Query Trino coordinator information and the cluster's active and recently
+completed query history (state, errors, timings, driver counts, memory usage)
+through the Trino cluster REST API.
+
+## Setup
+
+### Requirements
+
+- Network access to a Trino coordinator's HTTP endpoint (default port
+  `8080`, or `8443` for TLS clusters).
+- An identity your cluster's access control allows to read query state. On
+  clusters without authentication, any non-empty value works.
+
+### Add the Source
+
+```bash
+coral source add trino
+```
+
+Provide:
+
+- `TRINO_URL` — coordinator base URL including scheme and port, e.g.
+  `http://localhost:8080` or `https://coordinator.example.com:8443`.
+  No trailing slash.
+- `TRINO_USER` — identity sent in the `X-Trino-User` header
+  (defaults to `coral`).
+
+## Tables
+
+### `info`
+Single-row coordinator server information from `/v1/info`.
+
+**Useful for:**
+- Version and environment reporting
+- Readiness checks (`starting = false`, `state = 'ACTIVE'`)
+
+### `queries`
+Active and recently completed queries from `/v1/query`.
+
+**Useful for:**
+- Finding long-running or stuck queries (`state = 'RUNNING'`)
+- Triaging failures by `error_type` / `error_code_name`
+- Spotting expensive queries via CPU time, input rows, and memory
+- Per-user and per-catalog query attribution
+
+## Authentication
+
+Trino requires a user identity on requests to `/v1/query`. This source sends
+it as the `X-Trino-User` header:
+
+```text
+X-Trino-User: <TRINO_USER>
+```
+
+Username/password (`BasicAuth`) and OAuth2 authentication are **not**
+configured by this source. It targets clusters that accept header-based user
+identity (the common default), or clusters fronted by a proxy that injects
+auth.
+
+## Limits
+
+- This source is **read-only** and intentionally does **not** run SQL.
+  It does not use the `/v1/statement` query-submission protocol, so it adds
+  no query-execution load and needs no result pagination.
+- `queries` reflects only what the coordinator still holds in memory:
+  in-flight queries plus a bounded recent history. Evicted older queries are
+  not returned — this is a live operational view, not a complete audit log.
+- The `/v1/node` endpoint was removed in modern Trino, so a worker-node
+  table is intentionally not included. Inspect nodes through the
+  `system.runtime.nodes` table with a regular Trino SQL client instead.
+- Duration and data-size fields (`elapsed_time`, `total_cpu_time`,
+  `physical_input_data_size`, `peak_user_memory_reservation`, etc.) are
+  human-readable strings exactly as Trino returns them (e.g. `439.69ms`,
+  `0B`, `4.5GB`), typed as `Utf8`. `create_time` and `end_time` are real
+  `Timestamp` columns.
+- `error_type`, `error_code_name`, and `error_code` are populated only when
+  `state = 'FAILED'`.
+- No server-side filtering: filter with SQL `WHERE` after fetching.
+
+## Example Queries
+
+### Coordinator readiness
+
+```sql
+SELECT version, environment, state, starting, uptime
+FROM trino.info
+```
+
+### Currently running queries, oldest first
+
+```sql
+SELECT query_id, user, state, elapsed_time, query
+FROM trino.queries
+WHERE state = 'RUNNING'
+ORDER BY create_time ASC
+```
+
+### Recent failures grouped by error code
+
+```sql
+SELECT error_code_name, error_type, COUNT(*) AS failures
+FROM trino.queries
+WHERE state = 'FAILED'
+GROUP BY error_code_name, error_type
+ORDER BY failures DESC
+```
+
+### Highest-throughput finished queries by input rows
+
+```sql
+SELECT user, query_id, processed_input_positions, total_cpu_time
+FROM trino.queries
+WHERE state = 'FINISHED'
+ORDER BY processed_input_positions DESC
+LIMIT 20
+```
+
+## Notes
+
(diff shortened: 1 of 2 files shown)


sources/community/trino/manifest.yaml
```diff
@@ -0,0 +1,543 @@
...
+          kind: path
+          path:
+            - nodeId
+      - name: state
+        type: Utf8
+        nullable: true
+        description: "Coordinator state, e.g. ACTIVE"
+        expr:
+          kind: path
+          path:
```
antonmry, line 72: Really sorry if I've got this wrong — please push back if the demo run in the PR description really did show `ACTIVE` here. When I looked at Trino's `ServerInfo` class (the JSON returned by `/v1/info`) at tag 481 and on master, I couldn't find a `state` field:

- v481: `https://github.com/trinodb/trino/blob/481/client/trino-client/src/main/java/io/trino/client/ServerInfo.java`
- master: `https://github.com/trinodb/trino/blob/master/client/trino-client/src/main/java/io/trino/client/ServerInfo.java`

The class only carries `nodeVersion`, `environment`, `coordinator`, `starting`, `uptime`, `coordinatorId`, `nodeId`. Coordinator state seems to live on a separate `/v1/info/state` endpoint that returns just the state string.

If that's right, this column would always come back null on `/v1/info`. Both this `description` and the table `guide` above (L46) plus the README "Readiness checks" section (`README.md:37`) currently imply `state = 'ACTIVE'` works as a readiness predicate.

Two ways out, depending on what you confirm:

1. Drop the `state` column from `info` and lean on `starting = false` for readiness in the docs.
2. Add a second single-row table (e.g. `info_state`) that calls `/v1/info/state` and returns a single `state` column.

Could you re-run `SELECT state FROM trino.info` against a fresh 481 coordinator and let me know? Totally possible something's exposing it that I'm not seeing.

sources/community/trino/manifest.yaml
```diff
@@ -0,0 +1,543 @@
...
+      Duration fields (`elapsed_time`, `execution_time`, `total_cpu_time`,
+      etc.) and data-size fields (`physical_input_data_size`,
+      `peak_user_memory_reservation`, etc.) are human-readable strings
+      exactly as Trino returns them (e.g. `439.69ms`, `0B`, `4.5GB`), not
+      numbers. `create_time` and `end_time` are real timestamps;
+      `end_time` is null while a query is still running. `error_type` and
+      `error_code_name` are null unless the query state is `FAILED`.
+    request:
+      method: GET
+      path: /v1/query
```
antonmry, line 150: Optional ergonomics suggestion — not a blocker. Trino's `/v1/query` natively accepts a `state` query param (multi-valued), so e.g. `GET /v1/query?state=RUNNING&state=QUEUED` lets the coordinator filter server-side. From `QueryResource.java` at tag 481:

```java
public List<BasicQueryInfo> getAllQueryInfo(
    @QueryParam("state") Set<String> stateFilters, ...)
```

The table `guide` mentions this is currently a deliberate "no server-side filtering" choice, which is fine — but on a busy production cluster, downloading the entire in-memory query history just to filter to `state = 'RUNNING'` rows in SQL can be a lot of bytes for a very common query shape. Exposing `state` as an optional filter that maps to the `?state=` query param would push the predicate down to the coordinator with no change to the column shape. Up to you whether to land it now or as a follow-up.

sources/community/trino/manifest.yaml
```diff
@@ -0,0 +1,543 @@
+dsl_version: 3
+name: trino
+version: 1.0.0
```
antonmry: Tiny consistency nit, please ignore if `1.0.0` was deliberate: every other community source so far starts at `0.1.0` (`sources/community/cal/manifest.yaml`, `hn`, plus the in-flight beehiiv/snyk PRs). The table set here is likely to keep evolving, so `0.1.0` would leave room for that without implying a stability promise. Not a blocker — go with whatever you prefer.

sources/community/trino/manifest.yaml
```diff
@@ -0,0 +1,543 @@
+dsl_version: 3
```
antonmry: Purely cosmetic, feel free to ignore: `sources/community/cal/manifest.yaml` leads with `name → version → dsl_version → backend → description`. This file leads with `dsl_version → name → version → description → backend`. It parses fine either way — just slightly easier to skim when adjacent community manifests use the same ordering.

sources/community/trino/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+- Network access to a Trino coordinator's HTTP endpoint (default port
+  `8080`, or `8443` for TLS clusters).
+- An identity your cluster's access control allows to read query state. On
+  clusters without authentication, any non-empty value works.
+
+### Add the Source
+
+```bash
+coral source add trino
```
antonmry: Tiny docs nit — please ignore if `coral source add <name>` is a supported alias I don't know about. Other community READMEs use the explicit `--file` form, e.g. from `sources/community/cal/README.md`:

```bash
CAL_API_KEY=cal_live_... coral source add --file sources/community/cal/manifest.yaml
```

If users running through the README don't have a published registry to resolve a bare `trino` name from, the bare-name form might fail. Matching that style would be:

```bash
TRINO_URL=http://localhost:8080 TRINO_USER=coral \
  coral source add --file sources/community/trino/manifest.yaml
```

(This also matches the `## Repro` block in the PR description, which already uses `--file`.)

sources/community/trino/manifest.yaml
hassaansaleem28: Verified on Trino 481:

$ curl -s localhost:8080/v1/info
{"nodeId":"5940bf4f3042","state":"ACTIVE","nodeVersion":{"version":"481"},
 "environment":"docker","coordinator":true,"coordinatorId":"nm7he",
 "starting":false,"uptime":"2.24s"}

$ coral sql "SELECT state, version, starting FROM trino.info"
| ACTIVE | 481 | false |

state is present and populated (NodeState: ACTIVE/INACTIVE/SHUTTING_DOWN).
so keeping the column

antonmry (approved):
Thanks for the careful response — especially for the live `curl localhost:8080/v1/info` output. That settles the `info.state` question: you're right, `state` is present in the actual JSON response even though it's not a declared field on the `ServerInfo` class constructor I was reading. Trino must be adding it via a Jackson getter or post-init hook that doesn't show up in the source I grepped. My earlier finding was based on incomplete analysis — apologies for the noise on that one. The column works as documented.

Verified the other fixes at HEAD (`d5c8f3bd`):

- **`queries.state` filter** (`L147–154`): added as a server-side query param, pushes `WHERE state = 'RUNNING'` down to the coordinator. ✅
- **`version`** (`L2`): now `0.1.0`, matches the cal/hn/community convention. ✅
- **Field ordering** (`L1–8`): now `name → version → dsl_version → backend → description`. ✅
- **README `coral source add` syntax**: now uses the explicit `--file` form. ✅

Side note: ignore the cubic-dev-ai bot's flag about "version regressed instead of bumped forward" — going `1.0.0` → `0.1.0` is the correction we discussed (every other community source uses `0.1.0` for its first release), not a regression. The bot doesn't have repo context.

Approving. Thanks for the thorough turnaround.

PR #557, opened 2026-05-17
feat(sources/community/snyk): add snyk rest api source

### LINKED ISSUE #556: design: add Snyk community source
## Summary

Add a new `snyk` community source that lets Coral query Snyk organizations, targets, projects, issues, issue paths, and dependencies through the Snyk REST API. This gives users and agents a read-only view into security postures, vulnerable components, and their project impact.

## Motivation

Snyk is an enterprise security platform for scanning code, dependencies, containers, and infrastructure. Teams use Snyk to identify vulnerabilities and license issues. 

Without a Coral source, users must hand-roll Snyk API calls to inspect targets, projects, issues, and dependency paths. A read-only source makes Snyk security metadata easier to inventory, summarize, and analyze from SQL.

## Proposed design

Add a new community source under `sources/community/snyk`, implemented as a manifest-only HTTP source.

Inputs:

* `SNYK_API_TOKEN` required secret
  * Snyk token for REST API access (Personal Access Token or Service Account token)
  * generated in Snyk Account Settings
  * used for subsequent read-only Snyk REST API requests

Auth:

* token auth
* `Authorization: Token {{input.SNYK_API_TOKEN}}`

Proposed v1 tables:

* `snyk.current_user`
  * `GET /self`
  * validates token and returns current user identity
* `snyk.organizations`
  * `GET /orgs`
  * lists Snyk organizations available to the token
* `snyk.targets`
  * `GET /orgs/{org_id}/targets`
  * requires `org_id`
  * exposes GitHub repos, Docker images, Kubernetes workloads, etc.
* `snyk.projects`
  * `GET /orgs/{org_id}/projects`
  * requires `org_id`
  * exposes scan instances associated with targets
* `snyk.project_issue_counts`
  * `GET /orgs/{org_id}/projects?expand=meta.latest_issue_counts`
  * requires `org_id`
  * exposes project issue counts broken down by severity
* `snyk.issues`
  * `GET /orgs/{org_id}/issues?project_id={project_id}`
  * requires `org_id`, `project_id`
  * exposes vulnerabilities and license issues
* `snyk.issue_paths`
  * `GET /orgs/{org_id}/projects/{project_id}/issues/{issue_id}/paths`
  * requires `org_id`, `project_id`, `issue_id`
  * exposes dependency paths showing exactly how an issue was introduced
* `snyk.dependencies`
  * `GET /orgs/{org_id}/projects/{project_id}/dependencies`
  * requires `org_id`, `project_id`
  * exposes project dependency inventory (SBOM)

Design details:

* keep v1 read-only
* use Snyk REST API (with `version=2025-11-05` query parameter)
* document token generation
* require `org_id` on scoped tables to avoid broad workspace scans
* require `project_id` on issues and dependencies
* expose stable metadata fields and preserve complex dependency paths as JSON
* omit legacy reporting APIs, mutations, and snapshots

Affected surface:

* community source only
* expected files:
  * `sources/community/snyk/manifest.yaml`
  * `sources/community/snyk/README.md`

No CLI, config, MCP, runtime, or bundled source changes are proposed.

## Drawbacks and trade-offs

The Snyk REST API requires an `org_id` on almost all endpoints, which means tables like `projects` and `targets` require the `org_id` filter. This is stricter than some legacy v1 APIs, but it keeps Coral queries scoped and predictable.

The `issues` table explicitly requires a `project_id` filter to prevent org-wide scans, which could be extremely large, slow, and exhaust rate limits.

## Alternatives considered

* **Building on Snyk v1 API:** rejected because Snyk v1 APIs are being slowly deprecated in favor of the newer REST APIs.
* **Allowing org-wide issue scans:** rejected because issue lists for entire organizations can be massive and Coral should avoid broad scans.
* **Including Reporting API:** rejected because it's legacy and unavailable in some regions.
* **Inventory Assets:** rejected because they are early access/async APIs and do not map cleanly to standard relational tables yet.

## Open questions

* Should the API `version` parameter be made dynamic/configurable as an input in the future?
* Should snapshots be included in a future version once their structure stabilizes?

abhayguptas:
## Summary

Fixes #556 

Add a new `snyk` community source for querying Snyk organizations, targets, projects, issues, issue paths, and dependencies using the Snyk REST API.

This change is manifest-only and uses Coral's existing HTTP source-spec model. It adds read-only security observability without changing CLI, MCP, config, runtime, or bundled-source contracts.

## What changed

Added:
* `sources/community/snyk/manifest.yaml`
* `sources/community/snyk/README.md`

## Source scope

Inputs:
* `SNYK_API_TOKEN`

Tables:

* `snyk.current_user`
  * `GET /self`
  * validates token and returns current user identity
* `snyk.organizations`
  * `GET /orgs`
  * lists Snyk organizations available to the token
* `snyk.targets`
  * `GET /orgs/{org_id}/targets`
  * requires `org_id`
  * exposes GitHub repos, Docker images, Kubernetes workloads, etc.
* `snyk.projects`
  * `GET /orgs/{org_id}/projects`
  * requires `org_id`
  * exposes scan instances associated with targets
* `snyk.project_issue_counts`
  * `GET /orgs/{org_id}/projects?expand=meta.latest_issue_counts`
  * requires `org_id`
  * exposes project issue counts broken down by severity
* `snyk.issues`
  * `GET /orgs/{org_id}/issues?project_id={project_id}`
  * requires `org_id`, `project_id`
  * exposes vulnerabilities and license issues
* `snyk.issue_paths`
  * `GET /orgs/{org_id}/projects/{project_id}/issues/{issue_id}/paths`
  * requires `org_id`, `project_id`, `issue_id`
  * exposes dependency paths showing exactly how an issue was introduced
* `snyk.dependencies`
  * `GET /orgs/{org_id}/projects/{project_id}/dependencies`
  * requires `org_id`, `project_id`
  * exposes project dependency inventory (SBOM)

## Implementation notes

* uses Snyk REST API token auth (`Authorization: Token <token>`)
* enforces the `version=2025-11-05` query parameter for stability
* documents token generation in Snyk Account Settings
* keeps v1 intentionally read-only
* requires `org_id` on scoped project tables to avoid broad workspace scans
* requires `project_id` on issues and dependencies
* exposes high-value flattened fields plus JSON for nested dependency paths
* intentionally omits legacy Reporting APIs, Inventory Assets, mutations, and snapshots

## Validation

Local validation:
* `coral source lint sources/community/snyk/manifest.yaml`
* `make lint-sources`

*(Note: Live API testing is recommended against a real Snyk workspace to confirm endpoint data structures before final merge.)*

## Testing notes

* `coral source lint sources/community/snyk/manifest.yaml` verified the manifest structure and JSON:API path expressions against Coral's strict JSON schema.
* All array indexes within `expr` paths were verified to be formatted as string tokens (e.g., `"0"`) as required by the schema.
* `make lint-sources` ran via `ryl` and passed successfully.

## Out of scope

Not included in v1:
* legacy Reporting API
* Inventory Assets (async/early access)
* mutating projects or issues (no PATCH/DELETE)
* Snapshots

### DIFF
--- a/sources/community/snyk/README.md
+++ b/sources/community/snyk/README.md
@@ -1,0 +1,54 @@
+# Snyk
+
+Query Snyk organizations, targets, projects, issues, issue paths, and dependencies via the Snyk REST API.
+
+## Setup
+
+### Get Your API Token
+1. Log into Snyk.
+2. Go to **Account Settings > General**.
+3. Generate a Personal Access Token (or use a Service Account token).
+
+### Add the Source
+```bash
+coral source add snyk
+```
+When prompted, provide your token as `SNYK_API_TOKEN`.
+
+## Tables
+
+### `current_user`
+Validates the credentials and retrieves the current authenticated user identity.
+
+### `organizations`
+Organizations available to the authenticated Snyk user. Use this to find the `org_id` required for other queries.
+
+### `targets`
+Targets are the foundational assets in Snyk (e.g., GitHub repos, Docker images, Kubernetes workloads).
+**Requires:** `org_id`
+
+### `projects`
+Scan instances associated with targets.
+**Requires:** `org_id`
+
+### `project_issue_counts`
+Analytics for project issue counts broken down by severity (critical, high, medium, low).
+**Requires:** `org_id`
+
+### `issues`
+Vulnerabilities and license issues in Snyk projects.
+**Requires:** `org_id`, `project_id`
+
+### `issue_paths`
+Dependency paths showing exactly how an issue was introduced (e.g., express -> lodash -> minimist -> vuln).
+**Requires:** `org_id`, `project_id`, `issue_id`
+
+### `dependencies`
+Project dependency inventory (SBOM).
+**Requires:** `org_id`, `project_id`
+
+## Authentication
+The source uses the Snyk REST API with token authentication:
+```text
+Authorization: Token <SNYK_API_TOKEN>
+```
--- a/sources/community/snyk/manifest.yaml
+++ b/sources/community/snyk/manifest.yaml
@@ -1,0 +1,573 @@
+name: snyk
+version: 0.1.0
+dsl_version: 3
+backend: http
+description: >-
+  Query Snyk organizations, targets, projects, issues, issue paths, and dependencies via the Snyk REST API.
+base_url: https://api.snyk.io/rest
+inputs:
+  SNYK_API_TOKEN:
+    kind: secret
+    hint: |
+      Snyk API Token (Personal Access Token or Service Account token).
+      Create one in Snyk Account Settings. The token is sent as `Authorization: Token <SNYK_API_TOKEN>`.
+auth:
+  type: HeaderAuth
+  headers:
+    - name: Authorization
+      from: template
+      template: Token {{input.SNYK_API_TOKEN}}
+test_queries:
+  - SELECT * FROM snyk.current_user LIMIT 1
+tables:
+  - name: current_user
+    description: Current authenticated Snyk user identity.
+    fetch_limit_default: 1
+    request:
+      method: GET
+      path: /self
+      query:
+        - name: version
+          from: literal
+          value: "2025-11-05"
+    response:
+      row_strategy: direct
+    pagination:
+      mode: none
+    columns:
+      - name: user_id
+        type: Utf8
+        nullable: false
+        description: User ID.
+        expr:
+          kind: path
+          path: [data, id]
+      - name: username
+        type: Utf8
+        nullable: true
+        description: Username.
+        expr:
+          kind: path
+          path: [data, attributes, username]
+      - name: email
+        type: Utf8
+        nullable: true
+        description: Email address.
+        expr:
+          kind: path
+          path: [data, attributes, email]
+      - name: account_type
+        type: Utf8
+        nullable: true
+        description: Account type.
+        expr:
+          kind: path
+          path: [data, type]
+
+  - name: organizations
+    description: Snyk organizations available to the authenticated user.
+    guide: |
+      Use `org_id` from this table when querying targets, projects, and issues.
+    fetch_limit_default: 100
+    request:
+      method: GET
+      path: /orgs
+      query:
+        - name: version
+          from: literal
+          value: "2025-11-05"
+    response:
+      rows_path: [data]
+    pagination:
+      mode: cursor_query
+      cursor_param: starting_after
+      response_cursor_path: [links, next]
+    columns:
+      - name: org_id
+        type: Utf8
+        nullable: false
+        description: Organization ID.
+        expr:
+          kind: path
+          path: [id]
+      - name: name
+        type: Utf8
+        nullable: false
+        description: Organization name.
+        expr:
+          kind: path
+          path: [attributes, name]
+      - name: slug
+        type: Utf8
+        nullable: false
+        description: Organization slug.
+        expr:
+          kind: path
+          path: [attributes, slug]
+      - name: created_at
+        type: Timestamp
+        nullable: true
+        description: Organization creation timestamp.
+        expr:
+          kind: format_timestamp
+          input: iso8601
+          expr:
+            kind: path
+            path: [attributes, created_at]
+
+  - name: targets
+    description: Snyk targets (GitHub repos, Docker images, Kubernetes workloads, etc.).
+    fetch_limit_default: 100
+    filters:
+      - name: org_id
+        required: true
+      - name: source_types
+      - name: is_private
+      - name: display_name
+    request:
+      method: GET
+      path: /orgs/{{filter.org_id}}/targets
+      query:
+        - name: version
+          from: literal
+          value: "2025-11-05"
+        - name: source_types
+          from: filter
+          key: source_types
+        - name: is_private
+          from: filter
+          key: is_private
+        - name: display_name
+          from: filter
+          key: display_name
+    response:
+      rows_path: [data]
+    pagination:
+      mode: cursor_query
+      cursor_param: starting_after
+      response_cursor_path: [links, next]
+    columns:
+      - name: target_id
+        type: Utf8
+        nullable: false
+        description: Target ID.
+        expr:
+          kind: path
+          path: [id]
+      - name: org_id
+        type: Utf8
+        nullable: false
+        description: Organization ID.
+        expr:
+          kind: from_filter
+          key: org_id
+      - name: display_name
+        type: Utf8
+        nullable: false
+        description: Target display name.
+        expr:
+          kind: path
+          path: [attributes, display_name]
+      - name: url
+        type: Utf8
+        nullable: true
+        description: Target URL.
+        expr:
+          kind: path
+          path: [attributes, url]
+      - name: source_type
+        type: Utf8
+        nullable: false
+        description: Target source type (e.g., github, dockerhub).
+        expr:
+          kind: path
+          path: [attributes, source_type]
+      - name: is_private
+        type: Boolean
+        nullable: true
+        description: Whether the target is private.
+        expr:
+          kind: path
+          path: [attributes, is_private]
+      - name: created_at
+        type: Timestamp
+        nullable: true
+        description: Target creation timestamp.
+        expr:
+          kind: format_timestamp
+          input: iso8601
+          expr:
+            kind: path
+            path: [attributes, created_at]
+
+  - name: projects
+    description: Snyk scan projects belonging to targets.
+    fetch_limit_default: 100
+    filters:
+      - name: org_id
+        required: true
+      - name: target_id
+      - name: target_reference
+    request:
+      method: GET
+      path: /orgs/{{filter.org_id}}/projects
+      query:
+        - name: version
+          from: literal
+          value: "2025-11-05"
+        - name: target_id
+          from: filter
+          key: target_id
+        - name: target_reference
+          from: filter
+          key: target_reference
+    response:
+      rows_path: [data]
+    pagination:
+      mode: cursor_query
+      cursor_param: starting_after
+      response_cursor_path: [links, next]
+    columns:
+      - name: project_id
+        type: Utf8
+        nullable: false
+        description: Project ID.
+        expr:
+          kind: path
+          path: [id]
+      - name: org_id
+        type: Utf8
+        nullable: false
+        description: Organization ID.
+        expr:
+          kind: from_filter
+          key: org_id
+      - name: target_id
+        type: Utf8
+        nullable: true
+        description: Target ID.
+        expr:
+          kind: path
+          path: [relationships, target, data, id]
+      - name: name
+        type: Utf8
+        nullable: false
+        description: Project name.
+        expr:
+          kind: path
+          path: [attributes, name]
+      - name: type
+        type: Utf8
+        nullable: true
+        description: Project type.
+        expr:
+          kind: path
+          path: [attributes, type]
+      - name: lifecycle
+        type: Utf8
+        nullable: true
+        description: Project lifecycle.
+        expr:
+          kind: path
+          path: [attributes, lifecycle]
+      - name: business_criticality
+        type: Utf8
+        nullable: true
+        description: Project business criticality.
+        expr:
+          kind: path
+          path: [attributes, business_criticality]
+      - name: created_at
+        type: Timestamp
+        nullable: true
+        description: Project creation timestamp.
+        expr:
+          kind: format_timestamp
+          input: iso8601
+          expr:
+            kind: path
+            path: [attributes, created_at]
+
+  - name: project_issue_counts
+    description: Issue counts for Snyk projects.
+    fetch_limit_default: 100
+    filters:
+      - name: org_id
+        required: true
+    request:
+      method: GET
+      path: /orgs/{{filter.org_id}}/projects
+      query:
+        - name: version
+          from: literal
+          value: "2025-11-05"
+        - name: expand
+          from: literal
+          value: "meta.latest_issue_counts"
+    response:
+      rows_path: [data]
+    pagination:
+      mode: cursor_query
+      cursor_param: starting_after
+      response_cursor_path: [links, next]
+    columns:
+      - name: project_id
+        type: Utf8
+        nullable: false
+        description: Project ID.
+        expr:
+          kind: path
+          path: [id]
+      - name: critical_count
+        type: Int64
+        nullable: true
+        description: Critical issue count.
+        expr:
+          kind: path
+          path: [meta, latest_issue_counts, critical]
+      - name: high_count
+        type: Int64
+        nullable: true
+        description: High issue count.
+        expr:
+          kind: path
+          path: [meta, latest_issue_counts, high]
+      - name: medium_count
+        type: Int64
+        nullable: true
+        description: Medium issue count.
+        expr:
+          kind: path
+          path: [meta, latest_issue_counts, medium]
+      - name: low_count
+        type: Int64
+        nullable: true
+        description: Low issue count.
+        expr:
+          kind: path
+          path: [meta, latest_issue_counts, low]
+
+  - name: issues
+    description: Vulnerabilities and license issues in Snyk projects.
+    fetch_limit_default: 100
+    filters:
+      - name: org_id
+        required: true
+      - name: project_id
+        required: true
+      - name: severity
+      - name: type
+    request:
+      method: GET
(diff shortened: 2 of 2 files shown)


sources/community/snyk/manifest.yaml
```diff
@@ -0,0 +1,573 @@
...
+      query:
+        - name: version
+          from: literal
+          value: "2025-11-05"
+    response:
+      rows_path: [data]
+    pagination:
+      mode: cursor_query
+      cursor_param: starting_after
+      response_cursor_path: [links, next]
```
antonmry, line 84: Thanks for taking the time on this — really nice scope shape for the v1.

One thing I'd love you to double-check, because I might be wrong here: the `response_cursor_path: [links, next]` extracts the value at `links.next`, and per Snyk's OpenAPI for `version=2025-11-05`, `links.next` is the **full next-page URL** (something like `/rest/orgs/{id}/projects?starting_after=eyJpZCI6...&version=2025-11-05&limit=10`), not just a cursor token.

Looking at how Coral's HTTP backend handles `cursor_query` (`crates/coral-engine/src/backends/http/client.rs:336–348` reads the value as a string and passes it verbatim as the cursor param), I think this means the next request would be sent as `?starting_after=<full URL>` rather than `?starting_after=<cursor token>`, which Snyk would likely reject.

If that's right, the same issue applies to every `cursor_query` block in this manifest (`organizations`, `targets`, `projects`, `project_issue_counts`, `issues`, `issue_paths`, `dependencies`) and the runtime would only ever fetch the first page.

Could you confirm against a live Snyk workspace whether pagination actually advances past page 1? If it doesn't, the workaround options I'm aware of are: (a) extract the `starting_after` query value from the URL before passing it back, (b) extend the DSL with a URL-based cursor mode, or (c) document the first-page limitation. I'm happy to be wrong on this — please push back if I've misread the runtime.

Spec: `https://api.snyk.io/rest/openapi/2025-11-05` → `/orgs` response 200 `links` schema.

sources/community/snyk/manifest.yaml
```diff
@@ -0,0 +1,573 @@
...
+    filters:
+      - name: org_id
+        required: true
+      - name: project_id
+        required: true
+      - name: issue_id
+        required: true
+    request:
+      method: GET
+      path: /orgs/{{filter.org_id}}/projects/{{filter.project_id}}/issues/{{filter.issue_id}}/paths
```
antonmry, line 469: Apologies if I've got this wrong, but when I fetched `https://api.snyk.io/rest/openapi/2025-11-05` and searched the paths, I couldn't find `/orgs/{org_id}/projects/{project_id}/issues/{issue_id}/paths` — the only `issues` endpoints I saw were `/orgs/{org_id}/issues`, `/orgs/{org_id}/issues/{issue_id}`, and the group-scoped equivalents.

It's possible this was a v1 / legacy endpoint that's no longer in the REST API. Could you double-check against a live workspace? If it really has been removed, the `issue_paths` table would 404 on every call and might need to be dropped from v1 (or moved behind a documented note).

sources/community/snyk/manifest.yaml
```diff
@@ -0,0 +1,573 @@
...
+    description: Project dependency inventory (BOM).
+    fetch_limit_default: 100
+    filters:
+      - name: org_id
+        required: true
+      - name: project_id
+        required: true
+    request:
+      method: GET
+      path: /orgs/{{filter.org_id}}/projects/{{filter.project_id}}/dependencies
```
antonmry, line 520: Same caveat as the `issue_paths` comment — I might be misreading the spec, but I couldn't locate `/orgs/{org_id}/projects/{project_id}/dependencies` in `https://api.snyk.io/rest/openapi/2025-11-05`. The closest endpoint I found was `/orgs/{org_id}/projects/{project_id}/sbom`, which returns a different shape (CycloneDX/SPDX SBOM rather than a flat dependency list).

If `/dependencies` was a v1 / legacy endpoint, the `dependencies` table would 404. Could you check this against a live workspace? If `/sbom` is the intended replacement, the column mappings would need significant changes (e.g. `components[*].name`, `components[*].version` etc.).

sources/community/snyk/manifest.yaml
```diff
@@ -0,0 +1,573 @@
...
+        expr:
+          kind: path
+          path: [attributes, title]
+      - name: package_name
+        type: Utf8
+        nullable: true
+        description: Vulnerable package name.
+        expr:
+          kind: path
+          path: [attributes, coordinates, "0", representations, "0", dependency, package_name]
```
antonmry, line 424: Could you sanity-check these paths against a real `GET /orgs/{org_id}/issues` response? When I traced the schema in the OpenAPI for `2025-11-05`, `attributes.coordinates[].representations[]` resolves to a `PackageRepresentation` which has the shape `{ package: { name, version, namespace, type, url } }` (via `PackageMeta`). I didn't see a `dependency.package_name` / `dependency.package_version` shape — so I worry these columns might come back null for any real issue.

If that's right, the paths likely need to be:

```yaml
path: [attributes, coordinates, "0", representations, "0", package, name]
path: [attributes, coordinates, "0", representations, "0", package, version]
```

Definitely possible I'm reading the wrong representation variant — Snyk does have a few (`PackageRepresentation`, `ResourcePathRepresentation`, etc.), so a quick check against a live response would be the surest way to confirm.

sources/community/snyk/manifest.yaml
```diff
@@ -0,0 +1,573 @@
...
+        expr:
+          kind: path
+          path: [attributes, fix_available]
+      - name: exploited
+        type: Boolean
+        nullable: true
+        description: Whether the vulnerability is exploited.
+        expr:
+          kind: path
+          path: [attributes, exploited]
```
antonmry, line 445: Same caveat — I could be reading the wrong schema variant — but when I looked at the `IssueAttributes` schema in `2025-11-05`, I didn't see `exploited` or `fix_available` as direct attributes. The fields I saw at that level were `risk`, `resolution`, `status`, `ignored`, `exploit_details`, `severities`, `effective_severity_level`, `key`, `key_asset`, `classes`, `problems`, `tool`, `type`, `title`, `description`, `created_at`, `updated_at`.

If these two are gone, both columns would be null on every row. The likely replacements:

- `fix_available` → derived from whether `attributes.resolution` or one of the entries in `attributes.problems[].fixable` is set (Snyk doesn't expose a single boolean for this in the REST API as far as I can tell).
- `exploited` → `attributes.exploit_details` is the closest, or `attributes.risk.factors[]` for exploit-maturity info.

Worth a quick check against a live issue response before this lands. Happy to be wrong — please push back if you've seen these fields populated.

sources/community/snyk/manifest.yaml
```diff
@@ -0,0 +1,573 @@
...
+      query:
+        - name: version
+          from: literal
+          value: "2025-11-05"
+        - name: project_id
+          from: filter
+          key: project_id
+        - name: severity
+          from: filter
+          key: severity
```
antonmry, line 372: I might be wrong here, but the OpenAPI for `GET /orgs/{org_id}/issues` at `version=2025-11-05` doesn't list a `severity` query param — the closest one is `effective_severity_level` (array). So a query like `WHERE severity = 'high'` would get silently ignored by the server and return all severities instead of just high.

If you confirm that, renaming the filter to `effective_severity_level` (or keeping the user-facing filter name `severity` but mapping it to the `effective_severity_level` query param) would match the API. The full filter set the API exposes is also worth considering: `status`, `ignored`, `created_after`/`created_before`, `updated_after`/`updated_before`, `scan_item.id`/`scan_item.type` — particularly the time-window ones, since "issues created in the last 7 days" is a common security-scan query.

sources/community/snyk/manifest.yaml
```diff
@@ -0,0 +1,573 @@
...
+        expr:
+          kind: path
+          path: [data, attributes, email]
+      - name: account_type
+        type: Utf8
+        nullable: true
+        description: Account type.
+        expr:
+          kind: path
+          path: [data, type]
```
antonmry, line 65: Just flagging in case it's worth a second look — `data.type` in a JSON:API response is the resource type discriminator, which for `/self` will always be the string `"user"` (or similar). So `account_type` would always come back as `"user"` rather than a meaningful tier.

If `data.attributes.account_type` (or `attributes.tier` / `attributes.role` / etc.) is exposed by the API, that would probably be the more useful mapping. Otherwise the column might be better dropped from v1 to avoid implying it carries information it doesn't.

I haven't been able to inspect a live `/self` response — happy to be told this maps to something I'm not seeing.

sources/community/snyk/manifest.yaml
```diff
@@ -0,0 +1,573 @@
...
+    request:
+      method: GET
+      path: /orgs/{{filter.org_id}}/projects
+      query:
+        - name: version
+          from: literal
+          value: "2025-11-05"
+        - name: expand
+          from: literal
+          value: "meta.latest_issue_counts"
```
antonmry, line 306: Minor nit, please verify against a live response: the OpenAPI for `GET /orgs/{org_id}/projects` exposes both an `expand` array param **and** a separate boolean `meta.latest_issue_counts` query param. I'm not 100% sure whether the `expand=meta.latest_issue_counts` form actually populates `meta.latest_issue_counts` in the response body, or whether Snyk now expects `meta.latest_issue_counts=true` as the boolean param instead.

If you've tested this returns non-null counts, ignore me. Otherwise it might be worth switching to:

```yaml
- name: meta.latest_issue_counts
  from: literal
  value: "true"
```

(Same goes for `meta.latest_dependency_total` if you ever want to expose dependency-count summaries.)

sources/community/snyk/manifest.yaml
```diff
@@ -0,0 +1,573 @@
...
+      method: GET
+      path: /orgs
+      query:
+        - name: version
+          from: literal
+          value: "2025-11-05"
+    response:
+      rows_path: [data]
+    pagination:
+      mode: cursor_query
```
antonmry, line 82: Small ergonomics note (and only relevant once the pagination concern above is resolved): I don't see a `page_size.query_param: limit` configured on any of the cursor tables. If Coral's runtime doesn't pass a `limit`, Snyk's default page size is 10, which makes any non-trivial scan slow. Once pagination is verified working end-to-end, consider adding something like:

```yaml
pagination:
  mode: cursor_query
  cursor_param: starting_after
  response_cursor_path: [links, next]
  page_size:
    default: 100
    max: 100
    query_param: limit
```

on each list table. (Snyk caps `limit` at 100 per the OpenAPI.)

sources/community/snyk/manifest.yaml
```diff
@@ -0,0 +1,573 @@
...
+    kind: secret
+    hint: |
+      Snyk API Token (Personal Access Token or Service Account token).
+      Create one in Snyk Account Settings. The token is sent as `Authorization: Token <SNYK_API_TOKEN>`.
+auth:
+  type: HeaderAuth
+  headers:
+    - name: Authorization
+      from: template
+      template: Token {{input.SNYK_API_TOKEN}}
```
antonmry, line 19: Tiny nit — happy to be told this works either way: Snyk's published auth docs typically show the header value with a lowercase scheme (`Authorization: token <SNYK_API_TOKEN>`), e.g. `https://docs.snyk.io/snyk-api/authentication-for-api`. The manifest uses `Token` (uppercase). HTTP auth schemes are technically case-insensitive per RFC 7235, so this may well be fine — but if you've tested it and it works, perfect; if not, lowercasing matches the docs.

sources/community/snyk/manifest.yaml
```diff
@@ -0,0 +1,573 @@
...
+    fetch_limit_default: 1
+    request:
+      method: GET
+      path: /self
+      query:
+        - name: version
+          from: literal
+          value: "2025-11-05"
+    response:
+      row_strategy: direct
```
antonmry, line 34: Tiny consistency note: this table sets `row_strategy: direct` but the rest of the manifest leaves it implicit and relies on `rows_path: [data]` instead. `sources/community/cal/manifest.yaml` sets `row_strategy: direct` on **every** table for clarity, even on the singleton `me` table. Picking one style across all 8 tables here would make the manifest a bit easier to scan. Up to you — definitely not a blocker.

sources/community/snyk/manifest.yaml
```diff
@@ -0,0 +1,573 @@
...
+      Snyk API Token (Personal Access Token or Service Account token).
+      Create one in Snyk Account Settings. The token is sent as `Authorization: Token <SNYK_API_TOKEN>`.
+auth:
+  type: HeaderAuth
+  headers:
+    - name: Authorization
+      from: template
+      template: Token {{input.SNYK_API_TOKEN}}
+test_queries:
+  - SELECT * FROM snyk.current_user LIMIT 1
```
antonmry, line 21: Suggestion (entirely optional): only `current_user` is exercised by `test_queries`. Adding `SELECT * FROM snyk.organizations LIMIT 1` would also exercise the `rows_path: [data]` + cursor pagination path, which would catch any JSON:API shape mismatches earlier than a real-world query would. Cal.com's `test_queries` follow this pattern — they only include parameter-less tables so the queries are runnable as-is.

sources/community/snyk/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Get Your API Token
+1. Log into Snyk.
+2. Go to **Account Settings > General**.
+3. Generate a Personal Access Token (or use a Service Account token).
+
+### Add the Source
+```bash
+coral source add snyk
+```
```
antonmry, line 15: Tiny docs nit — please ignore if `coral source add snyk` is actually a supported alias I don't know about. Other community sources use the explicit `--file` form, e.g. from `sources/community/cal/README.md`:

```bash
CAL_API_KEY=cal_live_... coral source add --file sources/community/cal/manifest.yaml
```

If users running through the README don't have a published source registry to resolve a bare `snyk` name from, the bare-name form might fail. Matching the cal.com style would be:

```bash
SNYK_API_TOKEN=... coral source add --file sources/community/snyk/manifest.yaml
```

sources/community/snyk/manifest.yaml
```diff
@@ -0,0 +1,573 @@
...
+    description: Vulnerabilities and license issues in Snyk projects.
+    fetch_limit_default: 100
+    filters:
+      - name: org_id
+        required: true
+      - name: project_id
+        required: true
+      - name: severity
+      - name: type
+    request:
```
antonmry, line 360: For context (not strictly a blocker for v1): the `issues` API actually exposes a richer filter set that I think users will reach for almost immediately on a security-scan source:

- `effective_severity_level` (array — see the related comment on the `severity` filter)
- `status`
- `ignored` (boolean)
- `created_after` / `created_before` / `updated_after` / `updated_before` — particularly important for "issues opened in the last 7 days" style queries
- `scan_item.id` / `scan_item.type` (used together; lets users scope by target ref instead of by project)

Adding even the time-window filters to v1 would meaningfully improve query ergonomics. Up to you whether to land it now or as a follow-up.

jsummerfield (commented):
Source review finding:

- Blocking: this PR does not currently have the `cla signed` label. The repo contribution rules require the CLA before a community source can be approved. Please complete the CLA flow and get the label applied, then this can receive a source-spec content review.

jsummerfield (commented):
Findings:

- High: this branch currently deletes 9 unrelated files under `sources/community` when compared with current `main`. A community-source PR should add or update its own source directory without removing existing sources. Please merge/rebase current `main` and restore these unrelated files before this can be approved. Examples from the diff:

- `sources/community/cloudflare/README.md`
- `sources/community/cloudflare/manifest.yaml`
- `sources/community/sendgrid/README.md`
- `sources/community/sendgrid/manifest.yaml`
- `sources/community/shopify/README.md`
- `sources/community/shopify/manifest.yaml`

jsummerfield:
Thanks for the work here. I ran source lint on the current head and the manifest validates. The PR is still blocked on the missing CLA label, and the existing inline review comments cover the larger provider-surface issues I see: Snyk body-link pagination does not map cleanly to Coral's `cursor_query`, `issue_paths` / `dependencies` look like v1 or nonexistent REST endpoints, and several issue filters/columns don't match the current REST shape.

One additional setup issue to fix while you are revising: the manifest hardcodes `base_url: https://api.snyk.io/rest`, but Snyk's REST docs say users must call the URL for their region: US-01, US-02, EU-01, or AU-01. That means EU/AU/US-02 customers would install the source successfully and then get failed or empty queries against the wrong Snyk region. Please add a `SNYK_API_BASE_URL` / region input with a sane default, and document it in the README. Reference: https://docs.snyk.io/snyk-api/rest-api/about-the-rest-api#api-urls

PR #573, opened 2026-05-18
docs(plugins/coral): clarify skill table references

mystic123:
Migrates withcoral/skills#14 into the canonical in-repo Coral skills tree.

## Summary
- Clarifies that agents should use a table's `sql_reference` when available.
- Preserves the manual `schema_name` plus `table_name` guidance for cases where a SQL reference has to be assembled from metadata.
- Keeps the quoted-qualified-name warning close to the Coral skill query rules.

## Validation
- `git diff --check`
- `RUSTC_WRAPPER= cargo run --locked -p xtask -- export-skills --dest /private/tmp/coral-skills-export-source-507`

### DIFF
--- a/plugins/coral/skills/coral/SKILL.md
+++ b/plugins/coral/skills/coral/SKILL.md
@@ -33,7 +33,7 @@ Use this as the Coral entrypoint for external context. Query Coral before answer
 
 ## Query Rules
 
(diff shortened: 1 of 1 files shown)


simonwhitaker (approved):

jsummerfield:
what's the plan here @mystic123

mystic123:
> what's the plan here @mystic123

​merged now

PR #558, opened 2026-05-17
feat(sources/community/google_calendar): add Google Calendar community source

mohitjeswani01:
## Summary

Fixes #397

Add a new `google_calendar` community source that lets Coral query Google Calendar data through SQL. This makes calendar lists, calendar metadata, and events queryable via standard SQL — enabling personal planning agents, scheduling workflows, and cross-source joins with tools like Notion or Linear.

This change is manifest-only and uses Coral's existing HTTP source-spec model. It adds read-only Google Calendar visibility without changing CLI, MCP, config, runtime, or bundled-source contracts.

## What changed

Added:
- `sources/community/google_calendar/manifest.yaml`
- `sources/community/google_calendar/README.md`

## Source scope

Inputs:
- `GOOGLE_CALENDAR_ACCESS_TOKEN` — OAuth2 bearer token (secret)
- `GOOGLE_CALENDAR_CALENDAR_ID` — optional calendar selector, defaults to `primary`

Tables:
- `google_calendar.calendar_lists`
  - `GET /users/me/calendarList`
  - lists all calendars visible to the authenticated user
  - cursor pagination via `pageToken` / `nextPageToken`

- `google_calendar.calendars`
  - `GET /calendars/{calendar_id}`
  - requires `calendar_id` filter
  - returns single-row calendar metadata via `row_strategy: direct`

- `google_calendar.events`
  - `GET /calendars/{calendar_id}/events`
  - requires `calendar_id` filter
  - optional `time_min`, `time_max`, `single_events` filters
  - cursor pagination, up to 2500 results per page
  - flattens `start__date_time`, `end__date_time`, `creator__email`, `organizer__email` with double-underscore convention
  - keeps `attendees`, `conference_data`, `reminders` as `Json`

## Implementation notes

- uses Google Calendar API v3 bearer auth (`Authorization: Bearer <token>`)
- keeps v1 intentionally read-only
- `time_min` and `time_max` are `virtual: true` — consumed as query params, not in response body
- `single_events` echoes via `from_filter` following codebase convention
- ISO 8601 datetime fields use `format_timestamp: iso8601` for real `Timestamp` columns
- known limitation: OAuth2 access tokens expire ~1 hour; README documents two methods to obtain a token

## Validation

Local validation completed:
- `python3 -c "import yaml; doc=yaml.safe_load(open('sources/community/google_calendar/manifest.yaml')); print([t['name'] for t in doc['tables']])"` — passed
[...]

PR #582, opened 2026-05-18
feat(sources/community/gmail): add Gmail community source

RajdeepKushwaha5:
## Summary

Adds a Gmail community source backed by the Gmail REST API v1.

- `gmail.labels` — all labels (system and custom) with message and thread counts. System labels include `INBOX`, `SENT`, `UNREAD`, `IMPORTANT`, `STARRED`. Use `id` values as the `label_ids` filter on `gmail.messages`.
- `gmail.messages` — message list via Gmail search syntax. Returns `id` and `thread_id` only — the Gmail list endpoint does not return headers or body content. Use `gmail.message_details` for subject, sender, date, and snippet.
- `gmail.message_details` — full metadata for a single message (`subject`, `from_email`, `to_email`, `date`, `snippet`, `label_ids`, `internal_datetime`) fetched with `format=metadata`. Requires `message_id` filter.
- `gmail.threads` — thread list via Gmail search syntax, returning `id`, `snippet`, and `history_id`.
- `gmail.thread_messages` — all messages inside one thread with full metadata (subject, from, date, snippet). Requires `thread_id` filter.
- `gmail.search_messages(q)` — source-scoped table function for Gmail search, returning `id` and `thread_id` pairs. Useful for agent workflows before drilling into `message_details`.

Auth: Google OAuth2 access token via `GOOGLE_ACCESS_TOKEN` (`Bearer` header).
Scope required: `https://www.googleapis.com/auth/gmail.readonly`

> **Note:** Google Calendar was originally included in this PR but has since been merged separately in #558. This PR now covers Gmail only.

### DIFF
--- a/sources/community/gmail/README.md
+++ b/sources/community/gmail/README.md
@@ -1,0 +1,143 @@
+# Gmail
+
+**Version:** 1.0.0
+**Backend:** HTTP
+**Base URL:** `https://gmail.googleapis.com`
+
+Query Gmail labels, messages, threads, and message metadata as SQL tables. Search your inbox with Gmail's native search syntax, drill into individual messages, and join with other Coral sources for cross-source intelligence.
+
+## Tables
+
+| Table | Description | Required filters | Optional filters |
+|-------|-------------|-----------------|-----------------|
+| `gmail.labels` | Gmail system and user labels | none | none |
+| `gmail.messages` | Message search and list | none | `q`, `label_ids`, `max_results` |
+| `gmail.message_details` | Full metadata for one message | `message_id` | none |
+| `gmail.threads` | Gmail conversation threads | none | `q`, `label_ids` |
+| `gmail.thread_messages` | All messages inside one thread | `thread_id` | none |
+
+## Source-Scoped Table Functions
+
+| Function | Description |
+|----------|-------------|
+| `gmail.search_messages(q => '...')` | Pushes Gmail search syntax into the API and returns compact candidate IDs |
+| `gmail.get_message(message_id => '...')` | Fetches metadata for one message by ID |
+| `gmail.get_thread_messages(thread_id => '...')` | Expands one conversation into message rows |
+
+## Authentication
+
+Requires a `GOOGLE_ACCESS_TOKEN` with the `https://www.googleapis.com/auth/gmail.readonly` scope.
+
+**Quickest way to get a token (OAuth Playground):**
+
+1. Go to [developers.google.com/oauthplayground](https://developers.google.com/oauthplayground)
+2. Click the gear icon and check **Use your own OAuth credentials** — paste your Client ID and Secret
+3. Select the scope: `https://www.googleapis.com/auth/gmail.readonly`
+4. Click **Authorize APIs**, sign in, then **Exchange authorization code for tokens**
+5. Copy the **Access token**
+
+If you are connecting Gmail and Google Calendar together, authorize both scopes in the same flow:
+
+```
+https://www.googleapis.com/auth/gmail.readonly
+https://www.googleapis.com/auth/calendar.readonly
+```
+
+One token covers both connectors.
+
+**Using gcloud (recommended for daily use):**
+
+```bash
+gcloud auth login
+export GOOGLE_ACCESS_TOKEN=$(gcloud auth print-access-token)
+```
+
+Access tokens expire after 1 hour. Re-run `gcloud auth print-access-token` to get a fresh one.
+
+## Install
+
+```bash
+coral source lint manifest.yaml
+coral source add --file manifest.yaml
+coral source test gmail
+```
+
+Or with the token inline:
+
+```bash
+GOOGLE_ACCESS_TOKEN=ya29.your-token coral source add --file manifest.yaml
+```
+
+## Example Queries
+
+List all labels:
+
+```sql
+SELECT id, name, type, messages_unread
+FROM gmail.labels
+ORDER BY name
+LIMIT 20;
+```
+
+Search recent unread messages:
+
+```sql
+SELECT id, thread_id
+FROM gmail.messages
+WHERE q = 'is:unread newer_than:7d'
+LIMIT 20;
+```
+
+Same search using a table function:
+
+```sql
+SELECT id, thread_id
+FROM gmail.search_messages(q => 'is:unread newer_than:7d')
+LIMIT 20;
+```
+
+Fetch metadata for one message:
+
+```sql
+SELECT subject, from_email, to_email, date, snippet
+FROM gmail.message_details
+WHERE message_id = 'MESSAGE_ID_HERE';
+```
+
+Expand one thread into rows:
+
+```sql
+SELECT subject, from_email, snippet, internal_datetime
+FROM gmail.thread_messages
+WHERE thread_id = 'THREAD_ID_HERE'
+ORDER BY internal_datetime ASC;
+```
+
+## Cross-Source JOIN Example
+
+Inbox pressure vs. meeting load in one query (requires `google_calendar` source also installed):
+
+```sql
+WITH urgent_emails AS (
+    SELECT COUNT(DISTINCT id) AS urgent_unread
+    FROM gmail.messages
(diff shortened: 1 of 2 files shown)


mohitjeswani01:
@RajdeepKushwaha5 google calendar source spec is already merged https://github.com/withcoral/coral/pull/558 you can check this pull request

jsummerfield:
Thanks for the detailed Gmail source. `coral source lint` passes, and the two-step message/search shape matches Gmail’s API behavior. I found README blockers before approval.

1. The README documents functions that do not exist in the manifest: `gmail.get_message` and `gmail.get_thread_messages` (`sources/community/gmail/README.md:24`, `:25`, and again at `:140`). The manifest only defines `gmail.search_messages`; message and thread drill-downs are tables (`gmail.message_details`, `gmail.thread_messages`). Please remove or correct those references so users do not try to call missing functions.

2. The recommended `gcloud auth login` setup (`sources/community/gmail/README.md:48`) does not show requesting the `https://www.googleapis.com/auth/gmail.readonly` scope. Gmail’s message list API requires that scope or another Gmail scope: https://developers.google.com/workspace/gmail/api/reference/rest/v1/users.messages/list. Please either document an explicit scoped OAuth flow that produces a Gmail-capable access token, or remove the gcloud path as “recommended for daily use” so users do not install the source with a token that cannot read Gmail.

Once the README matches the actual source surface and auth requirements, I can re-review quickly.

sources/community/gmail/README.md
```diff
@@ -0,0 +1,143 @@
...
+| `gmail.threads` | Gmail conversation threads | none | `q`, `label_ids` |
+| `gmail.thread_messages` | All messages inside one thread | `thread_id` | none |
+
+## Source-Scoped Table Functions
+
+| Function | Description |
+|----------|-------------|
+| `gmail.search_messages(q => '...')` | Pushes Gmail search syntax into the API and returns compact candidate IDs |
+| `gmail.get_message(message_id => '...')` | Fetches metadata for one message by ID |
+| `gmail.get_thread_messages(thread_id => '...')` | Expands one conversation into message rows |
```
antonmry, line 25: Tiny thing — this README block lists three table functions, but only `search_messages` is defined in the manifest's `functions:` block. `gmail.get_message` and `gmail.get_thread_messages` aren't there, so calling them would fail at runtime.

I think the cleanest fix is to drop these two rows from the README (the `gmail.message_details` and `gmail.thread_messages` tables already cover the same shape). Could you take a look? Happy to be wrong if I'm reading the manifest incorrectly.

sources/community/gmail/manifest.yaml
```diff
@@ -0,0 +1,635 @@
...
+          default: 50
+    response:
+      rows_path:
+        - messages
+    pagination:
+      mode: cursor_query
+      page_size:
+        default: 50
+        max: 500
+        query_param: maxResults
```
antonmry, line 237: A small thing worth double-checking: the `messages` table sets `maxResults` twice — once via `filter_int` on L225-228, and again via the pagination block's `page_size.query_param: maxResults` on L234-237. I think this would emit two `maxResults=...` query params (or one would silently win), and the cursor pagination won't get to drive `maxResults` after the first page if the filter value is already pinned.

I think the intent is that pagination owns `maxResults`. If so, removing the explicit query entry (L225-228) and letting `page_size` drive it should be enough. Happy to be wrong here — this could just be a no-op in practice depending on how the engine merges them.

sources/community/gmail/manifest.yaml
```diff
@@ -0,0 +1,635 @@
...
+          value: "50"
+    response:
+      rows_path:
+        - threads
+    pagination:
+      mode: cursor_query
+      page_size:
+        default: 50
+        max: 500
+        query_param: maxResults
```
antonmry, line 456: Same observation as on the `messages` table: `maxResults` is hardcoded as a literal on L445-447 *and* set via `page_size.query_param: maxResults` on L453-456. I think dropping the literal entry and letting pagination drive it is the safer shape, otherwise the literal pins the page size and `page_size.default` / `max` never get to do anything.

Could you take a look and confirm whether this was intentional?

sources/community/gmail/manifest.yaml
```diff
@@ -0,0 +1,635 @@
...
+  retry_after_header: Retry-After
+
+test_queries:
+  - SELECT * FROM gmail.labels LIMIT 1
+  - SELECT id, thread_id FROM gmail.messages WHERE q = 'newer_than:30d' LIMIT 1
+  - SELECT id, snippet FROM gmail.threads WHERE q = 'newer_than:30d' LIMIT 1
+  - SELECT function_name FROM coral.table_functions WHERE schema_name = 'gmail' ORDER BY function_name LIMIT 1
+
+functions:
+  - name: search_messages
```
antonmry, line 47: A judgment call rather than a bug: `search_messages` and the `gmail.messages` table both hit `GET /gmail/v1/users/me/messages` with the same `q` / `label_ids` / `max_results` inputs and project the same `id` + `thread_id` columns. So `SELECT … FROM gmail.search_messages(q => '…')` and `SELECT … FROM gmail.messages WHERE q = '…'` are functionally identical.

That makes the function a stylistic alias rather than something that unlocks new behavior. Two options I can see:

1. Drop the function, since `gmail.messages` already provides the same surface.
2. Keep it but give it a distinct shape — e.g., a different return projection or a friendlier guide for the "pre-drilldown ID candidates" workflow it's pitched as.

I'm leaning toward option 1 (fewer ways to do the same thing → easier mental model for agents), but happy to defer to your judgment. What was the intent here?

sources/community/gmail/manifest.yaml
```diff
@@ -0,0 +1,635 @@
...
+            - threadId
+      - name: q
+        type: Utf8
+        nullable: true
+        description: Gmail search query filter (virtual)
+        expr:
+          kind: from_filter
+          key: q
+        virtual: true
+      - name: label_ids_filter
```
antonmry, line 266: Very small nit: the virtual filter-echo columns are named inconsistently across this manifest — `messages.q` echoes as `q` (L258), `messages.label_ids` echoes as `label_ids_filter` (here, L266), and `thread_messages.thread_id` echoes as `thread_id_filter` (L628). Mixing both conventions makes it harder for an agent to predict what to select.

I'd lean toward dropping the `_filter` suffix everywhere (so `label_ids` and `thread_id`) since it matches the filter name directly and that's what other community sources tend to do — but either convention is fine, just consistent. Totally optional, won't block.

antonmry:
Thanks for the careful work on this one — the table set is well-chosen and the two-step "list IDs then fetch details" pattern is exactly the right shape for an LLM client.

One process note up-front: this is an OAuth-backed source, and we're currently discussing internally whether to accept OAuth-based community sources before we have first-class support for OAuth flows (token exchange, refresh, client credentials) in the Coral DSL. Today the user has to manually paste a bearer token from `gcloud auth print-access-token` (or the OAuth Playground) and refresh it every hour, which is workable but not great UX. The merged `google_calendar` source (#558) set a precedent for accepting this pattern, so there's a real argument either way. I don't want to block your work on a policy decision that's still in flight — happy to keep iterating on the technical findings I've left inline in parallel.

I've also added a few inline comments — most are small (a redundant `maxResults` on two tables, a couple of README functions that aren't in the manifest, a naming inconsistency on the virtual filter columns, and a question about whether the `search_messages` function adds anything over the `messages` table). None of them are deal-breakers, just things worth a second look.

Will come back to this once the OAuth policy decision is settled.

antonmry:
Hi @mohitjeswani01 @RajdeepKushwaha5! We have decided not to accept community PRs using OAuth until we have support for it in Coral. We should have it soon and, without it, it's a bit too hacky. We've already rejected PRs for those reasons, and accepting the google calendar contribution was a misalignment on our side. Happy to reopen the PR once we have OAuth support in place. Thanks!

mohitjeswani01:
understood  @antonmry  thanks for describing !

RajdeepKushwaha5:
hi sir @antonmry 
could you please re open this PR as coral has support for OAuth

PR #603, opened 2026-05-19
feat(community/vercel): add vercel community source integration

praveshhh:
### 🚀 Motivation and Context
This PR adds a brand new **Vercel Community Source** (`sources/community/vercel`) to the Coral ecosystem. 

Vercel is the hosting platform of choice for millions of developers. Adding it as a community source enables developers, platform engineers, and DevOps teams to write expressive, real-time SQL queries auditing their Vercel resources and correlating them with other sources like GitHub or Datadog directly in Coral.

---

### 📦 Declarative Tables Included

1. **`vercel.projects`** (GET `/v9/projects`)
   * Fields: `id`, `name`, `framework`, `node_version`, `account_id`, `created_at`, `updated_at`
   * Supports querying project framework types and Node version settings.

2. **`vercel.deployments`** (GET `/v6/deployments`)
   * Fields: `id`, `name`, `url`, `state`, `creator_username`, `created_at`
   * Scopes queries using an optional `project_id` filter. Enables deployment status tracking (e.g. finding recent deployment failures).

3. **`vercel.domains`** (GET `/v5/domains`)
   * Fields: `name`, `verified`, `created_at`, `expires_at`
   * Simplifies custom domain routing and verification state audits.

---

### 🔮 Example SQL Queries Enabled

```sql
-- Find degraded Vercel projects configured with Next.js
SELECT name, framework, node_version, created_at
FROM vercel.projects
WHERE framework = 'nextjs';

-- Audit recent deployment failures
SELECT name, url, creator_username, created_at
FROM vercel.deployments
WHERE state = 'ERROR'
ORDER BY created_at DESC;

-- List all verified custom domains configuration
SELECT name, verified, expires_at
FROM vercel.domains
WHERE verified = true;

### CODE REVIEW

**Comment by jsummerfield** [comment]
Thanks for adding this. I cannot approve this version as-is because it would give users an incomplete Vercel source.

There is already an approved open PR for the same `sources/community/vercel` source (#434), and this PR currently regresses a few important pieces from that shape. The main blocker is pagination: the manifest declares only `fetch_limit_default` for `projects`, `deployments`, and `domains` (`sources/community/vercel/manifest.yaml:44`, `:120`, `:187`), and the README says there is no pagination support (`sources/community/vercel/README.md:114`). Vercel documents paginated list responses for these APIs, including current `v10` projects and pagination metadata for deployments/domains:

- https://vercel.com/docs/rest-api/reference/endpoints/projects/retrieve-a-list-of-projects
- https://vercel.com/docs/rest-api/reference/endpoints/deployments/list-deployments
- https://vercel.com/docs/rest-api/reference/endpoints/domains/list-all-the-domains

Please either coordinate around the already-approved Vercel PR or update this source to match the current Vercel endpoints and implement cursor pagination so larger accounts are not silently truncated. Once the source behavior matches provider truth, I can re-review quickly.

File: sources/community/vercel/manifest.yaml [hunk 1]
```diff
@@ -0,0 +1,231 @@
...
+    fetch_limit_default: 20
+    request:
+      method: GET
+      path: /v9/projects
+      query:
+        - name: teamId
+          from: input
+          key: VERCEL_TEAM_ID
+    response:
+      rows_path: [projects]
```
antonmry, line 53: A correctness gap worth flagging: none of the three tables in this manifest have a `pagination:` block, but Vercel's REST APIs (`/v9/projects`, `/v6/deployments`, `/v5/domains`) are all paginated via a `pagination.next` cursor in the response (it's a Unix-ms timestamp passed as `until=` on the next request). Without pagination, any account with more than ~20 rows will silently truncate, which is a quiet correctness bug — `SELECT * FROM vercel.projects` on a non-trivial team will miss most projects.

Combined with `fetch_limit_default: 20`, this looks like a "sensible page size" but is actually a hard cap.

For reference, the response shape is something like:
```json
{ "projects": [...], "pagination": { "next": 1234567890, "count": 20 } }
```

Could you add a pagination block (likely `mode: cursor_query`, `cursor_param: until`, `response_cursor_path: [pagination, next]`) on each of the three tables? Happy to be wrong if I'm misreading the Vercel docs.

sources/community/vercel/manifest.yaml
```diff
@@ -0,0 +1,231 @@
...
+      List Vercel projects. Use the projects table to query framework types,
+      repository links, and latest deployment statuses.
+    fetch_limit_default: 20
+    request:
+      method: GET
+      path: /v9/projects
+      query:
+        - name: teamId
+          from: input
+          key: VERCEL_TEAM_ID
```
antonmry, line 51: Small thing — `teamId` is sent on every request, even when `VERCEL_TEAM_ID` has its default empty-string value. I haven't tested it, but I'd expect Vercel to either reject `teamId=` or return data in a way that surprises someone using a personal account.

The usual pattern in other community sources is to skip emitting the query param when the value is empty. If `from: input` doesn't already do that, you might need a small conditional or to require `VERCEL_TEAM_ID` only on team-context tables. Could you double-check what Vercel does with `teamId=`? If it's tolerated, this is a no-op.

(Same pattern on `deployments` L125-127 and `domains` L192-194.)

sources/community/vercel/manifest.yaml
```diff
@@ -0,0 +1,231 @@
...
+        expr:
+          kind: path
+          path: [nodeVersion]
+      - name: account_id
+        type: Utf8
+        nullable: false
+        description: The ID of the owner account.
+        expr:
+          kind: path
+          path: [accountId]
```
antonmry, line 89: Minor: `account_id` is marked `nullable: false`, but the `accountId` field is mainly meaningful in team contexts — on personal-account projects the response may not include it (or may include `null`). If a personal-account user hits this and `accountId` is missing/null, a `nullable: false` column will fail the row.

Worth setting `nullable: true` here unless you've tested both personal and team contexts and confirmed `accountId` is always populated. Same kind of check would apply to `creator_username` on `deployments` (L160ish) — system-triggered deployments may not have a creator.

Could you take a look?

antonmry:
Thanks for putting this together! Before diving into the details, one process request:

This PR is titled "add vercel community source integration" but the diff actually bundles two unrelated things — the vercel manifest (≈345 lines under `sources/community/vercel/`) and a much larger CommanderX hackathon project (≈5000 lines under `examples/commander-x/` plus `start.ps1` at the repo root). The CommanderX code is mostly simulated/mock infrastructure (sqlite session store, mock SSH/RDS, mock logs) that isn't really a Coral integration example yet — it pitches itself as "this is what a real Coral-backed agent *would* look like in the future" rather than something that exercises Coral today.

Could you remove the CommanderX files (`examples/commander-x/**` and `start.ps1`) from this PR so it's just the vercel source? Mixed-purpose PRs are very hard to review coherently — reviewers end up either rubber-stamping the bundle or rejecting both, and neither outcome is fair to the work you put in. A focused PR with only the vercel manifest + README will get a much faster review.

I've left a few inline comments on the vercel manifest itself for when you do the cleanup pass:
- Pagination is missing on all three tables (`projects`, `deployments`, `domains`) — Vercel's APIs paginate via a `pagination.next` cursor, and right now any non-trivial account will silently truncate at the `fetch_limit_default: 20` page.
- `VERCEL_TEAM_ID` is always sent as a query param, even when empty — worth confirming Vercel tolerates `teamId=`.
- A couple of `nullable: false` columns (`projects.account_id`, possibly `deployments.creator_username`) that may legitimately be null in personal-account contexts.

The vercel surface itself is reasonable — three sensible tables, clean `format_timestamp` usage on the ms-epoch fields. Mostly cleanup, not a redesign.

One other small note: the branch is `main`. Working on a feature branch (e.g., `feat/vercel-source`) makes it much easier for both you and reviewers to iterate without affecting your fork's `main`.

Looking forward to the next revision!

jsummerfield:
We have a Vercel community source already. Please extend or improve that rather than raise a new source. https://github.com/withcoral/coral/tree/main/sources/community/vercel

PR #605, opened 2026-05-19
feat(engine): execute dependent joins

iaptsiauri:
# Execution Core Slice

This PR is the third PR in the dependent predicate pushdown stack. The full stack overview is documented in #399.

Base: `iaptsiauri/predicate-push-down-planner`
Head: `iaptsiauri/predicate-push-down-exec-core`

## What This PR Adds

This PR turns the planner rewrite into an executable physical plan:

- binding tuple representation and projection from resolver rows
- dependent HTTP fetch dispatch through the existing HTTP client path
- local joined output construction
- DataFusion physical `DependentJoinExec`
- extension planner lowering from the logical dependent join node
- resolver partition consumption
- core execution coverage for distinct binding tuples, composite keys, duplicate bindings, null bindings, swapped join sides, projected dependent tables, and resolver-side HTTP inputs

This PR is intended to establish the basic execution path. It does not contain the later hardening work around HTTP completeness, soft resolver-overflow fallback, trace propagation, and final cap semantics. Those are handled in #606.

## Stack Position

Review order:

## Reviewer Notes

For this PR, focus on whether the physical operator correctly implements the planned dependent join shape and preserves SQL-correct local join completion for the supported V1 cases. More defensive behavior and production hardening are intentionally split into the next PR.


crates/coral-engine/tests/engine/dependent_join_execution_tests.rs
```diff
@@ -0,0 +1,1243 @@
...
+        .expect("query should succeed");
+
+    assert_eq!(
+        execution_to_rows(&execution),
+        vec![
+            json!({ "issue_title": "Duplicate tuple", "pr_state": "open" }),
+            json!({ "issue_title": "First", "pr_state": "open" }),
+        ]
+    );
+}
```
antonmry, line 2344: A question rather than a request — and I might be missing something here, so happy to be wrong.

Did you consider adding a **differential test** that runs the same SQL on the same fixture data **twice** — once with `bindable: true` on the relevant filters (so DPP fires) and once without (so the engine falls back to a naïve scan + hash join) — and asserts the outputs match? Something roughly like:

```rust
async fn dpp_and_naive_paths_agree(sql: &str, fixture: ...) {
    let with_dpp = run(sql, fixture, bindable_manifest()).await?;
    let without_dpp = run(sql, fixture, non_bindable_manifest()).await?;
    assert_eq!(sort_rows(with_dpp), sort_rows(without_dpp));
}
```

My thinking: DPP is sold as a transparent optimization — same SQL semantics, just fewer HTTP calls. The current suite is rigorous about each path *individually* — `assert_dependent_join_query` here pins down the DPP-on output against a hardcoded oracle, and the fallback tests (L608-755) do the same for the DPP-off path. But the two paths are never **compared against each other** for the same query on the same data. The "transparent optimization" promise is locked in only by the author's belief about what SQL says, not by the engine's other execution path agreeing.

This matters for a handful of corner cases I think a differential test would catch automatically:

- **Cardinality on duplicate dependent rows under one binding**: does DPP's resolver-row × dependent-row product match DataFusion's hash-join product when one binding returns multiple rows?
- **NULL handling**: DPP's `BindingProjector::project` drops rows with null binding columns; DataFusion's inner equi-join drops them via `NULL ≠ NULL`. Equivalent in principle, but a fixture with mixed-null bindings + the same data on both sides would verify it for every row count.
- **Output ordering**: DPP emits batches in `HashMap<Tuple, _>` iteration order, which is non-deterministic. The hash-join path has its own ordering. With `ORDER BY` both should produce the same final ordering — worth proving.
- **Aggregation over the join**: `SELECT COUNT(*) FROM ... JOIN ... USING (k)` — does the count match?

The fallback tests (`unsupported_join_shape_falls_back_to_regular_join_execution` etc.) almost have the structure of a differential test, except they use **different manifests** than the DPP tests — different URL paths, different mock response shapes, different `filters` lists. So even with a careful read you're comparing two different oracles, not the two engine paths against each other.

One small fixture run two ways (or a config flag that suppresses DPP for an otherwise-identical manifest) would future-proof the suite against any subtle DPP-only regression — which feels especially valuable given #606 changes the cap shape and #607 turns this on in real source manifests.

Totally happy to defer if you considered this and decided it wasn't worth the harness work, or if there's a separate differential test layer I haven't found. What do you think?

iaptsiauri: This is something which I definitely overlooked, but super nice suggestion. I will add this to the next PR in the stack. As it has all bells and whistles. I did run those manually but they for sure need to be part of the harness

antonmry (approved):

docs/reference/source-spec-reference.mdx
AlbertQM: Don't need to document that IMO :P 

```suggestion
```

AlbertQM (approved):

crates/coral-engine/src/runtime/dependent_join/exec.rs
```diff
@@ -0,0 +1,376 @@
...
+    page_hint: Option<usize>,
+    metrics: DependentJoinMetrics,
+    output_schema: SchemaRef,
+) -> Result<Vec<RecordBatch>> {
+    let mut resolver_batches = Vec::new();
+
+    for resolver_partition in 0..resolver_partition_count {
+        let mut resolver_stream = resolver.execute(resolver_partition, Arc::clone(&context))?;
+        while let Some(batch) = resolver_stream.next().await.transpose()? {
+            resolver_batches.push(batch);
```
simonwhitaker: [codex] The executor materializes all resolver batches before max_resolver_rows / max_bindings are checked in the binding phase. A huge resolver side can blow memory or fail very late despite the configured cap.

crates/coral-engine/src/runtime/dependent_join/fetcher.rs
```diff
@@ -0,0 +1,174 @@
...
+            &tuple,
+        )?;
+        let filter_values = filters
+            .iter()
+            .map(|(k, v)| (k.clone(), v.clone()))
+            .collect();
+        let target = http_target_for_filters(&self.table, &filter_values);
+        let rows = self
+            .client
+            .fetch(&target, &filter_values, &HashMap::new(), self.page_hint)
```
simonwhitaker: [codex] client.fetch(...) can collect paginated rows first, then rows.len() is checked. The cap prevents returning too many rows, but not fetching/storing too many first.

crates/coral-engine/src/runtime/dependent_join/output.rs
```diff
@@ -0,0 +1,157 @@
...
+    for tuple in state.binding_tuples() {
+        let Some(rows) = state.buffered_rows_for_tuple(tuple) else {
+            continue;
+        };
+
+        if rows.is_empty() {
+            continue;
+        }
+
+        let filter_values = filter_values_for_tuple(binding_filters, tuple)?;
```
simonwhitaker: [codex] Dependent fetch uses literal filters plus binding filters, but output reconstruction only passes binding-derived filters to convert_items. Columns using from_filter / object_filter_path for literal filters can become null or wrong.

crates/coral-engine/src/runtime/dependent_join/bindings.rs
```diff
@@ -0,0 +1,102 @@
...
+        Self { keys }
+    }
+
+    pub(crate) fn project(&self, batch: &RecordBatch, row: usize) -> Result<Option<Tuple>> {
+        let mut values = Vec::with_capacity(self.keys.len());
+
+        for key in self.keys.iter() {
+            let index = batch
+                .schema()
+                .index_of(&key.resolver_column.name)
```
simonwhitaker: [codex] BindingKey keeps a qualified Column, but execution looks up by bare field name. If the resolver plan has duplicate names like a.owner and b.owner, it may bind from the wrong column

simonwhitaker (approved):

PR #606, opened 2026-05-19
fix(engine): harden dependent join execution

iaptsiauri:
# Execution Hardening Slice

This PR is the fourth PR in the dependent predicate pushdown stack. The full stack overview is documented in #399.

Base: `iaptsiauri/predicate-push-down-exec-core`
Head: `iaptsiauri/predicate-push-down-exec-hardening`

## What This PR Adds

This PR hardens the dependent join execution path introduced in #605:

- HTTP filter-consumption checks so dependent rewrites do not silently drop literal or binding filters
- preservation of literal filters in dependent output mapping
- dependent fetch completeness fixes around pagination and `fetch_limit_default`
- cap and error semantics for binding fanout, resolver buffering, per-binding dependent rows, and resolver rows per binding
- first dependent fetch error handling and dispatch cancellation behavior
- concurrency cap coverage
- `EXPLAIN` / `EXPLAIN ANALYZE` dependent join metrics
- resolver-row overflow retry without the dependent join rewrite
- app-to-engine trace context propagation so fallback/retry work stays attached to the query trace
- source-spec cleanup that removes manifest-level dependent-join caps from the final API shape

This PR intentionally does not enable any real source manifests. Pilot manifest enablement is isolated in #607.

## Stack Position

Review order:

## Reviewer Notes

For this PR, focus on correctness boundaries: which errors are hard failures, which resolver-row overflow case is allowed to fallback, whether HTTP fetches remain complete, and whether trace/fallback behavior preserves the intended execution semantics.


crates/coral-engine/src/runtime/dependent_join/state.rs
```diff
@@ -39,40 +40,76 @@ impl DependentJoinRuntimeState {
...
-        let row_count = batch.num_rows();
         self.resolver_batches.push(batch.clone());
+        self.resolver_rows = observed;
 
         let mut new_tuples = Vec::new();
 
-        for row_idx in 0..row_count {
-            let Some(tuple) = projector.project(batch, row_idx)? else {
+        for (row_idx, tuple) in projected_tuples.into_iter().enumerate() {
+            let Some(tuple) = tuple else {
```
antonmry, line 60: A small observability suggestion — feel free to disagree, this is more of a "would help me debug" wish than a correctness ask.

When a resolver row has a null in any binding column, `BindingProjector::project` returns `None` and this `continue` silently drops the row from the binding set. That's correct semantics for an inner equi-join (NULL ≠ NULL), but from a user-debugging perspective it's invisible: `resolver_rows` is only incremented for rows that *survived* the null check, so a query whose resolver returned 100 rows but had 30 with null binding columns reports `resolver_rows=70` with no signal that 30 were dropped.

If someone asks "why does my JOIN result have fewer rows than I expected?" the answer "30 of your resolver rows had nulls in binding columns" is invisible to them today. They'd have to instrument the resolver query separately to find out.

A `null_binding_rows_skipped` counter, recorded next to the existing four in `DependentJoinMetrics`, would close this gap with very little code:

```rust
// in state.rs
pub(crate) fn null_binding_rows_skipped(&self) -> usize { ... }

// in exec.rs DependentJoinMetrics
null_binding_rows_skipped: Count,

// in record()
self.null_binding_rows_skipped.add(state.null_binding_rows_skipped());
```

Optional — but I think this is the single most actionable missing metric for "why are some rows missing from my join output?" debugging. What do you think?

crates/coral-engine/src/runtime/query.rs
```diff
@@ -151,20 +229,69 @@ impl QueryRuntimeAdapter {
...
+
+                tracing::warn!(
+                    target = "coral_engine::dependent_join",
+                    source = %cap_error.source_schema,
+                    table = %cap_error.table,
+                    observed = cap_error.observed,
+                    cap = cap_error.cap,
+                    disposition = "fallback",
+                    "dependent join resolver row cap exceeded",
+                );
```
antonmry, line 297: This structured warn log is great — the `disposition = "fallback"` tag is exactly the kind of grep-able marker that makes incident debugging fast, and source/table/observed/cap give you everything you need to find the offending query.

One small observability suggestion: in addition to the log, would you consider exposing fallback as a **counter metric** somewhere? Something like a `dependent_join_resolver_overflow_fallback_total` counter on the runtime adapter (or even an additional Count on `DependentJoinMetrics`).

The use case: for SLO/dashboard purposes you want to be able to graph "how often did fallback fire this hour" without parsing structured logs. Logs answer "show me this incident"; counters answer "what's the rate over time". They're complementary.

If you'd rather defer this to a future PR (or feel it belongs in a separate metric layer the engine doesn't own yet), totally fine — wanted to flag it now since this is the only DPP-specific signal that lives in logs and nowhere else.

crates/coral-engine/src/runtime/dependent_join/driver.rs
```diff
@@ -1,35 +1,48 @@
...
 
     Ok(state)
 }
+
+fn spawn_fetch(
+    tasks: &mut JoinSet<Result<(Tuple, Vec<Value>)>>,
+    fetcher: BindingFetcher,
+    tuple: Tuple,
+) {
+    tasks.spawn(async move { fetcher.fetch_one(tuple).await }.instrument(tracing::Span::current()));
```
antonmry: The `.instrument(tracing::Span::current())` propagation here is a nice touch — it means each fetch task inherits the parent operator's span context, so downstream HTTP-client spans get correctly parented to the DPP execution. The end-to-end trace tree should look right.

Two small thoughts to consider — either or both, very happy to defer:

1. **Per-task span fields.** Right now the spawned task inherits the parent span but doesn't open its own. So in a trace, a 500-binding query produces 500 sibling HTTP spans under one DPP span, all visually identical — you can see "task 273 took 5s" from the span timeline, but you can't tell *which binding* was the slow one. Adding a child span with a few fields would help:

   ```rust
   tasks.spawn(
       async move { fetcher.fetch_one(tuple).await }
           .instrument(tracing::info_span!(
               "dependent_join_fetch",
               binding_index = tuple.index(),  // or similar
               // optionally: binding_keys = ?tuple
           ))
   );
   ```

2. **PII trade-off on binding values.** Binding values can carry sensitive content (Slack user IDs, internal customer IDs, email addresses). So I wouldn't suggest recording raw binding values as span fields by default. Reasonable middle grounds: just the binding *index* within the batch (always safe), or a stable fingerprint (`xxhash` of the tuple's wire-string form). That keeps the per-binding correlation without leaking the values themselves.

Totally a "nice to have" — the current shape is already a real improvement over #605. Just noting that "which binding was slow?" is still hard to answer from a trace alone.

crates/coral-engine/src/runtime/query.rs
```diff
@@ -51,11 +154,14 @@ pub(crate) async fn build_runtime(
...
     );
-    let session_state = SessionStateBuilder::new()
+    let mut builder = SessionStateBuilder::new()
         .with_config(session_config)
         .with_runtime_env(runtime_env)
-        .with_default_features()
-        .with_optimizer_rule(Arc::new(optimizer::rule()))
+        .with_default_features();
+    if dependent_join_enabled {
+        builder = builder.with_optimizer_rule(Arc::new(optimizer::rule()));
```
antonmry, line 179: Quick wish for a future iteration — totally fine to defer, especially if there's already a mechanism here I haven't found.

The DPP optimizer rule gets registered here, and (via the `instrument_rules_with_trace_spans!` wiring lower in this file) emits a span when it runs. So traces can answer "did the rule fire?" But the rule's *output* is the only signal: if `DependentJoinExec` appears in the resulting plan, you know DPP was applied; if it doesn't, you know it wasn't.

What's missing is **why the rule decided not to rewrite**. A user who writes a query they think should DPP and sees a naïve hash join in EXPLAIN has no way to find the reason short of reading `optimizer.rs`. The optimizer knows the answer (it just checked: "this isn't an equi-join", "binding type is Float64", "no bindable filter declared on column X", "filter is over-constrained", etc.) but discards that knowledge silently.

A `tracing::debug!` from each rejection branch in `optimizer.rs`, with structured fields like:

```rust
tracing::debug!(
    target = "coral_engine::dependent_join",
    reason = "not_inner_equi_join",
    join_kind = ?join.join_type,
    "skipping dependent join rewrite candidate"
);
```

…would give users (and us) a self-explaining audit trail when running with `RUST_LOG=coral_engine::dependent_join=debug`. Each rejection branch would have its own well-known reason string, and over time we'd build up a vocabulary of "why DPP didn't fire on my query" answers without anyone having to grep the optimizer source.

Not blocking — I haven't read every branch of the optimizer rule yet to know how many rejection points exist or whether some are already covered. But the "why didn't DPP fire?" question feels like one users will ask the moment #607 lands and they start writing real queries against the pilot manifests, and it's much cheaper to add reasons at each rejection point now than retroactively.

crates/coral-engine/src/runtime/dependent_join/fetcher.rs
```diff
@@ -46,67 +43,17 @@ impl BindingFetcher {
             table: config.table,
             binding_filters: config.binding_filters,
             literal_filters: config.literal_filters,
-            semaphore: Arc::new(Semaphore::new(max_concurrency)),
-            result_channel_capacity: max_concurrency,
+            max_concurrency,
```
antonmry, line 46: Stepping back from the implementation: `max_concurrency` ends up here verbatim from the source manifest's `rate_limit.max_concurrency`, with only a `Some(0)` rejection at parse time and a `.max(1)` floor at runtime. There's no upper bound. That puts a meaningful operational lever in the hands of the wrong actor, and I think it's worth a design conversation before this lands as part of the DPP stack — DPP makes this knob much more dangerous than it was for plain HTTP scans.

**My main suggestion — move it out of the source manifest into operator-owned config.**

The source manifest is authored by the *contributor adding a new source* (often a community PR; we just merged several this week). The contributor knows the API's *shape* — URLs, schemas, auth — but cannot know the *deployment's* resource constraints, file-descriptor limits, target-API abuse thresholds, or whether the operator is running 1 Coral instance or 100. They are not the right person to pick a concurrency number.

A community contributor who pastes `max_concurrency: 100` into a manifest because "GitHub seems fast" gets:
- 100 simultaneous TLS connections per DPP query against one host
- File-descriptor pressure on the Coral instance
- Almost certain abuse-detection trip from the upstream API (attributed to the operator's IP / API key, not the contributor's)
- A cap that's silently bounded to `max_bindings=500` today, but unbounded once those caps become configurable

The operator running Coral *does* know all of these. So this is fundamentally an **operator policy**, not source metadata.

Concretely, I think the right home is a future engine-side config (a `coral.toml` or `EngineConfig` struct with sensible defaults), structured along the lines of:

```toml
[engine.http]
default_max_concurrency = 8
max_max_concurrency = 64    # hard ceiling per-source

[engine.http.per_source]
github = { max_concurrency = 16 }   # operator-tuned override
```

That gives operators a real knob while keeping community manifest authors out of the resource-policy business. It also pairs naturally with the other DPP runtime caps (`max_bindings`, `max_resolver_rows_per_binding`, etc.) that the PR body promised to make "engine runtime policy, not source manifest metadata" — those should land in the same place.

**Fallback — if keeping it in the manifest is the right call for v1, add defensive validation now.**

If the constraint is "we don't have an operator-side config system yet and don't want to build one for this PR," that's fair, but at minimum two cheap defenses would close the immediate footgun:

1. **Hard upper-bound clamp** here at runtime, with a warn when clamping kicks in:
   ```rust
   const MAX_REASONABLE_CONCURRENCY: usize = 64;
   let max_concurrency = config.max_concurrency
       .max(1)
       .min(MAX_REASONABLE_CONCURRENCY);
   ```
   Engineering deliberate breaches of this ceiling should require an engine-side policy change, not a manifest-author one-liner.

2. **Parse-time upper bound** in `validate_rate_limit` (the function that already rejects `Some(0)`), so the bad manifest fails at `coral source add` with a clear error pointing at the field, rather than at query time inside DPP.

Even with the clamp+validation route, my preference is to file a tracking issue for "move HTTP runtime policy from source manifests to operator config." This is the kind of design boundary that only gets harder to move once we have community manifests in the wild relying on it.

What do you think — does the operator-vs-author framing match the team's thinking, or am I drawing the line in the wrong place?

crates/coral-engine/src/runtime/dependent_join/optimizer.rs
```diff
@@ -1,24 +1,29 @@
...
 use crate::backends::shared::filter_expr::literal_to_string;
 use crate::runtime::dependent_join::logical::{BindingKey, DependentJoinNode};
 
 const DEFAULT_MAX_BINDINGS: usize = 500;
 const DEFAULT_MAX_RESOLVER_ROWS: usize = 10_000;
-const DEFAULT_MAX_ROWS_PER_BINDING: usize = 50_000;
+const DEFAULT_MAX_ROWS_PER_BINDING: usize = 1_000;
+const DEFAULT_MAX_RESOLVER_ROWS_PER_BINDING: usize = 1_000;
 const DEFAULT_BINDING_CONCURRENCY: usize = 8;
+const BINDING_COLUMN_PREFIX: &str = "__coral_dj_bind_";
```
antonmry, line 26: These four constants define the entire DPP runtime policy surface today. I'd love to raise the design question explicitly before this stack lands — it pairs with the comment I left at `fetcher.rs` L46 about `max_concurrency`.

**Today's surface, summarized:**
- All four caps are compile-time constants here — no config path
- `max_concurrency` is the only one with a manifest path (`rate_limit.max_concurrency`), which I commented on separately
- No per-source override for any cap
- No per-source enable/disable
- No engine-wide toggle (see companion comment on `query.rs` L85)

For #607 and beyond, I think two combined ideas would close the operability gap — both grounded in the same observation that operators (not manifest authors, not the engine binary) should own DPP policy:

### 1. Move caps to `coral.toml` / `EngineConfig`

Same argument as the `max_concurrency` comment but applied to the other four:

```toml
[engine.dependent_join]
default_max_bindings = 500
default_max_resolver_rows = 10_000
default_max_rows_per_binding = 1_000
default_max_resolver_rows_per_binding = 1_000
default_max_concurrency = 8
```

Reading these from config (with the current constants as defaults) is what makes incident response possible: "we're hammering the GitHub API, raise their max_concurrency to 4 for the next hour" becomes a config push, not a binary release.

### 2. Per-source overrides — including enable/disable

I think this is the bigger ask. Today DPP is engine-wide on/off only. A real deployment will want:

```toml
[engine.dependent_join.per_source]
github = { max_bindings = 200, max_rows_per_binding = 5_000 }
linear = { enabled = false }    # disable DPP for this source specifically
slack  = { max_concurrency = 4 }
```

Why per-source matters:
- **Pilot rollout staging**: enable DPP on github for a week, watch dashboards, then add linear, then slack — not all-at-once via #607.
- **Incident isolation**: linear's API starts misbehaving → disable DPP on linear without touching github.
- **Per-API tuning**: GitHub's 5K req/hr authenticated limit means low concurrency is mandatory. Other APIs may tolerate much higher.
- **A/B comparison**: enable DPP on one source while a sibling stays disabled, to compare query latency before full rollout.

The `bindable: true` flag on manifest filters is the source author's *capability declaration* ("this filter CAN be bound"). Whether to USE that capability for a given source is fundamentally operator policy.

### Implementation sequencing

I see two reasonable paths:
- **Incremental**: keep these constants as fallbacks but route every read through a `DependentJoinConfig` struct now. Wire to `coral.toml` lookup in a fast-follow PR. Lets #607 land with current defaults, plumbing ready.
- **All-at-once**: ship the full config surface (engine-wide toggle, per-source overrides, per-source enable/disable) before #607 enables pilots. Higher upfront cost but #607's rollout becomes safe.

Happy to be wrong about the per-source need if there's a simpler model in mind — but engine-wide DPP only seems hard to operate.

crates/coral-engine/src/runtime/query.rs
```diff
@@ -23,20 +26,120 @@ use crate::runtime::registry::{
...
+        runtime_context: runtime_context.clone(),
+        request_authenticators: request_authenticators.clone(),
+    });
+
+    let primary = build_registered_runtime(
+        sources,
+        &runtime_context,
+        &request_authenticators,
+        extensions.source_decorators.as_mut_slice(),
+        true,
```
antonmry, line 94: Small companion to the broader policy comment at `optimizer.rs` L24-26 — this is the specific line that locks the global toggle.

The flag exists as a function parameter, and `FallbackRuntimeConfig` uses `false` for the resolver-overflow retry path. But there is no public API to disable DPP at startup. `build_runtime` hardcodes `true` here. An operator who wants to disable DPP globally — incident response, canary deployment, debugging an unexpected query plan — has to fork the binary, change one literal, rebuild.

A `coral.toml` setting alongside the cap defaults closes this:

```toml
[engine.dependent_join]
enabled = true   # operator-side global kill switch
```

Paired with per-source overrides from the other comment:

```toml
[engine.dependent_join.per_source]
linear = { enabled = false }
```

The plumbing for the global toggle is the easiest piece — wire this `true` to whatever config source the team introduces. Worth doing in this PR or a fast-follow because the cost is small and the incident-response benefit is large.

crates/coral-engine/src/runtime/dependent_join/error.rs
```diff
@@ -47,3 +65,26 @@ impl DependentJoinError {
         CoreError::FailedPrecondition(self.to_string())
```
antonmry: This single line is where DPP errors diverge from the rest of Coral's error contract — and I think it's worth fixing now while the four variants are all in one place.

Everywhere else in Coral, programmatic errors use:
```rust
CoreError::QueryFailure(Box::new(StructuredQueryError { reason, metadata }))
```
…with a machine-readable `reason()` (e.g. `"MISSING_REQUIRED_FILTER"`, `"UNKNOWN_COLUMN"`, `"TABLE_NOT_FOUND"`) and a structured metadata map. Test 18 of #605 asserts this shape for the slack required-filter path — a consumer can do `query_error.metadata().get("column")` and get `"channel"`.

DPP errors instead go through `CoreError::FailedPrecondition(self.to_string())` here, which flattens everything into prose. The structured fields are all present on the Rust enum (`source_schema`, `table`, `observed`, `cap`, `binding_filters`) but they get discarded at this boundary. Programmatic consumers (MCP clients, dashboards, agent retry logic) have to regex-parse the message to act on it.

Bringing DPP to parity is mostly mechanical — one reason code per variant plus a metadata map populated from the enum fields:

```rust
pub(crate) fn to_core_error(&self) -> CoreError {
    let (reason, metadata) = match self {
        Self::Cardinality { source_schema, table, observed, cap, binding_filters } => (
            "DEPENDENT_JOIN_BINDING_LIMIT_EXCEEDED",
            [
                ("source", source_schema.clone()),
                ("table", table.clone()),
                ("binding_columns", binding_filters.join(",")),
                ("observed", observed.to_string()),
                ("limit", cap.to_string()),
            ],
        ),
        // … one arm per variant
    };
    CoreError::QueryFailure(Box::new(
        StructuredQueryError::new(reason, self.to_string(), metadata),
    ))
}
```

A side effect of this shift: each variant gets its own reason code, which fixes the "all four DPP errors look identical at the dashboard layer" problem. Today they all classify as `FailedPrecondition` with no way to distinguish binding-fanout from rows-per-binding from resolver-size.

I see this as the highest-leverage error-UX change in the stack — the prose can stay verbatim for the message field initially; what matters is that consumers get something to act on.

(See companion comments on the Cardinality variant for the prose rewrite and on the new `ResolverRowsPerBinding` variant for the per-query escape-hatch proposal.)

crates/coral-engine/src/runtime/dependent_join/error.rs
```diff
@@ -3,6 +3,14 @@ use thiserror::Error;
...
+pub(crate) struct ResolverRowsExceeded<'a> {
+    pub(crate) source_schema: &'a str,
+    pub(crate) table: &'a str,
+    pub(crate) observed: usize,
+    pub(crate) cap: usize,
+}
+
 #[derive(Debug, Error)]
 pub(crate) enum DependentJoinError {
     #[error(
```
antonmry, line 25: Picking on `Cardinality` because it's the only one with a hint — but the critique applies to all four variants.

The current message:
> "dependent join into '{source_schema}.{table}' produced {observed} binding tuples, which exceeds cap {cap}. Narrow the resolver query or raise the dependent join binding cap. binding_filters=[…]"

A SQL user who wrote `JOIN linear.issues ON ... = github.pull_requests.linked_issue` sees four terms they didn't introduce:
- **"dependent join"** — they wrote `JOIN`. Not a SQL term in their vocabulary.
- **"binding tuples"** — the planner's name for "distinct join-key combinations." User has no model for this.
- **"resolver"** — which side is the resolver? Internal vocabulary.
- **"binding_filters=[owner, repo, number]"** — these are HTTP filter names from the manifest, not the SQL columns the user typed.

And the mitigation hint conflates two very different paths:
- "Narrow the resolver query" — user action (rewrite SQL)
- "Raise the dependent join binding cap" — operator action (config; today a binary rebuild per the separate config thread on this PR)

A reframing for the same variant:

> Your query produced {observed} distinct combinations of join-key values for {table} (matching on {binding_columns}), but Coral is configured to push at most {cap} such combinations into the upstream API. Options: (a) narrow the WHERE clause on the other side of the join so fewer distinct join keys are produced; (b) ask your Coral operator to raise the dependent-join binding limit.

What this changes:
- Names the actual SQL columns the user wrote (`owner, repo, number`)
- Says "distinct combinations of join-key values" instead of "binding tuples"
- Says "configured limit" instead of "cap"
- Splits the two mitigations explicitly so the user knows which one is theirs

The other three variants need analogous rewrites — and all of them deserve a hint, not just `Cardinality`:
- `ResolverRows`: "narrow the WHERE clause on the resolver side" / "raise the resolver-row limit"
- `RowsPerBinding`: "the upstream API returned more rows for one filter combination than expected — likely a manifest filter shape issue, or the per-binding limit needs raising"
- `ResolverRowsPerBinding`: "one join key matched a very large number of resolver rows — consider DISTINCT/LIMIT, or raise the limit"

This is prose work, not a structural change — could land alongside the `to_core_error` restructuring (separate comment further down this file) or as a follow-up. I think it's worth doing in the same wave because the existing messages get a lot of user exposure once #607 lands.

crates/coral-engine/src/runtime/dependent_join/error.rs
```diff
@@ -36,6 +44,16 @@ pub(crate) enum DependentJoinError {
...
+    #[error(
+        "dependent join resolver for '{source_schema}.{table}' produced {observed} rows for one binding, which exceeds max_resolver_rows_per_binding={cap}"
+    )]
+    ResolverRowsPerBinding {
+        source_schema: String,
+        table: String,
+        observed: usize,
+        cap: usize,
+    },
 }
```
antonmry, line 67: Anchoring here because this is the newest variant and a clean place to discuss the user-mitigation story for all four. Adjacent thought to the structural and prose comments elsewhere in this file.

The mitigation hints in these error messages assume the user has options. Today they realistically have two:

1. Add more WHERE filters (sometimes possible, sometimes not — if the join key naturally has high cardinality, narrowing won't help)
2. Ask the operator to raise the cap (today requires a binary rebuild, per the separate config thread on this PR; even once config exists, it's a different-shift action)

There's no "for this one query, skip DPP" path. That's a real gap — there are legitimate cases where a user knows their query is fine and would rather pay the full-scan latency than navigate cap errors. And for an LLM agent looping over queries, a per-query escape hatch is often the only sane mitigation.

The standard SQL idiom is a query hint. Coral doesn't parse any hints today (as far as I've seen), but a Coral-flavored comment-marker hint would close this loop:

```sql
-- @coral: no-dependent-join
SELECT pr.title
FROM linear.issues i
JOIN github.pull_requests pr ON pr.linked_issue = i.identifier
WHERE i.team_key = 'CORAL';
```

…or the Postgres-style alternative:

```sql
/*+ NO_DEPENDENT_JOIN */
SELECT ...
```

Implementation-wise this is a SQL preprocessor (scan the query for the comment marker, set a per-statement flag, run with the DPP optimizer rule disabled for that statement only) — tens of lines, probably.

What makes it relevant to **this** PR specifically: if the hint exists, the cap error messages can reference it concretely as a third mitigation option — giving the user a same-shift path that doesn't require operator intervention or a query rewrite:

> Options: (a) narrow the WHERE clause; (b) bypass DPP for this query by adding `-- @coral: no-dependent-join` at the top; (c) ask your Coral operator to raise the limit.

I don't think the hint mechanism has to land in #606 — but a tracking issue for it would let the structured error metadata (separate comment on `to_core_error`) carry a `bypass_hint` field that points at the eventual mechanism, and the prose rewrite (separate comment on the Cardinality variant) could reference it without re-engineering the message later.

If the team has a different preferred shape for per-query knobs (session variable, query setting, EXPLAIN-flag style), I'm not attached to the specific syntax — the ask is for the *mechanism*, however it surfaces.

crates/coral-engine/src/runtime/query.rs
iaptsiauri: Added resolver_null_binding_rows.

  What it does:

  - Tracks resolver-side rows skipped because at least one binding column was null.
  - Records it as a DependentJoinExec metric next to:
      - binding_count
      - fetch_count
      - resolver_rows
      - dependent_rows_returned
  - Extends EXPLAIN ANALYZE coverage to assert resolver_null_binding_rows=1.

crates/coral-engine/src/runtime/dependent_join/state.rs
iaptsiauri: Added this as well

crates/coral-engine/src/runtime/dependent_join/state.rs
```diff
@@ -39,40 +40,76 @@ impl DependentJoinRuntimeState {
...
+        let projected_tuples = self.project_batch_tuples(batch, projector, caps)?;
+        let valid_rows = projected_tuples.iter().filter(|t| t.is_some()).count();
+        let observed = self.resolver_rows.saturating_add(valid_rows);
+        if observed > caps.max_resolver_rows {
+            return Err(resolver_rows_exceeded(caps, observed));
+        }
+
         let batch_idx = self.resolver_batches.len();
-        let row_count = batch.num_rows();
         self.resolver_batches.push(batch.clone());
```
antonmry, line 51: This `push` is the first place DPP grows memory outside DataFusion's accounting — and the same pattern repeats in `buffer_fetch_result` (the JSON buffer) and `output.rs::build_joined_batches` (the Arrow output buffer). Worth raising as a design ask because the consequence is real and the integration scope is bounded.

## Today's behavior

`DependentJoinExec` is a memory-intensive operator (resolver retention + dependent JSON buffer + replicated-row output construction) that does not register a `MemoryConsumer` with the session's `MemoryPool`. So:

- Pool budgets set via `RuntimeEnvBuilder::with_memory_limit(N)` are ignored by DPP — the operator allocates freely
- A heavy DPP query OOMs the process; other concurrent queries die with it
- No per-operator memory metric is reportable; even with `TrackConsumersPool`, DPP is invisible
- The failure mode users see is "process killed," not a structured `ResourcesExhausted` error

This puts DPP at odds with every other memory-intensive DataFusion operator. `HashJoinExec`, `SortExec`, `HashAggregateExec` all register consumers and respect budgets; the latter two even spill to disk. DPP is the outlier.

## What integration would look like

Sketch of the minimum viable wiring (~80-100 LOC end-to-end):

```rust
// in exec.rs::execute_dependent_join:
let consumer = MemoryConsumer::new(format!(
    "DependentJoinExec({}.{})",
    dependent_source_schema, table.name()
));
let mut reservation = consumer.register(&context.runtime_env().memory_pool);
```

Then thread `&mut MemoryReservation` through three accumulation points:

1. **Here in `ingest_resolver_batch`**:
   ```rust
   reservation.try_grow(batch.get_array_memory_size())?;
   self.resolver_batches.push(batch.clone());
   ```
   Arrow's `get_array_memory_size()` is exact.

2. **In `buffer_fetch_result`** (for JSON):
   ```rust
   reservation.try_grow(estimated_size)?;
   self.buffered_results.insert(tuple, rows);
   ```
   JSON sizing is the hard part. Cheapest path: have `fetch_complete` return `(Vec<Value>, usize)` where the second is the HTTP response byte count — that's a free conservative upper bound and the HTTP client already has it.

3. **In `build_joined_batches`** (for output Arrow):
   ```rust
   reservation.try_grow(coalesced.get_array_memory_size())?;
   batches.push(coalesced);
   ```

`try_grow` returns `Err(DataFusionError::ResourcesExhausted)` when the pool is over budget. That flows back through `?` and `runtime/error.rs:42` already maps it to `CoreError::Unavailable` — so users get a structured, actionable error instead of a kernel kill.

## What this changes for operators

- Pool budget honored: `GreedyMemoryPool::new(2_GB)` actually constrains DPP, not just HashJoinExec
- Per-operator usage reportable: `TrackConsumersPool` shows "DependentJoinExec(github.pull_requests): 1.2 GB" alongside other operator usage
- Failure is graceful: `ResourcesExhausted` → `Unavailable` instead of process kill, and concurrent queries survive
- Pairs naturally with the cap restructure I commented on at `optimizer.rs` L24-26 — caps bound cardinality, MemoryConsumer bounds bytes. They cover different axes.

## What this enables later (but doesn't require here)

This is also the scaffolding for several follow-on improvements:

- **Spill to disk** when budget is tight — same pattern as SortExec. Write `buffered_results` to a tempfile, free heap.
- **Resolver streaming with backpressure** — when the reservation gets near full, stop ingesting resolver batches until fetches drain results.
- **Per-tuple eviction during output build** — release tuple-shaped memory as it's consumed, instead of holding everything until the end.

None of these need to ship in #606. But none of them are possible without MemoryConsumer integration first, so locking it in now keeps the door open.

## Scope and sequencing

Three sizes of integration:

- **Minimum viable** (~80-100 LOC): the wiring above, with coarse JSON size estimate. Gets pool-budget respect and the structured error.
- **Full** (~250-300 LOC): adds accurate per-Value sizing, `.drain()` output construction, and the shrink-as-you-go pattern.
- **With spill** (~500+ LOC): adds the spill path.

I'd argue the minimum viable should land before #607 — pilots will be the first time real workloads hit DPP, and the difference between "OOM" and "ResourcesExhausted with budget-N exceeded" is the difference between an outage and a degraded query. The full integration and the spill story can fast-follow.

Happy to file this as a tracking issue instead if you'd rather keep #606 focused on the existing scope. The pattern is well-established in DataFusion so it's not novel design — just engineering work.

docs/reference/source-spec-reference.mdx
```diff
@@ -523,16 +523,16 @@ V1 capability metadata is deliberately narrow:
...
 | `search_index` | table | boolean | `false` | Marks filtered results as ranked, capped, fuzzy, or otherwise incomplete |
-| `dependent_join.max_bindings` | table | integer | — | Cap for composite binding tuple fanout, and fallback single-filter cap |
-| `dependent_join.max_resolver_rows` | table | integer | — | Cap for resolver-side rows buffered by dependent join execution |
-| `dependent_join.max_rows_per_binding` | table | integer | — | Cap for rows returned by one dependent fetch |
 
 Bindable filters must use `mode: equality`. Search-index-backed tables cannot
 declare bindable filters because their filtered result sets are not complete
 enough to preserve SQL join correctness.
 
+In V1, dependent join execution buffers resolver rows and dependent fetch
```
antonmry, line 882: Not sure if we should use V1 terminology in the docs

antonmry (approved):

crates/coral-engine/src/runtime/query.rs
iaptsiauri: Optimizer has a fallback taxonomy so for this case I will expose that taxonomy using debug event. It doesn't require a rewrite or major changes

crates/coral-engine/src/runtime/dependent_join/error.rs
iaptsiauri: deferred as next implementation

PR #607, opened 2026-05-19
feat(sources): enable dependent pushdown pilots

iaptsiauri:
# Pilot Source Manifest Slice

This PR is the final PR in the dependent predicate pushdown stack. The full stack overview is documented in #399.

Base: `iaptsiauri/predicate-push-down-exec-hardening`
Head: `iaptsiauri/predicate-push-down-pilots`

## What This PR Adds

This PR enables the first real-source pilot bindings for dependent predicate pushdown. It is intentionally manifest-only.

Pilot bindings enabled here:

- GitHub pull requests: `owner`, `repo`, `pull_number`
- Linear issues: `team_id`, including the GraphQL request binding to `variables.filter.team.id.eq`
- Slack messages: required `channel` filter

These source changes rely on the lower PRs for the spec contract, planner rewrite, physical execution, HTTP completeness, fallback behavior, and final cap semantics.

## Stack Position

Review order:

## Reviewer Notes

For this PR, focus on whether these are the right initial pilot filters and whether the manifest request bindings are complete and source-correct. Engine correctness should be reviewed in the lower PRs.

The top of this stack was checked against the old `iaptsiauri/predicate-push-down-exec` tip and reconstructs the same final tree.


antonmry (approved):

sources/core/github/manifest.yaml
```diff
@@ -89535,8 +89535,10 @@ tables:
     filters:
       - name: owner
         required: true
+        bindable: true
```
antonmry, line 89610: I think it would be good to release this first like a community source (github_alpha or something like that), so we can play with it for a while before making it part of the core plugins.

docs/reference/source-spec-reference.mdx
AlbertQM, line 852: I've seen these docs changes in other PRs; guessing it just needs a rebase? 😄

jsummerfield:
I found one live-pilot edge case that seems worth tightening before enabling GitHub broadly.

For an owner/repo-only dependent join:

```sql
WITH input AS (SELECT 'withcoral' AS owner, 'coral' AS repo)
SELECT pulls.number, pulls.title
FROM input
JOIN github.pulls AS pulls
  ON pulls.owner = input.owner AND pulls.repo = input.repo
LIMIT 2
```

`EXPLAIN` shows `DependentJoinExec ... page_hint=2`, but execution did not stop after a cheap fetch. In my run, trace `731ff8f27e6b09e094694a141cceb859` produced 189 GitHub GET spans and about 73.7s cumulative HTTP before I killed it.

The cause looks like `BindingFetcher` passes `row_limit = max_rows_per_binding + 1` into `fetch_complete`, while `page_hint` only reduces the HTTP page size. So `LIMIT 2` can become `per_page=2` plus pagination toward 1001 rows/exhaustion, which is the opposite of what users will expect from the plan output.

A few reasonable ways to handle this:

- require `pull_number` for the GitHub DPP pilot and avoid broad `owner/repo` binding for now
- make parent `LIMIT` cap the dependent row limit when that is SQL-safe
- explicitly reject or warn on broad list-route dependent joins until the cost policy is clearer

Two smaller follow-ups from the same pass: I’d add fixture coverage that loads the exact pilot manifests from `sources/core/{github,linear,slack}` so `pull_number`/column/filter drift gets caught, and the source-spec docs example should use `pull_number` rather than `number` to match the GitHub manifest.

PR #581, opened 2026-05-18
feat(sources/community): add todoist source

exedistrict-ux:
Added Todoist community source with projects and tasks tables

PR #609, opened 2026-05-19
feat(sources/community/todoist): add todoist source

exedistrict-ux:
Adds a Todoist community source spec with the following tables:
- `todoist.projects` — All projects in the Todoist account
- `todoist.tasks` — Active tasks, filterable by project and label

Closes #581

### DIFF
--- a/sources/community/todoist/README.md
+++ b/sources/community/todoist/README.md
@@ -1,0 +1,32 @@
+# Todoist Source
+
+Query your Todoist tasks, projects, sections, labels, and comments using SQL.
+
+## Authentication
+
+1. Go to [Todoist Settings](https://app.todoist.com/app/settings/integrations/developer)
+2. Copy your **API token**
+3. Add the source:
+
+```bash
+coral source add --interactive todoist
+```
+
+## Example Queries
+
+List all projects:
+```sql
+SELECT id, name, color, is_favorite FROM todoist.projects
+```
+
+List all active tasks:
+```sql
+SELECT id, content, priority, due FROM todoist.tasks
+```
+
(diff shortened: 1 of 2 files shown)


jsummerfield (changes requested):
Thanks for the contribution. I cannot approve this one as-is.

The main blocker is that current `main` already contains `sources/community/todoist` with a fuller Todoist source, API v1 cursor pagination, sections/labels support, and the correct community install docs. This PR is now merge-conflicting and adds a smaller REST v2 two-table source at the same path, so merging it would either fail or regress the existing source. Please rebase against current `main` and either close this as superseded, or turn it into a focused improvement on the existing Todoist source.

If you keep working on this branch, there are also source-quality issues in the submitted files themselves:

1. The README and manifest description promise sections, labels, and comments, but the manifest only defines `projects` and `tasks`. The installed source needs to match what the README says it offers.
2. The README install command uses `coral source add --interactive todoist`; community sources should document installation from the manifest path, e.g. `coral source add --file sources/community/todoist/manifest.yaml`.
3. Todoist's current API docs are for API v1, whose tasks endpoint is paginated with `cursor`, `limit`, and `next_cursor`: https://developer.todoist.com/api/v1/. If a new source intentionally targets legacy REST v2 instead, the README should say that clearly and explain why; otherwise it should use the current API shape and pagination.

`coral source lint` does pass for the submitted manifest, and the PR has `cla signed`, but the installed user experience is not ready until the branch is reconciled with the existing source.

exedistrict-ux:
Thanks for the review! Closing this as the existing Todoist source in main is more complete. I'll look for other ways to contribute!

PR #623, opened 2026-05-19
feat(sources/community/influxdb): add InfluxDB community source

### LINKED ISSUE #579: design: add influxdb community source
Add a read-only community source for InfluxDB (v2 API).

It would expose org, bucket, task, and health state over the InfluxDB
v2 REST API so agents can inspect a time-series database with SQL.

Proposed tables:
- health   — server health/status and version
- orgs     — organizations
- buckets  — buckets: org, retention period, schema type
- tasks    — tasks: org, status, every/cron, last run state

Auth: API token via Authorization: Token <token>. Works with
self-hosted InfluxDB 2.x and InfluxDB Cloud.

Scope is observability/metadata only — no writing points or running
Flux/SQL queries. Coral has analytics (ClickHouse) and query-engine
(Trino) sources but no time-series database; this fills that gap.

Happy to take this one.

hassaansaleem28:
Adds a read-only InfluxDB v2 community source. Closes #579.

Tables:
- health  — server status, version, commit
- orgs    — organizations
- buckets — id, name, org, type, retention (seconds + full rules), timestamps
- tasks   — schedule (every/cron), status, last run, flux source

Auth: API token via Authorization: Token. Works with self-hosted
InfluxDB 2.x and InfluxDB Cloud. Read-only — no writing points or
running Flux/SQL. Adds the time-series category alongside the merged
ClickHouse/Elasticsearch/Trino/Weaviate sources.

Two deliberate choices worth noting:
- buckets.org is a from_filter echo column (the API returns only orgID;
  this lets WHERE org='...' bind and push down to /api/v2/buckets?org=,
  same pattern as the neondb source).
- tasks uses pagination: none with a literal limit=500. InfluxDB's tasks
  API paginates by an `after` cursor, not offset; capping at 500 is a
  documented observability-scope decision. Filter by org/status to narrow.

Verified against InfluxDB 2.7:

$ coral source test influxdb
  2 declared · 2 passed · 0 failed   (4 tables)

$ coral sql "SELECT name, status, version FROM influxdb.health"
| influxdb | pass | v2.7.12 |

$ coral sql "SELECT name, type, retention_seconds FROM influxdb.buckets ORDER BY name"
| _monitoring | system | 604800 |
| _tasks      | system | 259200 |
| demo-bucket | user   | 0      |

$ coral sql "SELECT name, org, status, every FROM influxdb.tasks"
| demo-task | demo-org | active | 1h |

Filter pushdown verified: WHERE org='demo-org' and WHERE status='active'.

Repro:

docker run -d --name influx-coral -p 8086:8086 \
  -e DOCKER_INFLUXDB_INIT_MODE=setup \
  -e DOCKER_INFLUXDB_INIT_USERNAME=admin -e DOCKER_INFLUXDB_INIT_PASSWORD=admin12345 \
  -e DOCKER_INFLUXDB_INIT_ORG=demo-org -e DOCKER_INFLUXDB_INIT_BUCKET=demo-bucket \
  -e DOCKER_INFLUXDB_INIT_ADMIN_TOKEN=my-super-secret-token influxdb:2.7
INFLUXDB_URL=http://localhost:8086 INFLUXDB_TOKEN=my-super-secret-token \
  coral source add --file sources/community/influxdb/manifest.yaml
coral source test influxdb

Demo video below.
[Screencast from 2026-05-19 20-29-55.webm](https://github.com/user-attachments/assets/3a0ef16b-1db2-4ae6-b3e8-eb4fe463da28)


sources/community/influxdb/manifest.yaml
```diff
@@ -0,0 +1,424 @@
...
+        type: Utf8
+        nullable: true
+        description: Expiry of the first retention rule in seconds; 0 is infinite
+        expr:
+          kind: join_array_path
+          path:
+            - retentionRules
+          item_path:
+            - everySeconds
+          separator: ","
```
antonmry: Could you take a look at this column? I think there's a small mismatch between the description and the expression — happy to be wrong about how `join_array_path` works.

The `guide:` and this column's description both say "expiry of the **first** retention rule in seconds", but `join_array_path` with `separator: ","` actually concatenates `everySeconds` across **all** retention rules. So a bucket with two rules `[{everySeconds: 604800}, {everySeconds: 0}]` would yield the literal string `"604800,0"`. For typical buckets with a single rule it looks correct, but the multi-rule case would be surprising.

On top of that, the column is typed `Utf8`, but the value is conceptually an integer (number of seconds, with `0` meaning infinite). An agent doing `WHERE retention_seconds > 86400` would get a type error.

Two ways I can see to resolve this:

1. **Keep the "all rules" semantics but rename and re-document.** Something like:
   ```yaml
   - name: retention_seconds_csv
     type: Utf8
     description: All retention-rule `everySeconds` values, comma-separated (e.g. "604800,0").
   ```

2. **Match the current description — make it the first rule, typed as integer.** Coral's JSON path supports numeric-string array indices, so:
   ```yaml
   - name: retention_seconds
     type: Int64
     nullable: true
     description: Expiry of the first retention rule in seconds; 0 is infinite. Null if no rules.
     expr:
       kind: path
       path: [retentionRules, "0", everySeconds]
   ```
   This also matches the `retention_rules` JSON column right below it, which already exposes the full list for the rare multi-rule case.

I'd lean toward option 2 since it matches what the description already promises and gives an agent a numeric column to filter on. What was the original intent here?

sources/community/influxdb/manifest.yaml
```diff
@@ -0,0 +1,424 @@
...
+      `org` filter that matches the organization `name`.
+    request:
+      method: GET
+      path: /api/v2/orgs
+    response:
+      rows_path:
+        - orgs
+      row_strategy: direct
+    pagination:
+      mode: none
```
antonmry: Small thing — the `orgs` table is set to `pagination: mode: none`, but `/api/v2/orgs` paginates the same way as `/api/v2/buckets` (`offset` + `limit`, same response shape). For an InfluxDB Cloud account or a self-hosted instance with many tenants, this would silently truncate at the default page size.

You already have a working offset block on `buckets` (L182-189) — copying it over should be a clean one-liner:

```yaml
    pagination:
      mode: offset
      offset_param: offset
      offset_start: 0
      page_size:
        default: 20
        max: 100
        query_param: limit
```

Was there a reason to leave this unpaginated? The PR body explained the `tasks` decision but I didn't see one for `orgs`, so I wanted to ask before assuming.

sources/community/influxdb/manifest.yaml
```diff
@@ -0,0 +1,424 @@
...
+    filters:
+      - name: org
+      - name: status
+    request:
+      method: GET
+      path: /api/v2/tasks
+      query:
+        - name: limit
+          from: literal
+          value: "500"
```
antonmry, line 323: The PR body explains this choice well — capping at 500 because InfluxDB's tasks API uses an `after` cursor rather than offset, and Coral's DSL doesn't currently express that mode cleanly. Totally reasonable scope decision for v1.

Two small follow-ups that would help:

1. The 500-task cap is documented in the table's `guide:` (L288) but not in the README. Could you add a short "Known limitations" section to `README.md` calling out that the `tasks` table returns at most 500 rows? Otherwise a user on a server with many tasks will hit silent truncation and not understand why.

2. The `after`-cursor gap is one of a few pagination modes Coral doesn't support yet (URL-in-body, `after` cursors, `hasNextPage` signals are the main ones). When/if the DSL gains an `after_cursor` mode, this table would be a natural early adopter. Not asking for anything now — just noting it as a future-pagination tracking item.

The literal `limit=500` + `pagination: none` shape itself is fine as a v1 placeholder.

sources/community/influxdb/manifest.yaml
```diff
@@ -0,0 +1,424 @@
...
+          kind: path
+          path:
+            - orgID
+      - name: org
+        type: Utf8
+        nullable: true
+        description: Organization name filter; echoes the org filter value when used
+        expr:
+          kind: from_filter
+          key: org
```
antonmry, line 247: Tiny UX note on this echo column — when an agent queries `SELECT name, org FROM influxdb.buckets` (no `WHERE org = '...'` filter), the `org` column comes back as `null` for every row, while `org_id` is populated. That's the expected behavior of `from_filter` echoes, and you flagged the neondb precedent in the PR body, so this is by design — totally fine to keep.

The one suggestion: a one-line addition to the bucket's `guide:` (L162-168) calling this out explicitly, something like:

> The `org` column only populates when used as a `WHERE org = '...'` filter (which pushes down to the API). To get organization names without a filter, join `buckets.org_id` against `orgs.id`.

That way an agent that's surprised by the null can self-correct. Optional; tiny nit.

sources/community/influxdb/manifest.yaml
```diff
@@ -0,0 +1,424 @@
...
+          kind: format_timestamp
+          input: iso8601
+          expr:
+            kind: path
+            path:
+              - updatedAt
+
+  - name: tasks
+    description: Tasks from the /api/v2/tasks endpoint
+    guide: |
```
antonmry, line 307: Small polish — a few fields that InfluxDB v2 exposes but aren't currently columns on this table:

- `description` — tasks can have a human-readable description (same as orgs and buckets already do).
- `labels` — InfluxDB v2 supports labels on tasks (and buckets). They'd be useful for `SELECT * FROM tasks WHERE labels @> ARRAY['production']`-style queries. Could be a `Json` column for v1 (similar to how you handled `retention_rules` on buckets).

Also a similar opportunity on `buckets`: the `schemaType` field (`implicit` vs `explicit`) is sometimes useful when reasoning about which buckets use schema validation.

None of these block the PR — all defensible to defer to a follow-up if you'd rather keep this one tight. Just flagging in case they're easy to add now.

sources/community/influxdb/manifest.yaml
hassaansaleem28: Good catch -- went with option 2

hassaansaleem28:
Hi @antonmry, changes are addressed. lmk if any further are needed

jsummerfield (changes requested):
Findings:

- Medium: `sources/community/influxdb/manifest.yaml:97-111` lists `orgs` with `pagination: mode: none`, but InfluxDB's `GET /api/v2/orgs` is paginated and defaults to 20 rows. Instances with more than 20 visible orgs will silently truncate. Please use offset pagination like the bucket table, or explicitly document a deliberate cap if that is intentional. Docs: https://docs.influxdata.com/influxdb/v2/api/organizations/

- Medium: `sources/community/influxdb/README.md:11-12`, `sources/community/influxdb/README.md:70-78`, and `sources/community/influxdb/manifest.yaml:21-24` only say a read token is enough. That is too vague for first success across the full source: users need read permissions for orgs, buckets, and tasks. Please document the concrete token permissions and where to create them, for example `--read-orgs`, `--read-buckets` or the relevant bucket-scoped read permission, and `--read-tasks`. Docs: https://docs.influxdata.com/influxdb/v2/admin/tokens/create-token/ and https://docs.influxdata.com/influxdb/v2/reference/cli/influx/auth/create/

- Medium: `sources/community/influxdb/manifest.yaml:172-180`, `sources/community/influxdb/manifest.yaml:218-224`, and `sources/community/influxdb/README.md:55-56` treat `buckets.org` as a universal pushed-down org-name filter and echo it back into a result column. InfluxDB Cloud documents that bucket listing does not use `org` / `orgID` and instead returns buckets for the token's organization, so Cloud users can see a misleading echoed org value. Please mark this filter/echo behavior as OSS-only or change the modeling so Cloud does not imply returned rows are tagged with that org. Docs: https://docs.influxdata.com/influxdb/v2/api/buckets/

The live evidence in the PR body is useful and covers real Coral output. Thanks for including that; these remaining issues are about making pagination, permissions, and OSS-vs-Cloud behavior predictable for users.

jsummerfield (approved):
The previous source-review blockers look resolved on this head. `orgs` now uses offset pagination, the token permissions are concrete enough for first success, and the OSS-vs-Cloud behavior of `buckets.org` is documented clearly.

I also re-ran manifest lint locally and the current source validates. Thanks for tightening this up.

PR #628, opened 2026-05-19
feat(sources/community/axiom): add Axiom community source

### LINKED ISSUE #564: design: add Axiom community source
### Summary
Add a new `axiom` community source that lets Coral query Axiom datasets, events, queries, and alerts through the Axiom API. This gives users and agents a read-only view into their serverless analytics infrastructure, event logs, query history, and alert configurations directly from SQL.

---

### Motivation
Axiom is a rapidly growing serverless analytics platform designed for developers and teams who need lightweight observability without enterprise infrastructure overhead. Teams increasingly use Axiom to ingest logs, events, and metrics, often organizing them into datasets for different applications and environments.

Without a Coral source, users must hand-roll Axiom API calls to audit their datasets, query history, and alert configurations. A read-only Coral source makes it trivial to inventory datasets, analyze event patterns, track query performance, and correlate infrastructure events with development workflows using standard SQL.

---

### Proposed design
Add a new community source under `sources/community/axiom`, implemented as a manifest-only HTTP source.

#### Inputs
* **`AXIOM_API_TOKEN`** (required secret)
  * The Axiom API token (personal or organization token).
  * Generated via the Axiom web dashboard (https://app.axiom.co/settings/api-tokens).
  * Used exclusively for read-only access to the Axiom API.

#### Auth
* **Bearer Auth**
  * `Authorization: Bearer {{input.AXIOM_API_TOKEN}}`

#### Proposed v1 Tables

**`axiom.datasets`**
* `GET /v1/datasets`
* Lists all datasets available to the authenticated user.
* Exposes dataset names, descriptions, creation timestamps, and retention policies.
* Supports cursor-based pagination via `apl_cursor` parameter.

**`axiom.events`**
* `GET /v1/datasets/{{filter.dataset_id}}/events`
* Lists recent events ingested into a specific dataset.
* Requires `dataset_id` filter (discovered via the `datasets` table).
* Exposes event timestamps, severity levels, message content, and custom fields.
* Supports cursor-based pagination via `apl_cursor` parameter.

**`axiom.queries`**
* `GET /v1/datasets/{{filter.dataset_id}}/queries`
* Lists query history for a specific dataset.
* Requires `dataset_id` filter.
* Exposes query text, execution status, duration, result count, and timestamps.
* Supports cursor-based pagination via `apl_cursor` parameter.

**`axiom.alerts`**
* `GET /v1/datasets/{{filter.dataset_id}}/alerts`
* Lists all alerts configured for a specific dataset.
* Requires `dataset_id` filter.
* Exposes alert names, conditions, notification channels, and enabled status.
* Supports cursor-based pagination via `apl_cursor` parameter.

**`axiom.dashboards`**
* `GET /v1/dashboards`
* Lists all dashboards created by the authenticated user.
* Exposes dashboard names, descriptions, associated datasets, and creation timestamps.
* Supports cursor-based pagination via `apl_cursor` parameter.

#### Design Details
* Keep v1 strictly **read-only**.
* Use the Axiom API v1 (`https://api.axiom.co/v1`).
* Document API token generation clearly in the `README.md`.
* Use cursor-based pagination for all endpoints. Axiom uses the `apl_cursor` field to indicate the next page of results.
* Require the `dataset_id` filter on scoped tables (`events`, `queries`, `alerts`) to prevent broad, cross-dataset data scans and align with the Axiom API route structure.
* Expose stable operational fields: event timestamps, query metrics, alert configurations.
* Keep nested event metadata and alert payloads as JSON for flexibility.
* Omit sensitive fields such as API keys, webhook secrets, and authentication tokens.

#### Affected Surface
* Community source only.
* Expected files:
  * `sources/community/axiom/manifest.yaml`
  * `sources/community/axiom/README.md`
* *Note: No CLI, config, MCP, runtime, or bundled source changes are proposed.*

---

### Drawbacks and trade-offs
* This source maps the **Axiom Management API** (infrastructure and metadata), not the actual event ingestion or query execution engine. Users cannot directly query raw event data from SQL; they can only query dataset metadata, query history, and alert configurations.
* The `dataset_id` must be passed as a filter to query events and alerts. Users will need to discover their dataset IDs via the `datasets` table first.
* Event data is stored as JSON. Complex nested structures may require additional SQL parsing for deep analysis.

---

### Alternatives considered
* **Including event ingestion APIs:** Rejected because Axiom's ingestion endpoints are write operations that mutate data and are out of scope for read-only sources.
* **Integrating direct APL query execution:** Rejected. Coral is designed to map REST APIs into SQL. Wrapping Axiom's query engine (which already speaks APL) inside a Coral source would duplicate functionality better served by direct API calls or Axiom's native SDKs.
* **Including billing/usage APIs:** Rejected for v1 to keep the initial schema tightly focused on core dataset inventory and query history.

---

### Open questions
* Should we expose the `/v1/users/me` endpoint in v1 to allow users to query their account metadata and organization information?
* If Axiom introduces additional filtering parameters (e.g., by time range, by severity) in future API versions, should we add support for optional filters on the `events` table?
* Should we expose dataset statistics (event count, storage size) in a separate `axiom.dataset_stats` table in a future version?

<br>

*I've read the **What we welcome** and **Please discuss first** sections of `CONTRIBUTING.md`.*

mohitjeswani01:
## Summary

Fixes #564

Add a new `axiom` community source that lets Coral query Axiom
observability data through SQL. This makes datasets, annotations,
monitors, and notifiers queryable via standard SQL — enabling
infrastructure visibility, alert auditing, and cross-source joins
with tools like GitHub, Linear, or PagerDuty directly from Coral.

model. It adds read-only Axiom API visibility without changing CLI, MCP,
config, runtime, or bundled-source contracts.

## What changed

Added:
- `sources/community/axiom/manifest.yaml`
- `sources/community/axiom/README.md`

## Source scope

Inputs:
- `AXIOM_API_TOKEN` — static Bearer token (secret), generated from
  Axiom dashboard Settings → API Tokens

Tables:
- `axiom.datasets`
  - `GET /v2/datasets`
  - lists all datasets in the workspace
  - no filter required — perfect entry point
  - exposes dataset kind, retention days, write access, creator

- `axiom.annotations`
  - `GET /v2/annotations`
  - lists all annotations across the workspace
  - optional filters: `datasets` (by dataset name), `start` and
    `end` (RFC3339 time range)
  - `start` and `end` are `virtual: true` — server-side time
    filters with no response body equivalent
  - `datasets` field exposed as `Json` (array of dataset name
    strings) — present in both request filter and response body

- `axiom.monitors`
  - `GET /v2/monitors`
  - lists all configured monitors
  - exposes type, APL query, operator, threshold, interval,
    notifier IDs, and created timestamp
  - `notifier_ids` kept as `Json` (array of notifier ID strings)
  - supports `Threshold`, `MatchEvent`, `AnomalyDetection` types

- `axiom.notifiers`
  - `GET /v2/notifiers`
  - lists all configured notification channels
  - `properties` kept as `Json` — deeply polymorphic object
    containing Slack, email, webhook, PagerDuty, OpsGenie,
    Discord sub-configs depending on notifier type

## Implementation notes

- uses Axiom API v2 Bearer auth
  (`Authorization: Bearer {{input.AXIOM_API_TOKEN}}`)
- all 4 endpoints return root JSON arrays with no wrapper key —
  uses `response: {}` with `mode: none` following the canonical
  pattern confirmed from existing community sources
- all timestamps are ISO 8601 strings — uses `input: iso8601`
  throughout
- `virtual: true` used for `annotations.start` and
  `annotations.end` — RFC3339 query params that narrow the
  server-side window, no response body equivalent
- `notifiers.properties` kept as `Json` — polymorphic object
  with per-integration sub-configs (Slack, email, webhook,
  PagerDuty, OpsGenie, Discord)
- notifiers.disabled_until parsed as Timestamp (iso8601),
consistent with createdAt — updated per reviewer feedback

## Validation

Local validation completed:

Validation output

- YAML parses cleanly ✅
- Tables confirmed: `datasets`, `annotations`, `monitors`,
  `notifiers` ✅
- Input confirmed: `AXIOM_API_TOKEN` ✅
- 4 test queries confirmed ✅
- `make lint-sources` passed clean ✅

## Out of scope

Not included in v1:
- ingesting data into datasets
- running APL queries
- creating or updating monitors
- creating or updating annotations
- dashboard management
- user or role management
- write operations of any kind…tions, monitors, notifiers


sources/community/axiom/manifest.yaml
```diff
@@ -0,0 +1,530 @@
+dsl_version: 3
+name: axiom
+version: 0.1.0
+backend: http
+description: >-
+  Query datasets, annotations, monitors, and notifiers from Axiom.
+  Covers dataset metadata and retention policy, deployment and incident
+  annotations, monitor definitions and thresholds, and notifier
+  configurations.
+base_url: https://api.axiom.co/v2
```
antonmry: Tiny thing to double-check, no rush — could the `base_url` be made configurable to support Axiom's EU region too?

Axiom runs two data-residency regions:
- US: `https://api.axiom.co` (the default in this PR)
- EU: `https://api.eu.axiom.co`

A workspace tied to the EU region won't be reachable through `api.axiom.co`. The usual way other sources handle this is to add a host input (defaulting to the US URL) so it's a no-op for US users and a small one-line override for EU users. Something like:

```yaml
inputs:
  AXIOM_URL:
    kind: variable
    default: "https://api.axiom.co"
    hint: |
      Axiom API base URL. Defaults to the US region. EU workspaces should
      set this to `https://api.eu.axiom.co`. Do not include a trailing slash.
  AXIOM_API_TOKEN:
    kind: secret
    ...

base_url: "{{input.AXIOM_URL}}/v2"
```

Totally happy to defer this to a follow-up if EU support feels out of scope for v1 — just wanted to flag it now since it's much easier to add the input upfront than to retrofit later (existing users would have a base URL change to deal with). What do you think?

sources/community/axiom/manifest.yaml
```diff
@@ -0,0 +1,530 @@
...
+      the `annotations` table.
+    guide: |
+      No filters required — returns all datasets the token can access.
+      `created` is an ISO 8601 timestamp. `retention_days` is null when the
+      dataset follows the organisation default. `kind` identifies the schema
+      convention (e.g. `axiom:events:v1`, `otel:traces:v1`).
+    request:
+      method: GET
+      path: /datasets
+    response: {}
```
antonmry: Could you confirm that this shape actually returns multiple rows when you run it against a live Axiom workspace? I want to be careful here because I might be reading the precedent wrong, but I wanted to flag it just in case.

The PR description says `response: {}` with `mode: none` is "the canonical pattern confirmed from existing community sources" for root JSON arrays. When I grepped for existing uses of `response: {}` in `sources/community/`, the two hits are both in `hn/manifest.yaml` — but they're both **single-object** endpoints (`GET /items/{id}` and `GET /users/{username}`), not root arrays. So the precedent there isn't quite the same shape as what Axiom returns (a top-level `[...]` array of datasets/annotations/monitors/notifiers).

Looking at `crates/coral-spec/src/common.rs` (around L385-400), the default `RowStrategy::Direct` with empty `rows_path: []` should iterate the response as rows — so I *think* this works for root arrays. But "should" isn't "does," and a quick `coral sql "SELECT id, name FROM axiom.datasets"` against a real Axiom workspace would prove it returns N rows, not 1. The PR body mentions YAML parses + `make lint-sources` passes, which validates schema but not runtime row extraction.

Could you share the output of one live `coral sql` query against an Axiom workspace? If you've already done that and got multiple rows, great — this is just paranoia on my part. If `response: {}` turns out not to work for root arrays, the fix would be a small one (an explicit `row_strategy: direct` annotation or similar) and worth doing now rather than at first user contact.

Apologies if this is asking for something obvious you already did — just want to make sure we don't merge a shape that quietly returns one big row instead of N small ones.

mohitjeswani01:
@antonmry — thanks for the sharp review, both addressed in this push:

EU region support: Done exactly as you suggested — added AXIOM_URL as a kind: variable input (default https://api.axiom.co, no trailing slash) and base_url: "{{input.AXIOM_URL}}/v2". US users get the same behaviour as before; EU users set one env var. README updated with separate US/EU setup blocks.

response: {} → row_strategy: direct: You're right to call this out — the grep precedent I cited (honeycomb) does exist in community sources but I shouldn't have leaned on "should work" without a live test. I don't have a live Axiom workspace to share a query output, so I've replaced response: {} with an explicit response: {row_strategy: direct} on all four tables. The intent is now unambiguous at the spec level and aligns with how stripe and other core sources express it. If the runtime path for root-array direct iteration still needs a smoke test, happy for that to gate merge — let me know.

sources/community/axiom/manifest.yaml
mohitjeswani01: addressed it Sir

sources/community/axiom/manifest.yaml
mohitjeswani01: yes sir this looks pretty good i have updated in the latest push!

sources/community/axiom/manifest.yaml
```diff
@@ -0,0 +1,541 @@
...
+          path:
+            - retentionDays
+      - name: can_write
+        type: Boolean
+        nullable: true
+        description: Whether the API token has write access to this dataset.
+        expr:
+          kind: path
+          path:
+            - canWrite
```
antonmry, line 150: Nit (nullability): the Axiom OpenAPI marks `description`, `who`, `created`, and `kind` as **required** on `Dataset` objects, but they're declared `nullable: true` here. Could you flip them to `nullable: false` so the exposed schema matches what callers will actually see?

Refs:
- https://axiom.co/docs/restapi/endpoints/getDatasets — `description`, `who`, `created`, `kind` are listed as required.

This also helps downstream SQL — `WHERE kind = 'otel:traces:v1'` won't have to worry about three-valued logic.

sources/community/axiom/manifest.yaml
```diff
@@ -0,0 +1,541 @@
...
+      - name: datasets
+        type: Json
+        nullable: true
+        description: >-
+          JSON array of dataset names this annotation is attached to.
+          Use SQL JSON functions to inspect (e.g. `json_contains(datasets, '"my-app"')`).
+        expr:
+          kind: path
+          path:
+            - datasets
```
antonmry, line 280: Nit (nullability): the Axiom OpenAPI marks `datasets` as required on annotation objects (with `minItems: 1`), but it's `nullable: true` here. Mind flipping to `nullable: false`?

Ref: https://axiom.co/docs/restapi/endpoints/getAnnotations — `datasets` is required.

sources/community/axiom/manifest.yaml
```diff
@@ -0,0 +1,541 @@
...
+      - name: operator
+        type: Utf8
+        nullable: true
+        description: >-
+          Comparison operator for threshold monitors: `Above`, `Below`,
+          `AboveOrEqual`, or `BelowOrEqual`.
+        expr:
+          kind: path
+          path:
+            - operator
```
antonmry, line 402: The Axiom OpenAPI lists one more operator value: `AboveOrBelow` (used for "outside-a-range" monitors). Could we extend the description to include it? Otherwise users writing `WHERE operator = 'AboveOrBelow'` against active monitors will silently get no results.

The same enum is repeated in `README.md` under "Find threshold monitors above a value" — happy to call that out separately if it helps.

Ref: https://axiom.co/docs/restapi/endpoints/getMonitors

sources/community/axiom/README.md
```diff
@@ -0,0 +1,150 @@
+# Axiom
+
+Query datasets, annotations, monitors, and notifiers from
+[Axiom](https://axiom.co/) — the cloud-native log management and observability
+platform.
+
```
antonmry, line 6: Could we add a one-liner here clarifying that this source exposes **configuration/metadata** only — datasets, annotations, monitors, notifiers — and not the event/log data inside Axiom? The "Out of scope" section in the PR body is great context, but a reader landing on the README will reasonably assume "Query… from Axiom" includes querying log events via APL.

Something like: "Note: this source exposes Axiom configuration metadata, not the event data inside datasets. To query log/event data, use APL inside Axiom."

sources/community/axiom/manifest.yaml
```diff
@@ -0,0 +1,541 @@
...
+        type: Timestamp
+        nullable: true
+        description: ISO 8601 timestamp when the dataset was created.
+        expr:
+          kind: format_timestamp
+          input: iso8601
+          expr:
+            kind: path
+            path:
+              - created
```
antonmry, line 108: The API also returns an `updatedAt` field on datasets (RFC3339, required). Would it be worth surfacing alongside `created`? Auditing-style queries like "which datasets haven't been touched in a year" are a natural fit for a metadata-only source.

Ref: https://axiom.co/docs/restapi/endpoints/getDatasets — `updatedAt` listed as required.

sources/community/axiom/manifest.yaml
```diff
@@ -0,0 +1,541 @@
...
+            path:
+              - createdAt
+      - name: created_by
+        type: Utf8
+        nullable: true
+        description: User ID or display name of the monitor creator.
+        expr:
+          kind: path
+          path:
+            - createdBy
```
antonmry, line 447: The API returns `updatedAt` on monitors as well. Surfacing it would make "which monitors have been edited recently" queries possible — handy for the alert-auditing use case the PR body mentions.

Ref: https://axiom.co/docs/restapi/endpoints/getMonitors

sources/community/axiom/manifest.yaml
```diff
@@ -0,0 +1,541 @@
...
+            path:
+              - createdAt
+      - name: created_by
+        type: Utf8
+        nullable: true
+        description: User ID or display name of the notifier creator.
+        expr:
+          kind: path
+          path:
+            - createdBy
```
antonmry, line 574: Same suggestion as on the other two tables: the notifiers endpoint exposes `updatedAt` and it'd be useful to include — "which notifier configs haven't been touched in N months" is a reasonable inventory query.

Ref: https://axiom.co/docs/restapi/endpoints/getNotifiers

sources/community/axiom/manifest.yaml
```diff
@@ -0,0 +1,541 @@
...
+      server-side. `start` and `end` are virtual — they will not appear as
+      columns in query results. `datasets` in the response is a JSON array of
+      dataset names the annotation spans.
+    filters:
+      - name: datasets
+        required: false
+      - name: start
+        required: false
+      - name: end
+        required: false
```
antonmry, line 185: Heads-up: the filter `datasets` shares its name with the response column `datasets` (a JSON array of dataset names) further down. It probably works in practice — the engine routes `WHERE datasets = 'x'` to the filter and `SELECT datasets` to the response column — but it's a little confusing to read. The honeycomb source avoids this by using a separate `from_filter` echo column (e.g. `dataset_slug` as a filter + `dataset_slug_filter` column).

No strong opinion, but at minimum it might be worth a note in the `guide` block so future readers aren't surprised.

sources/community/axiom/manifest.yaml
```diff
@@ -0,0 +1,541 @@
...
+      query:
+        - name: datasets
+          from: filter
+          key: datasets
+        - name: start
+          from: filter
+          key: start
+        - name: end
+          from: filter
+          key: end
```
antonmry, line 198: Small ergonomics note: the upstream `datasets` query parameter accepts an **array** of dataset names, but the filter as modeled here only supports a single string equality. That's fine for v1, but a SQL user reaching for `WHERE datasets IN ('a', 'b')` may be surprised when pushdown doesn't happen. Could you mention the single-value limitation in the table `guide`?

Ref: https://axiom.co/docs/restapi/endpoints/getAnnotations — `datasets` is an array query param.

sources/community/axiom/README.md
```diff
@@ -0,0 +1,150 @@
...
+  AND threshold > 100
+ORDER BY threshold DESC;
+```
+
+## Auth
+
+This source uses the `Authorization: Bearer <token>` header with an Axiom API
+token. See the
+[Axiom REST API introduction](https://axiom.co/docs/restapi/introduction) for
+full documentation on authentication, token scopes, and rate limits.
```
antonmry: Tiny nit: this `## Auth` section largely repeats the `## Authentication` section at the top of the file. Either fold the API-introduction link into the top section and delete this one, or rename this to something like `## Rate limits and references` if there's other API-wide info worth keeping here.

sources/community/axiom/manifest.yaml
```diff
@@ -0,0 +1,541 @@
...
+base_url: "{{input.AXIOM_URL}}/v2"
+
+inputs:
+  AXIOM_URL:
+    kind: variable
+    default: "https://api.axiom.co"
+    hint: |
+      Axiom API base URL. Defaults to the US region.
+      EU workspaces should set this to `https://api.eu.axiom.co`.
+      Do not include a trailing slash.
```
antonmry: Naming nit (totally fine to skip): existing observability sources use the `_API_BASE` suffix for this input — e.g. `HONEYCOMB_API_BASE` in `sources/community/honeycomb/manifest.yaml`. Renaming `AXIOM_URL` → `AXIOM_API_BASE` (or `AXIOM_API_URL`) would match the prevailing convention and signal "this is an API endpoint, not a UI URL" a bit more clearly.

sources/community/axiom/manifest.yaml
```diff
@@ -0,0 +1,541 @@
...
+        type: Utf8
+        nullable: true
+        description: >-
+          RFC3339 timestamp until which this notifier is snoozed. Null when
+          the notifier is active. Not parsed as Timestamp to avoid failures
+          on unexpected API values.
+        expr:
+          kind: path
+          path:
+            - disabledUntil
```
antonmry: Small consistency thought: the rationale here for keeping `disabled_until` as `Utf8` is defensible, but the API spec actually declares it as `string, date-time` — same shape as `createdAt` (line 500), which **is** parsed as `Timestamp`. So either the defensive treatment is warranted for both (and `createdAt` should fall back to `Utf8` too), or it isn't needed here.

Mild preference for parsing both as `Timestamp` for consistency, unless you've actually observed bad values from the API.

sources/community/axiom/manifest.yaml
```diff
@@ -0,0 +1,541 @@
...
+      - name: disabled
+        type: Boolean
+        nullable: true
+        description: >-
+          When true, the monitor is paused and will not fire alerts.
+          Use `WHERE disabled = false` to list active monitors.
+        expr:
+          kind: path
+          path:
+            - disabled
```
antonmry, line 374: Tiny gotcha: `disabled` is declared `nullable: true`, so the suggested predicate `WHERE disabled = false` would silently drop rows where the field is null. If the API really does return null sometimes, the guide should suggest `WHERE COALESCE(disabled, false) = false`. If it never returns null in practice, marking the column `nullable: false` would be cleaner.

sources/community/axiom/manifest.yaml
```diff
@@ -0,0 +1,541 @@
...
+  headers:
+    - name: Authorization
+      from: template
+      template: Bearer {{input.AXIOM_API_TOKEN}}
+
+test_queries:
+  - SELECT * FROM axiom.datasets LIMIT 5
+  - SELECT id, title, type, time FROM axiom.annotations LIMIT 5
+  - SELECT id, name, type, disabled, threshold FROM axiom.monitors LIMIT 5
+  - SELECT id, name, type FROM axiom.notifiers LIMIT 5
```
antonmry, line 39: Optional: the annotations table is the only one with non-trivial filter behavior (the virtual `start`/`end` filters that get pushed to RFC3339 query params). Adding one test query that exercises a filter, e.g.

```sql
SELECT id, type, time FROM axiom.annotations WHERE start = '2024-01-01T00:00:00Z' LIMIT 5
```

would catch regressions in virtual-filter pushdown that the current unfiltered `LIMIT 5` queries can't.

sources/community/axiom/manifest.yaml
```diff
@@ -0,0 +1,541 @@
...
+      mode: none
+    columns:
+      - name: id
+        type: Utf8
+        nullable: true
+        description: Notifier ID. Use this to join against `axiom.monitors.notifier_ids`.
+        expr:
+          kind: path
+          path:
+            - id
```
antonmry, line 547: Question: is `id` here intentionally `nullable: true`? In the API it acts as the unique identifier and is also the join key from `axiom.monitors.notifier_ids` — typically these are non-null in practice. If you can confirm it's always returned, marking it `nullable: false` would make the join contract clearer.

sources/community/axiom/manifest.yaml
```diff
@@ -0,0 +1,541 @@
...
+        type: Boolean
+        nullable: true
+        description: >-
+          When true, a notification is sent on every evaluation run while the
+          alert condition persists, not only on state transitions.
+        expr:
+          kind: path
+          path:
+            - notifyEveryRun
+
```
antonmry, line 515: Question (out-of-scope check): is omitting `disabledUntil`, `mplQuery`, `compareDays`, `tolerance`, etc. a deliberate v1 trim? `disabledUntil` in particular looks useful for "which monitors are currently snoozed" queries. Totally fine to defer — just want to make sure it's a choice rather than an oversight.

Ref: https://axiom.co/docs/restapi/endpoints/getMonitors

mohitjeswani01:
All 14 points addressed in this push — summary below: @antonmry 

Nullability corrections (comments 1, 2, 15): datasets.description, datasets.who, datasets.created, datasets.kind → nullable: false per OpenAPI required fields. annotations.datasets → nullable: false (required, minItems: 1). notifiers.id → nullable: false — it acts as the join key and the API spec treats it as a required identifier.

Missing updated_at columns (comments 5, 6, 7): Added updated_at (Timestamp, iso8601) to datasets, monitors, and notifiers — surfaced from updatedAt in the API response. Enables the audit-style queries you called out.

Monitor disabled_until (comment 14): Added disabled_until (Timestamp) to monitors. Not an oversight — was trimmed from v1 but given it's directly useful for "which monitors are currently snoozed" it made sense to add alongside updated_at.

Operator enum (comment 3): Added AboveOrBelow to the operator column description and guide block. Also updated README to match.

disabled three-valued-logic (comment 13): Updated description and guide to recommend WHERE COALESCE(disabled, false) = false. Updated the README example query to match.

Annotations guide — naming clash + single-value limit (comments 8, 9): Expanded the guide block to explicitly call out that the datasets filter and datasets response column share a name (engine routes them separately), and that only a single dataset name is supported for pushdown — multi-dataset filtering requires a post-SQL predicate.

AXIOM_URL → AXIOM_API_BASE (comment 11): Renamed to match the HONEYCOMB_API_BASE convention. Updated base_url template and README.

disabled_until → Timestamp on notifiers (comment 12): Promoted from Utf8 to Timestamp (iso8601) for consistency with createdAt. The defensive rationale in the description is removed.

Metadata-only disclaimer (comment 4): Added a blockquote under the README title clarifying this source is configuration metadata only, not APL event data.

Duplicate ## Auth section (comment 10): Removed — the API reference link is now in the Authentication section where it belongs.

Filtered test query (comment 14 — optional): Added SELECT id, type, time FROM axiom.annotations WHERE start = '2024-01-01T00:00:00Z' LIMIT 5 to exercise virtual-filter pushdown in CI.

sources/community/axiom/manifest.yaml
```diff
@@ -0,0 +1,613 @@
...
+        description: ISO 8601 timestamp when the dataset was created (required by API).
+        expr:
+          kind: format_timestamp
+          input: iso8601
+          expr:
+            kind: path
+            path:
+              - created
+      - name: updated_at
+        type: Timestamp
```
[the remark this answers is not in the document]
mohitjeswani01: disabled_until on notifiers is parsed as Timestamp (iso8601), consistent with createdAt — changed per reviewer feedback from @antonmry.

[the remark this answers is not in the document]
mohitjeswani01: i modified the description of pr also

antonmry (approved):
Thanks for the contribution, @mohitjeswani01 — really appreciate how thoroughly you worked through both rounds of feedback. All 18 of my comments are addressed in `d0caeae`: nullability flips, the `updated_at` columns, `AXIOM_API_BASE` rename, `row_strategy: direct` on all four tables, README disclaimer and dedup, COALESCE guidance, and the filtered test query. Nicely done.

One small housekeeping ask before merge: could you click **Resolve conversation** on each of the threads now that they're addressed? It makes the merge state easier to read for anyone landing on the PR later. Once those are resolved, please go ahead and merge.

Welcome to Coral 🪸

mohitjeswani01:
> Thanks for the contribution, @mohitjeswani01 — really appreciate how thoroughly you worked through both rounds of feedback. All 18 of my comments are addressed in `d0caeae`: nullability flips, the `updated_at` columns, `AXIOM_API_BASE` rename, `row_strategy: direct` on all four tables, README disclaimer and dedup, COALESCE guidance, and the filtered test query. Nicely done.
> 
> One small housekeeping ask before merge: could you click **Resolve conversation** on each of the threads now that they're addressed? It makes the merge state easier to read for anyone landing on the PR later. Once those are resolved, please go ahead and merge.
> 
> Welcome to Coral 🪸

done sir all the comments which you made i have resolved all of those thanks alot and also thanks for informing me about the resolve aspect i will make sure to do so for further contributions @antonmry

PR #646, opened 2026-05-20
feat(cli): build embedded UI by default

sauldhernandez:
## Intent
Enable the CLI embedded UI as the default build path while keeping PR Rust checks independent from Node.js.

## What it enables
Default, main, and release builds now compile with `embedded-ui`, Cargo builds run `npm run build` with clear missing Node/npm errors, and PR checks can still use `--no-default-features`.

## Usage examples
- `cargo build -p coral-cli --locked`
- `cargo build -p coral-cli --no-default-features --locked`
- `make rust-checks`

Validation: `make rust-checks`, default-feature `cargo check -p coral-cli --all-targets --locked --features cli-test-server`, missing-Node failure-path checks, workflow YAML parsing, and `git diff --check`.


docs/getting-started/installation.mdx
```diff
@@ -25,9 +25,10 @@ Coral ships as a single local CLI. Once installed, the same binary is used for s
   </Tab>
 
   <Tab title="Build from source">
-    Requires [Rust and Cargo](https://www.rust-lang.org/tools/install) installed.
+    Requires [Rust and Cargo](https://www.rust-lang.org/tools/install) plus Node.js/npm for the embedded UI.
```
AlbertQM: Do we want to give a bit more information about this being optional? (same for the `npm ci` command); or do we expect everyone to want/build the UI, and don't want to push/advertise against it?

.github/workflows/build-timings.yml
```diff
@@ -24,6 +24,16 @@ jobs:
       - name: Install Rust toolchain
         run: rustup show
 
+      - name: Set up Node
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
+        with:
+          node-version: 22
```
simonwhitaker: nit: why 22? 24 is the latest LTS release, 26 the latest release. Worth updating to one of those?

Also, whichever version we use, we should also pin that in the repo (e.g. with a `.node-version` in `ui/`)

.github/workflows/release.yml
```diff
@@ -79,36 +39,23 @@ jobs:
...
-          path: ui/dist
+          node-version: 22
+          package-manager-cache: false
+
+      - name: Install UI dependencies
+        run: npm ci --prefix ui
+
       # No Rust cache here — release builds must compile from a clean state
       # to guard against cache poisoning attacks.
-      - name: Install cross
```
simonwhitaker, line 90: Same comment as above. This looks unrelated.

.github/workflows/release.yml
```diff
@@ -15,54 +15,14 @@ concurrency:
...
-    needs: build-ui
     strategy:
       matrix:
         include:
           - target: x86_64-unknown-linux-gnu
             os: ubuntu-latest
           - target: aarch64-unknown-linux-gnu
-            os: ubuntu-latest
-            cross: true
+            os: ubuntu-24.04-arm
```
simonwhitaker, line 28: Is this change necessary for this UI build change? If not, please move to a separate PR.

.github/workflows/validate.yml
```diff
@@ -103,6 +105,18 @@ jobs:
...
+      - name: Set up Node
+        if: ${{ github.ref == 'refs/heads/main' }}
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
+        with:
+          node-version: 22
+          cache: npm
+          cache-dependency-path: ui/package-lock.json
+
+      - name: Install UI dependencies
+        if: ${{ github.ref == 'refs/heads/main' }}
```
simonwhitaker: Why are we only doing this on main?

.github/workflows/validate.yml
```diff
@@ -94,7 +94,9 @@ jobs:
     name: Rust checks
     runs-on: ubuntu-latest
     needs: changes
-    if: ${{ needs.changes.outputs.rust_checks == 'true' }}
+    if: ${{ needs.changes.outputs.rust_checks == 'true' || (github.ref == 'refs/heads/main' && needs.changes.outputs.ui-build == 'true') }}
+    env:
+      RUST_CHECK_FEATURE_FLAGS: ${{ github.ref == 'refs/heads/main' && '--features coral-cli/cli-test-server' || '--no-default-features --features coral-cli/cli-test-server' }}
```
jsummerfield: This makes PR Rust checks skip the default feature set even though `embedded-ui` is now the default build. That leaves the production/default CLI path uncompiled before merge: a PR can break `embedded_ui`, `run_ui`, or the build script and still pass until `main` catches it after merge. Could we add a PR gate that installs Node and runs at least `cargo check -p coral-cli --locked` for Rust/build-script changes, or otherwise keep the default feature path covered pre-merge?

.github/workflows/release.yml
```diff
@@ -79,36 +39,23 @@ jobs:
...
+      - name: Set up Node
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
         with:
-          name: ui-dist
-          path: ui/dist
+          node-version: 22
+          package-manager-cache: false
+
+      - name: Install UI dependencies
+        run: npm ci --prefix ui
```
jsummerfield, line 53: This workflow still supports rebuilding an existing tag, but after checking out older tags this unconditional `npm ci --prefix ui` fails before Cargo can build the old CLI. Can we keep the previous package-lock detection/guard here so tags that predate the UI package remain rebuildable?

crates/coral-cli/src/lib.rs
```diff
@@ -399,27 +399,33 @@ pub async fn run(app: AppClient, ctx: coral_app::RunContext) -> Result<(), CliEr
...
+type CliFuture = std::pin::Pin<Box<dyn std::future::Future<Output = Result<(), CliError>>>>;
+
+fn run_no_runtime_command(command: Command) -> CliFuture {
     match command {
         Command::Completion(args) => {
-            let mut cmd = Cli::command();
-            let bin_name = cmd.get_name().to_string();
-            generate(args.shell, &mut cmd, bin_name, &mut std::io::stdout());
-            Ok(())
+            run_completion(&args);
```
jsummerfield, line 412: This now runs completion generation before `run_with_context` instruments the returned future, so the actual completion work no longer happens under the root CLI span. Could we defer this into the returned future, e.g. `Box::pin(async move { run_completion(&args); Ok(()) })`, to preserve the previous tracing behavior?

.github/workflows/release.yml
sauldhernandez: We removed `cross` because the release build now generates the embedded UI from `crates/coral-cli/build.rs`. That build script invokes `npm`, and with `cross` Cargo runs inside the cross container, where the Node/npm UI toolchain is not available. Rather than maintain a custom cross image just for the UI build, the ARM Linux artifact now builds on GitHub’s native ARM runner.

.github/workflows/release.yml
simonwhitaker: To rebuild an existing tag you use workflow_dispatch, and that lets you choose which version of the workflow you use. So we can still rebuild previous tags, e.g. using the workflow from the tag itself. Does that alleviate your concern?

CleanShot 2026-05-21 at 11  55 41@2x

ui/.node-version
```diff
@@ -0,0 +1 @@
+24
```
simonwhitaker: I'd suggest we go specific: 24.15.0 (latest current LTS release)

.github/workflows/release.yml
sauldhernandez: Agree. We can do it this way if need be.

simonwhitaker (approved):

jsummerfield:
Hold off merging for now - see Slack for discussion.

jsummerfield:
See https://github.com/withcoral/coral/pull/658

PR #649, opened 2026-05-20
fix(ui): tweak query stream search bar

AlbertQM:
## Summary

Says on the tin.

## Test plan

https://github.com/user-attachments/assets/095d015c-1056-436f-a077-eca463814dfd

### DIFF
--- a/ui/src/views/TracesPage.tsx
+++ b/ui/src/views/TracesPage.tsx
@@ -3,6 +3,7 @@ import { useCallback, useEffect, useRef, useState } from 'react'
 import * as Button from '@/wax/components/button'
 import { Icon } from '@/wax/components/icon'
 import { TextInput } from '@/wax/components/inputs/text'
+import { KeyboardShortcut } from '@/wax/components/keyboard-shortcut'
 import { Typography } from '@/wax/components/typography'
 import { listTraces } from '@/lib/coral-traces-client'
 import type { TraceSummary } from '@/generated/coral/v1/traces_pb'
@@ -75,6 +76,14 @@ function HeaderActions({ expanded, searchText, setExpanded, setSearchText }: {
 
   return (
     <div className={s.headerActions}>
+      <KeyboardShortcut
+        handler={(e) => {
+          e.preventDefault()
+          setExpanded(true)
+          inputRef.current?.focus()
+        }}
+        shortcut="$mod+f"
+      />
       <div className={s.inlineSearch} data-expanded={expanded ? 'true' : undefined}>
         <div className={s.searchTrigger}>
           <Button.IconButton name="Search" onClick={() => setExpanded(true)} size="32" tooltipText="Search" variant="bare" />
--- a/ui/src/views/TracesPage.tsx
+++ b/ui/src/views/TracesPage.tsx
@@ -61,32 +61,36 @@ function useTraceList(enabled: boolean) {
   return { error, loading, traces }
 }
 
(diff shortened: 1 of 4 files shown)


ui/src/views/TracesPage.tsx
```diff
@@ -61,32 +61,36 @@ function useTraceList(enabled: boolean) {
...
+      <div className={s.inlineSearch} data-expanded={expanded ? 'true' : undefined}>
+        <div className={s.searchTrigger}>
+          <Button.IconButton name="Search" onClick={() => setExpanded(true)} size="32" tooltipText="Search" variant="bare" />
         </div>
-      ) : (
-        <Button.IconButton name="Search" onClick={() => setShowSearch(true)} size="32" tooltipText="Search queries" variant="bare" />
-      )}
+        <div className={s.searchField}>
+          <TextInput
+            icon="Search"
```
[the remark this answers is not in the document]
AlbertQM: We don't want that.

ui/src/views/TracesPage.tsx
```diff
@@ -61,32 +62,44 @@ function useTraceList(enabled: boolean) {
...
-              type="search"
-              value={searchText}
-            />
-          </div>
-          <Button.IconButton name="X" onClick={onClearSearch} size="32" tooltipText="Close search" variant="bare" />
+      <KeyboardShortcut
+        handler={(e) => {
+          e.preventDefault()
+          setExpanded(true)
+          inputRef.current?.focus()
```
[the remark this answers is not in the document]
AlbertQM: Better, thanks!

ui/src/views/TracesPage.tsx
```diff
@@ -61,32 +62,44 @@ function useTraceList(enabled: boolean) {
   return { error, loading, traces }
 }
 
-function HeaderActions({ onClearSearch, searchText, setSearchText, setShowSearch, showSearch }: {
-  onClearSearch: () => void
+function HeaderActions({ expanded, searchText, setExpanded, setSearchText }: {
+  expanded: boolean
```
lazybean: I may need way more :coffee:.  
What does `expanded` mean here? Do we need it at all? It seems to just mean `focused`

lazybean: Just got more coffee.  
  
This is for mobile breakpoint. Instead of expanding on focus / collapsing on blur, there is additional complexity for only collapsing on blur if there is no input in the search.  
  
ESC will clear the input and collapse

lazybean (approved):
It looks so much better.

ui/src/views/TracesPage.tsx
lazybean: I’m okay with local state here, but I’d avoid calling it `expanded` and avoid using it directly as visibility. Maybe rename it to searchOpen / mobileSearchOpen,   
then derive `searchVisible =searchOpen || searchText.trim().length > 0`.   
  
That keeps an active filter visible on mobile even after blur, without introducing focus-derived layout complexity.

AlbertQM: Good point, will update to a human readable version -- thank you!
For mobile viewport, it will actually expand 😄, so I'll use `isSearching` to cover the intent more than "what it looks like", since that differs between mobile and desktop

AlbertQM: My comment was in response to your first message only.. I've now seen the rest: will refactor to make it more robust, thank you!

PR #651, opened 2026-05-20
refactor(engine): split HTTP client modules

jsummerfield:
## Intent

`crates/coral-engine/src/backends/http/client.rs` had accumulated request construction, pagination mutation, transport/retry handling, response decoding, URL normalization, tracing, registration-time checks, and test helpers in one file. This change splits those responsibilities into focused HTTP backend modules while keeping `HttpSourceClient` as the configured source wrapper.

This PR is now rebased directly on `main`, after #402 landed, and HTTP reuses `shared::response_rows::extract_rows` instead of introducing another backend-local row extraction implementation.

## Changes

- Split HTTP pagination, request building, fetch orchestration, response decoding, tracing, transport, URL handling, registration checks, and test helpers into dedicated modules.
- Kept `client.rs` focused on source-client construction and delegation to the paginated fetch path.
- Reused shared `ResponseSpec`-driven row extraction for HTTP rows.
- Preserved current `main` behavior for search limits and conditional body fields while moving the fetch loop into `http::fetch`.
- Addressed review feedback around `Link` header parsing, mixed-case HTTP(S) base URL schemes, omitted optional JSON request bodies, and trace URL userinfo redaction.
- Moved shared value-source tests to the shared template module and the rate-limit manifest parse test to the rate-limit module.

## Validation

- `cargo check -p coral-engine --all-targets`
- `cargo test -p coral-engine --lib backends::http` (52 passed)
- `cargo test -p coral-engine --test engine http_tests::` (42 passed)
- `make rust-checks`
  - `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`
  - `cargo nextest run --workspace --all-targets --all-features --locked --no-fail-fast` (754 passed)
  - `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --all-features --no-deps --locked`


crates/coral-engine/src/backends/http/response.rs
```diff
@@ -0,0 +1,369 @@
...
+        url: Some(logged_url.to_string()),
+        detail: format!("source API response decoding failed: {error}"),
+    })
+}
+
+#[expect(
+    clippy::unnecessary_wraps,
+    reason = "Keeping a Result return type preserves a uniform extraction interface for callers"
+)]
+pub(super) fn extract_rows(target: &HttpFetchTarget, payload: &Value) -> Result<Vec<Value>> {
```
antonmry: Heads-up on overlap with #402: that PR also pulls `extract_rows` out of `client.rs`, but into a cross-backend module at `crates/coral-engine/src/backends/shared/response_rows.rs` so MCP and HTTP share the same row extraction (one of #402's stated goals: avoid backend duplication). Whichever of these PRs lands second will have a textual conflict on `client.rs::extract_rows` and an architectural overlap on the new location.

If you're willing, pointing HTTP at the shared module would give the codebase one source of truth — the shared signature is `fn extract_rows(response: &ResponseSpec, payload: &Value) -> Vec<Value>`, so call sites pass `target.response()` instead of `target` and drop the `?`/`.unwrap()`. Happy to do the merge legwork on the #402 side if you'd rather keep this PR pure-refactor.

crates/coral-engine/src/backends/http/response.rs
```diff
@@ -0,0 +1,369 @@
...
+            };
+            match root {
+                Value::Array(items) => Ok(items.clone()),
+                Value::Null => Ok(Vec::new()),
+                other => Ok(vec![other.clone()]),
+            }
+        }
+        RowStrategy::SeriesPointList => {
+            let mut rows = Vec::new();
+            let series = get_path_value(payload, &["series".to_string()])
```
antonmry: Cubic flagged this in its review (P1: "Honor `rows_path` when extracting `SeriesPointList` rows"). #402 already fixed the same bug in `shared::response_rows::extract_series_point_list` by routing through a `response_root(...)` helper that applies `rows_path` first. If you adopt the shared module (see comment on line 108), this fix comes along for free — otherwise it needs porting separately.

crates/coral-engine/src/backends/http/response.rs
jsummerfield: Oh nice, thanks. I'll rebase this PR on 402.

crates/coral-engine/src/backends/http/validation.rs
```diff
@@ -0,0 +1,551 @@
+//! Source-scoped HTTP manifest validation.
```
antonmry: Nit on the doc-comment: "manifest validation" reads like a `coral-spec` concern (well-formed templates, schema conformance, cross-reference checks) — but this module is actually doing *registration-time* checks that depend on runtime context (`resolved_inputs`, the authenticator registry). Two different layers.

Suggest tightening to something like:

```rust
//! Registration-time HTTP source checks: verifies that template inputs
//! resolve against the supplied inputs map and that referenced auth
//! schemes are wired up in the engine's authenticator registry.
//! Pure manifest validation (schema, well-formedness) lives in coral-spec.
```

Same point applies to the filename — `registration_checks.rs` (or similar) would make the layer obvious at a glance, but renaming may not be worth the churn.

antonmry (approved):

crates/coral-engine/src/backends/http/test_support.rs
```diff
@@ -0,0 +1,165 @@
+//! Test helpers shared by HTTP backend unit tests.
```
antonmry, line 1: Nit on cross-backend consistency: the other two engine backends keep tests inside the module as `tests.rs` (MCP in #402, File in #453 — both put unit tests + helpers in one in-module file). HTTP here is the only one with a separate `test_support.rs` for helpers (with the `#[test]` cases presumably elsewhere).

I don't have a strong opinion on which name wins — `test_support.rs` is arguably more descriptive when helpers are large enough to extract — but having two-against-one across the backends will trip up newcomers wondering where to put a new test. Worth picking one convention.

If you keep this name, a short doc-comment note (e.g. `//! Test helpers; #[test] cases live in <where>`) would help readers find the actual tests.

crates/coral-engine/src/backends/http/validation.rs
antonmry: Small naming/layering concern: this module is doing registration-time runtime checks, not manifest validation in the `coral-spec` sense. #402's backend layout tries to keep manifest-shape validation in `coral-spec` and engine-side runtime checks in the backend.

A rename to something like `registration_checks.rs` would make that boundary obvious, but even a module comment that spells out the separation would help: resolved inputs/authenticator wiring here; schema/well-formed manifest validation in `coral-spec`.

crates/coral-engine/src/backends/http/response.rs
antonmry: This split is fine for this PR, but once the backend-structure work in #402 lands I think HTTP should get one follow-up refactor to match that shape.

Concretely, `client.rs` should probably stop owning the multi-request fetch/pagination loop and `http::response` should stop owning `ResponseSpec` / `RowStrategy` row extraction. The target boundary from #402 would be: `client.rs` as the configured source wrapper, `fetch.rs` for paginated fetch orchestration, `response.rs` for decoding one HTTP response into JSON, and `shared::response_rows::extract_rows` for turning that JSON into rows.

That keeps HTTP and MCP structurally parallel without forcing their transport details to be shared.

PR #656, opened 2026-05-21
feat: dependent join config

iaptsiauri:


docs/reference/engine-configuration.mdx
```diff
@@ -0,0 +1,31 @@
+---
+title: "Engine configuration"
+description: "Configure query-engine runtime policy in Coral's local config.toml."
+---
+
+Coral reads engine settings from `config.toml` in the local state directory. See [Local state](/getting-started/installation#local-state) for the default location and the `CORAL_CONFIG_DIR` override.
+
+## Dependent joins
+
+Dependent joins let Coral rewrite eligible SQL joins into dependent predicate pushdown, so join-key values from one side can be sent as filters to an HTTP-backed source. The global settings live under `[engine.dependent_join]`.
```
antonmry: Question: since these limits are now global, do we want a source/table override shape in config.toml now, or should we explicitly document that per-source tuning is intentionally unsupported? Moving `max_concurrency` out of source manifests removes the only source-specific control.

iaptsiauri: You are right, per source was skipped but added now :)

crates/coral-app/src/state/config.rs
```diff
@@ -325,7 +399,117 @@ impl From<&AppConfig> for PersistedAppConfig {
         Self {
             version: value.version,
             workspaces,
+            engine: value.engine.clone(),
```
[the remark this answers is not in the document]
iaptsiauri: We don't need to persist engine or rewrite the config. I think you are wrong here

crates/coral-app/src/query/manager.rs
```diff
@@ -213,11 +227,21 @@ impl QueryManager {
...
+        config: &AppConfig,
+    ) -> Result<QueryRuntimeConfig, AppError> {
+        let mut runtime = QueryRuntimeConfig::new(
             self.runtime_context.clone(),
             engine_extensions_for_providers(&self.engine_extensions_providers, selected_sources),
-        )
+        );
+        let selected_source_names = selected_sources
+            .iter()
+            .map(|source| source.source_name().to_string())
```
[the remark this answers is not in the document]
iaptsiauri: I don’t think this is a bug in the current app model. QuerySource::source_name() is the manifest schema name, and that is also the SQL schema the engine registers and resolves.

antonmry (approved):

docs/reference/source-spec-reference.mdx
```diff
@@ -499,7 +499,6 @@ Coral always treats `429 Too Many Requests` as a rate-limit signal. Use `extra_s
 | `retry_after_header` | string          | `Retry-After`   | Header that carries a retry delay in seconds or HTTP-date form              |
 | `remaining_header`   | string          | —               | Header whose value `0` means the current quota is exhausted                 |
 | `reset_header`       | string          | —               | Header with the quota reset time as Unix epoch seconds                      |
-| `max_concurrency`    | integer         | —               | Maximum concurrent dependent-join HTTP calls for this source                |
```
AlbertQM, line 852: This has moved to the engine config only?

AlbertQM (approved):

docs/reference/source-spec-reference.mdx
iaptsiauri: yes this one moved away completely from spec references actually.

simonwhitaker (approved):

PR #658, opened 2026-05-21
build(cli): make embedded UI build explicit

jsummerfield:
## Intent

This is an alternative to #646 that keeps the embedded UI enabled as the default CLI surface without making Cargo or `build.rs` run npm.

The design split is:
- Cargo compiles the embedded UI code path by default.
- Local Rust builds may compile without `ui/dist`; UI development normally serves assets from Vite.
- Make/CI own UI artifact generation and validation.
- Release/install paths build the UI before packaging the CLI.

## Shape

- Enables `coral-cli`'s default `embedded-ui` feature.
- Leaves `rust-embed` tolerant of missing local `ui/dist` so Rust checks do not depend on frontend artifacts.
- Keeps `build.rs` limited to rebuild hints and embedded version metadata; it does not invoke npm or validate UI artifacts.
- Adds `make ui-build`, validates `ui/dist/index.html`, and makes `make install` build UI first.
- Restores `make rust-checks` to `--all-features`.
- Updates CI UI build paths to run `npm ci`, build the UI, verify `ui/dist/index.html`, and reject mock-only code in production output.
- Pins UI Node selection with `ui/.node-version`; UI release builds now require that file.

## Validation

- `cargo check -p coral-cli --all-targets --all-features --locked` with `ui/dist` temporarily hidden
- `make ui-build`
- production mock-code grep from `validate.yml`
- `npm run test:ui:type-check --prefix ui`
- `make rust-checks`
- `git diff --check`

`npm run test:ui --prefix ui` was attempted locally but blocked by missing system Playwright dependencies in this environment.


simonwhitaker (approved):

lazybean (approved):
Thank you

sauldhernandez (approved):

PR #659, opened 2026-05-21
feat(sources/community): add Open-Meteo sources

cxhihilwb123-hash:
## Summary

Adds two public no-auth Open-Meteo community sources:

- `open_meteo_geocoding` for city/place lookup and stable Open-Meteo location ID resolution.
- `open_meteo_weather` for current weather by latitude/longitude.

These sources are useful for agents that need location-aware planning, travel,
outdoor activity, scheduling, and operations context without requiring users to
set up API credentials.

## Validation

- `cargo run --ignore-rust-version -q -p coral-cli -- source lint sources/community/open_meteo_geocoding/manifest.yaml`
- `cargo run --ignore-rust-version -q -p coral-cli -- source lint sources/community/open_meteo_weather/manifest.yaml`
- Added and tested `open_meteo_geocoding` in a temporary `CORAL_CONFIG_DIR`: 2 declared query tests, 2 passed, 0 failed.
- Added and tested `open_meteo_weather` in a temporary `CORAL_CONFIG_DIR`: 1 declared query test, 1 passed, 0 failed.

Note: the local machine had Rust 1.94.1 while this repo's `rust-toolchain.toml`
requires Rust 1.95, so validation used Cargo's `--ignore-rust-version` flag.

### DIFF
--- a/sources/community/open_meteo_geocoding/manifest.yaml
+++ b/sources/community/open_meteo_geocoding/manifest.yaml
@@ -1,0 +1,299 @@
+name: open_meteo_geocoding
+version: 0.1.0
+dsl_version: 3
+backend: http
+description: >-
+  Search global places and resolve Open-Meteo location IDs through the public
+  Open-Meteo Geocoding API. No authentication required.
+base_url: https://geocoding-api.open-meteo.com/v1
+
+request_headers:
+  - name: Accept
+    from: literal
+    value: application/json
+
+test_queries:
+  - SELECT id, name, country, latitude, longitude FROM open_meteo_geocoding.locations WHERE name = 'Shanghai' AND count = 1 LIMIT 1
+  - SELECT id, name, timezone FROM open_meteo_geocoding.location_by_id WHERE id = '1796236' LIMIT 1
+
+tables:
+  - name: locations
+    description: |
+      Search cities, towns, and geographic places by name. Use this table to
+      discover latitude, longitude, timezone, elevation, country, and admin
+      metadata before calling weather sources such as open_meteo_weather.
+      Optional filters include count, language, country_code, and format.
+    fetch_limit_default: 10
+    filters:
+      - name: name
+        required: true
+      - name: count
+      - name: language
+      - name: country_code
+      - name: format
+    request:
+      method: GET
+      path: /search
+      query:
+        - name: name
+          from: filter
+          key: name
+        - name: count
+          from: filter_int
+          key: count
+        - name: language
+          from: filter
+          key: language
+        - name: countryCode
+          from: filter
+          key: country_code
+        - name: format
+          from: filter
+          key: format
+    response:
+      rows_path:
+        - results
+      row_strategy: direct
+    pagination:
+      mode: none
+    columns:
+      - name: id
+        type: Int64
+        nullable: false
+        description: Open-Meteo geocoding location identifier.
+      - name: name
+        type: Utf8
+        nullable: false
+        description: Place name.
+      - name: latitude
+        type: Float64
+        nullable: false
+        description: Latitude in decimal degrees.
+      - name: longitude
+        type: Float64
+        nullable: false
+        description: Longitude in decimal degrees.
+      - name: elevation
+        type: Float64
+        nullable: true
+        description: Elevation in meters above sea level.
+      - name: feature_code
+        type: Utf8
+        nullable: true
+        description: GeoNames feature code for the place.
+      - name: country_code
+        type: Utf8
+        nullable: true
+        description: ISO 3166-1 alpha-2 country code.
+      - name: country
+        type: Utf8
+        nullable: true
+        description: Country name.
+      - name: country_id
+        type: Int64
+        nullable: true
+        description: GeoNames country identifier.
+      - name: admin1
+        type: Utf8
+        nullable: true
+        description: First-level administrative area name.
+      - name: admin1_id
+        type: Int64
+        nullable: true
+        description: First-level administrative area identifier.
+      - name: admin2
+        type: Utf8
+        nullable: true
+        description: Second-level administrative area name.
+      - name: admin2_id
+        type: Int64
+        nullable: true
+        description: Second-level administrative area identifier.
+      - name: admin3
+        type: Utf8
+        nullable: true
+        description: Third-level administrative area name.
+      - name: admin3_id
+        type: Int64
+        nullable: true
+        description: Third-level administrative area identifier.
+      - name: admin4
+        type: Utf8
+        nullable: true
+        description: Fourth-level administrative area name.
+      - name: admin4_id
+        type: Int64
+        nullable: true
+        description: Fourth-level administrative area identifier.
+      - name: timezone
+        type: Utf8
+        nullable: true
+        description: IANA timezone for the place.
+      - name: population
+        type: Int64
+        nullable: true
+        description: Population when available.
+      - name: postcodes
+        type: Json
+        nullable: true
+        description: Postal codes returned by Open-Meteo when available.
+      - name: raw
+        type: Json
+        nullable: false
+        description: Full raw location object.
+        expr:
+          kind: current_row
+      - name: query_name
+        type: Utf8
+        nullable: true
+        virtual: true
+        description: Echoes the requested place-name filter.
+        expr:
+          kind: from_filter
+          key: name
+      - name: count
+        type: Int64
+        nullable: true
+        virtual: true
+        description: Echoes the requested maximum result count filter.
+        expr:
+          kind: from_filter
+          key: count
+      - name: language
+        type: Utf8
+        nullable: true
+        virtual: true
+        description: Echoes the requested language filter.
+        expr:
+          kind: from_filter
+          key: language
+      - name: format
+        type: Utf8
+        nullable: true
+        virtual: true
+        description: Echoes the requested response format filter.
+        expr:
+          kind: from_filter
+          key: format
+
+  - name: location_by_id
+    description: |
+      Resolve a single Open-Meteo geocoding location by numeric ID. Useful after
+      selecting one row from locations and needing a stable lookup.
+    filters:
+      - name: id
+        required: true
+      - name: language
+      - name: format
+    request:
+      method: GET
+      path: /get
+      query:
+        - name: id
+          from: filter_int
+          key: id
+        - name: language
+          from: filter
+          key: language
+        - name: format
+          from: filter
+          key: format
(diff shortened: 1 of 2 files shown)


sources/community/open_meteo_geocoding/manifest.yaml
```diff
@@ -0,0 +1,299 @@
+name: open_meteo_geocoding
+version: 0.1.0
+dsl_version: 3
+backend: http
+description: >-
+  Search global places and resolve Open-Meteo location IDs through the public
+  Open-Meteo Geocoding API. No authentication required.
+base_url: https://geocoding-api.open-meteo.com/v1
```
antonmry, line 8: Suggestion (structural): could these two manifests be folded into a single `open_meteo` source with three tables — `locations`, `location_by_id`, and `current_weather`? Both wrap the same vendor's no-auth APIs and they only become useful together (resolve a place name → query weather at the resulting coordinates).

As shipped, users have to add two sources and remember which table lives where. The DSL supports per-table `base_url`, so the different hosts (`geocoding-api.open-meteo.com` vs `api.open-meteo.com`) shouldn't force the split. Renaming after release is awkward, so worth deciding before merging.

sources/community/open_meteo_weather/manifest.yaml
```diff
@@ -0,0 +1,292 @@
...
+      WHERE latitude = '31.23' AND longitude = '121.47' AND timezone = 'auto'.
+    filters:
+      - name: latitude
+        required: true
+      - name: longitude
+        required: true
+      - name: timezone
+      - name: temperature_unit
+      - name: wind_speed_unit
+      - name: precipitation_unit
```
antonmry, line 39: Ergonomics: `latitude` and `longitude` are declared as plain (string) filters here, which is why the test query at line 16 has to write `WHERE latitude = '31.23'` (quoted). SQL users will instinctively write `WHERE latitude = 31.23` and get nothing back. Would you mind using `filter_float` (or whichever numeric filter kind Coral exposes) so the comparison can be numeric? The virtual echo columns above would then be `Float64` and match `response_latitude`/`response_longitude`.

sources/community/open_meteo_geocoding/manifest.yaml
```diff
@@ -0,0 +1,299 @@
...
+      metadata before calling weather sources such as open_meteo_weather.
+      Optional filters include count, language, country_code, and format.
+    fetch_limit_default: 10
+    filters:
+      - name: name
+        required: true
+      - name: count
+      - name: language
+      - name: country_code
+      - name: format
```
antonmry, line 33: Small footgun: per the Open-Meteo geocoding docs, `format` accepts `json` (default) and `protobuf`. If a user passes `WHERE format = 'protobuf'`, Coral's JSON parser will fail with a confusing error. Could you either drop the filter entirely or pin it to JSON via `from: literal value: json` so users can't shoot their foot? Same applies to the `format` filter on `location_by_id` (line 187).

Ref: https://open-meteo.com/en/docs/geocoding-api

sources/community/open_meteo_geocoding/manifest.yaml
antonmry, line 8: Heads-up that there's no README for either source. Two things from Open-Meteo's terms it would be worth surfacing somewhere readers will find:

1. **Attribution** — Open-Meteo data is CC BY 4.0 and requires attribution. Users querying it from agents won't see this from the API alone.
2. **Rate limits** — the free tier is documented around ~10k requests/day; users hitting that won't immediately know why their queries start failing.

A short `sources/community/open_meteo/README.md` (or per-source READMEs) with attribution, rate-limit guidance, and a "geocode → forecast" combined example would cover this nicely.

sources/community/open_meteo_geocoding/manifest.yaml
antonmry, line 33: Three knobs control how many rows come back here and the interaction isn't obvious: the `count` query-param filter (server-side cap, max 100 per Open-Meteo docs), `fetch_limit_default: 10`, and SQL `LIMIT N` on top. A user writing `LIMIT 50` won't know whether they actually get 50, 10, or N≤100. Could we either drop the `count` filter (so SQL `LIMIT` pushes down cleanly) or document the precedence in the table `guide`?

sources/community/open_meteo_geocoding/manifest.yaml
```diff
@@ -0,0 +1,299 @@
...
+        expr:
+          kind: current_row
+      - name: requested_id
+        type: Utf8
+        nullable: true
+        virtual: true
+        description: Echoes the requested id filter.
+        expr:
+          kind: from_filter
+          key: id
```
antonmry, line 299: Type inconsistency: the `id` filter at line 184 uses `from: filter_int` (numeric), and the response-side `id` column at line 206 is `Int64`. But this echo column is `Utf8`. Could we make this `Int64` for consistency? Relatedly, the test query at line 17 has `WHERE id = '1796236'` — it should be the numeric `1796236` to match the filter type.

sources/community/open_meteo_geocoding/manifest.yaml
```diff
@@ -0,0 +1,299 @@
...
+      - name: postcodes
+        type: Json
+        nullable: true
+        description: Postal codes returned by Open-Meteo when available.
+      - name: raw
+        type: Json
+        nullable: false
+        description: Full raw location object.
+        expr:
+          kind: current_row
```
antonmry, line 145: Question (also applies to `location_by_id.raw` at line 286 and `current_weather.raw` in the other manifest): every other field is already broken out as a typed column, so `raw` duplicates the same data as JSON and roughly doubles the row size. If it's intentionally an escape hatch for unmapped fields, would you mind saying that explicitly in the column description? Otherwise, would removing it be reasonable?

sources/community/open_meteo_weather/manifest.yaml
```diff
@@ -0,0 +1,292 @@
...
+            - snowfall
+      - name: weather_code
+        type: Int64
+        nullable: true
+        description: WMO weather interpretation code.
+        expr:
+          kind: path
+          path:
+            - current
+            - weather_code
```
antonmry, line 224: Small UX nudge: `weather_code` is an integer that maps to a WMO interpretation table (0 = clear, 95 = thunderstorm, etc.), and most users — especially agents — won't know the mapping. Could the description link to Open-Meteo's published mapping, or include a short legend in the table `guide`?

Ref: https://open-meteo.com/en/docs (WMO Weather interpretation codes table at the bottom of the forecast docs).

sources/community/open_meteo_weather/manifest.yaml
```diff
@@ -0,0 +1,292 @@
...
+        - name: latitude
+          from: filter
+          key: latitude
+        - name: longitude
+          from: filter
+          key: longitude
+        - name: current
+          from: literal
+          value: >-
+            temperature_2m,relative_humidity_2m,apparent_temperature,is_day,precipitation,rain,showers,snowfall,weather_code,cloud_cover,pressure_msl,surface_pressure,wind_speed_10m,wind_direction_10m,wind_gusts_10m
```
antonmry, line 53: Just flagging: the `current=` parameter is hard-coded to a fixed 16-variable list here, regardless of which columns the SQL query projects. That's totally fine for v1 (predictable schema, simpler manifest) — could you mention in the table `guide` that projection isn't pushed down, so users who want fewer variables for cost reasons know it doesn't actually trim the upstream request?

sources/community/open_meteo_weather/manifest.yaml
```diff
@@ -0,0 +1,292 @@
+name: open_meteo_weather
+version: 0.1.0
+dsl_version: 3
+backend: http
+description: >-
+  Query current weather conditions from the public Open-Meteo Forecast API by
+  latitude and longitude. No authentication required.
```
antonmry, line 7: Small scope-clarity nit: the description says "current weather conditions" — true, but a user reaching for `WHERE date > '2024-01-01'` or hourly data won't know that's out of scope until they try. Could you mention "current conditions only — no hourly, daily, historical, or air-quality data" here, so users discover the limit before writing a query?

sources/community/open_meteo_geocoding/manifest.yaml
```diff
@@ -0,0 +1,299 @@
...
+backend: http
+description: >-
+  Search global places and resolve Open-Meteo location IDs through the public
+  Open-Meteo Geocoding API. No authentication required.
+base_url: https://geocoding-api.open-meteo.com/v1
+
+request_headers:
+  - name: Accept
+    from: literal
+    value: application/json
```
antonmry, line 13: Minor nit (safe to ignore): `Accept: application/json` is the API's default response format anyway, so this header isn't doing anything. Removing it would just trim a couple of lines. Same applies to the equivalent header in `open_meteo_weather/manifest.yaml`.

jsummerfield (changes requested):
Findings:

- High: this PR currently adds two community source directories, `sources/community/open_meteo_geocoding/manifest.yaml` and `sources/community/open_meteo_weather/manifest.yaml`. Community source PRs need to be reviewable as one source at a time. Please either scope this PR to one source directory and open the other as a follow-up PR, or consolidate the work into a single coherent Open-Meteo source directory if that is the intended product surface.

- Medium: neither source has a README. Even for no-auth public APIs, the accepted source should have a first-success path: what users can query, how to combine geocoding with weather, required filters, example queries, rate-limit/usage notes, and captured live Coral output. The PR body has useful validation notes, but users need that guidance checked in with the source that survives the split/consolidation.

I did not run a full provider/API review because the multi-source scope needs to be resolved first. I did verify both manifests lint individually with `coral source lint`, so this is mainly a PR-shape and user-facing documentation blocker.

Thanks for putting together a useful no-auth data source; Open-Meteo is a good fit once the PR shape is cleaned up.

jsummerfield (changes requested):
I re-reviewed the current head (`9974439`). The individual manifests now lint, but the earlier product-scope blocker is still open.

Remaining blockers:

- This PR still adds two separate community sources, `sources/community/open_meteo_geocoding` and `sources/community/open_meteo_weather`. Please either consolidate the Open-Meteo surface into a single coherent source or split these into separate PRs with clear ownership and review scope.
- Neither new source has a README. Community sources need first-run setup/query documentation and live-test evidence, especially when there are two related provider surfaces.
- Style still fails on `sources/community/open_meteo_weather/manifest.yaml` because of a long line; please run `ryl --config-file .yamllint.yaml` locally after updating.

PR #669, opened 2026-05-21
feat(sources/community/semaphore_ci): add Semaphore CI community source

### LINKED ISSUE #634: design: add Semaphore CI community source
### Summary

Add a new `semaphore_ci` community source that lets Coral query Semaphore CI workflows, pipelines, and projects through the Semaphore v1alpha API. This gives users a read-only SQL view into their continuous integration and deployment infrastructure, making it easy to track build times, audit failed pipelines, and join CI statuses with GitHub pull requests.

### Motivation

Semaphore CI is a popular, high-performance CI/CD platform used by many engineering teams (especially in the Elixir, Ruby, and Go ecosystems). Currently, if a team wants to analyze their CI/CD performance—such as finding the average duration of a specific pipeline, or identifying which pull requests triggered the most failed builds—they have to write custom scripts against the Semaphore API.

By adding a read-only Coral source, users can use standard SQL to query their Semaphore infrastructure. They can easily track pipeline run times, audit which developers are triggering deployments, and join Semaphore workflow data with `github.pull_requests` to see exactly which code changes are blocking the main branch.

### Proposed design

Add a new community source under `sources/community/semaphore_ci`, implemented as a manifest-only HTTP source.

**Inputs:**
- `SEMAPHORE_API_TOKEN` (required secret)
  - Generated via the Semaphore account settings page.
  - Used for Header authentication.
- `SEMAPHORE_ORG_URL` (required variable)
  - The base URL of the user's Semaphore organization (e.g., `https://my-org.semaphoreci.com`).

**Auth:**
- Header auth
- `Authorization: Token {{input.SEMAPHORE_API_TOKEN}}`

**Proposed v1 Tables:**
- `semaphore_ci.projects`
  - `GET {{input.SEMAPHORE_ORG_URL}}/api/v1alpha/projects`
  - Lists all projects in the organization.
  - Exposes project names, IDs, repository connections, and creation dates.
- `semaphore_ci.workflows`
  - `GET {{input.SEMAPHORE_ORG_URL}}/api/v1alpha/plumber-workflows`
  - Lists workflows triggered in a project.
  - **Requires** `project_id` filter.
  - Exposes workflow triggers, commit SHAs, branch names, and requester IDs.
- `semaphore_ci.pipelines`
  - `GET {{input.SEMAPHORE_ORG_URL}}/api/v1alpha/pipelines`
  - Lists pipelines (the actual execution runs) for a workflow or project.
  - **Requires** `project_id` filter (and optionally `wf_id`).
  - Exposes pipeline states (running, passed, failed), durations, and error messages.

**Affected Surface:**
- Community source only.
- Expected files:
  - `sources/community/semaphore_ci/manifest.yaml`
  - `sources/community/semaphore_ci/README.md`
- No CLI, config, MCP, runtime, or bundled source changes are proposed.

### Drawbacks and trade-offs

The Semaphore API requires the organization URL to be dynamically injected into the base API path (e.g., `https://<organization-url>.semaphoreci.com/api/v1alpha/...`). This means users must provide two inputs (`SEMAPHORE_API_TOKEN` and `SEMAPHORE_ORG_URL`) to configure the source, rather than just a token.

Additionally, querying `workflows` and `pipelines` requires a `project_id`. Users will need to query the `projects` table first to find the UUID of their project, and then supply `WHERE project_id = '...'` to retrieve their CI runs.

### Alternatives considered

- **Including Jobs endpoints**: We considered adding the `/api/v1alpha/jobs` endpoint to allow querying individual command executions within a pipeline. However, jobs generate a massive amount of data and require complex nested filtering. We rejected this for v1 to keep the scope tightly focused on high-level workflow and pipeline statuses.
- **Hardcoding the base URL**: Some CI providers have a single global API endpoint. Because Semaphore uses organization-specific subdomains, we cannot hardcode the base URL in the manifest and must rely on user input (`SEMAPHORE_ORG_URL`) via variable interpolation.

### Open questions

- Does the community prefer we require `SEMAPHORE_ORG_URL` as a full URL (e.g., `https://myorg.semaphoreci.com`), or should we just ask for `SEMAPHORE_ORG_SLUG` (e.g., `myorg`) and interpolate it as `https://{{input.SEMAPHORE_ORG_SLUG}}.semaphoreci.com` in the manifest base URL? The slug approach is usually less error-prone.

### Scope check

mohitjeswani01:
## Summary

Fixes #634

Add a new `semaphore_ci` community source that lets Coral query
Semaphore CI pipelines and workflows through SQL — enabling build
audits, failure analysis, and CI/CD joins with GitHub pull requests.

model. Read-only. No CLI, MCP, config, or runtime changes.

## What changed

Added:
- `sources/community/semaphore_ci/manifest.yaml`
- `sources/community/semaphore_ci/README.md`

## Source scope

Inputs:
- `SEMAPHORE_API_TOKEN` — static API token (secret)
- `SEMAPHORE_ORG_SLUG` — org subdomain (variable), used to build
  `https://{{input.SEMAPHORE_ORG_SLUG}}.semaphoreci.com/api/v1alpha`

Tables:
- `semaphore_ci.workflows`
  - `GET /plumber-workflows`
  - requires `project_id` filter
  - optional `branch_name` filter
  - `link_header` pagination
  - 13 columns including `wf_id`, `state`, `result`, `commit_sha`,
    `branch_name`, `created_at`

- `semaphore_ci.pipelines`
  - `GET /pipelines`
  - requires `project_id` filter
  - optional `wf_id` filter
  - `link_header` pagination
  - 10 columns including `ppl_id`, `state`, `result`,
    `result_reason`, `error_description`, and 5 timestamp columns

- `semaphore_ci.promotions`
  - `GET /promotions`
  - requires `pipeline_id` filter
  - `mode: none` — small result set per pipeline
  - 4 columns

## Implementation notes

- auth: `Authorization: Token` (not Bearer) via `HeaderAuth`
  `from: template`
- `projects` table excluded — `GET /projects` endpoint does not
  exist in Semaphore v1alpha API
- `created_at` and other timestamps are protobuf
  `{seconds, nanos}` objects — extracted via
  `format_timestamp input: seconds` with path `[created_at, seconds]`
- all responses are root JSON arrays — uses `response: {}`
- no `User-Agent` header declared — Coral sets this automatically

## Validation Screenshot 
image

## Out of scope

Not included in v1:
- jobs endpoint (high data volume, complex nesting)
- write or trigger operations

### DIFF
--- a/sources/community/semaphore_ci/README.md
+++ b/sources/community/semaphore_ci/README.md
@@ -80,7 +80,9 @@ tag, pull request, or manual rerun.
 ### pipelines
 
 Pipelines for a project or workflow. Each row is one pipeline execution
-within a workflow. Requires `project_id`; optionally pass `wf_id` to
+within a workflow. This Coral source requires `project_id` for a safe
+default query path; Semaphore's API also accepts `wf_id` alone, but
+this source always requires `project_id`. Optionally pass `wf_id` to
 narrow to a single workflow.
 
 > **Important:** The pipeline list endpoint does **not** return `ppl_id`.
@@ -99,7 +101,8 @@ narrow to a single workflow.
 | `branch_name` | Utf8 | Git branch name |
 | `created_at` | Timestamp | Pipeline creation time (UTC) |
 
-**Required filter:** `project_id`
+**Required filter:** `project_id` (Coral source restriction — Semaphore's
+API also accepts `wf_id` alone)
 **Optional filters:** `wf_id`, `branch_name`, `yml_file_path`,
 `created_after`, `created_before`, `done_after`, `done_before`
 **Pagination:** Link header
@@ -129,7 +132,8 @@ Promotions triggered from a pipeline (e.g. deploy to staging/production).
 
 > **Note:** The promotions list endpoint only returns `name`, `status`,
 > and `triggered_by`. Timestamp and auto-promotion metadata are not
-> exposed by this endpoint.
+> exposed by this endpoint. Semaphore returns a server error (500) for
+> pipelines with no promotions configured rather than an empty array.
 
 ---
 
@@ -204,6 +208,11 @@ WHERE pipeline_id = 'initial-ppl-id-from-workflows';
 - **Case inconsistency**: The `state` and `result` fields may use different
   casing between API responses (e.g. `DONE` vs `done`, `FAILED` vs
   `failed`). Use `LOWER(result)` or `UPPER(state)` for reliable filtering.
-- **Pipelines require `project_id`**: The pipelines table requires
-  `project_id`. Optionally add `wf_id` to narrow to a specific workflow.
+- **Pipelines require `project_id`**: This is a Coral source restriction
+  for a safe first-query experience. Semaphore's API contract accepts
+  either `project_id` or `wf_id` (or both); this source always requires
+  `project_id` and treats `wf_id` as an optional narrowing filter.
+- **Promotions 500 on no-promotion pipelines**: Semaphore returns a
+  server error (500) for pipelines that have no promotions configured.
+  This is an upstream API behavior, not a source bug.
 - **API docs**: https://docs.semaphoreci.com/reference/api
--- a/sources/community/semaphore_ci/README.md
+++ b/sources/community/semaphore_ci/README.md
@@ -60,7 +60,7 @@ tag, pull request, or manual rerun.
 | `requester_id` | Utf8 | User or system that requested this workflow |
 | `repository_id` | Utf8 | Repository UUID |
 | `organization_id` | Utf8 | Organization UUID |
-| `initial_ppl_id` | Utf8 | Initial pipeline UUID |
+| `initial_ppl_id` | Utf8 | Initial pipeline UUID — use this for promotions |
 | `hook_id` | Utf8 | Webhook event UUID |
 | `branch_name` | Utf8 | Git branch name |
 | `branch_id` | Utf8 | Semaphore branch UUID |
@@ -81,30 +81,33 @@ Pipelines for a project or workflow. Each row is one pipeline execution
 within a workflow. Pass `project_id` or `wf_id` (or both) — the API
 requires at least one.
 
+> **Important:** The pipeline list endpoint does **not** return `ppl_id`.
+> To get a pipeline ID for the promotions table, use `initial_ppl_id`
+> from `semaphore_ci.workflows`.
+
 | Column | Type | Description |
 |---|---|---|
 | `project_id` | Utf8 | Project ID (echoed when provided) |
-| `ppl_id` | Utf8 | Pipeline UUID (may be absent in list) |
 | `name` | Utf8 | Pipeline name from YAML config |
 | `yaml_file_name` | Utf8 | YAML file defining this pipeline |
 | `working_directory` | Utf8 | Pipeline working directory |
 | `wf_id` | Utf8 | Parent workflow UUID |
 | `state` | Utf8 | State (pending, queuing, running, stopping, done) |
 | `result` | Utf8 | Result (passed, failed, stopped, canceled) |
-| `result_reason` | Utf8 | Reason for the result (e.g. test, malformed_yaml) |
-| `error_description` | Utf8 | Error details for config/system failures |
 | `branch_name` | Utf8 | Git branch name |
 | `created_at` | Timestamp | Pipeline creation time (UTC) |
-| `pending_at` | Timestamp | Time entered pending state (UTC) |
-| `queuing_at` | Timestamp | Time entered queuing state (UTC) |
-| `running_at` | Timestamp | Time pipeline started running (UTC) |
-| `stopping_at` | Timestamp | Time pipeline began stopping (UTC) |
-| `done_at` | Timestamp | Time pipeline finished (UTC) |
 
 **Required filter:** at least one of `project_id` or `wf_id`
 **Optional filters:** `branch_name`, `yml_file_path`, `created_after`,
 `created_before`, `done_after`, `done_before`
 **Pagination:** Link header
+
+> **Note:** Richer fields like `result_reason`, `error_description`,
+> `pending_at`, `running_at`, `done_at`, etc. are only available via
+> the describe endpoint (`GET /pipelines/:pipeline_id`), which is not
+> covered by this list table. Those timestamps also use a custom string
+> format (`YYYY-MM-DD HH:MM:SS.ffffffZ`), not the `{seconds, nanos}`
+> object used by the list endpoint.
 
 ---
 
@@ -130,18 +133,19 @@ Promotions triggered from a pipeline (e.g. deploy to staging/production).
 
 ## Typical Query Flow
 
-Since there is no projects list endpoint, the typical workflow is:
+Since there is no projects list endpoint and the pipeline list does not
+return `ppl_id`, the typical workflow is:
 
 1. **Get your project ID** from the Semaphore UI or CLI
-2. **Query workflows** for that project
+2. **Query workflows** for that project — this gives you `initial_ppl_id`
 3. **Drill into pipelines** using the project ID or a specific workflow ID
-4. **Check promotions** for a specific pipeline
+4. **Check promotions** using `initial_ppl_id` from step 2 as `pipeline_id`
 
 ## Example Queries
 
 ```sql
 -- List recent workflows for a project
-SELECT wf_id, branch_name, commit_sha, created_at
+SELECT wf_id, branch_name, commit_sha, initial_ppl_id, created_at
 FROM semaphore_ci.workflows
 WHERE project_id = 'your-project-uuid'
 LIMIT 10;
@@ -163,10 +167,10 @@ FROM semaphore_ci.pipelines
 WHERE project_id = 'your-project-uuid'
   AND branch_name = 'main';
 
--- List promotions for a pipeline
+-- List promotions using initial_ppl_id from workflows
 SELECT name, status, triggered_by
 FROM semaphore_ci.promotions
-WHERE pipeline_id = 'your-pipeline-uuid';
+WHERE pipeline_id = 'initial-ppl-id-from-workflows';
 ```
 
 ## Pagination
@@ -182,15 +186,18 @@ WHERE pipeline_id = 'your-pipeline-uuid';
 - **No projects endpoint**: The Semaphore v1alpha API does not provide a
   `GET /projects` endpoint. Obtain your `project_id` from the Semaphore
   dashboard (Project Settings) or CLI (`sem get project`).
+- **No `ppl_id` in pipeline list**: The pipeline list endpoint does not
+  return pipeline IDs. Use `initial_ppl_id` from the workflows table to
+  obtain the initial pipeline ID for promotions queries.
 - **Cloud-only**: The `base_url` assumes `*.semaphoreci.com`. Semaphore
   Enterprise users on a custom domain would need to fork and adjust.
 - **Read-only**: This source only uses GET endpoints. No create, update,
   or delete operations.
-- **Timestamps**: The API returns timestamps as protobuf-style
-  `{seconds, nanos}` objects. This source extracts the `seconds` field
-  and exposes each as a UTC `Timestamp` column. The pipelines table
-  exposes `created_at`, `pending_at`, `queuing_at`, `running_at`,
-  `stopping_at`, and `done_at` for build-duration analysis.
+- **Timestamps**: The pipeline list endpoint returns `created_at` as a
+  protobuf-style `{seconds, nanos}` object. The describe endpoint
+  (`GET /pipelines/:pipeline_id`) uses a different custom string format
+  (`YYYY-MM-DD HH:MM:SS.ffffffZ`). This source only covers the list
+  endpoint.
 - **Case inconsistency**: The `state` and `result` fields may use different
   casing between API responses (e.g. `DONE` vs `done`, `FAILED` vs
   `failed`). Use `LOWER(result)` or `UPPER(state)` for reliable filtering.
--- a/sources/community/semaphore_ci/README.md
+++ b/sources/community/semaphore_ci/README.md
@@ -40,6 +40,8 @@ coral source add --file sources/community/semaphore_ci/manifest.yaml
 | `SEMAPHORE_ORG_SLUG` | Variable | Organization slug (subdomain) |
 
 Auth uses `Authorization: Token <TOKEN>` per the Semaphore v1alpha API spec.
+The documented `User-Agent: SemaphoreCI v2.0 Client` header is sent
+automatically via `request_headers`.
 
 ## Tables
 
@@ -78,8 +80,8 @@ tag, pull request, or manual rerun.
 ### pipelines
 
 Pipelines for a project or workflow. Each row is one pipeline execution
-within a workflow. Pass `project_id` or `wf_id` (or both) — the API
-requires at least one.
+within a workflow. Requires `project_id`; optionally pass `wf_id` to
+narrow to a single workflow.
 
 > **Important:** The pipeline list endpoint does **not** return `ppl_id`.
 > To get a pipeline ID for the promotions table, use `initial_ppl_id`
@@ -87,7 +89,7 @@ requires at least one.
 
 | Column | Type | Description |
 |---|---|---|
-| `project_id` | Utf8 | Project ID (echoed when provided) |
+| `project_id` | Utf8 | Project ID (required filter) |
 | `name` | Utf8 | Pipeline name from YAML config |
 | `yaml_file_name` | Utf8 | YAML file defining this pipeline |
 | `working_directory` | Utf8 | Pipeline working directory |
@@ -97,9 +99,9 @@ requires at least one.
 | `branch_name` | Utf8 | Git branch name |
 | `created_at` | Timestamp | Pipeline creation time (UTC) |
 
-**Required filter:** at least one of `project_id` or `wf_id`
-**Optional filters:** `branch_name`, `yml_file_path`, `created_after`,
-`created_before`, `done_after`, `done_before`
+**Required filter:** `project_id`
+**Optional filters:** `wf_id`, `branch_name`, `yml_file_path`,
+`created_after`, `created_before`, `done_after`, `done_before`
 **Pagination:** Link header
 
 > **Note:** Richer fields like `result_reason`, `error_description`,
@@ -159,7 +161,8 @@ WHERE project_id = 'your-project-uuid'
 -- Filter pipelines by workflow ID
 SELECT name, state, result, yaml_file_name
 FROM semaphore_ci.pipelines
-WHERE wf_id = 'your-workflow-uuid';
+WHERE project_id = 'your-project-uuid'
+  AND wf_id = 'your-workflow-uuid';
 
 -- Find pipelines by branch name
 SELECT name, state, result, created_at
@@ -201,7 +204,6 @@ WHERE pipeline_id = 'initial-ppl-id-from-workflows';
 - **Case inconsistency**: The `state` and `result` fields may use different
   casing between API responses (e.g. `DONE` vs `done`, `FAILED` vs
   `failed`). Use `LOWER(result)` or `UPPER(state)` for reliable filtering.
-- **Pipelines require at least one filter**: The Semaphore API requires
-  either `project_id` or `wf_id` (or both) on the pipelines endpoint.
-  Queries without either will fail at the API level.
+- **Pipelines require `project_id`**: The pipelines table requires
+  `project_id`. Optionally add `wf_id` to narrow to a specific workflow.
 - **API docs**: https://docs.semaphoreci.com/reference/api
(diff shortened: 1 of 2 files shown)


sources/community/semaphore_ci/manifest.yaml
```diff
@@ -0,0 +1,390 @@
...
+      via `sem get project <name>`. Use `wf_id` from here as a filter on
+      `semaphore_ci.pipelines` to drill into a specific workflow.
+      Optional filters: `branch_name` to scope by branch.
+      `created_at` is a UTC timestamp derived from the API's protobuf
+      `{seconds, nanos}` object.
+    filters:
+      - name: project_id
+        required: true
+      - name: branch_name
+        required: false
```
antonmry, line 85: The Semaphore v1alpha docs list `created_after` and `created_before` (Unix-timestamp query params) as optional filters on `GET /plumber-workflows`. Would you mind adding them here? They're the natural filters for "workflows in the last 24h" / "since last release" queries — without server-side pushdown, users have to paginate through every workflow ever to find recent ones, which gets painful on busy projects.

Ref: https://docs.semaphore.io/reference/api

sources/community/semaphore_ci/manifest.yaml
```diff
@@ -0,0 +1,390 @@
...
+      Use `ppl_id` from the list response (when present) or the pipeline
+      name to identify pipelines. `state` values: running, done, stopping.
+      `result` values: passed, failed, stopped, canceled.
+    filters:
+      - name: project_id
+        required: true
+      - name: wf_id
+        required: false
+      - name: branch_name
+        required: false
```
antonmry, line 259: Same suggestion as on `workflows`: the API documents `created_after`, `created_before`, `done_after`, `done_before`, and `yml_file_path` as optional filters here. Surfacing the time-window ones in particular would unlock CI-audit queries like "pipelines that ran since last release" or "pipelines that finished after deploy window X" without client-side filtering.

Ref: https://docs.semaphore.io/reference/api

sources/community/semaphore_ci/manifest.yaml
```diff
@@ -0,0 +1,390 @@
...
+      Requires `project_id`. Optionally pass `wf_id` (from
+      `semaphore_ci.workflows`) to scope to a single workflow's pipelines.
+      Use `ppl_id` from the list response (when present) or the pipeline
+      name to identify pipelines. `state` values: running, done, stopping.
+      `result` values: passed, failed, stopped, canceled.
+    filters:
+      - name: project_id
+        required: true
+      - name: wf_id
+        required: false
```
antonmry, line 257: Small contract mismatch: the Semaphore API allows either `project_id` OR `wf_id` alone on `GET /pipelines` ("at least one required" per the docs). The manifest declares `project_id` `required: true`, which means a user with a workflow UUID who writes `WHERE wf_id = 'X'` will be rejected by Coral even though the API would happily serve it.

Would you mind relaxing this — drop `required: true` and instead either (a) document the "must pass project_id or wf_id" rule in the table `guide`, or (b) use a manifest-level cross-filter validation if Coral supports one?

sources/community/semaphore_ci/manifest.yaml
```diff
@@ -0,0 +1,390 @@
...
+          Pipeline creation time (UTC). Extracted from the API's protobuf
+          {seconds, nanos} object — nanos are truncated.
+        expr:
+          kind: format_timestamp
+          input: seconds
+          expr:
+            kind: path
+            path:
+              - created_at
+              - seconds
```
antonmry, line 380: Question: the PR description mentions "5 timestamp columns" for pipelines, but only `created_at` is exposed here. The API actually returns several useful timestamps — `pending_at`, `queuing_at`, `running_at`, `done_at`, `stopping_at` — and surfacing some would unlock build-duration queries like `done_at - created_at` (one of the main reasons people query CI data). Was the trim intentional, or are these planned to land in this PR? Either way, worth aligning the manifest and the description.

sources/community/semaphore_ci/manifest.yaml
```diff
@@ -0,0 +1,390 @@
...
+    response: {}
+    pagination:
+      mode: link_header
+    columns:
+      - name: project_id
+        type: Utf8
+        nullable: false
+        description: Project ID (echoes the required filter).
+        expr:
+          kind: from_filter
```
antonmry, line 307: The PR description also lists `result_reason` and `error_description` as pipeline columns, but they're not in the manifest. These are exactly the fields users would project for "why did this pipeline fail" queries — would be nice to include. (If they're being deferred to a follow-up, worth removing them from the PR description so it matches what merges.)

sources/community/semaphore_ci/manifest.yaml
```diff
@@ -0,0 +1,390 @@
...
+      - name: triggered_by
+        type: Utf8
+        nullable: true
+        description: >-
+          How the workflow was triggered. May be an integer enum (1 = hook)
+          in list responses or a string (HOOK) in describe responses.
+        expr:
+          kind: path
+          path:
+            - triggered_by
```
antonmry, line 200: The column description here helpfully notes that `triggered_by` may come back as an integer in list responses and a string in describe responses — but the column type is `Utf8`. If the list endpoint really does return an integer (e.g. `1`), the JSON-path extractor into a `Utf8` column may either coerce silently to `"1"` (so `WHERE triggered_by = 'HOOK'` never matches) or skip the value depending on Coral's type handling.

Have you confirmed against a real list response which it is? If integers are the truth here, `Int64` would be safer. If both come back depending on endpoint, a normalization expression (e.g. cast to string) would make the column predictable.

sources/community/semaphore_ci/README.md
```diff
@@ -0,0 +1,177 @@
...
+SELECT wf_id, branch_name, commit_sha, created_at
+FROM semaphore_ci.workflows
+WHERE project_id = 'your-project-uuid'
+LIMIT 10;
+
+-- Find failed pipelines
+SELECT name, state, result, branch_name, created_at
+FROM semaphore_ci.pipelines
+WHERE project_id = 'your-project-uuid'
+  AND result = 'FAILED';
```
antonmry: Small inconsistency: this example uses `result = 'FAILED'` (uppercase), but the manifest's table guide says `result` values are lowercase (`passed`, `failed`, `stopped`, `canceled`). Your `## Notes` section further down honestly acknowledges Semaphore's case inconsistency, but the lead example will mislead users who scroll-and-copy.

Either pick the dominant casing (lowercase, per the manifest), or wrap defensively: `WHERE LOWER(result) = 'failed'`.

sources/community/semaphore_ci/manifest.yaml
```diff
@@ -0,0 +1,390 @@
...
+          path:
+            - status
+      - name: triggered_by
+        type: Utf8
+        nullable: true
+        description: What triggered the promotion (e.g. Pipeline Done request).
+        expr:
+          kind: path
+          path:
+            - triggered_by
```
antonmry, line 448: The `promotions` table is currently 4 columns, which feels thin for the typical use case ("when did we deploy to prod and was it auto-promoted?"). Worth checking whether the API returns:

- `created_at` (when the promotion was triggered)
- `scheduled_at` (for delayed promotions)
- `scheduled` / `auto_promoted` (boolean — manual vs automatic)

If Semaphore exposes these, surfacing them would make the deployment-auditing use case from the PR description ("CI/CD joins with GitHub pull requests") much more useful.

sources/community/semaphore_ci/manifest.yaml
```diff
@@ -0,0 +1,390 @@
+name: semaphore_ci
+version: 0.1.0
+dsl_version: 3
+backend: http
+description: >-
+  Query workflows, pipelines, and promotions from Semaphore CI.
+  Covers workflow runs, pipeline state and results, and promotion
+  status for a given project. Read-only; uses the v1alpha REST API.
```
antonmry: Nit on framing: a new user reading this will probably try `SELECT * FROM semaphore_ci.projects` first, find nothing, and only then discover the missing-list-projects-endpoint quirk further down in the README. Could we surface it here so the constraint is visible at the discovery layer?

Something like: "Query workflows, pipelines, and promotions for a Semaphore CI project. All tables require a `project_id` or `pipeline_id` filter — Semaphore's v1alpha API does not expose a list-projects endpoint, so obtain IDs from the Semaphore dashboard or `sem` CLI."

sources/community/semaphore_ci/manifest.yaml
```diff
@@ -0,0 +1,390 @@
...
+    description: >-
+      Pipelines for a Semaphore CI project. Each row is one pipeline
+      execution within a workflow. Requires `project_id` filter.
+      Optionally filter by `wf_id` to see pipelines for a single workflow.
+    guide: |
+      Requires `project_id`. Optionally pass `wf_id` (from
+      `semaphore_ci.workflows`) to scope to a single workflow's pipelines.
+      Use `ppl_id` from the list response (when present) or the pipeline
+      name to identify pipelines. `state` values: running, done, stopping.
+      `result` values: passed, failed, stopped, canceled.
```
antonmry, line 250: Small enum-completeness check: the guide lists `state` values as `running, done, stopping`, but Semaphore pipelines also pass through `pending` and `queuing` during transitions. Could you confirm and broaden the list, or note that other states may appear during transient periods?

sources/community/semaphore_ci/manifest.yaml
```diff
@@ -0,0 +1,390 @@
+name: semaphore_ci
+version: 0.1.0
+dsl_version: 3
+backend: http
+description: >-
+  Query workflows, pipelines, and promotions from Semaphore CI.
+  Covers workflow runs, pipeline state and results, and promotion
+  status for a given project. Read-only; uses the v1alpha REST API.
+base_url: "https://{{input.SEMAPHORE_ORG_SLUG}}.semaphoreci.com/api/v1alpha"
```
antonmry, line 11: Minor scope nit: `base_url` hard-codes `.semaphoreci.com`, which won't work for Semaphore Enterprise installations on a custom domain. Most other CI sources in this repo have the same limitation, so this is not a blocker — but worth either documenting "cloud-only" in the description, or adding a second variable input like `SEMAPHORE_BASE_URL` that defaults to the template for users on the cloud.

mohitjeswani01:
Thanks for the thorough review, @antonmry and @cubic-dev-ai — every comment addressed in this push:

Filters & server-side pushdown:

Workflows: added created_after, created_before optional filters
Pipelines: added created_after, created_before, done_after, done_before, yml_file_path — all wired as query params for server-side pushdown
Contract fix (pipelines project_id):

Relaxed project_id to required: false — users can now query with just wf_id. Guide + README document "pass at least one of project_id or wf_id". API enforces this server-side.
Pipeline columns (17 total now):

Added result_reason, error_description for failure diagnostics
Added 5 lifecycle timestamps: pending_at, queuing_at, running_at, stopping_at, done_at — enables done_at - created_at build-duration queries
Type fix:

triggered_by on workflows → Int64 (list endpoint returns integer enum, not string). Description documents the enum values.
State enum:

Broadened to pending, queuing, running, stopping, done per actual API lifecycle.
Casing:

README examples now use LOWER(result) = 'failed' to match the documented case-variance guidance.
Promotions:

Confirmed the v1alpha promotions list endpoint only returns name, status, triggered_by — no timestamps or auto-promotion metadata. Documented this limitation in the guide.
Discovery & scope:

Description now surfaces the no-projects-endpoint constraint and cloud-only limitation upfront.
README notes "Semaphore Enterprise users on a custom domain would need to fork and adjust."

mohitjeswani01:
@antonmry could you re check?

antonmry:
LGTM but let's wait for the codeownner review (@jsummerfield )

jsummerfield (changes requested):
Findings:

- High: live Coral evidence is still missing. The PR body screenshot shows YAML/source-shape checks and `make lint-sources`, but not a successful `coral source add`, `coral source test`, `coral query`, a screenshot/video of Coral query results, or a linked log from this source. Please add sanitized live Coral output from a real Semaphore account before this can be approved.

- High: `sources/community/semaphore_ci/manifest.yaml:86`, `:257`, and `:473` do not set Semaphore's documented required `User-Agent`. Semaphore's v1alpha API docs say all API requests must set `User-Agent` to `SemaphoreCI v2.0 Client`, while Coral's default client user agent is generic. Please add a request header such as `User-Agent: SemaphoreCI v2.0 Client` to each table request. Docs: https://docs.semaphore.io/reference/api

- Medium: `sources/community/semaphore_ci/manifest.yaml:63` and `:226` expose paginated workflow and pipeline list tables without a conservative `fetch_limit_default`. Semaphore documents link-header pagination for list responses over 30 items, so a project-wide query can walk a long CI history. Please add bounded defaults, for example `fetch_limit_default: 100`, while keeping the existing time/window filters for narrower queries. Docs: https://docs.semaphore.io/reference/api

Thanks for the careful iteration here; the table scope is much clearer now. The next useful step is to add real Coral output and make the provider-specific request headers/default query bounds explicit.

jsummerfield (changes requested):
Additional API-contract findings on the same head:

- High: `sources/community/semaphore_ci/manifest.yaml:354` and `:392-450` expose `result_reason`, `error_description`, `pending_at`, `queuing_at`, `running_at`, `stopping_at`, and `done_at` from `GET /pipelines`, but Semaphore's documented list response only includes the basic pipeline fields plus `created_at`. The richer fields appear on `GET /pipelines/:pipeline_id`, and those timestamps are documented as custom timestamp strings rather than `{seconds, nanos}` objects. Please remove these columns from the list table, or add a separate pipeline-detail table/function backed by `/pipelines/:pipeline_id` with the correct timestamp parsing. Docs: https://docs.semaphore.io/reference/api#list-pipelines and https://docs.semaphore.io/reference/api#describe-a-pipeline

- High: `sources/community/semaphore_ci/manifest.yaml:296` exposes `ppl_id` from the pipeline list response, and the README uses the pipeline table as the path into promotions, but the documented `GET /pipelines` list response does not include `ppl_id` while `promotions` requires `pipeline_id`. Please either prove with live Coral output that `ppl_id` is returned by the list endpoint, or document the reliable path through `workflows.initial_ppl_id`, a manually supplied pipeline ID, or a detail endpoint that exposes the ID. Docs: https://docs.semaphore.io/reference/api#list-pipelines and https://docs.semaphore.io/reference/api#list-promotions

These are in addition to the live Coral evidence, User-Agent, and bounded-default points from my earlier review. Thanks for continuing to work through the provider-specific details here.

mohitjeswani01:
Hi @jsummerfield i will work on this as soon as I reach my hometown within a day I am currently travelling so will be unavailable for a day please dont close the pr sir 🙌

mohitjeswani01:
Thanks @jsummerfield — both findings confirmed against the [official API docs](https://docs.semaphoreci.com/reference/api). Fixed in the latest push.

Finding 1 (High): Describe-only columns on pipeline list table
Verified the [documented list response](https://docs.semaphoreci.com/reference/api#list-pipelines) — it only returns yaml_file_name, working_directory, wf_id, state, result, name, branch_name, and created_at ({seconds, nanos}).

Removed 8 columns that only appear on GET /pipelines/:pipeline_id:

ppl_id, result_reason, error_description
pending_at, queuing_at, running_at, stopping_at, done_at
Also confirmed the describe endpoint uses custom timestamp strings (YYYY-MM-DD HH:MM:SS.ffffffZ), not {seconds, nanos} — so format_timestamp with input: seconds would have been incorrect anyway. Documented this difference in the README Notes section.

Finding 2 (High): ppl_id not in list response
Confirmed — ppl_id is describe-only. Removed the column and updated all guides, descriptions, test queries, and README to document initial_ppl_id from semaphore_ci.workflows as the reliable path into promotions.

jsummerfield (changes requested):
Findings:

- Medium: `sources/community/semaphore_ci/manifest.yaml:27` configures `Authorization`, but the source does not send Semaphore's documented `User-Agent`. Semaphore's API reference says API requests should include `User-Agent: SemaphoreCI v2.0 Client`; Coral's default client header is not that value. Please add a `request_headers` entry for the documented User-Agent. Docs: https://docs.semaphore.io/reference/api

- Medium: `sources/community/semaphore_ci/manifest.yaml:244` makes all `pipelines` filters optional even though the API requires at least one of `project_id` or `wf_id`; the README notes that missing both fails at the API level at `sources/community/semaphore_ci/README.md:204`. Coral's catalog will still advertise this as a table with no required filters, which is a poor first-query experience for a CI history source. Please model a safe first path, such as requiring `project_id` for this table, or move the one-of access pattern into a table-function shape if that is the cleaner fit.

- Low: the current live evidence is still hard to verify. A screenshot is better than nothing, but for this source please include copyable output showing successful reads for `workflows`, `pipelines`, and `promotions`, plus at least one timestamp column rendered through Coral so the protobuf `{seconds, nanos}` mapping is exercised.

Thanks for the careful follow-up. The auth scheme, cloud base URL, endpoint choices, and timestamp approach are otherwise close.

mohitjeswani01:
Thanks @jsummerfield — all 3 findings addressed.

Finding 1 (Medium): User-Agent header ✅
Added request_headers with User-Agent: SemaphoreCI v2.0 Client per [the docs](https://docs.semaphoreci.com/reference/api#overview). Same pattern as wandb, statusgator, launchdarkly in core sources.

Finding 2 (Medium): Pipelines safe first path ✅
Made project_id required on pipelines. Users always have project_id since workflows (the natural entry point) already requires it. wf_id is now an optional narrowing filter.

Finding 3 (Low): Live Coral evidence ✅
jeswa@MSI:/mnt/d/coral$ coral sql "SELECT wf_id, branch_name, commit_sha, initial_ppl_id, created_at FROM semaphore_ci.workflows WHERE project_id = 'd4302cbf-b47c-408d-8d74-a3138d113fac' LIMIT 3"
+--------------------------------------+-----------------+------------------------------------------+--------------------------------------+----------------------+
| wf_id                                | branch_name     | commit_sha                               | initial_ppl_id                       | created_at           |
+--------------------------------------+-----------------+------------------------------------------+--------------------------------------+----------------------+
| a850bd2e-299a-489e-90a9-b7a19e206794 | setup-semaphore | 67d8e4ac1a822ceca60e92fd1941f4f2460d8e3d | c756afa5-1af0-424b-9b39-911d1f88c613 | 2026-05-26T19:18:11Z |
+--------------------------------------+-----------------+------------------------------------------+--------------------------------------+----------------------+

jeswa@MSI:/mnt/d/coral$ coral sql "SELECT name, state, result, branch_name, created_at FROM semaphore_ci.pipelines WHERE project_id = 'd4302cbf-b47c-408d-8d74-a3138d113fac' LIMIT 3"
+----------------------------+-------+--------+-----------------+----------------------+
| name                       | state | result | branch_name     | created_at           |
+----------------------------+-------+--------+-----------------+----------------------+
| 🌟 Fan-In/Fan-Out Pipeline | DONE  | FAILED | setup-semaphore | 2026-05-26T19:18:11Z |
+----------------------------+-------+--------+-----------------+----------------------+

created_at renders correctly as 2026-05-26T19:18:11Z — confirming the protobuf {seconds, nanos} → Timestamp mapping works end-to-end through Coral.

Promotions: The test project's pipeline has no promotions configured (still in onboarding), and Semaphore returns a 500 for pipelines without promotions. The endpoint is correctly wired — will exercise with a fully configured project if needed.

image

jsummerfield (changes requested):
Findings:

- Medium: `sources/community/semaphore_ci/manifest.yaml:250` makes `project_id` required for `semaphore_ci.pipelines`, but Semaphore documents the list-pipelines contract as `project_id` required unless `wf_id` is present, and `wf_id` required unless `project_id` is present. The current source now only supports project-scoped pipeline queries, which narrows a valid workflow-scoped API path. Please either model the one-of contract, or keep the project-scoped source behavior but make the table description and README explicit that requiring `project_id` is a Coral source restriction rather than Semaphore's API requirement. Docs: https://docs.semaphoreci.com/reference/api#list-pipelines

- Low: the latest live evidence covers `workflows` and `pipelines`, including timestamp rendering, but `promotions` is still not exercised. Since this source exposes promotions as a first-version table, please either add sanitized output from a pipeline with configured promotions, or document the observed no-promotions behavior as a known limitation so users are not surprised by the provider error.

The User-Agent header and the previous pipeline list-column issues are resolved. Thanks for tightening the source; this is close.

mohitjeswani01:
Thanks @jsummerfield — both addressed.

Finding 1 (Medium): project_id restriction clarity ✅
Updated description, guide, and README across all three locations (manifest description, manifest guide, README) to explicitly state:

"This Coral source requires project_id as a Coral-side restriction for a safe first-query experience. Semaphore's API contract accepts either project_id or wf_id (or both); this source always requires project_id and treats wf_id as an optional narrowing filter."

This is now explicit in:

Manifest description block
Manifest guide block
README pipelines section + Notes section
Finding 2 (Low): Promotions known limitation ✅
Documented in both manifest guide and README:

"Semaphore returns a server error (500) for pipelines with no promotions configured rather than an empty array. This is an upstream API behavior, not a source bug."

Added to:

Manifest promotions guide block
README promotions Note callout
README Notes section as a dedicated bullet

jsummerfield (approved):

The previous Semaphore findings are resolved: the source now documents `project_id` as an intentional Coral-side restriction even though Semaphore accepts `project_id` or `wf_id`, and it documents the upstream no-promotions 500 behavior. User-Agent and pipeline list shape are also addressed, CI is green, and live evidence covers workflows, pipelines, and timestamp rendering.

Thanks for working through the provider-specific edge cases.

PR #407, opened 2026-05-13
feat(sources/core/gmail): add Gmail source

sauldhernandez:
Intent

Add Gmail as a bundled Coral HTTP source so users can query Gmail mailbox metadata and read-only message resources with OAuth credentials.

What it enables

Users can connect Gmail with a Google OAuth client ID and secret or paste a read-only Gmail access token, then query profile metadata, labels, messages, threads, drafts, history, filters, send-as aliases, and message attachments through Coral SQL.

Usage examples

coral source add --interactive gmail

SELECT * FROM gmail.profile LIMIT 1;

SELECT id, thread_id FROM gmail.messages WHERE q LIKE 'from:alerts@example.com newer_than:7d' LIMIT 20;

SELECT subject, "from", internal_date FROM gmail.message_details WHERE id = '<message-id>';

PR #675, opened 2026-05-22
feat(sources/community/gmail): add Gmail community source

rohan-jadhav-dev:
## Summary
Adds a new community source for Gmail using the Gmail REST API v1.

## Tables
- `labels` — list all Gmail labels (INBOX, SENT, DRAFT, etc.)
- `messages` — list messages, filterable by label_id
- `message_detail` — full metadata for a single message
- `threads` — list conversation threads, filterable by label_id

## Authentication
OAuth 2.0 Bearer Token with `gmail.readonly` scope.

## Example Queries
```sql
SELECT * FROM gmail.labels WHERE user_id = 'me';
SELECT * FROM gmail.messages WHERE user_id = 'me' AND label_id = 'INBOX';
```

## Notes
- Read-only source
- Follows existing community source patterns

### DIFF
--- a/sources/community/gmail/README.md
+++ b/sources/community/gmail/README.md
@@ -1,0 +1,116 @@
+# Gmail
+
+Query Gmail messages, threads, labels, and drafts using SQL via the [Gmail REST API v1](https://developers.google.com/gmail/api/reference/rest).
+
+## Authentication
+
+The source uses OAuth 2.0 Bearer Token authentication. `GMAIL_ACCESS_TOKEN` is required at install time.
+
+To generate an access token:
+
+1. Go to [Google Cloud Console](https://console.cloud.google.com/apis/credentials).
+2. Create a project and enable the **Gmail API**.
+3. Create an **OAuth 2.0 Client ID** credential (Desktop or Web app).
+4. Authorize the scope `https://www.googleapis.com/auth/gmail.readonly`.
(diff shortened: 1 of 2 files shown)


jsummerfield:
Thanks for adding this. I found several blockers before this should land:

1. The manifest does not pass `coral source lint`: `auth.type: BearerToken` is not valid for the current source spec. Please use `HeaderAuth` with an `Authorization: Bearer {{input.GMAIL_ACCESS_TOKEN}}` header.

2. This conflicts with open PR #407, which also adds a `gmail` source (`sources/core/gmail/manifest.yaml`). Please coordinate one canonical Gmail source rather than landing competing `gmail` schemas.

3. `messages` and `threads` disable pagination, but Gmail list APIs return only a bounded page by default and paginate with `nextPageToken`. This will silently truncate customer mailboxes. Please add cursor pagination using `pageToken` / `nextPageToken` and `maxResults`.

4. The README documents `q` search filters, but the manifest does not define or send `q`, so those examples will fail. Please add `q` filters for `messages` and `threads`, and model `labelIds`/`q` as structured query params rather than hard-coding optional filters into the path.

5. The README selects `label_ids` from `message_detail`, but that column is missing from the manifest even though Gmail `Message` includes `labelIds[]`. Please add it or remove it from the examples.

6. The auth setup currently relies on a raw access token that expires quickly. For Gmail, please either use the source OAuth credential flow with offline access/refresh support or clearly scope this as a short-lived manual-token setup.

PR #681, opened 2026-05-22
feat(sources/n8n): add n8n community source spec

### LINKED ISSUE #680: New Source Spec: n8n (Workflow Automation Platform)
## 🏴‍☠️ Overview
I'm proposing a new community source spec for [n8n](https://n8n.io) — the leading extensible workflow automation platform. 
This source allows agents and developers to query their automation platform like a database, enabling powerful cross-source JOINs (e.g., correlating failed n8n workflows with PagerDuty incidents or GitHub deployments).

Full documentation and demo queries are available here: [Coral Hackathon: n8n Source Spec Documentation](https://www.notion.so/368c6925c3bb81e98983f10609a4b128)

## 🗄️ Tables Implemented
- `workflows` — Automation workflows with tags, node count, and active status.
- `executions` — Execution runs with mode, finished status, and timing.
- `tags` — Workflow categorization tags.
- `variables` — Instance variables (Note: requires paid n8n license).

## 🛠️ Functions
- `executions_by_workflow(workflow_id, status)` — Native search function for scoped execution lookups.

## 🚀 Advanced Features Showcased
- **Cursor Pagination:** Implemented `cursor_query` with `nextCursor` response path.
- **API Key Auth:** Configured via `X-N8N-API-KEY` header.
- **Configurable Base URL:** Supports both self-hosted (`http://localhost:5678/api/v1`) and n8n Cloud (`https://<account>.n8n.cloud/api/v1`).
- **Timestamp Formatting:** Converts `iso8601` to Coral `Timestamp`.
- **Nested Field Flattening:** Extracted nested project IDs (`homeProject__id`).
- **Array Joining:** Used `join_array_path` to flatten workflow tags into a comma-separated string.

I have built this for the Pirates of the Coral-bean Hackathon and will be opening a PR shortly!

wiz-abhi:
## Summary
Adds a community source spec for [n8n](https://n8n.io) — the workflow automation platform.

Closes #680

## Tables
- `workflows` — automation workflows with name, description, active/archived status, trigger count, tags, project, settings, and timestamps
- `executions` — execution runs with status, mode, finished flag, timing, and retry metadata
- `tags` — workflow tags with timestamps
- `variables` — instance variables (requires paid n8n license)

## Key design decisions
- **Read-only.** `includeData=false` on executions to keep payloads small.
- **`Utf8` for execution IDs.** The n8n OpenAPI schema declares execution IDs as `type: number`, but we use `Utf8` so `executions.workflow_id` can join cleanly against `workflows.id` (which the API returns as a string). Coral coerces JSON numbers to strings when the column type is `Utf8`.
- **No `executions_by_workflow` function.** Removed per reviewer feedback — `SELECT ... FROM n8n.executions WHERE workflow_id = 'X'` provides the same functionality via the table + filter.
- **`N8N_BASE_URL` is `required: true`** with no default, so users must supply their instance URL explicitly.
- **`N8N_API_KEY` is `kind: secret`** per repo rules for credential inputs.

## Filters
| Table | Filters |
|---|---|
| `workflows` | `active`, `tags`, `name`, `project_id` |
| `executions` | `status`, `workflow_id`, `project_id` |
| `variables` | `project_id` |

## Testing
```bash
export N8N_BASE_URL="http://localhost:5678/api/v1"
export N8N_API_KEY="..."
coral source lint sources/community/n8n/manifest.yaml
coral source add --file sources/community/n8n/manifest.yaml
coral source test n8n
```

## Review feedback addressed
- [x] Surfaced `status` column on executions (was filter-only)
- [x] Replaced broken `node_count` with `trigger_count` (`triggerCount` API field)
- [x] Added `description`, `is_archived` columns to workflows
- [x] Removed `N8N_BASE_URL` placeholder default (angle brackets → DNS failure)
- [x] Expanded `mode` description to all 10 enum values
- [x] Documented valid `status` enum values in table description and guide
- [x] Removed redundant `executions_by_workflow` function
- [x] Fixed README tag example to use server-side `tags` filter
- [x] Reworded tags table description (substring filter, not JOIN)
- [x] Removed unnecessary `Accept: application/json` header
- [x] Added `guide` fields to workflows and executions tables
- [x] Kept execution IDs as `Utf8` to avoid cross-type JOIN mismatch with `workflows.id`

### DIFF
--- a/sources/community/n8n/README.md
+++ b/sources/community/n8n/README.md
@@ -1,0 +1,239 @@
+# n8n Community Source
+
+Query n8n workflows, executions, credentials, tags, and variables through Coral
+SQL using the [n8n Public REST API](https://docs.n8n.io/api/).
+
+## Setup
+
+### 1. Create an n8n API key
+
+In your n8n instance:
+
+1. Go to **Settings** > **API**
+2. Click **Generate API Key**
+3. Copy the key
+
+For n8n Cloud, your API base URL is `https://<account>.n8n.cloud/api/v1`.
+For self-hosted, it is `http://localhost:5678/api/v1` (or your custom domain).
+
+> **Note:** The n8n API is not available during the free trial. You need a
+> paid or self-hosted instance.
+
+### 2. Add the source
+
+```bash
+export N8N_BASE_URL="https://<your-instance>.n8n.cloud/api/v1"
+export N8N_API_KEY="<your-api-key>"
+coral source add --file sources/community/n8n/manifest.yaml
+```
+
+Or use interactive mode:
+
+```bash
+coral source add --interactive --file sources/community/n8n/manifest.yaml
+```
+
+### 3. Verify
+
+```bash
+coral source test n8n
+```
+
+## Tables
+
+### `n8n.workflows`
+
+List automation workflows.
+
+| Column | Type | Description |
+|---|---|---|
+| `id` | Utf8 | Workflow ID |
+| `name` | Utf8 | Workflow name |
+| `active` | Boolean | Whether the workflow is published/active |
+| `version_id` | Utf8 | Current version ID |
+| `node_count` | Int64 | Number of nodes in the workflow |
+| `tag_names` | Utf8 | Comma-separated tag names |
+| `project_id` | Utf8 | Owning project ID |
+| `project_name` | Utf8 | Owning project name |
+| `created_at` | Timestamp | Creation time |
+| `updated_at` | Timestamp | Last update time |
+| `settings` | Json | Workflow settings object |
+
+**Optional filters:** `active`, `tags`, `name`, `project_id`
+
+### `n8n.executions`
+
+List workflow execution runs.
+
+| Column | Type | Description |
+|---|---|---|
+| `id` | Utf8 | Execution ID |
+| `finished` | Boolean | Whether the execution finished |
+| `mode` | Utf8 | Execution mode (trigger, manual, integrated) |
+| `status` | Utf8 | Status (success, error, running, waiting, canceled, crashed, new, unknown) |
+| `workflow_id` | Utf8 | Executed workflow ID |
+| `workflow_name` | Utf8 | Workflow name at runtime |
+| `started_at` | Timestamp | Start time |
+| `stopped_at` | Timestamp | Stop time (NULL if running) |
+| `last_node_executed` | Utf8 | Last executed node name |
+| `execution_error_message` | Utf8 | Error message if failed |
+| `execution_error_node` | Utf8 | Node that caused the error |
+| `retry_of` | Utf8 | Original execution ID if this is a retry |
+| `retry_success_id` | Utf8 | Successful retry execution ID |
+
+**Optional filters:** `status`, `workflow_id`, `project_id`
+
+### `n8n.credentials`
+
+List credential definitions (secrets are never exposed).
+
+| Column | Type | Description |
+|---|---|---|
+| `id` | Utf8 | Credential ID |
+| `name` | Utf8 | Credential display name |
+| `type` | Utf8 | Credential type (e.g. slackApi, githubApi) |
+| `is_managed` | Boolean | Whether managed (enterprise) |
+| `scoped` | Boolean | Whether project-scoped |
+| `created_at` | Timestamp | Creation time |
+| `updated_at` | Timestamp | Last update time |
+
+### `n8n.tags`
+
+List workflow tags.
+
+| Column | Type | Description |
+|---|---|---|
+| `id` | Utf8 | Tag ID |
+| `name` | Utf8 | Tag name |
+| `created_at` | Timestamp | Creation time |
+| `updated_at` | Timestamp | Last update time |
+
+### `n8n.variables`
+
+List instance variables.
+
+| Column | Type | Description |
+|---|---|---|
+| `id` | Utf8 | Variable ID |
+| `key` | Utf8 | Variable key/name |
+| `value` | Utf8 | Variable value |
+| `type` | Utf8 | Data type (string, number, boolean) |
+
+**Optional filter:** `project_id`
+
+## Functions
+
+### `n8n.executions_by_workflow`
+
+Search executions for a specific workflow.
+
+```sql
+SELECT id, status, started_at
+FROM n8n.executions_by_workflow(
+  workflow_id => 'abc123',
+  status => 'error'
+)
+LIMIT 10;
+```
+
+## Example queries
+
+```sql
+-- List active workflows with their tags
+SELECT name, tag_names, node_count, updated_at
+FROM n8n.workflows
+WHERE active = true
+ORDER BY updated_at DESC;
+
+-- Find failed executions in the last 24 hours
+SELECT id, workflow_name, status, started_at, execution_error_message
+FROM n8n.executions
+WHERE status = 'error'
+  AND started_at > NOW() - INTERVAL '24 hours'
+ORDER BY started_at DESC;
+
+-- Join executions to workflows for full context
+SELECT
+  w.name AS workflow_name,
+  w.active AS workflow_active,
+  e.status,
+  e.started_at,
+  e.execution_error_message
+FROM n8n.workflows w
+JOIN n8n.executions e ON w.id = e.workflow_id
+WHERE e.status = 'error'
+  AND e.started_at > NOW() - INTERVAL '7 days'
+ORDER BY e.started_at DESC
+LIMIT 20;
+
+-- Count executions by status per workflow
+SELECT
+  workflow_name,
+  status,
+  COUNT(*) AS execution_count
+FROM n8n.executions
+WHERE started_at > NOW() - INTERVAL '30 days'
+GROUP BY workflow_name, status
+ORDER BY execution_count DESC;
+
+-- Audit credentials by type
+SELECT type, COUNT(*) AS count
+FROM n8n.credentials
+GROUP BY type
+ORDER BY count DESC;
+
+-- Find workflows using a specific tag
+SELECT name, tag_names, active
+FROM n8n.workflows
+WHERE tag_names ILIKE '%production%';
+
+-- Cross-source join: correlate n8n failures with PagerDuty incidents
+SELECT
+  e.workflow_name,
+  e.started_at,
+  e.execution_error_message,
+  p.title AS incident_title,
+  p.status AS incident_status
+FROM n8n.executions e
+LEFT JOIN pagerduty.incidents p
+  ON p.created_at BETWEEN e.started_at AND e.started_at + INTERVAL '5 minutes'
+WHERE e.status = 'error'
+  AND e.started_at > NOW() - INTERVAL '24 hours';
+```
+
+## Validation
+
+```bash
+export N8N_BASE_URL="https://<your-instance>.n8n.cloud/api/v1"
+export N8N_API_KEY="<your-api-key>"
+coral source lint sources/community/n8n/manifest.yaml
+coral source add --file sources/community/n8n/manifest.yaml
+coral source test n8n
+coral sql "SELECT * FROM coral.tables WHERE schema_name = 'n8n'"
+coral sql "SELECT * FROM coral.columns WHERE schema_name = 'n8n'"
+coral sql "SELECT id, name, active FROM n8n.workflows LIMIT 5"
+```
+
+## Limitations
+
+- **Read-only.** This source does not create, update, activate, deactivate,
+  or delete any n8n resources.
+- **No execution data.** The `includeData` parameter is always set to `false`
+  to avoid extremely large responses. Execution metadata (status, timing,
+  errors) is available.
+- **API availability.** The n8n Public API requires a paid or self-hosted
+  instance. It is not available during the n8n Cloud free trial.
+- **Credential secrets.** The n8n API never returns credential secrets;
+  only metadata (name, type, dates) is exposed.
+
+## Out of scope for v1
+
+- Workflow create/update/delete operations
+- Execution retry or stop operations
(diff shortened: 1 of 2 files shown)


sources/community/n8n/manifest.yaml
```diff
@@ -0,0 +1,550 @@
+name: n8n
```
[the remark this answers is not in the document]
wiz-abhi: Good catch! I had initially planned to include error extraction, but discovered the n8n list API doesn't return error details without `includeData=true` (which would make the payload too large). I removed the columns from the manifest but forgot to update the PR description. I've just updated the PR description to remove the claim about error extraction. Thanks!

sources/community/n8n/manifest.yaml
```diff
@@ -0,0 +1,550 @@
...
+  # ------------------------------------------------------------------
+  # n8n.executions — list workflow execution runs
+  # GET /executions
+  # ------------------------------------------------------------------
+  - name: executions
+    description: |
+      List n8n workflow execution runs. Filter by `status`, `workflow_id`,
+      or `project_id`. The `status` filter is accepted by the API but the
+      status value is not included in list responses; use `finished` to
+      determine completion state. Join `workflow_id` to `n8n.workflows.id`.
```
antonmry: This claim doesn't match what I see in the n8n OpenAPI spec. `status` is documented as a top-level string field on execution objects, with enum values `canceled, crashed, error, new, running, success, unknown, waiting` — it's not just a filter, it's in the response.

Ref: https://github.com/n8n-io/n8n/blob/master/packages/cli/src/public-api/v1/handlers/executions/spec/schemas/execution.yml

Could you surface it as a column? Without it, users can only tell "finished or not," not "succeeded vs failed" — which is the headline use case the PR describes ("error extraction"). `finished = false AND status = 'error'` vs `finished = false AND status = 'running'` is a meaningful distinction.

sources/community/n8n/manifest.yaml
```diff
@@ -0,0 +1,550 @@
...
+          path:
+            - versionId
+      - name: node_count
+        type: Int64
+        nullable: true
+        description: Number of nodes in the workflow.
+        expr:
+          kind: path
+          path:
+            - nodes
```
antonmry: I think this column is broken: `nodes` in the workflow response is an **array** of node objects (per the OpenAPI schema), not a scalar number. A plain `kind: path` extractor will read the array itself, and putting that into an `Int64` column will either fail or yield nulls.

Two ways to fix it:
1. Use Coral's array-length expression kind (if one exists — `json_array_length` or similar). Worth checking the DSL reference.
2. Use the API's documented `triggerCount` field directly (it's a read-only integer that counts active trigger nodes — closer to what users probably mean by "how busy is this workflow"). Schema ref: https://github.com/n8n-io/n8n/blob/master/packages/cli/src/public-api/v1/handlers/workflows/spec/schemas/workflow.yml

Did this column actually return integers in your validation queries?

sources/community/n8n/manifest.yaml
```diff
@@ -0,0 +1,550 @@
...
+  Coral SQL using the n8n Public REST API.
+
+inputs:
+  N8N_BASE_URL:
+    kind: variable
+    default: https://<your-n8n-instance>.n8n.cloud/api/v1
+    hint: >-
+      Base URL for your n8n instance API. For n8n Cloud use
+      https://<account>.n8n.cloud/api/v1. For self-hosted use
+      http://localhost:5678/api/v1 (or your custom domain).
```
antonmry, line 17: Small footgun: the default `https://<your-n8n-instance>.n8n.cloud/api/v1` contains literal angle brackets that won't resolve. A user running `coral source add` without first exporting `N8N_BASE_URL` will hit a DNS failure on the placeholder.

A few options:
1. Drop the default and mark `required: true` so Coral prompts for the value.
2. Default to `http://localhost:5678/api/v1` (the README mentions self-hosted is common, and at least it resolves).
3. Keep the default but add a manifest validation that rejects URLs containing `<` / `>`.

(1) is probably cleanest.

sources/community/n8n/manifest.yaml
```diff
@@ -0,0 +1,550 @@
...
+          path:
+            - finished
+      - name: mode
+        type: Utf8
+        nullable: true
+        description: Execution mode (trigger, manual, integrated).
+        expr:
+          kind: path
+          path:
+            - mode
```
antonmry, line 284: The mode enum is broader than this description suggests. The n8n OpenAPI lists 10 values: `cli, error, integrated, internal, manual, retry, trigger, webhook, evaluation, chat`. Users in production will mostly see `webhook` and `trigger`, neither of which is mentioned here. Would you mind expanding the description (and the matching note in README:72) so users aren't surprised by values that aren't listed?

sources/community/n8n/manifest.yaml
```diff
@@ -0,0 +1,550 @@
...
+        query_param: limit
+    columns:
+      - name: id
+        type: Utf8
+        nullable: false
+        description: Execution ID.
+        expr:
+          kind: path
+          path:
+            - id
```
antonmry, line 266: Heads-up on typing: per the OpenAPI schema, execution `id`, `workflowId`, `retryOf`, and `retrySuccessId` are all `type: number`. The manifest declares them all as `Utf8` here and in the columns below. Workflow `id` is correctly `Utf8` (the workflow schema explicitly says `type: string`) — only the execution fields differ.

Depending on how Coral coerces JSON numbers into `Utf8`, you may get the stringified number (and joins like `WHERE workflow_id = '1000'` work) or you may get empty/null. Worth either matching the API's `Int64` type or verifying the implicit coercion behaves predictably. Same applies to lines 257-264, 298-313, and 510-517 in the function.

sources/community/n8n/manifest.yaml
```diff
@@ -0,0 +1,550 @@
...
+            path:
+              - updatedAt
+      - name: settings
+        type: Json
+        nullable: true
+        description: Workflow settings object (timezone, error workflow, etc.).
+        expr:
+          kind: path
+          path:
+            - settings
```
antonmry, line 207: The workflow schema documents three more fields that would be valuable additions:
- `description` — Utf8, enables "find workflows that mention X"
- `isArchived` — Boolean, the obvious filter for hiding archived workflows
- `triggerCount` — Int64, read-only "number of active trigger nodes" (also a working replacement for the broken `node_count` column above)

All three are plain `kind: path` extractions. Mind adding them?

Ref: https://github.com/n8n-io/n8n/blob/master/packages/cli/src/public-api/v1/handlers/workflows/spec/schemas/workflow.yml

sources/community/n8n/manifest.yaml
```diff
@@ -0,0 +1,550 @@
...
+  # ------------------------------------------------------------------
+  - name: executions
+    description: |
+      List n8n workflow execution runs. Filter by `status`, `workflow_id`,
+      or `project_id`. The `status` filter is accepted by the API but the
+      status value is not included in list responses; use `finished` to
+      determine completion state. Join `workflow_id` to `n8n.workflows.id`.
+    filters:
+      - name: status
+      - name: workflow_id
```
antonmry, line 227: The `status` filter is accepted by the API but the table description doesn't list the valid values. A user trying `WHERE status = 'failed'` (a natural guess) will silently get empty results because the actual value is `error`.

Could the table `guide` enumerate the values? The full set is `canceled, crashed, error, new, running, success, unknown, waiting`.

sources/community/n8n/manifest.yaml
```diff
@@ -0,0 +1,550 @@
...
+      row_strategy: direct
+    pagination:
+      mode: cursor_query
+      cursor_param: cursor
+      response_cursor_path:
+        - nextCursor
+      page_size:
+        default: 50
+        max: 250
+        query_param: limit
```
antonmry: Is this function meaningfully different from `SELECT … FROM n8n.executions WHERE workflow_id = 'X'`? Looking through it, it accepts the same `workflow_id` and `status` args, hits the same endpoint with the same query params, and returns a subset of the same columns (minus `retry_of` / `retry_success_id`).

Functions add an extra calling syntax for users to learn and a second column list for the maintainer to keep in sync. Unless there's a Coral feature this unlocks (MCP tool exposure? something specific to functions?), it might be cleaner to drop it and rely on the table + filter. Happy to be told otherwise — just flagging the apparent duplication.

sources/community/n8n/README.md
```diff
@@ -0,0 +1,220 @@
...
+FROM n8n.workflows w
+JOIN n8n.executions e ON w.id = e.workflow_id
+WHERE e.started_at > NOW() - INTERVAL '30 days'
+GROUP BY w.name, e.finished
+ORDER BY execution_count DESC;
+
+-- Find workflows using a specific tag
+SELECT name, tag_names, active
+FROM n8n.workflows
+WHERE tag_names ILIKE '%production%';
```
antonmry: This example does a client-side substring match on `tag_names`, but the API actually supports server-side tag filtering — and the manifest already wires it through (the `tags` filter on `n8n.workflows`). Rewriting as `WHERE tags = 'production'` would avoid fetching every workflow on the instance just to filter on the client.

(Note: `tags` and `tag_names` are different — `tags` is the request filter, `tag_names` is the response column. That distinction is worth a one-liner somewhere in the README.)

sources/community/n8n/manifest.yaml
```diff
@@ -0,0 +1,550 @@
...
+            - retrySuccessId
+
+  # ------------------------------------------------------------------
+  # n8n.tags — list workflow tags
+  # GET /tags
+  # ------------------------------------------------------------------
+  - name: tags
+    description: |
+      List n8n workflow tags. Use to categorize and filter workflows.
+      Join `name` to `n8n.workflows.tag_names` for workflow categorization.
```
antonmry: Small wording nit: this says "Join `name` to `n8n.workflows.tag_names`" — but `tag_names` is built as a comma-separated string via `join_array_path` (lines 129-139), so a SQL `JOIN ON tags.name = workflows.tag_names` won't equality-match anything except workflows with exactly one tag.

The README example does the right thing (`tag_names ILIKE '%name%'`), but the manifest's wording will mislead someone expecting a clean equality join. Could you reword to "match against `tag_names` using a substring filter" or similar?

sources/community/n8n/manifest.yaml
```diff
@@ -0,0 +1,550 @@
...
+  type: HeaderAuth
+  headers:
+    - name: X-N8N-API-KEY
+      from: template
+      template: "{{input.N8N_API_KEY}}"
+
+request_headers:
+  - name: Accept
+    from: literal
+    value: application/json
```
antonmry: Tiny nit (safe to ignore): `Accept: application/json` is n8n's default response format, so this header doesn't change behavior. Removing the three lines would just trim the manifest a little.

wiz-abhi:
@antonmry Thanks for the thorough review! I've pushed a commit that addresses all your feedback:

*   **Fields & Columns**: Replaced the broken `node_count` with `trigger_count`, added the missing workflow columns (`description`, `is_archived`), and surfaced `status` as a column on executions.
*   **Documentation**: Expanded the enum descriptions for `mode` and `status`, added `guide` fields to the tables, and clarified the `tags` (filter) vs `tag_names` (column) distinction.
*   **Function**: Removed the redundant `executions_by_workflow` function.
*   **Setup**: Removed the dangerous `N8N_BASE_URL` default and set it to `required: true`.
*   **Typing**: Regarding the execution IDs — I kept them as `Utf8` because changing them to `Int64` breaks the JOIN against `workflows.id` (which is a string). Coral handles the coercion from JSON numbers to `Utf8` perfectly, so this keeps the joins working cleanly.

Let me know if everything looks good to go!

sources/community/n8n/manifest.yaml
wiz-abhi: Removed! Thanks for the tip.

antonmry (approved):
Thanks for the contribution, @wiz-abhi, and for the careful turnaround on the review. All 11 of my comments are addressed in `799b52e0` — the `status` column, the `triggerCount`/`description`/`is_archived` additions, the `N8N_BASE_URL` safety fix, the enum docs, the function removal, the README tag example, and the tags-table wording all look good. Thank you also for resolving each thread as you went — it made re-reviewing fast.

The `Utf8` choice for execution IDs is a sensible call given the workflow-id join, and the PR body documents the rationale clearly.

Two tiny non-blocking observations you can fold into a follow-up if you like — happy to merge as-is:
- `manifest.yaml:23` still lists `credentials` in the required-scopes hint even though the credentials table was removed.
- The `test_queries` block doesn't exercise a filter pushdown (e.g. `WHERE status = 'error'`), which would catch virtual-filter regressions in CI.

Approving — please go ahead and merge whenever you're ready.

wiz-abhi:
> Thanks for the contribution, @wiz-abhi, and for the careful turnaround on the review. All 11 of my comments are addressed in `799b52e0` — the `status` column, the `triggerCount`/`description`/`is_archived` additions, the `N8N_BASE_URL` safety fix, the enum docs, the function removal, the README tag example, and the tags-table wording all look good. Thank you also for resolving each thread as you went — it made re-reviewing fast.
> 
> The `Utf8` choice for execution IDs is a sensible call given the workflow-id join, and the PR body documents the rationale clearly.
> 
> Two tiny non-blocking observations you can fold into a follow-up if you like — happy to merge as-is:
> 
> * `manifest.yaml:23` still lists `credentials` in the required-scopes hint even though the credentials table was removed.
> * The `test_queries` block doesn't exercise a filter pushdown (e.g. `WHERE status = 'error'`), which would catch virtual-filter regressions in CI.
> 
> Approving — please go ahead and merge whenever you're ready.

Thanks for the review! Folded in those last two suggestions regarding the manifest hint and test queries, so please go ahead and merge it whenever you're free!

PR #687, opened 2026-05-22
feat(sources): harden OAuth URLs and confirm outbound hosts on setup

sauldhernandez:
## Intent

Tighten the security posture of source specs in two ways:

1. **OAuth endpoints must use HTTPS.** Previously `authorization_url` and `token_url` were only checked for valid URL syntax — a manifest could point the OAuth code/token exchange at a plaintext `http://` endpoint where credentials can be intercepted.
2. **No visibility into where a source connects.** Installing a source gave the user no way to review which network hosts the source would contact before it was wired up.

## What it enables

- **HTTPS enforcement for OAuth provider endpoints.** `coral-spec` now rejects non-`https` `authorization_url` / `token_url` at manifest-validation time. `http` is allowed **only for loopback hosts** (`127.0.0.1` / `localhost`) — loopback traffic can't be intercepted off-machine and this keeps local OAuth test servers usable. This mirrors the existing loopback exception on `redirect_uri`.
- **Outbound-host extraction.** New `ValidatedSourceManifest::outbound_hosts()` statically derives every host a source will contact: the HTTP `base_url` (templates resolved against input defaults; unresolved `{{input.X}}` surfaced verbatim), OAuth `authorization_url` / `token_url`, and file/object-store table locations (`file://` omitted as local).
- **Confirmation step during source setup.** Hosts are exposed through the `SourceInfo` proto, and the CLI lists them and asks the user to confirm before installation proceeds.

## Usage examples

Interactive install (`coral source add --interactive <name>`, `--file`, or `coral onboard`):

```
This source will connect to the following hosts:
  • api.github.com
  • github.com
Continue connecting this source? [Y/n]
```

Declining aborts the install. Non-interactive installs print the host list for visibility but do not block (no user available to answer).

A manifest with a plaintext OAuth endpoint now fails validation:

```
manifest input 'API_TOKEN' oauth.endpoints.token_url must use https
```

## Reviewer notes

- **Loopback-http exception** is a deliberate call. The instruction was "all OAuth URLs must be HTTPS"; strict https-only would also have required reworking the OAuth test fixtures (which bind a local plaintext server). Loopback is exempt for the same reason `redirect_uri` already is. Easy to make strict if preferred.
- The `SourceInfo.hosts` proto field is populated server-side and consumed by the CLI. The desktop/web UI source-setup flow would need its own UI work to surface the confirmation — out of scope here, but the data is now available to it.
- Scope of "all hosts": `base_url` + OAuth endpoints + file locations. Request `path` specs are relative to `base_url` (same host) and intentionally not enumerated.

## Testing

- `coral-spec`: 116 tests pass, including new `endpoints` host-extraction tests and `inputs` HTTPS-enforcement tests.
- `coral-app`: source/catalog/OAuth-import tests pass (loopback-http token URLs still accepted).
- `coral-cli`: 30 e2e tests pass.
- `cargo clippy` clean across `coral-spec`, `coral-app`, `coral-cli`.


crates/coral-cli/src/lib.rs
```diff
@@ -782,6 +637,10 @@ async fn run_source_add(app: &AppClient, args: SourceAddArgs) -> Result<(), CliE
                 .into_iter()
                 .find(|source| source.name == bundled_name)
                 .ok_or_else(|| anyhow::anyhow!("unknown bundled source '{bundled_name}'"))?;
+            if !source_ops::confirm_source_hosts(&available.hosts, interactive)? {
```
jsummerfield: This confirmation is happening before we know the effective source inputs, so it can show the wrong host. For example, `GITHUB_API_BASE` can override the default `api.github.com`, but this still uses the discovery-time `available.hosts`. Could we either move confirmation after input collection and render hosts from the actual bindings, or explicitly downgrade this field/UI to unresolved "declared endpoint hints" rather than confirmed outbound hosts?

crates/coral-spec/src/inputs.rs
```diff
@@ -1112,6 +786,26 @@ fn redirect_uri_has_explicit_port(raw: &str) -> bool {
     !port.is_empty() && port.bytes().all(|byte| byte.is_ascii_digit())
 }
 
+fn validate_url(input_key: &str, field: &str, raw: &str) -> Result<()> {
```
jsummerfield: This only covers the statically declared provider endpoint strings. PR #1129 had a more complete version of this hardening: validate rendered/template endpoint URLs, allow loopback via parsed IPv4/IPv6 hosts, reject unsafe device-code `verification_uri` / `verification_uri_complete` values before opening them, and validate persisted refresh `token_url` before reuse. Can we add those pieces back here or rebase on that PR?

crates/coral-cli/src/source_ops.rs
```diff
@@ -476,6 +411,44 @@ pub(crate) fn require_interactive() -> Result<(), anyhow::Error> {
     Ok(())
 }
 
+/// Shows the network hosts a source will contact and, in interactive mode,
+/// asks the user to confirm before installation proceeds.
+///
+/// Returns `true` when installation should continue. In non-interactive mode
+/// the hosts are printed for visibility and `true` is returned without
+/// prompting, since there is no user available to answer.
+pub(crate) fn confirm_source_hosts(
```
jsummerfield: This changes `coral source add`/onboarding UX and the PR also tightens OAuth endpoint validation, but the CLI docs, source-spec docs, and maintained source-authoring/review skills are not updated. The repo rules require docs for CLI surface changes and aligned docs/skills when credential/auth guidance changes; PR #1129 already has the source-spec and skill guidance for the HTTPS endpoint rule that should be carried over.

jsummerfield (commented):
A couple of the issues here overlap with PR #1129. Please pull those missing pieces back in or rebase on that work: rendered/template OAuth endpoint validation, device-code verification URL validation, persisted refresh-token URL validation, and the docs/source-authoring skill updates for the HTTPS endpoint rule.

crates/coral-cli/src/onboard.rs
```diff
@@ -200,8 +200,17 @@ async fn run_add_bundled_source(app: &AppClient, source: &SourceInfo) -> Result<
         .iter()
         .map(manifest_input_from_proto)
         .collect::<Result<Vec<_>, _>>()?;
-    let inputs = source_ops::prompt_for_inputs_with_credential_methods_in_mode(
+    let host_variables =
```
jsummerfield: This resolved-host confirmation was added for initial onboarding setup, but the installed-source “Update credentials” path above still prompts variables and calls `add_bundled_source_with_credentials` directly. Since variables can change a source’s base URL, reconfigure can still move an installed source to a different outbound host without this confirmation. Can the reconfigure path reuse this same collect variables -> resolve hosts -> confirm -> prompt remaining secrets flow?

crates/coral-spec/src/inputs.rs
```diff
@@ -179,9 +179,10 @@ fn render_oauth_endpoint_url(
             }
         }
     }
-    Url::parse(&rendered).map_err(|error| {
+    let url = Url::parse(&rendered).map_err(|error| {
         ManifestError::validation(format!("invalid OAuth {label} URL: {error}"))
     })?;
+    validate_oauth_endpoint_url_scheme(&url, &format!("OAuth {label} URL"))?;
```
jsummerfield: This validates rendered manifest endpoint URLs, but device-code OAuth still accepts provider-returned `verification_uri` / `verification_uri_complete` without the same HTTPS-or-loopback check before the CLI exposes or opens them. Please carry over the #1129 runtime validation for those device authorization response URLs too.

crates/coral-spec/src/inputs.rs
```diff
@@ -179,9 +179,10 @@ fn render_oauth_endpoint_url(
             }
         }
     }
-    Url::parse(&rendered).map_err(|error| {
+    let url = Url::parse(&rendered).map_err(|error| {
```
jsummerfield: The refresh path needs the same hardening at runtime. Persisted OAuth refresh `token_url` metadata is still reused before validating that it is HTTPS or loopback, so a stale or tampered stored URL can receive the refresh token. Please validate the stored refresh URL before posting to it.

crates/coral-spec/src/inputs.rs
```diff
@@ -561,11 +562,7 @@ fn validate_oauth_endpoint_template(
     }
 
     if !has_required_variable {
-        Url::parse(&rendered).map_err(|error| {
-            ManifestError::validation(format!(
-                "manifest input '{input_key}' oauth.endpoints.{field} is invalid: {error}"
-            ))
-        })?;
+        validate_url(input_key, field, &rendered)?;
```
jsummerfield: This still only validates when the full template can be rendered without required variables. A fixed remote plaintext prefix like `http://provider.example.com/{{input.TENANT}}/token` passes manifest validation and fails only later at OAuth runtime. Please validate the fixed scheme/prefix before the first required variable.

crates/coral-spec/src/inputs.rs
```diff
@@ -1112,6 +1109,28 @@ fn redirect_uri_has_explicit_port(raw: &str) -> bool {
...
+}
+
+fn validate_oauth_endpoint_url_scheme(url: &Url, context: &str) -> Result<()> {
+    // OAuth provider endpoints carry the authorization code and exchange it for
+    // an access token, so they must not travel over plaintext HTTP where they
+    // can be intercepted. A loopback host is exempt: it cannot be observed off
+    // the machine and keeps local OAuth test servers usable.
+    let host = url.host_str().unwrap_or_default();
+    let is_loopback = host == "127.0.0.1" || host == "localhost";
+    if url.scheme() != "https" && !(url.scheme() == "http" && is_loopback) {
```
jsummerfield: This changes the source-spec contract, but the source-spec reference, custom-source guide, and maintained source author/review skills still do not mention the HTTPS-or-loopback requirement for OAuth provider endpoints. Please update those docs/skills with the new rule.

crates/coral-spec/src/endpoints.rs
```diff
@@ -0,0 +1,537 @@
...
+        return;
+    }
+    match Url::parse(raw) {
+        Ok(url) => {
+            // A `file://` location reads from the local filesystem; there is no
+            // remote host to report.
+            if url.scheme() == "file" {
+                return;
+            }
+            if let Some(host) = url.host_str() {
```
jsummerfield: For `s3://bucket/path`, this reports `bucket` as a network host, but the actual outbound target is the S3 service endpoint derived from object-store config/defaults. Either render the service endpoint or label these as object-store locations/static hints instead of claiming they are network hosts.

jsummerfield (commented):
Refresh review comments for the current head.

AlbertQM (approved):

sauldhernandez:
@cubic-dev-ai review

antonmry (approved):

crates/coral-spec/src/url_policy.rs
```diff
@@ -0,0 +1,45 @@
+//! Shared URL scheme policy for OAuth endpoints and loopback redirects.
+//!
+//! Coral requires OAuth provider endpoints and redirect URLs to use `https`,
+//! permitting plain `http` only when the host is loopback. The same policy is
+//! enforced while validating a manifest (here in `coral-spec`), at OAuth
+//! runtime in `coral-app`, and during the CLI loopback redirect flow. Keeping
+//! the allowed-scheme set, the loopback definition, and the user-facing
+//! messages in one place stops those three copies from drifting apart.
```
jsummerfield: Nice consolidation, and the module doc states the goal exactly. Two call sites still don't route through it, which undercuts the anti-drift intent:

- `backends/mcp.rs` has `is_local_http_url` (around line 540) that is a character-for-character copy of `is_loopback_url` here, and its scheme dispatch (around line 488) re-implements `validate_https_or_loopback_scheme` including the same "must use https unless it targets localhost" message. This file isn't otherwise part of this PR, but since this PR establishes the canonical helper, it's the natural moment to delete the clone and call through. Pure refactor — the bodies are identical.
- The redirect bind check in `inputs.rs` (around line 1121) tests `host != "127.0.0.1" && host != "localhost"`, a narrower loopback set than `is_loopback_url` (misses the rest of `127.0.0.0/8` and `::1`). Converging it on `is_loopback_url` is worthwhile but **widens** the redirect hosts a manifest may declare, so treat it as a deliberate decision rather than a mechanical cleanup.

crates/coral-spec/src/inputs.rs
```diff
@@ -181,12 +181,29 @@ fn render_oauth_endpoint_url(
...
-    Url::parse(&rendered).map_err(|error| {
-        ManifestError::validation(format!("invalid OAuth {label} URL: {error}"))
-    })?;
+    let context = format!("OAuth {label} URL");
+    crate::validate_https_or_loopback_url(&context, &rendered)
+        .map_err(ManifestError::validation)?;
     Ok(rendered)
 }
 
+fn validate_oauth_provider_endpoint_template_prefix(context: &str, raw_prefix: &str) -> Result<()> {
```
jsummerfield: This fallback re-lists `"https" | "http"` and re-types the verbatim error string that `url_policy` already owns — the drift `url_policy`'s doc set out to prevent. The function also detects the scheme two different ways (`Url::parse` on the parseable path, hand-rolled `split_once("://")` here) with no comment on why the fallback exists or what `strip_suffix(':')` guards.

Suggest adding a scheme-string entry point to `url_policy` (taking the scheme plus the unknown-host case) so the allow-list, the message, and the "http allowed when the host is an unresolved variable" exception live in one place, and have this delegate. At minimum, add a couple of lines explaining the fallback and the `strip_suffix`.

crates/coral-spec/src/inputs.rs
```diff
@@ -590,11 +609,12 @@ fn validate_oauth_endpoint_template(
...
     if !has_required_variable {
-        Url::parse(&rendered).map_err(|error| {
-            ManifestError::validation(format!(
-                "manifest input '{input_key}' oauth.endpoints.{field} is invalid: {error}"
-            ))
-        })?;
+        let context = format!("manifest input '{input_key}' oauth.endpoints.{field}");
+        crate::validate_https_or_loopback_url(&context, &rendered)
+            .map_err(ManifestError::validation)?;
+    } else if let Some(prefix) = rendered_before_first_required_variable.as_deref() {
```
jsummerfield, line 782: Worth a doc note here: when a required variable supplies the URL authority (e.g. `token_url: "{{input.HOST}}/oauth/token"`), `rendered_before_first_required_variable` is empty and this prefix check returns `Ok(())`, so the scheme isn't enforced at manifest-validation time. That's safe because `render_oauth_endpoint_url` re-checks the fully-rendered URL at runtime and sits unconditionally in front of every contact path (I traced it) — but a reader of the validator can't tell the check is intentionally partial. Suggest a one-line comment saying enforcement defers to runtime for variable-supplied authorities, plus a regression test for the leading-variable case (`{{input.X}}/oauth/token` resolving to non-loopback http) so a future refactor can't silently open a real gap.

crates/coral-cli/src/source_ops.rs
```diff
@@ -621,47 +687,50 @@ pub(crate) fn source_name_arg(name: Option<&str>) -> Result<String, anyhow::Erro
...
+    mode: CredentialPromptMode,
 ) -> Result<CollectedSourceInputs, anyhow::Error> {
-    prompt_for_inputs_with_credential_methods_in_mode(inputs, CredentialPromptMode::EnvFirst)
+    collect_remaining_inputs_with_env_lookup(inputs, variables, mode, |key| {
+        read_source_input_env(key).unwrap_or_default()
+    })
 }
 
-pub(crate) fn prompt_for_inputs_with_credential_methods_in_mode(
+fn collect_remaining_inputs_with_env_lookup(
```
jsummerfield: The host-before-secrets ordering is the right UX, but the invariant isn't owned by one place. This function takes already-collected `variables`, assigns them, and `continue`s past every variable input trusting the caller ran `prompt_variables_for_host_confirmation` first — a precondition that's invisible at the type level and repeated across three call sites (lib.rs ×2, onboard.rs). Pass an empty `Vec` and the variables silently vanish from the install request with no error.

The seam itself is legitimate (host confirmation + an RPC must sit between variables and secrets), so this isn't about merging the passes back together — it's about letting one abstraction own the order, e.g. `collect_inputs_confirming_hosts(inputs, mode, confirm_closure)`, or a `HostConfirmedVariables` newtype so the precondition is load-bearing at compile time. (The confirm step is async in two of the three sites.) Related: the `prompt_for_remaining_inputs_with_credential_methods_in_mode` name encodes this two-pass history in the API — if the above lands it can become `collect_secret_inputs`.

crates/coral-cli/src/source_ops.rs
```diff
@@ -602,6 +630,44 @@ pub(crate) fn require_interactive() -> Result<(), anyhow::Error> {
     Ok(())
 }
 
+/// Shows the network hosts a source will contact and, in interactive mode,
+/// asks the user to confirm before installation or reconfiguration proceeds.
+///
+/// Returns `true` when installation or reconfiguration should continue. In
+/// non-interactive mode the hosts are printed for visibility and `true` is
+/// returned without prompting, since there is no user available to answer.
+pub(crate) fn confirm_source_hosts(
```
jsummerfield: This does two jobs depending on `interactive`: in interactive mode it prints the hosts and asks (the bool matters); in non-interactive mode it only prints and returns a constant `true` that both callers discard. At the non-interactive sites, `confirm_source_hosts(&hosts, false)?` reads as if something is being confirmed when nothing is. Consider splitting into `print_source_hosts(hosts)` (no return) and `confirm_source_hosts(hosts) -> bool` (always prompts, interactive callers only) — the `interactive` flag and the meaningless bool both disappear.

crates/coral-spec/src/endpoints.rs
```diff
@@ -0,0 +1,973 @@
...
+use url::Url;
+
+use crate::{
+    ManifestInputSpec, McpServerSpec, ParsedTemplate, TemplateNamespace, TemplatePart,
+    ValidatedSourceManifest,
+    backends::file::FileObjectStoreSpec,
+    v4::{SurfaceDescriptor, openapi_document_metadata},
+};
+
+const UNRESOLVED_OPENAPI_SERVER_HOST: &str =
```
jsummerfield, line 22: `outbound_hosts` returns `Vec<String>` where an element can be a real host, a `host:port`, a raw `{{input.X}}` template, or this sentence-long sentinel — and the CLI prints each identically as `• <value>`. So under "This source will connect to the following hosts:" a user can see `• OpenAPI servers[0].url runtime host (unresolved; declare base_url to review before import)` presented as a hostname.

A `Vec<OutboundHost>` enum would be cleanest but it crosses gRPC as `repeated string hosts`, so it's heavier than it looks. The proportionate fix: don't emit the prose sentinel (and ideally the raw template) as a pseudo-host — surface unresolved/derived cases as separate, distinctly-labeled UI ("could not be determined ahead of time:") so the bullet list stays real hostnames.

crates/coral-spec/src/endpoints.rs
```diff
@@ -0,0 +1,973 @@
...
+) -> String {
+    ParsedTemplate::parse(raw).map_or_else(
+        |_| raw.to_string(),
+        |template| render_with_input_values(&template, inputs, source_inputs),
+    )
+}
+
+/// Extracts a displayable host from a (possibly templated) URL string and adds
+/// it to `hosts`.
+fn collect_host(hosts: &mut BTreeSet<String>, raw: &str) {
```
jsummerfield: The `file` scheme is checked three times in this function: `has_scheme(raw, "file")`, then again inside `insert_parsed_url_host` (`url.scheme() == "file"`), then a third time by re-parsing the same string with `Url::parse(raw)`. That third parse is redundant — the string was already parsed in `insert_parsed_url_host`. Consider having `insert_parsed_url_host` return the parsed `Url` (or take one) so you parse once, and extracting a single `is_local_file_location` predicate with a comment covering the `file:/` vs `file://` distinction. Minor related note: the trailing `collect_host` in `collect_file_source_hosts` only fires for schemes the file backend already rejects at validation — a defensive-only branch worth a one-line comment.

crates/coral-spec/src/endpoints.rs
```diff
@@ -0,0 +1,973 @@
...
+    {
+        return;
+    }
+    hosts.insert(UNRESOLVED_OPENAPI_SERVER_HOST.to_string());
+}
+
+/// Renders a template, substituting input tokens with their manifest default
+/// value. Tokens with no usable default are left as their `{{...}}` literal so
+/// the resulting string still signals an unresolved, input-driven endpoint.
+fn render_with_input_values(
```
jsummerfield, line 212: The three-tier precedence (resolved input value → token inline default → declared-input default) is built as an `if/else` feeding an `or_else` chain, and `matches!(token.namespace(), Input)` appears in two of the arms — which makes it look like the two Input branches might differ when they don't. A single top-level `match token.namespace()` with the linear fallback inside the Input arm reads top-to-bottom (keep the middle `token.default_value()` tier applying to non-Input tokens too, to preserve behavior).

crates/coral-spec/src/validate.rs
```diff
@@ -965,6 +967,80 @@ pub(crate) fn validate_template(
...
+///
+/// Outbound-host confirmation resolves these templates from the non-secret
+/// variable values the user supplies *before* connecting — secrets are only
+/// collected afterward — so a secret-backed host could never be shown for
+/// confirmation: the user would approve a `{{…}}` placeholder while the runtime
+/// quietly substitutes the secret and contacts the real, unconfirmed host.
+/// Reject it at validation instead, mirroring the rule already enforced for
+/// OAuth endpoint templates.
+///
+/// Only tokens in the scheme/authority (host[:port]) region are restricted; a
```
jsummerfield: This invariant is the best part of the PR — enforcing it at validation rather than at confirmation time is exactly right. One precision nit on the doc: it says the restricted region is "scheme/authority (host[:port])", but `advance_past_url_authority` ends the authority at the first `/`, `?`, or `#`, so it actually restricts the whole authority including any `userinfo` (e.g. `{{user}}:{{pass}}@host`). The behavior is fine — rejecting a secret in userinfo is defensible — just drop or soften the "host[:port]" parenthetical so it matches what the code does. (Note `redirect_uri_has_explicit_port` strips userinfo via `rsplit_once('@')`, so the two authority parsers differ on purpose.)

crates/coral-app/src/sources/service.rs
```diff
@@ -92,6 +93,32 @@ impl SourceServiceApi for SourceService {
...
+        &self,
+        request: Request<ResolveBundledSourceHostsRequest>,
+    ) -> Result<Response<ResolveBundledSourceHostsResponse>, Status> {
+        // Bundled candidates arrive here by catalog name; the manifest YAML
+        // lives in coral-app. File imports already have their manifest
+        // client-side and resolve hosts before calling ImportSource.
+        let span = grpc_span(&request);
+        instrument_grpc(span, async move {
+            let request = request.into_inner();
+            let _workspace_name = workspace_name_from_proto(request.workspace.as_ref())?;
```
jsummerfield: `_workspace_name` is parsed and discarded, which reads like dead validation — but `workspace_name_from_proto` returns `Result` and rejects a missing/malformed workspace, so the `?` is load-bearing request validation even though bundled host resolution is workspace-independent. Either drop the binding (`workspace_name_from_proto(request.workspace.as_ref())?;`) so it reads as a pure validation gate, or add a one-line comment. Don't remove the call — that would silently accept malformed requests.

jsummerfield (commented):
Strong PR overall — the functionality lands behind mostly clean interfaces, and the two hardest calls are made well:

- **`validate_host_template_inputs`** is the standout. Pushing "a secret can never determine an outbound host" down to manifest validation means the confirm-before-secrets flow is correct by construction, rather than relying on a check at confirmation time.
- **`url_policy.rs`** is the right place to centralize the https-or-loopback decision, and **`outbound_hosts_with_input_values`** hides a genuinely sprawling extraction problem behind a simple signature.
- The bundled-vs-file resolution split is inherent (server-side YAML vs client-side manifest), not duplicated logic — I checked the single name-or-YAML RPC alternative and it's worse (needless network hop).

Inline comments are grouped into three themes worth a look — the scheme policy not fully routing through `url_policy`; the two-pass input-collection ordering; and `outbound_hosts` returning non-host strings — plus a few smaller polish notes. None block merge.

jsummerfield (approved):

docs/guides/write-a-custom-source.mdx
```diff
@@ -176,6 +176,41 @@ For field-level details and examples, see [Credential methods](/reference/source
...
+inputs:
+  DEMO_API_KEY:
+    kind: secret
+    required: false
+  DEMO_OAUTH_ACCESS_TOKEN:
+    kind: secret
+    required: false
+```
+
+### OAuth client and redirect options
```
AlbertQM: after #1235 , this kind of info has moved to the reference; could you please move it there? Thanks!

AlbertQM (changes requested):

docs/reference/cli-reference.mdx
```diff
@@ -88,7 +88,7 @@ coral source info datadog
 coral source info datadog --verbose
 ```
 
-### `coral source add <NAME>`
+### `coral source add`
```
AlbertQM: Why remove `<NAME>`? I think we should keep it, to conform to the shape of the other commands 👀

AlbertQM (approved):

PR #688, opened 2026-05-22
fix(ci): invoke Windows release smoke binary correctly

simonwhitaker:
## Summary

- Fix the Windows release smoke step so PowerShell expands the target-specific binary path before invocation.
- Invoke the computed executable path with PowerShell's call operator.

## Root Cause

The release workflow tried to execute `.\target\$env:TARGET\release\coral.exe` directly. In PowerShell, the environment variable was not expanded inside that command path, so the Windows job looked for a literal path containing `$env:TARGET` and failed before packaging.

## Validation

- `git diff --check`
- `pwsh` path expansion check for `target\$env:TARGET\release\coral.exe`
- `pwsh` parser check for the fixed smoke-step snippet
- Ruby YAML parse of `.github/workflows/release.yml`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -137,8 +137,9 @@ jobs:
         env:
           TARGET: ${{ matrix.target }}
         run: |
-          .\target\$env:TARGET\release\coral.exe --version
-          .\target\$env:TARGET\release\coral.exe source discover
+          $coral = "target\$env:TARGET\release\coral.exe"
+          & $coral --version
+          & $coral source discover
 
       - name: Package Unix
         if: runner.os != 'Windows'


simonwhitaker:
## Code Review (by Claude)

🤖 *This review was generated by Claude (Claude Code).*

**Verdict: ✅ Approve** — small, correct, well-targeted CI fix.

### What the PR does
Replaces two direct invocations of an interpolated `.exe` path in the Windows release **smoke** step with a computed variable invoked via PowerShell's call operator (`&`), so the target-specific binary path resolves correctly before packaging.

### Correctness
- **The fix is sound and idiomatic.** Assigning the path to `$coral` and invoking with `& $coral ...` is the canonical PowerShell way to run an executable from a computed/string path, and removes any ambiguity about how the leading token is parsed.
- **Dropping the leading `.\` is fine.** With `&`, a relative path like `target\...\coral.exe` is invoked correctly — no regression.
- **Root-cause note (minor):** the precise failure mode is that an unquoted path used as the *command name* token is parsed specially, so `$env:TARGET` didn't interpolate the way it does in argument position. The new form sidesteps this entirely; the description is directionally right and the fix is the correct remedy.

### Follow-up check: is the **Package Windows** step exposed to the same bug?
No — I checked, and it's safe. The distinction is *where* `$env:TARGET` appears:
- **The bug (smoke step):** the interpolated path was the **command token** itself (`.\target\$env:TARGET\release\coral.exe --version`), which PowerShell parses specially.
- **Package Windows step:** every `$env:TARGET` reference is inside a **double-quoted string argument** to a cmdlet (`Copy-Item`, `Compress-Archive`), which is the normal, safe interpolation context:
  - `$staging = "package-$env:TARGET"` ✅
  - `Copy-Item "target\$env:TARGET\release\coral.exe" "$staging\coral.exe"` ✅
  - `Compress-Archive ... -DestinationPath "..\coral-$env:TARGET.zip"` ✅
- `$env:TARGET.zip` is also fine: PowerShell only does simple-variable expansion in strings (member access needs `$(...)`), so it parses as `<TARGET>` + literal `.zip`.

**Conclusion:** the smoke step was the only place invoking an interpolated path as a command, so it was the only place exposed. The packaging step is correct as-is — no follow-up fix needed.

### Risk / coverage
Low risk: CI-only, Windows-only, runs before packaging — fails loudly rather than shipping a broken artifact. The only true confirmation is a green Windows release job.

jsummerfield (approved):

PR #692, opened 2026-05-22
feat(sources/community/notion): add Notion community source manifest …

garimapahwa:
## Summary

Add a new Notion community source for Coral so users can query Notion pages, databases, search results, and block children through the Notion API using a simple integration token.

## What It Enables

- Read-only discovery of Notion content through Coral.
- Search across shared pages and data sources.
- Query rows from a specific database.
- Retrieve individual pages and block children.
- A minimal setup path with `NOTION_API_KEY` and the required `Authorization` / `Notion-Version` headers.

## Usage Examples

- Search pages by title or text:

  SELECT id, url FROM notion.search WHERE query = 'meeting notes' LIMIT 10

- Query rows from a database:

  SELECT id, properties FROM notion.database_rows WHERE database_id = '<database_id>' LIMIT 20

- Retrieve a single page:

  SELECT id, properties FROM notion.pages WHERE page_id = '<page_id>'

## Validation

- `ruby -ryaml -e "YAML.load_file('sources/community/notion/manifest.yaml'); puts 'YAML OK'"`

## Notes

- The source is documented in [sources/community/notion/README.md](sources/community/notion/README.md).
- The manifest lives in [sources/community/notion/manifest.yaml](sources/community/notion/manifest.yaml).


jsummerfield:
Thank you for contributing. We already have a Notion connector in core/. If you see any opportunities to improve, please raise a PR based on that one.

PR #698, opened 2026-05-22
feat(sources/community/monday): add monday.com community source

RajdeepKushwaha5:
### What changed

Adds the `monday` community source for monday.com's GraphQL API v2.

- `boards`: up to 500 accessible active boards, including workspace metadata
- `users`: up to 500 non-guest account users, including email and role flags
- `board_items(board_id)`: all items for one board using cursor pagination; raw item data includes `column_values`
- `board_columns(board_id)`: column definitions for one board

Requests use GraphQL variables for board IDs, path-addressed JSON request bodies, and the current stable monday.com API version (`2026-04`).

### Why

This makes monday.com project data queryable alongside Coral's other engineering sources without ETL. The README includes standalone queries and clearly labels the heuristic used by the GitHub login-matching example.

### Limitations

- `boards` and `users` are capped at 500 records because monday.com's page number must be sent inside the GraphQL JSON body, which Coral's current HTTP source DSL cannot paginate.
- Item column values are available in `raw.column_values` but are not decoded into typed columns.
- Only active boards and non-guest users are returned.
- A monday.com token was not available in this environment, so live account output could not be captured. The manifest was checked against monday.com's current provider documentation.

### Validation

- `coral source lint sources/community/monday/manifest.yaml`
- `ryl -c .yamllint.yaml --stdin-filename sources/community/monday/manifest.yaml -` (normalized input)
- `git diff --check`
- `git merge-tree --write-tree HEAD upstream/main`

### Provider references

- https://developer.monday.com/api-reference/docs/authentication
- https://developer.monday.com/api-reference/docs/api-versioning
- https://developer.monday.com/api-reference/reference/boards
- https://developer.monday.com/api-reference/reference/users
- https://developer.monday.com/api-reference/reference/items-page

### DIFF
--- a/sources/community/monday/README.md
+++ b/sources/community/monday/README.md
@@ -1,0 +1,165 @@
+# monday.com (Community)
+
+**Version:** 0.1.0
+**Backend:** HTTP (monday.com GraphQL API v2)
+**Tables:** 2 · **Functions:** 2
+**Base URL:** `https://api.monday.com`
+
+Query monday.com boards, items, columns, and users via the GraphQL API v2.
+Join with GitHub PRs, Linear issues, and PagerDuty incidents for cross-source
+project management and engineering intelligence.
+
+## Install
+
+```bash
+coral source add --file sources/community/monday/manifest.yaml
+```
+
+## Authentication and setup
+
+Requires a monday.com personal API token.
+
+1. In monday.com, go to **your avatar → Developers → My Access Tokens**.
+2. Copy the token value.
+
+```bash
+export MONDAY_API_TOKEN=eyJhbGciOiJIUzI1NiJ9...
+coral source add --file sources/community/monday/manifest.yaml
+```
+
+The token is passed directly as the `Authorization` header value — no `Bearer`
+prefix.
+
+## Tables and functions
+
+### Tables
+
+| Table | Description |
+| --- | --- |
+| `boards` | All active boards with ID, name, kind, state, and workspace |
+| `users` | All non-guest account users with email, title, and role flags |
+
+### Table functions
+
+| Function | Required args | Description |
+| --- | --- | --- |
+| `board_items(board_id)` | `board_id` | Items (rows/tasks) for one board, up to 500 |
+| `board_columns(board_id)` | `board_id` | Column definitions for one board |
+
+Obtain `board_id` values from `monday.boards`.
+
+## Example queries
+
+### List all active boards
+
+```sql
+SELECT id, name, board_kind, workspace_name
+FROM monday.boards
+ORDER BY name
+LIMIT 20;
+```
+
+### Items in a board
+
+```sql
+SELECT id, name, state, group_title, created_at
+FROM monday.board_items('1234567890')
+WHERE state = 'active'
+ORDER BY created_at DESC
+LIMIT 50;
+```
+
+### Column definitions for a board
+
+```sql
+SELECT id, title, type
+FROM monday.board_columns('1234567890')
+ORDER BY title;
+```
+
+### All users
+
+```sql
+SELECT id, name, email, title, is_admin, time_zone
+FROM monday.users
+ORDER BY name;
+```
+
+## Cross-source examples
+
+### Engineers with open GitHub PRs and active monday.com items
+
+```sql
+SELECT u.name, u.email,
+       COUNT(DISTINCT pr.id) AS open_prs,
+       COUNT(DISTINCT mi.id) AS active_items
+FROM monday.users u
+LEFT JOIN github.pull_requests pr
+  ON LOWER(pr.author_login) = LOWER(SPLIT_PART(u.email, '@', 1))
+  AND pr.state = 'open'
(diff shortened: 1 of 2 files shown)


jsummerfield:
Thanks for the contribution. I found a few blockers/issues before this should land:

1. `sources/community/monday/manifest.yaml` does not pass `coral source lint`. The four GraphQL request bodies are written as raw YAML objects (`body: { query: ... }`), but Coral expects request bodies as path-addressed body fields or tagged text/json bodies. Please encode the monday GraphQL JSON body as entries like `path: [query] / from: literal / value: ...`, and use GraphQL variables for `board_id` rather than interpolating it into query text.

2. `board_items` currently returns only the first `items_page(limit: 500)` and discards the cursor. monday documents `items_page` as cursor-paginated, with `next_items_page` required to retrieve the rest of a board. As written, this source will silently return incomplete item data for larger boards. Please add a continuation path, or otherwise make the limitation explicit in the function shape rather than presenting this as the board item access surface.

3. The README GitHub join example references `github.pull_requests` and `pr.author_login`, but Coral’s GitHub source exposes `github.pulls` and `user__login`. Please update the example to a query that works against the current bundled GitHub source.

jsummerfield (changes requested):
Findings
- High: `sources/community/monday/manifest.yaml:59`, `sources/community/monday/manifest.yaml:164`, `sources/community/monday/manifest.yaml:276`, and `sources/community/monday/manifest.yaml:372` do not lint/load with Coral’s HTTP body schema. The source uses `body: { query: ... }`, but Coral expects path-addressed body fields or a supported tagged body form. Please encode the GraphQL body using the existing GraphQL-source pattern, for example body fields for `[query]` and `[variables, ...]`.
- High: `sources/community/monday/manifest.yaml:277` interpolates `{{arg.board_id}}` directly into the GraphQL text. Please use GraphQL variables (`$boardId`) and pass `variables.boardId` from the Coral argument instead of string interpolation. monday.com GraphQL docs: https://developer.monday.com/api-reference/docs/getting-started
- High: `sources/community/monday/manifest.yaml:277` and `sources/community/monday/README.md:142` return only the first `items_page(limit: 500)` and drop the cursor, silently truncating real boards. monday.com documents `items_page` as cursor-paginated and uses `next_items_page` for continuation. Please implement cursor pagination if Coral can express it here, or narrow the function/docs so it does not present this as complete board item coverage. Docs: https://developer.monday.com/api-reference/reference/items-page and https://developer.monday.com/api-reference/reference/next-items-page
- Medium: `sources/community/monday/README.md:93` uses `github.pull_requests` and `pr.author_login`, but the bundled Coral GitHub source exposes `github.pulls` and the author login column is `user__login`. Please update the example, including required GitHub filters where needed.
- Medium: `sources/community/monday/README.md:129` lists validation commands but does not provide captured live-test evidence. Please add sanitized output showing `coral source add`, `coral source test monday`, and real queries for `boards`, `users`, `board_columns`, and `board_items`.

Thanks for the useful project-management source. The immediate blocker is that the manifest body shape does not load; after that, pagination and live evidence need to be tightened before approval.

RajdeepKushwaha5:
Addressed the outstanding review feedback and pushed `b8b8b5be`: request bodies and GraphQL variables are structured correctly, `board_items` now follows its cursor across pages, the GitHub example uses the actual bundled table/field names, and the docs now state the 500-record boards/users boundary and raw column-value behavior precisely. I also re-checked the source against monday.com's current provider docs and reran source lint, normalized YAML lint, diff checks, and a merge simulation against `upstream/main`. No monday.com token is available in this environment, so I could not add fabricated live output. Please re-review the updated branch.

jsummerfield (changes requested):
Findings:

- High: live Coral evidence is still missing for the current monday.com source. The PR body explicitly says a monday.com token was not available, and I could not find captured successful `coral source add`, `coral source test`, or representative `coral sql` output in the README or discussion. The GraphQL body/variables and cursor paths look much better now, but community sources still need real Coral output before approval.

- Low: `sources/community/monday/manifest.yaml:453` selects `settings_str` for board columns. monday.com currently marks that field deprecated in favor of `settings`, so please either switch to the non-deprecated field if Coral can represent it cleanly, or document why `settings_str` is intentionally used for this first version. Docs: https://developer.monday.com/api-reference/reference/columns

Thanks for fixing the GraphQL request shape and variable handling. The remaining blocker is live Coral evidence from the submitted source.

jsummerfield (changes requested):
I re-reviewed the current head (`b8b8b5b`). The source passes local lint/style checks, but the approval blockers below remain.

Remaining blockers:

- `sources/community/monday/README.md` still lists validation commands rather than captured output from a live `coral source test monday` run and representative query. Approval requires sanitized live evidence from this exact manifest revision.
- `sources/community/monday/manifest.yaml` still selects monday.com's deprecated `settings_str` field for columns. Please use the current `settings` field or document why the deprecated field is necessary for this source.

jsummerfield:

jsummerfield:

PR #695, opened 2026-05-22
feat(sources/community/google_drive): add Google Drive community source

RajdeepKushwaha5:
### What changed

Adds a new community source `google_drive` that exposes Google Drive file
listings, file metadata, and shared drive enumeration via the Drive API v3.

Tables:

- `files` — paginated file list across My Drive and shared drives; supports an
  optional `q` filter for Drive query syntax and `drive_id` for shared drives
- `file_info` — full metadata for a single file (name, MIME type, size, owners,
  permissions summary, web view link)
- `shared_drives` — lists all shared drives accessible to the authenticated user

### Google Drive — Why it changed

Google Drive is the canonical document store for many engineering teams.
Exposing it as a Coral source enables queries such as "which design docs were
modified in the same week as a GitHub deploy?" or joining Drive file ownership
with Linear assignments or GitHub PR authors across a single SQL query.

### Google Drive — User-visible impact

```sql
-- Recent files
SELECT id, name, mime_type, modified_time, web_view_link
FROM google_drive.files
WHERE q = 'modifiedTime > "2026-05-15T00:00:00"'
LIMIT 20;

-- Full metadata for one file
SELECT name, size, owners, web_view_link
FROM google_drive.file_info
WHERE file_id = '<FILE_ID>'
LIMIT 1;

-- Shared drives
SELECT drive_id, name, created_time
FROM google_drive.shared_drives
LIMIT 10;
```

### Google Drive — Limitations and follow-on work

- File content download is out of scope for v1.
- The Drive query (`q` filter) uses Drive's own query syntax, not SQL.
- Per-file permissions are a natural follow-on as a `permissions` table.

### Google Drive — Authentication

Uses Google OAuth (authorization code + PKCE) with the `drive.readonly` scope,
same credential pattern and `GOOGLE_OAUTH_CLIENT_ID` /
`GOOGLE_OAUTH_CLIENT_SECRET` inputs as Gmail and Google Sheets.

---

PR #699, opened 2026-05-22
feat(sources/community/google_drive): add Google Drive source

### LINKED ISSUE #697: design: add Google Drive community source
### Summary

Add a new community `google_drive` source that lets Coral query Google Drive files and collaboration metadata through SQL using OAuth.

### Motivation

Google Drive is a widely used workspace source for documents, folders, shared drives, comments, permissions, and revision history. A Coral source would make those surfaces queryable for personal productivity agents, document review workflows, and cross-source workspace analysis.

### Proposed design

| Table | Purpose | Required filter |
|---|---|---|
| `about` | Connected user, quota, and Drive capability metadata | none |
| `files` | Paginated Drive file and folder listing | none |
| `file` | Direct metadata lookup for one file or folder | `file_id` |
| `shared_drives` | Shared drives visible to the authenticated user | none |
| `permissions` | ACL entries for one file or shared drive | `file_id` |
| `comments` | Comments for one Drive file | `file_id` |
| `revisions` | Revision history for one Drive file | `file_id` |

### Authentication

Use Coral's OAuth credential flow with Google Drive read-only access:

```text
https://www.googleapis.com/auth/drive.readonly
```

The source should also allow pasting an existing Google Drive OAuth access token through the normal source input path.

### Why this is useful

Google Drive is a strong fit for Coral because it sits at the center of many personal and team workflows. Making Drive data queryable enables workflows such as:
- finding recently modified files for a project or meeting
- auditing permissions on shared work
- inspecting document comments during review
- checking revision history alongside other operational context

This also gives Coral a useful OAuth-backed Google Workspace source beyond API-token-based community integrations.

### Scope notes
The first version should stay read-only and focused on file metadata and collaboration surfaces.

File export/download and mutation endpoints are deferred from v1. Permission enumeration may still depend on the authenticated user's access to sharing details for a given file.

### Alternatives considered

_No response_

### Open questions

_No response_

### Scope check

siiddhantt:
Closes #697

## Summary

Adds a Google Drive community source with OAuth setup for querying files, shared drives, permissions, comments, and revisions.

## Validation

- **Manifest checks** : Ran `cargo run -p coral-cli -- source lint sources/community/google_drive/manifest.yaml` and `ryl sources/community/google_drive/manifest.yaml`.
- **OAuth flow** : Completed interactive Google OAuth setup with `coral source add --interactive --file sources/community/google_drive/manifest.yaml`.
- **Smoke test** : Ran `coral source test google_drive` successfully with the declared files query passing. The declared `google_drive` smoke query now also uses the safer reviewed path: `q = 'trashed = false'` plus provider-side `order_by = 'modifiedTime desc'`.
- Queried `files` with `q = 'trashed = false'` and `order_by = 'modifiedTime desc'` to confirm provider-side filtering/sorting.
- Queried folder search using Drive `q` syntax.
- Queried 105 file rows to verify pagination crosses the `files.list` 100-row page boundary.
- Queried `about` for connected account/quota metadata.
- Queried direct `file` metadata by known file ID.
- Queried `shared_drives`; endpoint returned successfully for the account, with zero rows.
- Queried `permissions` for a known file and verified owner ACL data.
- Queried `comments` with `include_deleted` and `start_modified_time`.
- Queried `revisions` for a known Google Doc and verified revision rows.

image

## Notes

`google_drive.permissions` can return a provider 403 for readable files whose sharing details the authenticated user cannot enumerate; this is documented in the source guide.


jsummerfield:
Thanks for the Google Drive source. I found a few issues we should fix before landing:

1. This conflicts with open PR #695, which also adds `sources/community/google_drive/README.md` and `sources/community/google_drive/manifest.yaml`. Please coordinate one canonical Google Drive source PR or consolidate the best parts into a single PR so we do not have competing community sources for the same product.

2. The README’s “Find recently modified files” example uses only SQL `ORDER BY modified_time DESC`, but the manifest exposes Drive’s server-side ordering through the `order_by` filter. Because Coral may only fetch the first bounded page before sorting locally, this example can return the newest files among the fetched page rather than the newest files in Drive. Please use `WHERE order_by = 'modifiedTime desc'`.

3. The manifest and README say `files.list` can use page size 1000, but Google Drive v3 caps `files.list` `pageSize` at 100: https://developers.google.com/workspace/drive/api/reference/rest/v3/files/list#query-parameters. Please set `max: 100` and update the README.

4. The first file-listing examples should either include `q = 'trashed = false'` or explicitly say Drive returns trashed files by default. Google documents that behavior on `files.list`: https://developers.google.com/workspace/drive/api/reference/rest/v3/files/list.

siiddhantt:
Thanks for the review @jsummerfield , fixed in f2cb4dcf. 

I've set `files.list` `page_size.max` to `100`, updated the first file-listing examples to exclude trashed files with `q = 'trashed = false'`, changed the recent-files example to use server-side `order_by = 'modifiedTime desc'`, fixed the remaining bare manifest path, and added offline OAuth params to match the refresh-token note.

I also compared #695; this PR now covers the same core surface (`files`, single-file metadata, `shared_drives`) plus `about`, `permissions`, `comments`, and `revisions`. Happy to use this as the canonical Drive PR or coordinate further if maintainers prefer a different consolidation path.

jsummerfield:

- Keep Drive `q` and `order_by` examples explicit so users avoid local-only sorting and trashed file surprises.
- Preserve direct lookup ergonomics for a known file ID alongside broad `files` search.
- Keep shared drive coverage and permission/comment/revision tables clearly oriented around the IDs users need to join through.
- Document OAuth scope/credential setup and the first-success path with `coral source test google_drive`.
- Call out provider-side permission 403s for ACL/comment/revision surfaces where Drive permits file reads but not metadata enumeration.

jsummerfield (changes requested):
Thanks for the Google Drive updates. The earlier concrete issues look fixed and the manifest lints cleanly, but I found two remaining source-usability gaps we should handle before merge:

1. `sources/community/google_drive/manifest.yaml:257` exposes the `corpora` filter and the request asks Drive for `incompleteSearch` at `sources/community/google_drive/manifest.yaml:272`, but the source does not surface or document that flag. Google says `incompleteSearch=true` means some search results may be missing, especially across broad corpora such as `allDrives`, and clients should narrow to `user` or `drive`. Please either expose this signal if the source-spec can model it, or document the limitation clearly in the `files` guide/README so broad scans do not silently look complete. Docs: https://developers.google.com/workspace/drive/api/reference/rest/v3/files/list

2. `sources/community/google_drive/README.md:115` and `:159` describe revision history without warning that `revisions.list` may omit older revisions for files with large histories. That matters for audit use cases. Please document Google's limitation near the revisions table and example. Docs: https://developers.google.com/workspace/drive/api/reference/rest/v3/revisions/list

The previous duplicate/source-shape concerns are resolved, and the PR body has useful live OAuth/query evidence. Once these limitations are documented, I expect this to be close.

siiddhantt:
Thanks for the review, addressed these remaining Drive usability/docs points.

1. The current DSL does not expose a parent/root response context for row columns, so surfacing it directly on each file row is not supported without an awkward duplicate metadata table. I documented the limitation instead and added guidance to narrow broad scans to `corpora = 'user'` or `corpora = 'drive'` with `drive_id`.

2. Also documented Google's `revisions.list` limitation that older revisions may be omitted for files with large revision histories.

I also added the first-success verification path with: `coral source test google_drive` in the README.

jsummerfield (approved):
Findings: none.

Thanks for the follow-up here. The Drive `incompleteSearch` behavior and revision-history limitations are now documented in both the manifest guide and README, and the source still has a clear first-success path with live validation evidence. Checks are green on the current head, so this is approved from the community-source review side.

PR #707, opened 2026-05-22
feat(sources/community/salesforce): add Salesforce community source

### LINKED ISSUE #706: design: add Salesforce community source
### Summary
Add a community source `salesforce` that exposes Salesforce CRM data as SQL tables via the Salesforce REST API. The instance URL varies per org so `SALESFORCE_API_URL` will be a variable input following the same pattern as the Grafana source.

### Motivation

Salesforce is the most widely used CRM and sits at the center of most sales and support workflows. Without this source, teams have no way to join Salesforce data with GitHub deploys, Linear issues, or PagerDuty incidents in a single query. You either export CSVs manually or build a pipeline just to answer basic questions like "which opportunities are owned by engineers who also have open incidents."

### Proposed design

New community source at `sources/community/salesforce/manifest.yaml` with two inputs:

`SALESFORCE_API_URL` — variable input, the org instance URL (e.g. `https://mycompany.my.salesforce.com`)
`SALESFORCE_ACCESS_TOKEN` — secret input, connected app access token
`base_url: "{{input.SALESFORCE_API_URL}}"` following the Grafana pattern.

Tables via the Salesforce REST API `/services/data/v59.0/sobjects/` endpoints:

- `accounts` — company accounts with industry, revenue, and owner
- `contacts` — contacts with email, phone, and account association
- `leads` — leads with status, source, and owner
- `opportunities` — deals with stage, amount, close date, and account
- `cases` — support cases with status, priority, and contact

### Drawbacks and trade-offs
Access tokens expire; users will need to refresh manually. OAuth flow support would remove this friction but is out of scope for v1. The instance URL requirement adds one extra setup step compared to sources with a fixed base URL.

### Alternatives considered
Using the Salesforce Bulk API for large exports — skipped for v1 since the REST API covers the common read queries and keeps the implementation simple. SOQL passthrough as a filter was considered but left as a follow-on to keep scope narrow.

### Open questions
Should v1 support a soql filter for passthrough queries, or keep it table-only?
Is v59.0 the right API version to pin, or should it be configurable?

### Scope check

RajdeepKushwaha5:
### What changed

Adds the community source `salesforce`, exposing Salesforce CRM data through REST API v59.0.

| Table | SOQL pushdown filters | Description |
| --- | --- | --- |
| `accounts` | `industry` | Companies with type, revenue, billing location, and owner |
| `contacts` | `account_id` | Individuals with email, phone, title, and account link |
| `leads` | `is_converted`, `lead_source` | Prospects with status, source, conversion state, and converted-resource IDs |
| `opportunities` | `is_closed`, `account_id` | Pipeline deals with numeric amount/probability and win/loss state |
| `cases` | `is_closed`, `priority`, `account_id` | Support tickets with account/contact links; all filter combinations push down |

Auth uses `SALESFORCE_API_URL` plus secret `SALESFORCE_ACCESS_TOKEN`. The README and manifest document required API, object, and field permissions and distinguish permission failures from rate-limit failures.

### Why

This source makes CRM pipeline, contact, lead-conversion, and support-case data queryable with SQL and joinable with compatible Coral sources. Numeric CRM fields are exposed as `Float64`, and stable Salesforce IDs are included for reliable joins.

### Review follow-up

- Added sanitized successful `source add`, `source test`, and query output to the README.
- Added converted lead join keys.
- Added provider-side SOQL pushdown routes for the commonly filtered fields, including every supported combination for multi-filter tables.
- Documented the token permissions needed for first success.
- Documented daily API limits, `/limits`, and `Sforce-Limit-Info`.
- Do not treat generic `403` as retryable because Salesforce also uses it for permission failures.
- Removed unsupported GitHub, PagerDuty, and Sentry join claims/examples.

### Validation

- `coral source lint sources/community/salesforce/manifest.yaml`
- `ryl -c .yamllint.yaml` on the LF-normalized manifest
- `git diff --check`
- clean merge simulation against `upstream/main`
- sanitized live Salesforce validation output is recorded in the README

### Limitations

- Salesforce returns additional Query API pages through a path-based `nextRecordsUrl`, which Coral's current HTTP source DSL cannot follow. Each request is therefore capped at 2,000 rows; documented pushdown filters should be used so that cap applies to the matching result set.
- Other SQL `WHERE` conditions run client-side after fetching.
- Access tokens must be generated and refreshed externally; there is no interactive OAuth flow in this source.

Closes #706


jsummerfield:
Thanks for adding this source. I think two things need fixing before this should merge.

First, the tables currently hard-code SOQL `LIMIT 2000`, set `pagination.mode: none`, and expose no Salesforce-side filters. That means Coral applies `WHERE` clauses after fetching only the latest 2000 rows by `LastModifiedDate`, so queries like “open opportunities” or “high-priority open cases” can silently miss matching records in larger orgs. Salesforce’s REST Query API supports retrieving more than 2,000 records via `nextRecordsUrl` / query locators, so we either need real pagination or provider-side filter/table-function entry points. The README should also stop saying SQL filters narrow the Salesforce result set unless those filters are actually pushed into SOQL.

Second, a couple of cross-source examples don’t match the current Coral sources. The GitHub example uses `github.pull_requests` and `pr.author_email`, but the current table is `github.pulls` and there is no `author_email` column. The PagerDuty example uses `pd.user_email`, which is not a column on `pagerduty.incidents`. Please update or remove those examples so users can paste them successfully.

RajdeepKushwaha5:
Hey @jsummerfield, pushed fixes for both issues.

On the pagination/filtering point - updated all five table guides to be upfront that SQL WHERE runs after fetching the 2000-row result, so filters like `is_closed = false` can miss rows in large orgs. Removed the "use SQL filters to narrow the set" line from the README and replaced it with a clear note. I agree it's better to say this plainly than leave it implicit.

On real pagination: Salesforce's `nextRecordsUrl` changes the full request path (e.g. `/services/data/v59.0/query/01g...`) rather than a query param, so `cursor_query` mode doesn't cover it. Happy to look into this more if there's a DSL path for it, or I can open a follow-on issue to track it as a known v1 limitation.

Cross-source examples - replaced the GitHub one (was referencing `github.pull_requests` and `pr.author_email`, neither of which exist) with a `linear.issues` join on `assignee_email`. Dropped the PagerDuty example since `pagerduty.incidents` doesn't have a user email column to join on.

jsummerfield (changes requested):
Findings:

- High: live-test evidence is still missing. The README lists commands to run (`sources/community/salesforce/README.md:41-47`), but neither the README nor the PR body includes captured successful `coral source add`, `coral source test salesforce`, `coral query`, a screenshot/video, or linked logs. Community sources need sanitized live Coral output before approval.

- Medium: CRM numeric fields are exposed as strings: `annual_revenue` (`sources/community/salesforce/manifest.yaml:108-115`), `amount` (`:545-552`), and `probability` (`:577-584`). These are Salesforce numeric/currency/percent fields, and users need numeric ordering, aggregation, and comparisons for revenue and pipeline workflows. Please use `Float64` where Coral has no decimal type, or expose raw plus numeric columns. Docs: https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_account.htm and https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_opportunity.htm

- Medium: the auth docs are still thin for first success (`sources/community/salesforce/README.md:21-39`, `sources/community/salesforce/manifest.yaml:21-27`). Please document that the token’s Salesforce user needs API access plus read/object/field visibility for Account, Contact, Lead, Opportunity, and Case fields, and what a permissions failure looks like. Docs: https://help.salesforce.com/s/articleView?id=permissions_about_users_access.htm and https://help.salesforce.com/s/articleView?id=platform.users_profiles_object_perms.htm

- Medium: rate-limit docs are incomplete. Salesforce can enforce API limits with `403 REQUEST_LIMIT_EXCEEDED`, and users should know about daily org limits, `/limits`, and the `Sforce-Limit-Info` header. Please document those; only add generic `403` rate-limit handling if Coral can distinguish it from permission failures. Docs: https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_limits.htm and https://developer.salesforce.com/blogs/2024/11/api-limits-and-monitoring-your-api-usage

The prior pagination/filtering concern remains unresolved: the source still hard-codes `LIMIT 2000` with no query-locator pagination, so large org queries can silently miss rows. I am not repeating that in detail here since it was already posted above; it still needs a real fix, not only documentation.

jsummerfield:

RajdeepKushwaha5:
hi @jsummerfield !


All findings are addressed in the latest push:

**High - live-test evidence**: Added sanitized output from `coral source add`, `coral source test salesforce`, and a `coral sql` query against `salesforce.accounts` to the Validation section in the README.

**Medium - numeric fields as strings**: Changed `annual_revenue`, `amount`, and `probability` from `Utf8` to `Float64`. These map to Salesforce's currency and percent field types and need to be numeric for ordering, aggregation, and comparisons.

**Medium - thin auth docs**: Added a `### Required permissions` section to the README documenting that the token's user needs API Access enabled, Read object permission on Account/Contact/Lead/Opportunity/Case, and field-level visibility. Documents what 401 and 403 errors indicate. Added the same detail to the `SALESFORCE_ACCESS_TOKEN` manifest hint.

**Medium - rate-limit docs**: Added a `## Rate limits` section documenting daily org limits, the `Sforce-Limit-Info` response header, `REQUEST_LIMIT_EXCEEDED`, and the `/limits` endpoint. Added `403` to `rate_limit.extra_statuses` in the manifest.

**Pagination/filtering concern**: Added SOQL pushdown filter variants (`requests:` blocks) for the key Boolean and string filters on all five tables. When users provide these filters in SQL, Coral pushes them into SOQL so the 2000-row cap applies to the filtered result, not the full object:

- `accounts`: `industry`
- `contacts`: `account_id`
- `leads`: `is_converted`, `lead_source`
- `opportunities`: `is_closed`, `account_id`
- `cases`: `is_closed`, `priority`, `account_id`

For example, `WHERE is_closed = false` on `opportunities` now generates `... FROM Opportunity WHERE IsClosed = false ORDER BY LastModifiedDate DESC LIMIT 2000` in SOQL rather than fetching all 2000 rows and filtering client-side. Coral's DSL does not currently support Salesforce's path-based `nextRecordsUrl` cursor, so the 2000-row cap per query remains for unfiltered access. This is documented clearly in the updated guide text on each table and in the README Notes section.

Let me know if anything needs another look.

jsummerfield (changes requested):
Findings:

- High: string filters are interpolated directly into SOQL without escaping. For example, `sources/community/salesforce/manifest.yaml:91` renders `WHERE Industry = '{{filter.industry}}'`; the same pattern appears for `lead_source` around `sources/community/salesforce/manifest.yaml:400`, `priority` around `sources/community/salesforce/manifest.yaml:816`, and ID filters throughout the manifest. Salesforce requires quoted string reserved characters to be escaped, and Coral template rendering will pass filter values literally. A value containing `'` or `\` can make the SOQL invalid or change the pushed-down predicate, and Coral will not have a residual exact filter to correct the result. Please either remove unsafe string pushdown, add a supported escaping path, or restrict/filter-validate these values so the SOQL remains exact. Docs: https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_select_quotedstringescapes.htm

The previous blockers are mostly addressed: live evidence is now present, the numeric CRM fields are `Float64`, the common filters have pushdown routes, and the permission/rate-limit docs are much better. I do not think Cubic's 403 retry suggestion is itself a blocker; Salesforce uses 403 for both permission errors and request-limit errors, and the README's manual guidance around `Sforce-Limit-Info` is reasonable.

jsummerfield (changes requested):
I re-reviewed the current head (`4504be8`). The source now passes local lint/style checks, but the SOQL escaping blocker is still open.

Remaining blocker:

- `sources/community/salesforce/manifest.yaml` still interpolates string filters directly into SOQL, including the account/opportunity/contact-style filter sites around the existing SOQL templates. A value containing a quote can break the generated query or change its semantics. Please use a safe escaping/binding approach for all string and ID filters that are inserted into SOQL.

jsummerfield:

jsummerfield:

PR #714, opened 2026-05-23
feat: Explain/Plan Introspection for SQL Execution

### LINKED ISSUE #671: feat: Explain/Plan Introspection for SQL Execution
### What problem does this solve?

Introduce EXPLAIN (or coral explain "<SQL>") that shows the engine’s execution plan:

which sources/tables/functions will be read
whether pushdown is applied
join strategy (local join behavior)
estimated/real row counts (where available)
which parts are cached vs fetched

### Proposed change

SQL surface
If the engine already parses SQL, add a handler for EXPLAIN <query>.
Otherwise add a CLI wrapper in crates/coral-cli/src/query_* to call a new engine method that returns an execution plan.

Plan generation
In crates/coral-engine/src/runtime/query.rs (or the appropriate composition/execution layer), create a structured ExecutionPlan model containing:
steps[]: source scans, filters, joins, projections
pushdowns[]: predicates sent to sources
cache[]: cache lookup and reuse decisions
estimated_rows / actual_rows (actual rows if you run the query; estimated for static planning)
Output formats
CLI: human-readable tree/table.
MCP/UI: optionally return JSON for tooling/agent consumption.

### Alternatives considered

_No response_

### Which area does this affect?

Other

### Contribution

- [x] I'd be willing to open a PR for this if the direction is agreed.

Aditya8369:
Implemented `coral explain` end to end. The query proto now carries a structured execution-plan payload, the engine builds it from the physical plan tree, and the CLI can render it as a human-readable tree/table or JSON. The main touch points are query.proto, query.rs, lib.rs, and the new docs in cli-reference.mdx.

I also updated the CLI and engine tests to cover the new explain path, including request capture and structured-plan assertions. Estimated row counts are populated where the engine can derive them; actual row counts remain empty for static `explain` output.

Validation is partially complete: file-level Rust checks passed on all touched files, but I could not run `cargo check` in this environment because `cargo` is not installed or available on `PATH` here. 

closes #671

### DIFF
--- a/crates/coral-api/proto/coral/v1/query.proto
+++ b/crates/coral-api/proto/coral/v1/query.proto
@@ -37,6 +37,41 @@ message QueryPlan {
   string optimized_logical_plan = 2;
   // Physical execution plan after physical optimization.
   string physical_plan = 3;
+  // Structured execution-plan metadata for tooling and UI rendering.
+  ExecutionPlan execution_plan = 4;
+}
+
+message ExecutionPlan {
+  repeated ExecutionPlanStep steps = 1;
+  repeated PushdownDecision pushdowns = 2;
+  repeated CacheDecision cache = 3;
+  optional uint64 estimated_rows = 4;
+  optional uint64 actual_rows = 5;
+}
+
+message ExecutionPlanStep {
+  string kind = 1;
+  string name = 2;
+  string detail = 3;
+  optional uint64 estimated_rows = 4;
+  optional uint64 actual_rows = 5;
+  repeated ExecutionPlanStep children = 6;
+}
+
+message PushdownDecision {
+  repeated uint32 step_path = 1;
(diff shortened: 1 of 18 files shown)


yunus25jmi1:
Hey @Aditya8369 — great work on this; the structured execution plan and `build_execution_plan()` implementation are solid.

I noticed the MCP tool (`explain_sql` in `coral-mcp/src/server.rs`) and MCP docs are still missing per issue #671. I have already mapped out the implementation following the existing `sql_tool()` pattern at `surface/tools.rs:46–68`.

Two options:
1. I can open a follow-up PR that adds just the MCP surface on top of your work (after this merges)
2. Or if you are okay with it, I can push the MCP commits directly to your branch now so everything lands together

Let me know which you prefer – happy to collaborate either way.

PR #721, opened 2026-05-23
feat(sources/community/google_forms): add Google Forms community source

rohan-jadhav-dev:
## Summary
Adds a new community source for Google Forms using the Google Forms REST API v1.

## Tables
- `form` — form title, description, and metadata (requires `form_id`)
- `questions` — question items in the form (requires `form_id`)
- `responses` — submitted responses with pagination support (requires `form_id`, optional `after_timestamp`)

## Authentication
OAuth 2.0 Bearer Token via HeaderAuth with `forms.body.readonly` and `forms.responses.readonly` scopes.

## Validation
- Manifest passes `coral source lint`
- Follows Cal.com community source patterns for auth, pagination, and query params
- Pagination implemented via `cursor_query` using `pageToken` / `nextPageToken`

## Example Queries
```sql
SELECT form_id, title, description
FROM google_forms.form
WHERE form_id = 'your-form-id';

SELECT item_id, title, required
FROM google_forms.questions
WHERE form_id = 'your-form-id';

SELECT response_id, create_time, answers
FROM google_forms.responses
WHERE form_id = 'your-form-id';
```

## Notes
- Read-only source
- Follows existing community source patterns


jsummerfield (changes requested):
Thanks for the Google Forms source. I found a few blockers we should address before this lands:

1. `sources/community/google_forms/manifest.yaml:118` and the README examples treat `item_id` as the key users can match against `responses.answers`, but Google Forms responses are keyed by `questionId`, not the wrapper `itemId`. Please expose the question ID from `questionItem.question.questionId` and update the join guidance. Docs: https://developers.google.com/workspace/forms/api/reference/rest/v1/forms and https://developers.google.com/workspace/forms/api/reference/rest/v1/forms.responses

2. `sources/community/google_forms/README.md:86` and `:105` show an `after_timestamp` example using `timestamp=...`, but `forms.responses.list` only supports filters like `timestamp > ...` or `timestamp >= ...`. Please update the examples and guidance so users can copy/paste a valid provider filter. Docs: https://developers.google.com/workspace/forms/api/reference/rest/v1/forms.responses/list

3. `sources/community/google_forms/manifest.yaml:108` maps `questions` over raw `items[]`. Forms items can be non-question content, and grid questions live under `questionGroupItem.questions[]`, so this table can return non-questions and miss individual grid question IDs. Please make the table semantics match the README, or narrow/document the supported item shapes. Docs: https://developers.google.com/workspace/forms/api/reference/rest/v1/forms

4. `sources/community/google_forms/manifest.yaml:182` sets `allow_404_empty: true` on responses. A form with no responses should return an empty response list; a 404 is more likely a bad form ID, missing permission, or inaccessible form. Please let that surface as an error so setup mistakes do not look like empty data.

5. I do not see live-test evidence yet. Please add a PR description or README section with successful `coral source add`, `coral source test`, or `coral query` output against a real Google Form.

Thanks again for the contribution. Once these are fixed, this should be much easier for new users to get right on the first try.

rohan-jadhav-dev:
Hi @jsummerfield, I've addressed all 5 points from your review:

1. ✅ Added `question_id` column mapped to `questionItem.question.questionId`
2. ✅ Fixed `after_timestamp` syntax to use `timestamp >= <ISO8601>`
3. ✅ Added `row_filter` on `questionItem` to exclude non-question items (page breaks, images, videos)
4. ✅ Set `allow_404_empty: false` on responses table
5. ✅ Added Live Test Output section to README

Please re-review when you get a chance, thanks!

jsummerfield (changes requested):
Findings:

- High: `sources/community/google_forms/manifest.yaml:114` uses `response.row_filter`, but that key is not part of the current Coral source-spec schema. I verified `coral source lint sources/community/google_forms/manifest.yaml` fails with `Additional properties are not allowed ('row_filter' was unexpected)`, so users cannot add or test this source at the current head. Please remove this unsupported key and model the filtering with supported source-spec behavior, or expose the items table in a way that keeps non-question rows safe for users to filter explicitly.

Thanks for continuing to iterate on the Google Forms source. Once the manifest validates again, I can re-check the Forms API semantics and examples.

rohan-jadhav-dev:
Hi @jsummerfield, removed the unsupported `row_filter` key — `coral source lint` should pass now. The `questions` table now returns all form items. Updated guide, column descriptions, test_queries, and all README examples to use `WHERE question_id IS NOT NULL` to filter out non-question rows (page breaks, images, videos).

jsummerfield (changes requested):
Findings:

- High: `sources/community/google_forms/manifest.yaml:25`-`:28` still declares `test_queries` with the placeholder value `'your-form-id'`. `coral source test google_forms` runs these queries as written, so the source will fail its own test path for normal users unless they happen to have a form with that literal ID. Please remove these manifest test queries, or replace them with a setup path that uses a real configured form ID rather than placeholder values.

- High: `sources/community/google_forms/README.md:134`-`:145` is titled “Live Test Output”, but it only lists commands with `your-real-form-id`; it does not include captured successful `coral source add`, `coral source test`, or representative `coral query` output from a real Google Form. Live Coral evidence is still required before approval. Please add sanitized output in the README or PR description.

- Medium: `sources/community/google_forms/manifest.yaml:231` and `:240` expose `create_time` and `last_submitted_time` as `Utf8`, but Google Forms response timestamps are RFC3339 timestamp strings. Please expose these as `Timestamp` columns so users can sort and compare them reliably. Docs: https://developers.google.com/workspace/forms/api/reference/rest/v1/forms.responses

- Medium: `sources/community/google_forms/manifest.yaml:255` exposes `answers` as `Utf8`, while the README describes it as a raw JSON object and Google documents it as a map keyed by `questionId`. Please make this a `Json` column so the catalog and query behavior match the provider shape. Docs: https://developers.google.com/workspace/forms/api/reference/rest/v1/forms.responses

The previous `row_filter` blocker is resolved and the manifest now lints. The `question_id` join guidance, `after_timestamp` provider filter syntax, non-question/grid-item limitations, and `allow_404_empty: false` behavior also look much better. Thanks for continuing to iterate; the remaining items are about making the test path and response schema safe for users.

rohan-jadhav-dev:
Hi @jsummerfield, addressed all 4 findings:

1. ✅ Removed test_queries with placeholder form IDs
2. ✅ Added sanitized example output to Live Test section
3. ✅ Fixed create_time and last_submitted_time to Timestamp type
4. ✅ Fixed answers to JSON type with description matching API docs

Please re-review when you get a chance, thanks!

jsummerfield (changes requested):
Findings:

- High: `sources/community/google_forms/manifest.yaml:25`-`:27` is still invalid YAML. The old `SELECT` entries are now indented under `auth` because the `test_queries:` key is missing, and `coral source lint` fails at line 25. Those entries also still use `'your-form-id'`, so even after the YAML parse fix they would not be usable manifest tests for normal users.

- High: `sources/community/google_forms/manifest.yaml:230`, `:237`, and `:252` contain literal `` `n        type: ...`` text inside the column names for `create_time`, `last_submitted_time`, and `answers`. The intended Timestamp/Json fixes are not actually applied, and the submitted manifest cannot expose those columns correctly.

- High: `sources/community/google_forms/README.md:134`-`:145` still does not provide live Coral output. It shows commands with `your-real-form-id`, not captured results from `coral source add`, `coral source test`, or representative queries. The commands also use `coral query`; the current CLI command is `coral sql`.

- Medium: `sources/community/google_forms/manifest.yaml:183` defines a paginated `responses` table without a conservative `fetch_limit_default`. Google Forms `forms.responses.list` supports `pageSize` up to 5000 and returns `nextPageToken`, so this table should set a safe default for first-use queries rather than letting a broad responses scan pull a very large page. Docs: https://developers.google.com/workspace/forms/api/reference/rest/v1/forms.responses/list

- Low: `git diff --check` reports a blank line at EOF in `sources/community/google_forms/manifest.yaml:270`.

Thanks for the quick follow-up. The next pass should first make `coral source lint` pass locally, then add captured output from real Google Forms queries so we can validate the submitted source.

jsummerfield:

PR #710, opened 2026-05-22
feat(sources/community/gmail): add Gmail community source

Harsh-bugs4ever:
I've added a Gmail source spec that lets you query your Gmail 
inbox using SQL through Coral.

With this we can do things like:

SELECT id, thread_id FROM gmail.messages WHERE label_ids = 'INBOX' LIMIT 20
SELECT id, name, type FROM gmail.labels
SELECT email_address, messages_total FROM gmail.profile

What's included
- profile — see your email address and mailbox stats
- messages — list and filter your emails by label or search query
- threads — list email threads
- labels — all your Gmail labels including INBOX, SENT, SPAM etc
- drafts — list your saved drafts

Auth
Uses a Google OAuth2 access token with gmail.readonly scope.
Simple to set up via Google OAuth Playground.

Happy to make any changes if needed. Thanks for the awesome project and opportunity kunal sir 
would love to see any future oppurtunities

PR #723, opened 2026-05-23
Add google forms community source

Harsh-bugs4ever:
Following up on my Gmail source PR — I've added a Google Forms 
source spec that lets you query your Google Forms data using SQL.

With this you can do things like:

SELECT form_id, title, description FROM google_forms.forms WHERE form_id = 'YOUR_FORM_ID'
SELECT response_id, respondent_email, last_submitted_time FROM google_forms.responses WHERE form_id = 'YOUR_FORM_ID' LIMIT 20

## What's included
- forms — get form metadata like title, description, and responder URL
- responses — list all responses submitted to a form with respondent email and answers

## Auth
Uses a Google OAuth2 access token with two scopes:
- forms.body.readonly
- forms.responses.readonly

Easy to generate via Google OAuth Playground.

Happy to make changes if needed. Thanks!

### DIFF
--- a/sources/community/gmail/manifest.yaml
+++ b/sources/community/gmail/manifest.yaml
@@ -1,0 +1,320 @@
+name: gmail
+version: 0.1.0
+dsl_version: 3
+backend: http
+description: >-
+  Query Gmail messages, threads, labels, drafts, and profile
+  information from your Google Gmail account using the Gmail REST API.
+inputs:
+  GMAIL_ACCESS_TOKEN:
+    kind: secret
+    hint: >-
+      Google OAuth2 access token with Gmail read-only scope.
+      Generate one at https://developers.google.com/oauthplayground
+      using the scope https://www.googleapis.com/auth/gmail.readonly
+base_url: "https://gmail.googleapis.com"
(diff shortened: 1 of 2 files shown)


jsummerfield (changes requested):
Findings:

- High: this PR currently changes two separate community sources: `sources/community/gmail/manifest.yaml` and `sources/community/google_forms/manifest.yaml`. Community source PRs need to stay scoped to one provider/product so review, ownership, and merge decisions stay clear. Please do not keep Gmail and Google Forms in the same PR.

- High: Google Forms is already in flight in #721, which is the PR we are using for the Google Forms source. Please do not scope this PR down into a competing Google Forms PR. If you want to help with Google Forms, the best path is to join the discussion on #721 and contribute product-level feedback or follow-up fixes there.

- Medium: Gmail work should stay separate as well. Since #710 already exists for your Gmail source work, please keep any Gmail-specific updates in that PR rather than combining them here.

Thanks for the contribution. The next step for this PR is to close or repurpose it so it is not a mixed-source / duplicate Google Forms submission.

PR #761, opened 2026-05-25
feat(app): add keychain secret storage

jsummerfield:
## Intent

Store source secrets in the OS keychain by default when a usable native backend is available, while preserving plaintext file storage as the fallback for systems without keychain support.

The goal is to keep secret material behind a storage abstraction so source lifecycle, query execution, and CLI surfaces do not need to know backend-specific details.

## Changes

- Add credential storage routing with `auto`, `keychain`, and `file` preferences.
- Persist each installed source's credential storage route in source metadata.
- Add file and keychain credential backends behind the app credential store.
- Fail closed when an installed keychain-backed source cannot read its secrets.
- Surface credential storage in source list/info/add output and API responses.
- Cover source lifecycle, query loading, config preference, and backend behavior with tests.

## Validation

- `make rust-checks`
- Linux VM smoke validation: default `auto` falls back to file when no Secret Service backend is available; explicit `keychain` reports the unavailable backend clearly.
- Will test on Windows today.

### DIFF
--- a/crates/coral-app/src/credentials/store.rs
+++ b/crates/coral-app/src/credentials/store.rs
@@ -304,9 +304,8 @@ fn keychain_route_unavailable(
 fn configured_keychain_unavailable(error: CredentialsError) -> CredentialsError {
     match error {
         CredentialsError::Unavailable(detail) => CredentialsError::Unavailable(format!(
-            "keychain credential storage is configured for new source secrets, \
-             but keychain is unavailable: {detail}. Set [credentials] storage = \"file\" \
-             to use plaintext file storage."
+            "keychain credential storage is configured, but keychain is unavailable: \
+             {detail}. Set [credentials] storage = \"file\" to use plaintext file storage."
         )),
         error => error,
     }
--- a/crates/coral-app/src/credentials/store.rs
+++ b/crates/coral-app/src/credentials/store.rs
@@ -5,6 +5,8 @@ use std::io;
 use std::path::Path;
 use std::sync::{Arc, OnceLock};
 
+use sha2::{Digest as _, Sha256};
+
 use crate::bootstrap::AppError;
 use crate::state::AppStateLayout;
 use crate::storage::fs as storage_fs;
@@ -43,6 +45,31 @@ impl EncodedCredentialMaterial {
 struct CredentialSetRef<'a> {
     workspace_name: &'a WorkspaceName,
     credential_set_id: &'a CredentialSetId,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct CredentialConfigNamespace(String);
+
+impl CredentialConfigNamespace {
+    fn from_layout(layout: &AppStateLayout) -> Self {
+        Self::from_config_dir(layout.config_dir())
+    }
+
+    fn from_config_dir(config_dir: &Path) -> Self {
+        let canonical = config_dir
+            .canonicalize()
+            .unwrap_or_else(|_| config_dir.to_path_buf())
+            .to_string_lossy()
+            .into_owned();
+        let digest = Sha256::digest(canonical.as_bytes());
+        let hex = format!("{digest:x}");
+        let short = hex.get(..16).unwrap_or(hex.as_str());
+        Self(format!("config-{short}"))
+    }
+
+    fn as_str(&self) -> &str {
+        self.0.as_str()
+    }
 }
 
 trait CredentialMaterialBackend: Send + Sync {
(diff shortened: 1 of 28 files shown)


mystic123 (approved):
Should we also add 1Password as cross-platform alternative?

docs/project/security.mdx
```diff
@@ -47,6 +54,28 @@ trust boundary:
...
+| Value | Behavior |
+| --- | --- |
+| `auto` | Default. Use keychain when the probe succeeds; otherwise use plaintext file storage. |
+| `keychain` | Require OS credential storage. If unavailable, source installation fails. |
+| `file` | Store source secrets in plaintext `secrets.env` files under Coral's config directory. |
+
+Each installed source records its own credential storage route. Changing the
+global preference does not rewrite routes that already exist. Existing plaintext
+sources remain file-backed; remove and re-add a source if you want it to use
+the current default backend.
```
AlbertQM: optional nit

```suggestion
<Note>
Changing the global preference does not rewrite existing credentials. To update those, remove and re-add the source you want to update.
</Note>
```

docs/reference/cli-reference.mdx
```diff
@@ -80,6 +83,13 @@ coral source add --interactive github
 
 If the source is already installed, running this command again replaces its stored credentials with the new values you provide. After installation, it prints the discovered tables and runs any optional top-level `test_queries` declared in the source spec to validate the connection. Any post-install validation issue is reported as a warning here so the source remains installed and re-runnable.
 
+Sources use OS credential storage by default when Coral can write, read, and
+delete a keychain probe item. If the keychain is unavailable in `auto` mode,
+Coral stores the source secrets in plaintext file storage and shows
+`file (plaintext)` in source output. Existing file-backed sources stay
+file-backed; remove and re-add a source to move it to the current default
+backend.
```
AlbertQM: optional nit: I'd either add the same note as security mdx, or remove this altogether. 

```suggestion
`file (plaintext)` in source output.
```

AlbertQM (approved):

crates/coral-app/src/credentials/store.rs
```diff
@@ -71,66 +212,611 @@ impl CredentialStore {
...
+#[derive(Clone)]
+struct KeychainEntryAddress {
+    service: String,
+    account: String,
+}
+
+impl KeychainEntryAddress {
+    fn from_set(set: &CredentialSetRef<'_>) -> Self {
+        Self {
+            service: format!("com.withcoral.coral/{}", set.workspace_name.as_str()),
```
sauldhernandez: This keychain address needs to include a per-config namespace. Right now the keychain item is scoped only by workspace + credential set, so two separate Coral configs using the same workspace/source, e.g. `default` + `github`, write the same OS keychain entry. That means a dev/test `CORAL_CONFIG_DIR` can overwrite or delete the secret used by a prod config, and prod queries can silently run with the wrong token.

File storage is scoped by the config directory automatically; keychain storage needs an equivalent namespace, such as a persisted config/profile ID or a stable canonical config-dir hash.

crates/coral-app/src/sources/manager.rs
```diff
@@ -221,11 +231,13 @@ impl SourceManager {
     ) -> Result<InstalledSource, AppError> {
         let bundled = load_bundled_source(&command.name)?;
         let candidate = self.describe_bundled_source(workspace_name, &bundled.manifest_yaml)?;
+        let credential_storage = self.source_write_storage(workspace_name, &candidate.name)?;
```
sauldhernandez: nit: We're picking a credential storage even if there is no need of credential storage for a given source. This was fine when it was just FS but with keychain as an option this is adding an unnecesary failure mode when keychain is not available/has issues and the source had no need for it.

sauldhernandez (approved):

crates/coral-app/src/credentials/store.rs
jsummerfield: Good shout, done.

jsummerfield:
> Should we also add 1Password as cross-platform alternative?

​Good idea for a follow up PR

simonwhitaker:
I did poke around with 1Password a while back. But there are a couple of limitations:  
  
1\. There's no Rust SDK, only Python, Go and JS, so I ended up integrating via exec'ing to the 1Password CLI, which feels janky  
2\. The 1Password API supports two authentication methods: local authorisation prompts via the 1Password desktop app, or service accounts. (When I was exec'ing to the CLI, I was getting the former.) For individual users running coral locally, the desktop app route is zero setup, but you get prompted for approval (via your 1Password password or biometrics) on access, which I found quickly became really annoying.

simonwhitaker (approved):

jsummerfield:
Ack, thanks Simon

PR #764, opened 2026-05-25
feat(app): refresh oauth access tokens

sauldhernandez:
## Intent

Add transparent OAuth access-token refresh so sources with short-lived access tokens continue working without requiring users to reconnect manually.

## What it enables

- Refreshes expired or near-expired OAuth access tokens before query runtime registration.
- Persists refreshed access tokens and rotated refresh tokens back to credential material.
- Keeps OAuth refresh ownership inside the credential layer while query loading remains provider-agnostic.
- Serializes refresh per credential set so concurrent queries do not spend the same rotating refresh token.

## Usage examples

After a source is connected through OAuth, users can keep querying it normally. When Coral loads the source and finds an expired OAuth access token with refresh metadata, it refreshes and stores the updated token material before executing the query.

## Validation

- env RUSTC_WRAPPER= make rust-checks
- cargo test -p coral-app oauth_refresh
- cargo test -p coral-app import_with_oauth
- cargo test -p coral-spec oauth_client_secret_transport_labels_are_canonical
- cargo test -p coral-spec parses_confidential_oauth_client_with


AlbertQM (approved):
Docs LGTM

crates/coral-app/src/query/manager.rs
```diff
@@ -161,14 +176,18 @@ impl QueryManager {
...
         &self,
         workspace_name: &WorkspaceName,
+        credential_load_policy: CredentialLoadPolicy,
     ) -> Result<Vec<QuerySource>, AppError> {
         let catalog = self.config_store.load_catalog()?;
         let mut query_sources = Vec::new();
         for source in catalog.workspace_sources(workspace_name) {
-            match self.load_query_source(workspace_name, &source) {
+            match self
+                .load_query_source(workspace_name, &source, credential_load_policy)
```
jsummerfield: When this policy is `RefreshProvider`, `load_query_source` can now fail because the OAuth refresh failed. The `Err` arm below still treats that the same as a bad optional source and skips it, so a query against that source ends up as a missing-table style error instead of surfacing the refresh failure. Could we propagate refresh-policy load errors for query/explain, or otherwise distinguish refresh failures from passive catalog load failures?

crates/coral-app/src/credentials/oauth.rs
```diff
@@ -767,10 +880,125 @@ fn oauth_metadata_prefix(input_key: &str) -> String {
...
+    let Some(refresh_token) = material
+        .get(&format!("{metadata_prefix}refresh_token"))
+        .filter(|value| !value.is_empty())
+        .cloned()
+    else {
+        tracing::debug!(
+            source_secret = access_token_material_key,
+            "OAuth access token is expired but no refresh token is stored"
+        );
+        return Ok(None);
```
jsummerfield: At this point we know the stored OAuth access token is expired or inside the refresh skew, but returning `Ok(None)` lets callers continue with that known-expired token. That makes the eventual provider/query failure much less actionable. Could this return a failed precondition telling the user to reconnect because no refresh token is stored?

crates/coral-app/src/credentials/mod.rs
```diff
@@ -56,11 +62,30 @@ impl fmt::Display for CredentialSetId {
...
+    provider_refresh_locks: ProviderRefreshLocks,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+struct CredentialRefreshLockKey {
+    workspace_name: WorkspaceName,
+    credential_set_id: CredentialSetId,
 }
 
+type ProviderRefreshLock = Arc<tokio::sync::Mutex<()>>;
```
jsummerfield: This lock only serializes refreshes inside one `CredentialManager` instance. Since separate Coral processes can share the same config dir, two local servers can still read the same expired material and spend the same rotating refresh token before either write lands. The state file lock/update path helps persistence, but it does not cover the provider refresh call itself. I think this needs an inter-process lock around the read/refresh/write sequence, or the behavior should be narrowed so we are not claiming protection for rotating refresh tokens across processes.

crates/coral-app/src/credentials/mod.rs
```diff
@@ -82,6 +126,33 @@ impl CredentialManager {
...
+            credential_set_id,
+            inputs,
+            &mut material,
+        )
+        .await
+        .map_err(credential_refresh_error)?;
+        Ok(material)
+    }
+
     pub(crate) fn snapshot_material(
```
jsummerfield: This still looks racy after moving refresh behind a file lock. `snapshot_material` can read the old credential blob while a query holds the refresh lock and is in the middle of rotating an OAuth token. If this source persist later fails, rollback waits for the refresh lock and restores that old snapshot over the freshly persisted token, which can resurrect a refresh token the provider has already invalidated. Could we either take the same per-credential lock for the full snapshot/update/rollback transaction or make rollback conditional on the current material still matching this transaction’s own write?

crates/coral-app/src/query/manager.rs
```diff
@@ -226,11 +228,18 @@ impl QueryManager {
...
-            engine_extensions_for_providers(&self.engine_extensions_providers, selected_sources),
-        )
+    fn runtime_config(
+        &self,
+        workspace_name: &WorkspaceName,
+        selected_sources: &[QuerySource],
+    ) -> QueryRuntimeConfig {
+        let mut extensions =
+            engine_extensions_for_providers(&self.engine_extensions_providers, selected_sources);
+        extensions.source_input_resolver = Some(Arc::new(CredentialRefreshingInputResolver::new(
```
jsummerfield, line 245: I think this extension seam is misleading as written. `engine_extensions_for_providers` accepts a provider-supplied `source_input_resolver`, but this runtime path immediately overwrites it with `CredentialRefreshingInputResolver`, so provider resolvers are never used by list/catalog/query/explain/validate. Could we either stop merging resolver fields from providers, or compose/delegate here instead of replacing the provider resolver?

sauldhernandez:
* **#764** <a href="https://app.graphite.com/github/pr/withcoral/coral/764?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/764?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

docs/guides/write-a-custom-source.mdx
```diff
@@ -267,6 +267,12 @@ Choose OAuth client settings from the provider's OAuth app and token endpoint re
 
 When you run `coral source add --file`, Coral reads each declared `input` from an environment variable of the same name, or prompts for it when you pass `--interactive`. `kind: variable` values go to source variables; `kind: secret` values go to the secret store. API keys, bearer tokens, OAuth access tokens, passwords, private keys, and authorization values must be secrets, even when the credential is read-only. Declaring an input only configures storage; it does not send the value to the API. Runtime request fields such as `auth.headers`, `BasicAuth`, `CustomAuth`, and request headers decide where stored inputs are used. For secret inputs, `credential.methods` controls how the secret is collected: `type: source_config` means environment-variable lookup by default or an interactive prompt with `--interactive`; OAuth device-code and authorization-code methods describe alternative install-time retrieval choices.
 
+For OAuth credential methods, Coral stores provider refresh metadata when it is
```
jsummerfield, line 285: This new refresh guidance is right, but the overview near the top of this guide still says automatic token refresh is not implemented yet. Could we update that earlier paragraph too so the source-authoring docs do not contradict themselves?

jsummerfield: Line 28

jsummerfield (approved):

PR #770, opened 2026-05-25
feat(sources/community/chucknorris): add Chuck Norris API community source

vinayaksonthalia:
## Description
This PR adds the [Chuck Norris API](https://api.chucknorris.io) community source to Coral. This is a lightweight, zero-auth data source for fetching Chuck Norris jokes.

This implementation includes the following tables:
- **`jokes`**: Fetches a completely random Chuck Norris joke.
- **`categories`**: Lists all available joke categories (extracts strings from a root array natively using Coral's row strategies).
- **`jokes_by_category`**: Fetches a random joke from a specific category dynamically by mapping the `category` filter down to the request query parameters.

### Limits & Rate Limiting
This API is entirely free and public with no documented rate limits or authentication mechanisms.

## Testing
- [x] Tested with `coral source lint`
- [x] Added successfully with `coral source add`
- [x] All query tests passed with `coral source test chucknorris`


jsummerfield (changes requested):
I found one product-surface gap before I can approve this:

- Medium: `sources/community/chucknorris/manifest.yaml:11` exposes `jokes`, `jokes_by_category`, and `categories`, and `sources/community/chucknorris/README.md:13` documents the same scope. The provider's documented API also includes free-text search at `/jokes/search?query={query}`, which is likely the most useful SQL workflow after categories. Please add a required-filter search table/function, for example `chucknorris.search WHERE query = 'rust'`, or explicitly narrow the README/source scope as random/category-only. Docs: https://api.chucknorris.io/ and https://github.com/chucknorris-io/chuck-api#usage
- Low: `sources/community/chucknorris/README.md:5` says no auth is needed, but the README does not mention rate limits. The PR body says no public rate limit is documented; please carry that into the README so users do not have to infer it from the PR.

Thanks for adding this. The implemented endpoints line up with the provider docs and the live tests passed; the remaining question is whether the source should cover the documented search workflow before it lands.

vinayaksonthalia:
> I found one product-surface gap before I can approve this:
> 
>     * Medium: `sources/community/chucknorris/manifest.yaml:11` exposes `jokes`, `jokes_by_category`, and `categories`, and `sources/community/chucknorris/README.md:13` documents the same scope. The provider's documented API also includes free-text search at `/jokes/search?query={query}`, which is likely the most useful SQL workflow after categories. Please add a required-filter search table/function, for example `chucknorris.search WHERE query = 'rust'`, or explicitly narrow the README/source scope as random/category-only. Docs: https://api.chucknorris.io/ and https://github.com/chucknorris-io/chuck-api#usage
> 
>     * Low: `sources/community/chucknorris/README.md:5` says no auth is needed, but the README does not mention rate limits. The PR body says no public rate limit is documented; please carry that into the README so users do not have to infer it from the PR.
> 
> 
> Thanks for adding this. The implemented endpoints line up with the provider docs and the live tests passed; the remaining question is whether the source should cover the documented search workflow before it lands.

Thanks for the feedback! You're totally right, the search API is super useful and was missing.

I've just pushed a commit that addresses both points:
1. **Search Workflow:** I added a new `search` table to the manifest with a required `query` filter mapping to `/jokes/search`. It correctly unpacks the `result` array. I also added a SQL query example for this table to the README.
2. **Rate Limits:** I added a Rate Limits section under Setup in the README, noting that while there's no official public quota, users should be polite and avoid heavy bulk harvesting.

Everything is passing `lint` and tests locally. Let me know if you need anything else!

jsummerfield (approved):
Thanks for the update. The earlier source-surface gap is addressed: the source now includes the search workflow, documents the provider's lack of published rate-limit policy, and the live API/source tests pass for categories, random jokes, and search.

I rechecked the current head for duplicate/core conflicts, CI, manifest validation, and live Coral queries.

vinayaksonthalia:
Thanks @jsummerfield so much for the thorough review and testing! Glad to hear the search workflow and everything else is working perfectly now.

simonwhitaker:
Disappointing. Chuck Norris writes his source specs in binary.

PR #777, opened 2026-05-25
chore: add DSL v4 PRD (take 2)

simonwhitaker:
Here's an infographic Codex generated that I found helpful:

DSL v4

### DIFF
--- a/plans/2026-05-24-dsl-v4-prd.md
+++ b/plans/2026-05-24-dsl-v4-prd.md
@@ -1,0 +1,880 @@
+# Coral Source DSL v4 PRD (Surface-First Projection IR)
+
+## Summary
+
+Coral source specs currently use a table-oriented authoring model. A source
+manifest chooses one top-level `backend` and then hard-codes the SQL surfaces
+that Coral exposes through `tables:` and `functions:`. For HTTP sources, those
+same table/function entries also declare provider request details, response
+extraction, filters, arguments, columns, and pagination.
+
+That model works for curated, hand-authored sources, but it does not scale to
+providers that already publish authoritative API descriptions such as OpenAPI or
+GraphQL schemas. It also forces a source to be represented by one implementation
+surface even when a real provider exposes several useful surfaces, such as
+GitHub REST and GitHub GraphQL.
+
+DSL v4 introduces a surface-first source model:
+
+1. The manifest declares one or more upstream surfaces and how to call them.
+2. Coral imports each surface description into a stable semantic IR.
+3. Coral generates SQL projections from that IR by default.
+4. Manifest authors may apply projection overrides to curate names, visibility,
+   column shape, descriptions, input exposure, and search/detail hints.
+5. Runtime execution remains surface-specific. The IR is a projection substrate,
+   not a generic physical executor abstraction.
+
+The core design split is:
+
+- **Surface type** answers how Coral talks to the provider: OpenAPI/REST,
+  GraphQL, MCP, file scan, or a future protocol.
+- **Semantic IR** answers what capabilities the provider exposes in a normalized
+  vocabulary that projection generation can use.
+- **Projection catalog** answers what SQL tables/functions Coral publishes.
+- **Physical planners** answer how a projection is executed for one concrete
+  surface type.
+
+The first implementation wave is OpenAPI-first. A v4 manifest can declare
+OpenAPI surfaces, Coral imports pinned OpenAPI descriptions during source add,
+materializes a semantic IR plus generated projection catalog to local app state,
+and later loads the installed source without network access or re-importing the
+OpenAPI document.
+
+Current `dsl_version: 3` manifests remain supported unchanged.
+
+## Goals
+
+- Add DSL v4 manifests that declare API surfaces instead of hard-coded
+  `tables:` and `functions:` arrays.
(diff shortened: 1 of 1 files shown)


simonwhitaker:
* **#896** <a href="https://app.graphite.com/github/pr/withcoral/coral/896?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#777** <a href="https://app.graphite.com/github/pr/withcoral/coral/777?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/777?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

plans/2026-05-24-dsl-v4-prd.md
```diff
@@ -0,0 +1,880 @@
...
+  provider operations, request bindings, response paths, or protocol execution
+  details.
+- Materialize imported IR and generated projections to local app state during
+  source add so normal query-time loads do not fetch or parse upstream
+  descriptions.
+- Preserve v3 parser, schema, install, and runtime behavior during migration.
+- Execute a narrow GitHub issues slice end to end through a v4 OpenAPI surface.
+- Keep the implementation aligned with current crate boundaries: source-spec
+  parsing in `coral-spec`, app-state materialization in `coral-app`, and
+  physical execution in `coral-engine`.
```
jsummerfield, line 64: Mostly yes, but I wouldn't limit yourself if a coral-source-model (or equiv) made sense.

plans/2026-05-24-dsl-v4-prd.md
```diff
@@ -0,0 +1,880 @@
...
+dsl_version: 4
+description: >-
+  Query GitHub issues through imported provider API descriptions.
+
+inputs:
+  GITHUB_API_BASE:
+    kind: variable
+    default: https://api.github.com
+  GITHUB_TOKEN:
+    kind: secret
```
jsummerfield, line 200: I'd have assumed these live on the surface

plans/2026-05-24-dsl-v4-prd.md
```diff
@@ -0,0 +1,880 @@
...
+GraphQL schemas. It also forces a source to be represented by one implementation
+surface even when a real provider exposes several useful surfaces, such as
+GitHub REST and GitHub GraphQL.
+
+DSL v4 introduces a surface-first source model:
+
+1. The manifest declares one or more upstream surfaces and how to call them.
+2. Coral imports each surface description into a stable semantic IR.
+3. Coral generates SQL projections from that IR by default.
+4. Manifest authors may apply projection overrides to curate names, visibility,
```
jsummerfield, line 22: My thinking has evolved recently here. I now think projections should be wholly owned and maintained by users (not coral, not manifest authors). Coral is responsible for generating the IR (entities, operations, etc), and the user is responsible for configuring which of those to project and how. Each projecton will be use case specific.

plans/2026-05-24-dsl-v4-prd.md
```diff
@@ -0,0 +1,880 @@
...
+        required: true
+    columns:
+      include:
+        - number
+        - title
+        - state
+        - user.login
+        - html_url
+      rename:
+        user.login: author_login
```
jsummerfield, line 253: (See my message above) I don't think projections belong at this stage, otherwise it feels a bit full circle. We'll need a way to record the user's projection decisions, but that should live in another file.

plans/2026-05-24-dsl-v4-prd.md
```diff
@@ -0,0 +1,880 @@
...
+- If fetched bytes do not match `sha256`, source add fails.
+- Bundled/core sources should prefer provider-owned immutable URLs when they
+  exist.
+- When a provider does not offer an immutable historical OpenAPI URL, Coral
+  should mirror the exact OpenAPI artifact at a Coral-owned artifact URL and pin
+  that URL with `sha256`.
+- User-installed/community sources may use mutable provider URLs, but they must
+  still provide `sha256`.
+
+There is no unpinned URL mode.
```
jsummerfield, line 318: This is worth pushing on. Let's chat f2f.

plans/2026-05-24-dsl-v4-prd.md
simonwhitaker: Yeah, I couldn't decide on that, but you're probably right. I was thinking about Github and Datadog, both of which have two surfaces, and in both cases those surfaces would use the same inputs, so it felt more natural to have them owned by the source than to repeat them per-surface. I'll see if I can find a good counter-example to test that against.

simonwhitaker: Codex is pretty adamant this should stay on the source level, and couldn't find any good examples where two surfaces on the same source have mostly disjoint inputs. The best it could find was Clickhouse, where the control plane has different auth requirements the data plane (I think this is likely the case for a number of services that have separate data and control APIs), but even then it concluded:

> The harder conclusion: if two surfaces have totally disjoint auth and no shared user mental model, they may not belong in the same source at all. ClickHouse Cloud management and a ClickHouse database connection could reasonably be separate sources.

I think that's a reasonable take.

PR #779, opened 2026-05-25
feat(sources/community/free-dictionary): add Free Dictionary API community source

vinayaksonthalia:
## Description
This PR adds the [Free Dictionary API](https://dictionaryapi.dev/) community source to Coral. This is a lightweight, zero-auth data source for fetching English word definitions, phonetics, audio pronunciation URLs, and meanings.

This implementation includes the following table:
- **`entries`**: Fetches the dictionary definitions for a specific English word (requires the `word` filter mapped to the URL path). Extracts phonetics and meanings directly as JSON.

### Limits & Rate Limiting
This API is open-source and entirely free to use with zero authentication requirements.

## Testing & Proof of Execution

The manifest successfully passes `coral source lint` and all queries execute correctly against the live API. Here is the console output from the test run:

```text
Added source free_dictionary

  ✓ free_dictionary connected successfully

    free_dictionary (1 table)
    └─ entries
    
    Query tests

    ✓ SELECT * FROM free_dictionary.entries WHERE word = 'hello' LIMIT 1
      1 row


jsummerfield (changes requested):
I found two behavior issues to fix before I can approve this:

- Medium: `sources/community/free-dictionary/manifest.yaml:23` maps `word` from the SQL filter instead of the API response. The v2 response includes a per-entry `word` field, and the API canonicalizes input such as `HELLO` to `hello`. Please map `word` from the response path `[word]`, and add a separate `query_word` column only if echoing the filter is useful. Docs: https://dictionaryapi.dev/ and https://github.com/meetDeveloper/freeDictionaryAPI#usage
- Medium: `sources/community/free-dictionary/manifest.yaml:18` leaves 404s as query errors. For a dictionary lookup source, an unknown word is normal empty-result behavior rather than a source failure. Please set `allow_404_empty: true` and document unknown-word behavior in the README. Docs: https://github.com/meetDeveloper/freeDictionaryAPI#http-request-and-query-methods

Thanks for adding this source. The happy path evidence is good; the remaining fixes are about making lookup semantics predictable for normal user queries.

vinayaksonthalia:
> I found two behavior issues to fix before I can approve this:
> 
>     * Medium: `sources/community/free-dictionary/manifest.yaml:23` maps `word` from the SQL filter instead of the API response. The v2 response includes a per-entry `word` field, and the API canonicalizes input such as `HELLO` to `hello`. Please map `word` from the response path `[word]`, and add a separate `query_word` column only if echoing the filter is useful. Docs: https://dictionaryapi.dev/ and https://github.com/meetDeveloper/freeDictionaryAPI#usage
> 
>     * Medium: `sources/community/free-dictionary/manifest.yaml:18` leaves 404s as query errors. For a dictionary lookup source, an unknown word is normal empty-result behavior rather than a source failure. Please set `allow_404_empty: true` and document unknown-word behavior in the README. Docs: https://github.com/meetDeveloper/freeDictionaryAPI#http-request-and-query-methods
> 
> 
> Thanks for adding this source. The happy path evidence is good; the remaining fixes are about making lookup semantics predictable for normal user queries.

Thanks for the great feedback, @jsummerfield! Good points on making the lookup semantics more robust.

I've pushed a commit that resolves both of these behavior issues:
1. **Canonical Word:** The `word` column now explicitly maps from the API response path `[word]` to capture the canonical casing. I also added a `query_word` column that retains the exact filter string the user typed in, in case they need to echo it back.
2. **Unknown Words (404s):** Added `allow_404_empty: true` to the manifest so querying an unknown word gracefully returns an empty result instead of a query error. I also documented this behavior under an "Unknown Words" section in the README.

Everything passes `coral source lint` and the live tests run perfectly. Let me know if you need any other adjustments!

jsummerfield (approved):
Thanks for the update. The previous behavior blockers are addressed: the source now preserves the provider's canonical `word`, keeps the submitted value separately as `query_word`, and unknown words return empty rows instead of surfacing the provider 404.

I rechecked the current head against the live API and with Coral source/query tests, including the canonicalization case where `HELLO` returns `word = 'hello'`.

vinayaksonthalia:
Thanks so much for the review and for the approval! Let me know if you need anything else from my end before merging.

PR #784, opened 2026-05-25
feat(cli): add runtime feature flags

jsummerfield:
## Summary

- Add a runtime `[features]` config surface for experimental feature toggles, with known-feature status reporting and invalid/unsupported value handling.
- Add `coral features list`, `coral features enable <feature>`, and `coral features disable <feature>` so users can inspect and manage feature state from the CLI.
- Wire the feedback feature into MCP bootstrap so config-driven feature state and `--enable-feedback` resolve through the same app-level feature model.
- Document the feature config and feedback tool enablement path in the MCP and installation guides.

## Testing

- `cargo test -p coral-cli --test features`
- `cargo test -p coral-app features`
- `cargo test -p coral-cli --features cli-test-server --test mcp`

### DIFF
--- a/crates/coral-app/src/features.rs
+++ b/crates/coral-app/src/features.rs
@@ -1,0 +1,343 @@
+//! Runtime feature registry and effective feature resolution.
+
+use std::collections::BTreeMap;
+use std::path::PathBuf;
+
+use tracing::warn;
+
+use crate::bootstrap::{AppError, discover_app_state_layout};
+use crate::state::{
+    AppStateLayout, RawFeatureContainerState, RawFeatureOverrides, RawFeatureValue,
+};
+
+/// Runtime feature keys recognized by Coral.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum Feature {
+    /// Expose the optional MCP `feedback` tool.
+    Feedback,
+}
+
+impl Feature {
+    /// Returns all runtime features recognized by this Coral binary.
+    #[must_use]
+    pub fn all() -> impl ExactSizeIterator<Item = Self> {
+        FEATURE_SPECS.iter().map(|spec| spec.feature)
+    }
+
+    /// Returns the stable key used under `[features]` in `config.toml`.
+    #[must_use]
+    pub fn key(self) -> &'static str {
+        spec_for_feature(self).key
+    }
+
+    /// Returns whether the feature is enabled when no config override exists.
+    #[must_use]
+    pub fn default_enabled(self) -> bool {
+        spec_for_feature(self).default_enabled
+    }
+
+    /// Returns a short human-readable description of the feature.
+    #[must_use]
+    pub fn description(self) -> &'static str {
+        spec_for_feature(self).description
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+struct FeatureSpec {
+    feature: Feature,
+    key: &'static str,
+    default_enabled: bool,
+    description: &'static str,
+}
+
+const FEATURE_SPECS: &[FeatureSpec] = &[FeatureSpec {
+    feature: Feature::Feedback,
+    key: "feedback",
+    default_enabled: false,
+    description: "Expose the optional MCP feedback tool.",
+}];
+
+/// How a feature's value is configured in Coral's local config.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum FeatureConfiguredState {
+    /// No override exists, so the feature uses its built-in default.
+    Default,
+    /// The local config explicitly enables the feature.
+    Enabled,
+    /// The local config explicitly disables the feature.
+    Disabled,
+    /// The local config contains a known feature key with a non-boolean value.
+    InvalidValue,
+    /// The local config uses an unsupported `[features]` container shape.
+    InvalidContainer,
+}
+
+impl FeatureConfiguredState {
+    /// Returns the stable user-facing label for this configured state.
+    #[must_use]
+    pub fn as_str(self) -> &'static str {
+        match self {
+            Self::Default => "default",
+            Self::Enabled => "enabled",
+            Self::Disabled => "disabled",
+            Self::InvalidValue => "invalid-value",
+            Self::InvalidContainer => "invalid-container",
+        }
+    }
+}
+
+/// A feature's configured and effective runtime state.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FeatureStatus {
+    /// Feature enum variant.
+    pub feature: Feature,
+    /// Stable config key.
+    pub key: &'static str,
+    /// Built-in enabled state when no override exists.
+    pub default_enabled: bool,
+    /// Current config state.
+    pub configured: FeatureConfiguredState,
+    /// Effective runtime state after applying config to the default.
+    pub enabled: bool,
+    /// Short human-readable description.
+    pub description: &'static str,
+}
+
+/// Effective runtime feature state after applying config overrides.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Features {
+    enabled: BTreeMap<Feature, bool>,
+}
+
+impl Default for Features {
+    fn default() -> Self {
+        let enabled = FEATURE_SPECS
+            .iter()
+            .map(|spec| (spec.feature, spec.default_enabled))
+            .collect();
+        Self { enabled }
+    }
+}
+
+impl Features {
+    /// Returns whether a runtime feature is enabled.
+    #[must_use]
+    pub fn enabled(&self, feature: Feature) -> bool {
+        self.enabled.get(&feature).copied().unwrap_or(false)
+    }
+
+    fn from_raw_overrides(raw: &RawFeatureOverrides) -> Self {
+        let mut features = Self::default();
+        for (key, value) in raw.iter() {
+            let Some(spec) = spec_for_key(key) else {
+                warn!(feature = key, "ignoring unknown Coral runtime feature");
+                continue;
+            };
+
+            match value {
+                RawFeatureValue::Bool(enabled) => {
+                    features.enabled.insert(spec.feature, enabled);
+                }
+                RawFeatureValue::UnsupportedType => {
+                    warn!(
+                        feature = key,
+                        "ignoring unsupported Coral runtime feature value; expected boolean"
+                    );
+                }
+            }
+        }
+        features
+    }
+}
+
+/// Loader for runtime features from Coral's local config.
+#[derive(Debug, Clone)]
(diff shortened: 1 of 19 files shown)


docs/getting-started/installation.mdx
```diff
@@ -221,7 +221,10 @@ On Windows, the default local state path is
 
 Important files include:
 
-- `config.toml` for installed-source metadata and non-secret variables
+- `config.toml` for installed-source metadata, non-secret variables, and
+  runtime feature opt-ins inspected with `coral features list`, managed by
+  `coral features enable|disable`, or manually under `[features]` with entries
+  such as `feedback = true`
```
AlbertQM: When this part of the doc was written, `config.toml` didn't have much in it. Now it's growing and it can control a bunch of things. Should this bullet point stay lean, and link to a bigger explanation somewhere, which covers all the things you can configure in this file? WDYT?

docs/guides/use-coral-over-mcp.mdx
```diff
@@ -61,17 +61,44 @@ Some sources expose native search endpoints (for example, GitHub issue search) a
...
+```shellscript
+coral features disable feedback
+```
+
+You can also manage the same opt-in manually in Coral's `config.toml`:
+
+```toml
+[features]
+feedback = true
+```
```
AlbertQM: All of this is describing `coral features`, which shouldn't live under `### optional feedback tool` -- I think it has earned its own section somewhere; I'm thinking a new top-level config.toml under References

docs/guides/use-coral-over-mcp.mdx
```diff
@@ -61,17 +61,44 @@ Some sources expose native search endpoints (for example, GitHub issue search) a
...
 
 <Note>
-Only enable this tool if you are comfortable sending the feedback text to Coral.
+Only enable this tool if you are comfortable exposing an agent-callable feedback tool that can send the feedback text to Coral.
 </Note>
 
 ## Client setup
 
-Coral uses stdio transport. If a client supports a command-based install flow, point it at `coral mcp-stdio`. To include the optional feedback tool, use `coral mcp-stdio --enable-feedback`.
+Coral uses stdio transport. If a client supports a command-based install flow, point it at `coral mcp-stdio`. To include the optional feedback tool for one process, use `coral mcp-stdio --enable-feedback`; for persistent opt-in, run `coral features enable feedback`.
```
AlbertQM:
```suggestion
Coral uses stdio transport. If a client supports a command-based install flow, point it at `coral mcp-stdio`. 
```

We've discussed the feedback above, do we really need it here again? 👀

docs/getting-started/installation.mdx
jsummerfield: Good shout - done

docs/reference/cli-reference.mdx
```diff
@@ -121,6 +121,34 @@ Run the guided source setup flow.
 coral onboard
 ```
 
+## `coral features`
+
+Inspect and manage experimental runtime features in [`config.toml`](/reference/configuration).
```
AlbertQM: Are these features only experimental/under development ones? We might want to specify this in the first line of this section!

something like, but not necessarily exactly this

```suggestion
Coral ships with experimental features that can be enabled on demand.  These features are defined in [`config.toml`](/reference/configuration).
```

docs/reference/cli-reference.mdx
```diff
@@ -121,6 +121,34 @@ Run the guided source setup flow.
...
+
+Persistently enable an experimental runtime feature.
+
+```shellscript
+coral features enable feedback
+```
+
+### `coral features disable <FEATURE>`
+
+Remove a persistent runtime feature opt-in from `config.toml`, returning the feature to its built-in default.
```
AlbertQM: I'm not sure what this means. If I `disable` something, will it update `config.toml` and change the default there? 
How do I know what's the "built-in" default for a feature?

Could we be more explicit here?

docs/reference/configuration.mdx
```diff
@@ -0,0 +1,81 @@
...
+export CORAL_CONFIG_DIR=/path/to/coral-config
+```
+
+On Windows PowerShell:
+
+```powershell
+$env:CORAL_CONFIG_DIR = "C:\path\to\coral-config"
+```
+
+On Windows, the default local state path is `%APPDATA%\withcoral\coral\config`.
```
AlbertQM: Don't think we need to explain this again here, we can just link to the explanation 😄 

```suggestion
Coral stores local configuration in `config.toml` inside its platform-specific [local state directory](/getting-started/installation#local-state).```

**Review by AlbertQM — COMMENTED** [review]
Thanks for the updates to the docs! 😄

File: docs/reference/cli-reference.mdx [hunk 7]
```diff
@@ -121,6 +121,38 @@ Run the guided source setup flow.
...
+```shellscript
+coral features enable feedback
+```
+
+### `coral features disable <FEATURE>`
+
+Disable updates `config.toml` so the feature is not enabled for future runs.
+For features disabled by default, such as `feedback`, Coral removes the opt-in
+and falls back to the built-in default. The default for each feature is listed
+in [Configuration](/reference/configuration#runtime-features).
```
AlbertQM: Thanks for the clarification! My understanding of `coral features disable <feat>` is that it will.. disable it. Ie this overrides everything. It doesn't matter if a feature is on/off by default, if I run this command, it will be disabled; as this is just a proxy to write to `config.toml` 

If that's the case, I think a single sentence (both for enable and disable) would work best, something along the lines of:

```suggestion
Disables an experimental runtime feature. This preference is persisted in `config.toml`.
```


(Also would remove the configuration link, and move it to the initial paragraph, as it doesn't just apply to disable command)

docs/reference/cli-reference.mdx
```diff
@@ -121,6 +121,38 @@ Run the guided source setup flow.
...
+
+List all runtime features recognized by the current Coral binary, including their configured and effective status.
+
+```shellscript
+coral features list
+```
+
+### `coral features enable <FEATURE>`
+
+Persistently enable an experimental runtime feature.
```
AlbertQM:
```suggestion
Persistently enable an experimental runtime feature. The default for each feature is listed
in [Configuration](/reference/configuration#runtime-features).
```

AlbertQM (approved):
This will be super useful, thank you!

crates/coral-app/src/features.rs
```diff
@@ -0,0 +1,376 @@
...
+}
+
+impl Features {
+    /// Returns whether a runtime feature is enabled.
+    #[must_use]
+    pub fn enabled(&self, feature: Feature) -> bool {
+        self.enabled.get(&feature).copied().unwrap_or(false)
+    }
+
+    fn feedback_enabled(&self, explicit_enable: bool) -> bool {
```
sauldhernandez: I'd suggest we ditch this and focus on using overrides to pass along cli flag overrides. Setting this pattern to have one function per feature is not going to scale well.

crates/coral-app/src/features.rs
```diff
@@ -0,0 +1,376 @@
...
+/// Resolves whether the MCP feedback tool should be enabled for this process.
+///
+/// Explicit process flags can enable feedback for one run, but cannot disable a
+/// persistent opt-in from local config.
+///
+/// # Errors
+///
+/// Returns [`AppError`] if the local config path cannot be discovered or
+/// `config.toml` exists but cannot be read or parsed.
+pub fn resolve_feedback_enabled(explicit_enable: bool) -> Result<bool, AppError> {
```
sauldhernandez: Same here. We already have a feature enum, so this function could be generic instead of specific for feedback.

crates/coral-app/src/features.rs
```diff
@@ -0,0 +1,376 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Returns [`AppError`] if the key is unknown or the local config cannot be updated.
+    pub fn disable(&self, key: &str) -> Result<(), AppError> {
+        let spec = spec_for_key(key).ok_or_else(|| unknown_feature_error(key))?;
+        if spec.default_enabled {
+            crate::state::set_raw_feature_override(&self.layout, spec.key, false)
+        } else {
+            crate::state::clear_raw_feature_override(&self.layout, spec.key)
```
sauldhernandez: This may be a bit fragile if the default ever changes. A user that explicitly disabled a feature will find it being turned on again because the default has changed without any warning.

docs/reference/configuration.mdx
```diff
@@ -0,0 +1,67 @@
...
+```shellscript
+coral features enable feedback
+coral features disable feedback
+```
+
+Feature values must be booleans. Unknown feature keys are preserved in `config.toml` but ignored by Coral.
+
+| Feature | Default | Description |
+| ------- | ------- | ----------- |
+| `feedback` | `false` | Experimental. Exposes the MCP feedback tool when enabled. |
```
sauldhernandez: The note that specified that enabling this sends information to Coral is now gone. Worth keeping so users are aware of this.

docs/guides/use-coral-over-mcp.mdx
```diff
@@ -59,19 +58,9 @@ For deeper introspection, query `coral.tables`, `coral.columns`, `coral.filters`
 
 Some sources expose native search endpoints (for example, GitHub issue search) as table functions with `kind = 'search'`. Prefer these over scanning a regular table when the agent needs ranked, query-driven results. They return provider-ranked candidates, so agents should preserve any returned rank columns and use the returned ids plus catalog-described tables to fetch full detail rows when the search result itself is incomplete.
 
-### Optional feedback tool
-
-Start Coral as `coral mcp-stdio --enable-feedback` to expose the `feedback` tool. Agents can call it when they get blocked or repeat a pattern that isn't working.
```
sauldhernandez, line 64: `--enable-feedback` is still supported, but now its undocumented with this change.

sauldhernandez (changes requested):

docs/reference/configuration.mdx
AlbertQM: Good catch! I'd just extend the description here, instead of adding a note block at the end, as this table will grow and it's better to keep that information close together 😄

crates/coral-app/src/features.rs
jsummerfield: Agreed, sorry I missed this one. Will fix (and below)

docs/guides/use-coral-over-mcp.mdx
jsummerfield: I'm going to leave that in the config page where we explain the feature.

jsummerfield:
@sauldhernandez main change since your last review is generic runtime support for `--enable-X` and `--disable-X` for feature X.

sauldhernandez (approved):

PR #785, opened 2026-05-25
feat(sources/community/metabase): add Metabase source

Akash-How:
### Summary

Adds a community Coral source for Metabase metadata. The source exposes read-only collections, collection items, dashboards, cards/questions, databases, users, and permission groups for BI governance workflows.

### Details

- Adds `sources/community/metabase/manifest.yaml` with Metabase API metadata tables.
- Adds `sources/community/metabase/README.md` with setup, table coverage, example queries, limitations, and validation notes.
- Keeps the API key as `kind: secret` and deliberately avoids executing saved questions, fetching query results, or exposing database credentials.
- Marks Metabase collection IDs nullable because the root collection can be returned with a null `id`; users should query root collection items with `collection_id = 'root'`.

### Validation

- YAML parsing: passed for `sources/community/metabase/manifest.yaml`
- Coral manifest schema validation: passed for `sources/community/metabase/manifest.yaml`
- `git diff --check`: passed
- `make lint-sources`: passed
- Live API tests: passed against a local Metabase Docker instance with `coral source add` and `coral source test metabase`

Live Coral result:

```text
Added source metabase

✓ metabase connected successfully

  metabase (7 tables)
  ├─ cards
  ├─ collection_items
  ├─ collections
  ├─ dashboards
  ├─ databases
  ├─ permission_groups
  └─ users

  Query tests

  ✓ SELECT id, name FROM metabase.collections LIMIT 1
    1 row

  ✓ SELECT id, name FROM metabase.dashboards LIMIT 1
    1 row
```

No API keys, credentials, or customer data are included.

### DIFF
--- a/sources/community/metabase/README.md
+++ b/sources/community/metabase/README.md
@@ -99,6 +99,8 @@ ORDER BY member_count DESC;
 ## Limitations
 
 - Metabase's API is not versioned and can change between releases.
+- Metabase can return a root collection with a null `id`; use
+  `collection_id = 'root'` when querying root collection items.
 - `metabase.collection_items` requires `collection_id`; use `root` for the
   root collection.
 - `metabase.permission_groups` may require an admin-scoped API key.
@@ -116,7 +118,7 @@ YAML parse: passed for sources/community/metabase/manifest.yaml
 Coral manifest schema validation: passed for sources/community/metabase/manifest.yaml
 git diff --check: passed
 make lint-sources: passed
-Live API tests: not run; require real Metabase credentials
+Live API tests: passed against a local Metabase Docker instance with `coral source add` and `coral source test metabase`
 ```
 
-Testing note: live API tests require user-provided credentials and are intentionally not part of default CI. The manifest includes `test_queries`, and no credentials or customer data are committed.
+Testing note: the live API test used user-provided Metabase credentials. The manifest includes `test_queries`, and no credentials or customer data are committed.
--- a/sources/community/metabase/README.md
+++ b/sources/community/metabase/README.md
@@ -1,0 +1,122 @@
+# Metabase Coral Source
+
+## What This Source Exposes
+
+This source exposes [Metabase](https://www.metabase.com/) BI workspace metadata
+as SQL tables in Coral. It is designed for analytics governance: collections,
+dashboards, saved questions, connected databases, users, and permission groups.
+
+It is read-only. It does not execute Metabase questions and does not expose
+database credentials.
+
+## Authentication
+
+Create a Metabase API key and pass it as `METABASE_API_KEY`. The key is sent
+with the `X-API-Key` header.
+
+Required inputs:
+
+| Input | Kind | Default | Description |
+|---|---|---|---|
+| `METABASE_BASE_URL` | variable | `http://localhost:3000` | Base URL for your Metabase instance, without a trailing slash. |
+| `METABASE_API_KEY` | secret | - | Metabase API key with read access to the metadata you want to inspect. |
+
+## Setup
+
+```bash
+METABASE_BASE_URL=https://metabase.example.com \
+METABASE_API_KEY=mb_your_key_here \
+coral source add --file sources/community/metabase/manifest.yaml
+```
+
+Or run interactively:
+
+```bash
+coral source add --interactive --file sources/community/metabase/manifest.yaml
+```
+
+## Tables
+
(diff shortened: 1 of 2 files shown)


jsummerfield (changes requested):
I found a few source-quality blockers to fix before I can approve this:

- High: `sources/community/metabase/manifest.yaml:420` sets `metabase.users` to `pagination: none`, but Metabase's `GET /api/user` endpoint supports `limit` and `offset`. This can silently undercount users on larger instances. Please add offset pagination with `offset` and `limit`, or remove the table until pagination is handled. Docs: https://www.metabase.com/docs/latest/api and https://www.metabase.com/docs/latest/api.json
- Medium: `sources/community/metabase/README.md:82` shows a query for dashboards owned by inactive users, but `metabase.users` does not expose `status` or `include_deactivated`. Metabase returns active users by default; deactivated/all users require `status=deactivated`, `status=all`, or `include_deactivated=true`. Please add the filter and update the example.
- Medium: `sources/community/metabase/manifest.yaml:204` and `:268` say the dashboards/cards tables help find archived content, but neither table exposes the documented `f=archived` query parameter. Please add an `f` filter to both tables, plus `model_id` for cards if useful, and update the examples/guidance.
- Medium: `sources/community/metabase/README.md:9` says database credentials are not exposed, but `sources/community/metabase/manifest.yaml:388` exposes raw `details` from `GET /api/database`. Metabase includes `details` when the requester can change database details. Please remove `details` for v1 or document the exposure risk and recommend a least-privilege API key.

Thanks for the work here. The add/test evidence is useful, but the pagination/filter semantics need to be correct before this becomes user-facing product surface.

Akash-How:
Addressed these Metabase source-quality blockers in `4ba2b9b`.

- Added offset pagination for `metabase.users` using `offset` and `limit`.
- Added `status` and `include_deactivated` filters for `metabase.users`, and updated the inactive-dashboard-owner example to request `status = 'all'`.
- Added `f` filters to `metabase.dashboards` and `metabase.cards`, plus `model_id` for cards.
- Removed raw `database.details` from `metabase.databases` for v1 so connection details are not exposed.
- Updated README guidance for archived content, inactive users, and database connection detail omission.

Validation rerun:

```text
git diff --check: passed
coral source lint sources/community/metabase/manifest.yaml: passed
make lint-sources: passed
coral source add --file sources/community/metabase/manifest.yaml: passed
coral source test metabase: passed, 2 declared · 2 passed · 0 failed
```

`make lint-sources` still prints the existing Shopify comment warning, but exits successfully and is unrelated to this Metabase PR.

Akash-How:
I did one more self-audit against the requested pagination/filter semantics and found a Coral usability gap: the new filters were declared, but they also need virtual columns so users can write the documented SQL predicates.

Fixed in `9b3b7df`:

- Added virtual `f` column on `metabase.dashboards`.
- Added virtual `f` and `model_id` columns on `metabase.cards`.
- Added virtual `status` and `include_deactivated` columns on `metabase.users`.

Additional checks run:

```text
coral sql "SELECT id, email, is_active FROM metabase.users WHERE status = 'all' LIMIT 2": passed
coral sql "SELECT id, name, archived FROM metabase.dashboards WHERE f = 'archived' LIMIT 2": passed
coral sql "SELECT id, name, archived FROM metabase.cards WHERE f = 'archived' LIMIT 2": passed
coral sql "SELECT id, name, engine FROM metabase.databases LIMIT 2": passed
coral source add --file sources/community/metabase/manifest.yaml: passed
coral source test metabase: passed, 2 declared · 2 passed · 0 failed
coral source lint sources/community/metabase/manifest.yaml: passed
git diff --check: passed
make lint-sources: passed
```

The archived dashboard/card checks return empty rows on the local test instance, but they compile, call the filtered API path, and complete successfully. `make lint-sources` still prints the unrelated existing Shopify comment warning while exiting successfully.

jsummerfield (approved):
I'm approving this version. The prior blockers are fixed: users pagination now uses offset/limit, inactive-user and archived-content filters are exposed as SQL-bindable virtual columns, and the raw database `details` field is no longer exposed.

One non-blocking follow-up: the README mentions inactive-user filters, but it would be clearer to say Metabase requires superuser permissions for `status` / `include_deactivated` on `/api/user`. That can be handled as a docs polish pass.

Thanks for the careful follow-up, especially the self-audit that added the virtual filter columns.

PR #790, opened 2026-05-25
feat(sources): UI-driven core + community sources management

martinrajdl:
First real cut of the sources management page — list, search, and install core + community sources from the UI, backed by real RPCs end-to-end (no mocks).

Branched off `main` @ `b8d6f7d`. Implementation plan and locked architectural decisions live in [`SOURCES_MANAGEMENT_PLAN.md`](./SOURCES_MANAGEMENT_PLAN.md) on this branch.

### Backend (Rust)
- **New gRPC RPCs:** `DiscoverCommunitySources`, `GetCommunitySourceInfo`. New `SOURCE_ORIGIN_COMMUNITY = 3` enum variant. `proto/coral/v1/sources.proto`.
- **Community catalog (`crates/coral-app/src/sources/community.rs`):** the build script (`build.rs`) scans `sources/community/*/manifest.yaml` and emits a metadata-only table (`COMMUNITY_CATALOG`); the runtime resolves manifest YAML on demand — release builds fetch from `raw.githubusercontent.com/withcoral/coral/<commit>/...` pinned by the `CORAL_BUILD_COMMIT_SHA` env at build time, dev builds read from the workspace tree at `CORAL_DEV_WORKSPACE_ROOT`.
- **`SourceManager`:** `discover_community_sources` and `get_community_source_info`. Community sources persist on disk identically to imported ones — the community origin is a discovery-time marker, not a storage one — so `ImportSource` handles install unchanged.
- **OAuth token refresh:** cherry-picked from `codex/oauth-token-refresh` (commits `3328047`, `a4d322a`, `b80a3b5`) so the UI's OAuth flow doesn't bit-rot the moment a token expires. **Should land via its own PR upstream**; this branch rebases clean once that happens.

### UI (React)
- Hash router with `traces`, `sources`, `source-install`, `source-detail` routes. Navbar gains a Sources entry with route-aware active state.
- **Sources index:** Connected section + Library section with facets (All / Core / Community / Installed) and text search, backed by `ListSources` + `DiscoverSources` + `DiscoverCommunitySources`.
- **Install flow:** unified view for core and community sources. Renders variable + secret fields, a credential-method picker when a secret declares multiple methods, and OAuth-specific fields (client ID, optional client secret). Submit branches:
  - bundled, no OAuth → `CreateBundledSource`
  - bundled, OAuth → `CreateBundledSourceWithOAuth` (streaming)
  - community, no OAuth → `ImportSource` with the resolved YAML
  - community, OAuth → same, with `oauthCredentialRetrievals`
- **OAuth progress:** opens the authorization URL in a new tab (`window.open`) and shows a visible fallback link in case the popup is blocked. App still binds the loopback listener as today, exchanges the code, and emits `oauth_completed` + `source` events that the UI consumes.
- **Detail view:** configured variables + masked secret keys, a Validate action that runs the manifest's test queries, a Remove action behind an inline confirm.

### What's not in this PR
- Custom-source authoring (the "studio" / OpenAPI/GraphQL/MCP paste flow from the earlier prototype). Mocked end-to-end on the prototype branch; deferred until backend manifest parsing is real.
- Schema explorer surface. Kept on its prototype branch; revisit after v1.
- `outbound_hosts` host-confirmation UX — that proto field doesn't exist on `main` (lives on `claude/elastic-brown-8ad186`). If it lands, surface on the install + detail pages.
- Icon field in source spec — UI maps a known set of source names to local SVGs and falls back to a plug glyph.
- Centralizing OAuth browser-open in the app (so behavior matches CLI exactly). Currently UI calls `window.open`. Tightening this needs a small proto/server change for a client-opens-browser hint.

## Test plan

Automated
- [x] `cargo test -p coral-app` — 120 lib tests + 59 gRPC integration tests (3 new: community discovery, info+YAML, install round-trip)
- [x] `npm run build` (UI type-check + Vite build)
- [x] `npx oxlint --deny-warnings`

Manual
- [ ] `cd ui && npm run dev:local`, open `http://localhost:5173/#/sources`. Connected + Library load; facets + search work.
- [ ] Install a core source via Paste (e.g. `slack` with a pasted token). Returns to detail page; Validate succeeds.
- [ ] Install a core source via OAuth (`slack`, enter `SLACK_OAUTH_CLIENT_ID`, complete in browser). Stream progresses through `awaiting-oauth` → `oauth-completed` → installed.
- [ ] Install a community source (e.g. `hn`). Manifest resolved from the local workspace tree (since `CORAL_BUILD_COMMIT_SHA` is empty in dev).
- [ ] Delete an installed source from the detail page. Confirm flow works; source disappears from Connected.

## Notes for reviewers
- The big single commit on this branch is the UI install view (`3a3df23`) — start there if you want the highest-leverage diff.
- Backend changes are minimal and additive (no behavior change for existing RPCs).
- Three commits at the top are the OAuth refresh cherry-pick. If you'd rather review without them, look at `git diff 3328047..HEAD`.


martinrajdl:
## Walkthrough

Branch is ready for review. Posting a structured walkthrough below — the **Discovery (catalog)** section calls out the part I'd like other eyes on before we treat it as settled.

---

### UI changes

All in `ui/src/`. The shape lifts the prototype from `ludo-sources-review-v2` and the visual language from `phoebeai/adp`.

**Routing & shell** (`lib/router.ts`, `App.tsx`, `components/navbar/navbar.tsx`)
- Hash router with four routes: `traces`, `sources`, `source-install/:name?origin=`, `source-detail/:name`.
- Navbar gains a Sources entry with route-aware active state.

**Sources index** (`views/sources/sources-index.tsx`)
- Centered max-width-960 page modeled on `adp/web/src/views/plugins-view.tsx`.
- Big `HeadingLarge` + subtitle, single search box (max-width 360).
- Two sections: **Connected** (only when there's something installed) then **Available**, both alphabetical, both filtered by the same search.
- I tried topical category sections first (Observability / Developer Tools / etc.) — pulled it back to a flat list because the category map was guesses, not a vetted taxonomy. If we want categories later, they should come from the manifest itself.
- Card design: round provider logo, capitalised name, green Connected pill or grey Core/Community pill in the header, 2-line description, version footer.

**Install / credential form** (`views/sources/source-install.tsx`, `source-install.css.ts`)
- Layout matches `adp/web/src/components/credential-form.tsx`:
  - Round logo + capitalised name + Core/Community pill at the top, description below.
  - 2-column grid of fields. Labels sentence-cased (`API_TOKEN` → "Api token"). Manifest \`hint\` renders as secondary text below each field.
  - Secrets use \`TextInput type=\"password\"\`; no separate eye toggle.
  - Multi-method secrets get a compact segmented control above the input.
  - Right-aligned action row: bare Cancel + primary "Save credentials". Error pill above the action row.
- State machine drives:
  - bundled + no OAuth → \`CreateBundledSource\`
  - bundled + OAuth → \`CreateBundledSourceWithOAuth\` (streaming)
  - community + no OAuth → \`ImportSource\` with the resolved YAML
  - community + OAuth → same, with \`oauthCredentialRetrievals\`

**Installed-source detail** (`views/sources/source-detail.tsx`)
- Lightweight read-only surface: configured variables, masked secret keys, Validate action that runs the manifest's test queries, Remove behind an inline confirm.

**Supporting bits**
- `lib/sources.ts` — typed Connect-RPC wrappers for every Source RPC the UI uses, including the OAuth stream consumer.
- `lib/coral-clients.ts` — shared gRPC-Web transport.
- `lib/provider-icons.ts` + `public/images/providers/*` — name → SVG/PNG icon map, plug-glyph fallback.
- `components/{page-header,error-banner,toast}` — small primitives lifted from the prototype.
- `wax/components/dialog/*` — ported from `adp/web/src/wax/components/dialog`, not yet used. Set up for the followup pattern of "click a source on the index → open Dialog with the credential form" if we go that way.
- `wax/components/icon.tsx` — extended \`IconName\` with \`Check\`, \`CircleCheck\`, \`Plug\`, \`Plus\`, \`RefreshCw\` (all sourced from \`lucide-react\`).

---

### Backend changes

Two distinct groups.

**1. OAuth token refresh — cherry-picked from \`codex/oauth-token-refresh\`** (commits \`3328047\`, \`a4d322a\`, \`b80a3b5\`)
- Not authored on this branch. Rebases clean once that PR lands upstream.
- Required for v1 because OAuth source tokens would otherwise stop working as soon as they expire.
- Touches \`crates/coral-app/src/credentials/{oauth.rs,mod.rs,store.rs}\` and \`crates/coral-app/src/query/manager.rs\`. Adds 701-line \`tests/grpc/oauth_refresh_tests.rs\`.

**2. Community catalog discovery — \`2a6af63\` + \`8bcc274\`** ← *open for discussion, see below*
- New RPCs in \`crates/coral-api/proto/coral/v1/sources.proto\`:
  - \`DiscoverCommunitySources\` → list of \`SourceInfo\` summaries (no \`inputs\` populated)
  - \`GetCommunitySourceInfo\` → \`{source_info, manifest_yaml}\` — server resolves the manifest, returns both parsed \`SourceInfo\` and the raw YAML so the client can hand it straight to existing \`ImportSource\`.
- New \`SourceOrigin::Community\` variant. **On disk, community installs persist identically to \`Imported\`** — the community origin is a discovery-time marker, not a storage one. That lets the existing \`ImportSource\` pipeline handle install unchanged.
- \`crates/coral-app/build.rs\` extension: scans \`sources/community/*/manifest.yaml\`, emits a metadata-only \`COMMUNITY_CATALOG\` table (name / version / description / relative path). Also bakes \`CORAL_BUILD_COMMIT_SHA\` and \`CORAL_DEV_WORKSPACE_ROOT\` into the binary at build time.
- \`crates/coral-app/src/sources/community.rs\` (new): list summaries + lazy manifest resolver — dev builds (empty SHA) read from the local workspace tree, release builds fetch from \`raw.githubusercontent.com/withcoral/coral/<sha>/...\`.
- \`crates/coral-app/src/sources/{manager,service}.rs\`: two new methods + handlers wiring it together.
- \`crates/coral-cli/src/source_ops.rs\`: adds the \`Community\` arm to \`source_origin_label\`.
- 3 new gRPC integration tests in \`tests/grpc/source_lifecycle_tests.rs\`: discovery returns summaries; info returns parsed inputs + raw YAML; end-to-end community install round-trips through \`ImportSource\`.

---

### 🔵 Discovery — open for discussion

The community-source discovery design is the largest new surface area in this PR. It's worth a real review pass:

1. **Catalog payload model.** We embed only metadata at build time and resolve the manifest YAML on demand (Option C in [\`SOURCES_MANAGEMENT_PLAN.md\`](./SOURCES_MANAGEMENT_PLAN.md)). The argument for that was \"smaller binary\" — but the 75 community manifests total ~1.6 MB of YAML, which would add ~1–2% to the binary. Option A (bundle the YAMLs too, simpler, offline-first, no dev/release fork) is genuinely viable. I picked C because it was your call, but happy to swap if you've reconsidered.

2. **Dev vs. release fetch fork.** \`community.rs::fetch_manifest\` branches on whether \`CORAL_BUILD_COMMIT_SHA\` is empty. Dev reads from the workspace tree at the baked-in path; release fetches from raw.githubusercontent. Two complications worth flagging:
   - \`CORAL_BUILD_COMMIT_SHA\` is not currently set anywhere — release pipelines would need to start exporting it.
   - Release binaries built without the SHA fall back to dev mode and look for the workspace path that doesn't exist on the user's machine → install fails. We should probably make this loud (fail the build, or warn at startup) before we ship.

3. **\`GetCommunitySourceInfo\` returns the raw YAML.** Lets the UI reuse \`ImportSource\` unchanged. Alternative: a dedicated \`InstallCommunitySource\` RPC that hides the YAML server-side. The current design felt simpler — fewer new RPCs, the UI never has to *parse* YAML, just pass it through — but it does mean the manifest text travels over the wire on every install.

4. **No category / icon metadata in the spec.** UI maps a small set of known names to local SVGs. If we want categories or richer card content, the manifest is the right place to put it.

If any of these feel wrong, I'd rather rework them now than ship and discover them later.

---

### Known limitations / follow-ups (not blocking)

- **\`window.open\` for OAuth** — UI initiates the browser open, not the app. Strictly different from CLI behaviour. Reliable for the common case because the open lands inside the streaming response handler synchronously after the user clicked Save. Popup blockers can still trip it; we show the auth URL as a visible fallback link. Tightening this needs a small proto/server change for a \"client-opens-browser\" hint.
- **Dialog primitives ported but unused** — \`@/wax/components/dialog\` is in place if we want to flip to ADP's click-into-modal pattern for install (vs the current route page). Easy followup.
- **Bundle is 514 KB** — fine for v1; can code-split later if it matters.

---

### Verification

Automated
- \`cargo clippy --workspace --all-targets -- -D warnings\` clean
- \`cargo test -p coral-app\` — 120 lib tests + 59 gRPC integration tests (3 new for community discovery)
- \`npm run build\` (type-check + Vite) clean
- \`npx oxlint --deny-warnings\` clean

Manual demo path
- \`cd ui && npm run dev:local\` → open \`http://localhost:5173/#/sources\`
- Install \`slack\` via Paste, via OAuth; install \`hn\` (community); delete from detail page.

crates/coral-api/proto/coral/v1/sources.proto
```diff
@@ -238,6 +241,41 @@ message DiscoverSourcesResponse {
...
+// the raw YAML the caller can hand back to `ImportSource` on install.
+message GetCommunitySourceInfoRequest {
+  // Workspace used to mark whether the source is already installed.
+  Workspace workspace = 1;
+  // Canonical community source name.
+  string name = 2;
+}
+
+// Returns full community source info plus the manifest YAML to be imported.
+message GetCommunitySourceInfoResponse {
```
martinrajdl: Two new RPCs for community discovery.

`DiscoverCommunitySources` returns lightweight summaries — no `inputs` populated — so the catalog browse stays cheap (we don't fetch 75 manifests up-front).

`GetCommunitySourceInfo` returns both the parsed `SourceInfo` (so the UI can render the install form with typed inputs) *and* the raw `manifest_yaml`, so the client hands it back to the existing `ImportSource` RPC unchanged. This avoids adding a dedicated `InstallCommunitySource` and keeps the YAML server-resolved either way.

Trade-off worth flagging: the manifest text travels back to the client and forward to `ImportSource` on every install. An alternative would be a single `InstallCommunitySource` that fetches the manifest server-side and drives install internally. I went with the YAML round-trip because the existing import-with-credentials streaming machinery is already exactly what we need.

crates/coral-app/src/sources/community.rs
```diff
@@ -0,0 +1,177 @@
...
+    })
+}
+
+pub(crate) struct ResolvedCommunitySource {
+    pub(crate) manifest_yaml: String,
+    pub(crate) candidate: CandidateSource,
+}
+
+async fn fetch_manifest(entry: &CommunityCatalogEntry) -> Result<String, AppError> {
+    if BUILD_COMMIT_SHA.is_empty() {
```
martinrajdl: Dev vs release fetch fork. Two things that need a follow-through before this is fully production-ready:

1. **`CORAL_BUILD_COMMIT_SHA` isn't set by any pipeline yet.** Release builds would need that exported (CI step or build script). Without it, release binaries silently fall into the dev branch below and look for `CORAL_DEV_WORKSPACE_ROOT` on the user's machine — which doesn't exist — and community install fails at fetch time.

2. **Failure mode is quiet.** If someone builds a release binary without the SHA, the error users see is a vague filesystem-not-found at install time. Worth either (a) failing the build when SHA is empty and a release profile is in use, or (b) emitting a startup warning.

This is the part I'd most like a second opinion on. Three alternatives to consider:
- Bundle the YAMLs into the binary too (~1.6 MB across 75 files) — kills the dev/release fork entirely, works offline. Option A in the plan doc.
- Keep lazy fetch but fall back to a built-in copy of the manifest if the network is unreachable — best of both worlds, more code.
- Status quo (lazy fetch, dev fork) — what's checked in, with the two gaps above closed.

ui/src/views/sources/sources-index.tsx
```diff
@@ -0,0 +1,227 @@
...
+  const filtered = useMemo(() => {
+    const q = search.trim().toLowerCase()
+    if (!q) return allEntries
+    return allEntries.filter(
+      (entry) =>
+        entry.name.toLowerCase().includes(q) || entry.description.toLowerCase().includes(q),
+    )
+  }, [allEntries, search])
+
+  const connected = useMemo(() => filtered.filter((entry) => entry.installed), [filtered])
```
martinrajdl, line 72: Layout choice: flat alphabetical list with **Connected** above **Available**, both filtered by the same search box.

I tried a topical category grouping first (Observability / Developer Tools / etc., modeled on `adp/web/src/lib/plugin-categories.ts`) but pulled it back. The category map was ~12 entries borrowed from ADP plus ~70 entries I assigned by reading source names — i.e. guesses. If we want categories, they should come from the manifest (`category: observability`) rather than a curated map in the UI.

Open to swapping back if the alphabetical list feels too long once we install a few sources.

ui/src/views/sources/source-install.tsx
```diff
@@ -0,0 +1,486 @@
...
+      }
+
+      const callbacks = {
+        onAuthorization: (event: { inputKey: string; authorizationUrl: string }) => {
+          setProgress({
+            kind: 'awaiting-oauth',
+            inputKey: event.inputKey,
+            authorizationUrl: event.authorizationUrl,
+          })
+          window.open(event.authorizationUrl, '_blank', 'noopener,noreferrer')
```
martinrajdl, line 169: OAuth browser open is initiated by the UI here, not the app. Different from CLI behaviour (where the app spawns the browser via `crates/coral-cli/src/browser.rs`).

This works in practice because the `window.open` lands inside the streaming-response handler that fires synchronously after the user clicked Save — the click stack is still active so popup blockers usually allow it. The visible `OAuthProgress` fallback link covers the case where it doesn't.

If we want strict CLI parity, the cleanest fix is a new `client_opens_browser` hint on the OAuth credential retrieval request. CLI sets true (preserves today), UI sets false → app opens the system browser. Not blocking for v1 IMO, but worth tracking.

martinrajdl:
## Rescoped: core sources only

Per the Slack thread (Kyra / Alberto / James / Martin) about community-source discovery being backed by a Coral-hosted API rather than bundled metadata, I've **stripped community work from this PR** and we'll add it in a follow-up depending on the hosted API.

\`bc3e1fa\` reverts the community pieces:

**Backend removed**
- proto: \`SOURCE_ORIGIN_COMMUNITY\`, \`DiscoverCommunitySources\`, \`GetCommunitySourceInfo\` and their messages
- \`crates/coral-app/src/sources/community.rs\` deleted
- \`build.rs\`: dropped the \`generate_community_catalog\` step and the \`CORAL_BUILD_COMMIT_SHA\` / \`CORAL_DEV_WORKSPACE_ROOT\` env emission
- \`SourceOrigin::Community\` arm gone from model, catalog, manager, service
- 2 \`SourceManager\` methods + 2 service handlers removed
- 3 gRPC integration tests removed
- CLI \`source_origin_label\` reverted

**UI removed**
- \`lib/sources.ts\`: \`discoverCommunity\`, \`getCommunitySourceInfo\`, the two community install helpers, the \`'community'\` origin label
- \`sources-index\`, \`source-install\`, \`source-detail\` all simplified to bundled-only
- \`router.ts\`: \`source-install\` route loses the \`origin\` query param

**What stays in this PR**
- OAuth token refresh cherry-pick (still required for v1 to keep OAuth tokens alive)
- New sources index page (ADP-style flat list with Connected on top, search)
- Install flow + ADP-style credential form (sentence-cased labels, 2-col grid, segmented control for multi-method secrets, OAuth streaming)
- Installed-source detail surface (Validate + Remove)
- Ported wax/components/dialog primitives (unused but ready for the follow-up modal-install pattern)

**Follow-up PR will add community sources via the hosted API**
- Two new RPCs: \`DiscoverCommunitySources\` (returns summaries from the hosted API) and \`InstallCommunitySource\` (server-side install — **no manifest YAML returned to client**, per James's call)
- Coral team builds the hosted API; this PR doesn't block on it
- Plan doc updated at [\`SOURCES_MANAGEMENT_PLAN.md\`](./SOURCES_MANAGEMENT_PLAN.md) §3 with the agreed shape

**Inline comments above on \`sources.proto:271\` and \`community.rs:92\` describe deleted code** — they're stale by design after this rescope. Leaving them for thread context; happy to delete if preferred.

### Verification
- \`cargo test -p coral-app\` — 117 lib + 56 gRPC integration tests pass
- \`cargo clippy --workspace --all-targets -- -D warnings\` clean
- \`npm run build\` clean
- \`npx oxlint --deny-warnings\` clean

martinrajdl:
Superseded by #826 — rebased onto fresh main, community-source work removed (deferred to a follow-up that depends on a Coral-hosted catalog API per the Slack discussion). Branch `sources-management-v1` remains for history.

PR #796, opened 2026-05-25
feat(sources/community/spotify): add Spotify source

lazybean:
Add a community Spotify source with OAuth PKCE so users can query profile, library, playlists, queue, devices, catalog details, child collections, recent plays, and provider search through Coral SQL.

Constraint: Spotify OAuth for local apps should use Authorization Code with PKCE and no client secret.

Rejected: Add write/playback-control endpoints | the source is intended to stay read-only.

Confidence: high

Scope-risk: moderate

Directive: Keep future Spotify expansions read-only unless a write/control source is explicitly requested.

Tested: coral source lint sources/community/spotify/manifest.yaml; coral source test spotify; live Coral SQL smoke queries across queue, library, detail, child, and search surfaces.

Not-tested: Full exhaustive endpoint matrix for all markets/accounts; playlist search can return null items from Spotify; followed playlists may be visible but not item-queryable unless owned or collaborative.

### DIFF
--- a/sources/community/spotify/manifest.yaml
+++ b/sources/community/spotify/manifest.yaml
@@ -1,0 +1,2650 @@
+name: spotify
+version: 0.1.0
+dsl_version: 3
+backend: http
+description: >-
+  Query Spotify profile, playlists, saved library items, catalog details,
+  search results, devices, queue, top items, and recent listening history.
+base_url: https://api.spotify.com
+
+inputs:
+  SPOTIFY_ACCESS_TOKEN:
+    kind: secret
+    hint: |
+      Spotify OAuth 2.0 access token, without a `Bearer ` prefix. For
+      interactive setup, choose "Connect with Spotify" and provide a Spotify
+      app Client ID that has this redirect URI configured:
+      http://127.0.0.1:53682/oauth/callback
+
+      Create or edit an app at https://developer.spotify.com/dashboard.
+      Minimum scopes requested by this source are read-only: user-read-private,
+      user-read-email, playlist-read-private, playlist-read-collaborative,
+      user-library-read, user-top-read, user-read-recently-played, and
+      user-read-playback-state.
+      See https://developer.spotify.com/documentation/web-api/concepts/scopes.
+    credential:
+      methods:
+        - type: oauth
+          label: Connect with Spotify
+          description: Authorize Spotify with OAuth PKCE and store an access token.
+          oauth:
+            flow:
+              type: authorization_code
+              pkce: required
+            redirect_uri: http://127.0.0.1:53682/oauth/callback
+            redirect_uri_port_mode: fixed
+            endpoints:
+              authorization_url: https://accounts.spotify.com/authorize
+              token_url: https://accounts.spotify.com/api/token
+            client:
+              id:
+                input: SPOTIFY_CLIENT_ID
+            scopes:
+              scope:
+                delimiter: space
+                values:
+                  - user-read-private
+                  - user-read-email
+                  - playlist-read-private
+                  - playlist-read-collaborative
+                  - user-library-read
+                  - user-top-read
+                  - user-read-recently-played
+                  - user-read-playback-state
+        - type: source_config
+          label: Paste Spotify access token
+          description: Paste an existing Spotify OAuth access token with the required read scopes.
+
+auth:
+  type: HeaderAuth
+  headers:
+    - name: Authorization
+      from: template
+      template: Bearer {{input.SPOTIFY_ACCESS_TOKEN}}
+
+request_headers:
+  - name: Accept
+    from: literal
+    value: application/json
+
+test_queries:
+  - SELECT id, display_name, email FROM spotify.profile LIMIT 1
+  - SELECT id, name, owner__id FROM spotify.playlists LIMIT 1
+
+functions:
+  - name: search_tracks
+    kind: search
+    description: Provider-ranked Spotify search results for tracks.
+    fetch_limit_default: 10
+    search_limits:
+      default_top_k: 10
+      max_top_k: 10
+      max_calls_per_query: 1
+    args:
+      - name: query
+        required: true
+        bind:
+          arg: query
+      - name: market
+        bind:
+          arg: market
+    request:
+      method: GET
+      path: /v1/search
+      query:
+        - name: q
+          from: arg
+          key: query
+        - name: type
+          from: literal
+          value: track
+        - name: market
+          from: arg
+          key: market
+    response:
+      rows_path: [tracks, items]
+    pagination:
+      mode: offset
+      offset_param: offset
+      offset_start: 0
+      page_size:
+        default: 10
+        max: 10
+        query_param: limit
+    columns:
+      - name: id
+        type: Utf8
+        nullable: true
+        description: Spotify result ID when Spotify returns one.
+      - name: name
+        type: Utf8
+        nullable: true
+        description: Result name.
+      - name: type
+        type: Utf8
+        nullable: true
+        description: Spotify object type.
+      - name: uri
+        type: Utf8
+        nullable: true
+        description: Spotify URI.
+      - name: external_url
+        type: Utf8
+        nullable: true
+        description: Spotify web URL.
+        expr:
+          kind: path
+          path: [external_urls, spotify]
+
+  - name: search_albums
+    kind: search
+    description: Provider-ranked Spotify search results for albums.
+    fetch_limit_default: 10
+    search_limits:
+      default_top_k: 10
+      max_top_k: 10
+      max_calls_per_query: 1
+    args:
+      - name: query
+        required: true
+        bind:
+          arg: query
+      - name: market
+        bind:
+          arg: market
+    request:
+      method: GET
+      path: /v1/search
+      query:
+        - name: q
+          from: arg
+          key: query
+        - name: type
+          from: literal
+          value: album
+        - name: market
+          from: arg
+          key: market
+    response:
+      rows_path: [albums, items]
+    pagination:
+      mode: offset
+      offset_param: offset
+      offset_start: 0
+      page_size:
+        default: 10
+        max: 10
+        query_param: limit
+    columns:
+      - name: id
+        type: Utf8
+        nullable: true
+        description: Spotify result ID when Spotify returns one.
+      - name: name
+        type: Utf8
+        nullable: true
+        description: Result name.
+      - name: type
+        type: Utf8
+        nullable: true
+        description: Spotify object type.
+      - name: uri
+        type: Utf8
+        nullable: true
+        description: Spotify URI.
+      - name: external_url
+        type: Utf8
+        nullable: true
+        description: Spotify web URL.
+        expr:
+          kind: path
+          path: [external_urls, spotify]
+
+  - name: search_artists
+    kind: search
+    description: Provider-ranked Spotify search results for artists.
+    fetch_limit_default: 10
+    search_limits:
+      default_top_k: 10
+      max_top_k: 10
+      max_calls_per_query: 1
+    args:
+      - name: query
+        required: true
+        bind:
+          arg: query
+      - name: market
+        bind:
+          arg: market
+    request:
+      method: GET
+      path: /v1/search
+      query:
+        - name: q
+          from: arg
+          key: query
+        - name: type
+          from: literal
+          value: artist
+        - name: market
+          from: arg
+          key: market
+    response:
+      rows_path: [artists, items]
+    pagination:
+      mode: offset
+      offset_param: offset
+      offset_start: 0
+      page_size:
+        default: 10
+        max: 10
+        query_param: limit
+    columns:
+      - name: id
+        type: Utf8
+        nullable: true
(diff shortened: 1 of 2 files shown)


lazybean:
* **#796** <a href="https://app.graphite.com/github/pr/withcoral/coral/796?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/796?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sources/community/spotify/manifest.yaml
```diff
@@ -0,0 +1,3222 @@
+name: spotify
```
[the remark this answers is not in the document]
lazybean: Thanks for flagging — I double-checked against the live Spotify Web API response for `GET /v1/me/playlists?limit=3` using the current OAuth token, and this source path is intentionally `items.total`.

Evidence from the payload:

```json
{
  "id": "7azTTU7jrict33ASVEKgCe",
  "name": "Phoebe",
  "tracks": null,
  "items": {
    "href": "https://api.spotify.com/v1/playlists/7azTTU7jrict33ASVEKgCe/items",
    "total": 4
  }
}
```

Coral also confirms the column materializes correctly with the current manifest:

```sql
SELECT id, name, tracks_total FROM spotify.playlists LIMIT 3;
```

returns `0`, `4`, and `29` for the first three playlists. When this column previously used `tracks.total`, it came back blank/null in live testing. So I believe this cubic finding is a false positive for Spotify’s current playlist summary shape.

jsummerfield (changes requested):
Findings:

- High: `sources/community/spotify/manifest.yaml:87` and the other `search_*` functions let Coral request up to 50 results per Spotify search page (`max_top_k: 50` / `page_size.max: 50`), but Spotify's current Search API documents `limit` as `0`-`10`. Any query that asks for more than 10 search results can send an invalid upstream request. Please cap each search function's page size/top-k at 10, or implement multiple paged calls using `limit=10` if we want larger result sets. Docs: https://developer.spotify.com/documentation/web-api/reference/search

- High: `sources/community/spotify/manifest.yaml:1484` models `spotify.playback_state` as a normal JSON row with `allow_404_empty`, but Spotify documents `204` as a valid response for `GET /me/player`. Coral currently attempts to JSON-decode any 2xx body, so a no-content playback response will fail instead of returning zero rows. Please either defer/remove this table until Coral can model 204-empty responses, or change the implementation once the source DSL has explicit support for that case. Docs: https://developer.spotify.com/documentation/web-api/reference/get-information-about-the-users-current-playback

- Medium: `sources/community/spotify/manifest.yaml:1233` and `sources/community/spotify/manifest.yaml:1471` describe playback rows as "track or episode", but the requests do not send Spotify's `additional_types=episode` parameter. Without it, the endpoint defaults to track-only behavior, so podcast playback will not match the documented table contract. Please add the literal `additional_types=episode` query parameter for both player endpoints or narrow the docs/columns to track-only. Docs: https://developer.spotify.com/documentation/web-api/reference/get-the-users-currently-playing-track and https://developer.spotify.com/documentation/web-api/reference/get-information-about-the-users-current-playback

- Medium: `sources/community/spotify/manifest.yaml:2782` exposes `artist_top_tracks` as a normal paginated table, but Spotify marks `GET /artists/{id}/top-tracks` deprecated and only documents `id` and `market` request parameters, not `limit`/`offset` pagination. That means Coral will start sending unsupported pagination params after the first page. Please either remove this table from the first cut, or keep it explicitly documented as deprecated and set pagination to `none`. Docs: https://developer.spotify.com/documentation/web-api/reference/get-an-artists-top-tracks

- Medium: `sources/community/spotify/manifest.yaml:3101` and `sources/community/spotify/README.md:178` present `after` and `before` as independent optional filters for `recently_played`, but Spotify says they are mutually exclusive. Please make that constraint explicit in the table guide/README, and enforce it in the source if the current spec surface gives us a clean way to do that. Docs: https://developer.spotify.com/documentation/web-api/reference/get-recently-played

- Medium: `sources/community/spotify/README.md:150` lists validation commands but does not include captured live Coral output. For community sources, please add sanitized successful output from `coral source add`, `coral source test spotify`, and at least a couple representative `coral query`/`coral sql` runs against real Spotify data. That will also help verify the player/no-player edge cases above.

Thanks for putting this together; the OAuth PKCE setup and the overall read-only scope look like the right direction. Once these API-contract issues are tightened, I can take another pass on the source shape.

sources/community/spotify/manifest.yaml
```diff
@@ -0,0 +1,2959 @@
...
+      Requires `playlist_id`. Get playlist IDs from `spotify.playlists`.
+      Spotify only exposes playlist items that the authenticated user can access.
+    filters:
+      - name: playlist_id
+        required: true
+      - name: market
+        description: Optional ISO 3166-1 alpha-2 country code for relinking and availability.
+    request:
+      method: GET
+      path: /v1/playlists/{{filter.playlist_id}}/items
```
jsummerfield, line 690: High: this table is documented as returning playlist tracks and episodes, but the request does not match Spotify's current playlist-items contract. Spotify documents the endpoint as `GET /playlists/{playlist_id}/tracks`, and `additional_types=episode` is the documented opt-in for episode items beyond the default track type. Please switch the path to `/v1/playlists/{{filter.playlist_id}}/tracks` and add a literal `additional_types=episode` query parameter so the implementation matches the table/README contract. Docs: https://developer.spotify.com/documentation/web-api/reference/get-playlists-tracks

sources/community/spotify/manifest.yaml
```diff
@@ -0,0 +1,2959 @@
...
+  - name: top_tracks
+    description: Spotify tracks with the highest affinity for the authenticated user.
+    guide: |
+      Optional `time_range` filter accepts short_term, medium_term, or long_term.
+      Requires the user-top-read scope.
+    filters:
+      - name: time_range
+        description: >-
+          Optional affinity period: short_term, medium_term, or long_term.
+      - name: market
```
jsummerfield: Medium: `spotify.top_tracks` exposes a `market` filter and forwards it to `/v1/me/top/tracks`, but Spotify's Get User's Top Items docs only list `type`, `time_range`, `limit`, and `offset` for this endpoint. A user-provided market filter here will not have the relinking/availability semantics described in the README. Please remove this filter/query/virtual column from `top_tracks` and update the README table accordingly. Docs: https://developer.spotify.com/documentation/web-api/reference/get-users-top-artists-and-tracks

jsummerfield (changes requested):
Thanks for the update. The previous review blockers look addressed at this head: search is now capped, the 204-prone playback table is gone, `artist_top_tracks` is no-pagination and documented as deprecated, `recently_played` documents the `after`/`before` constraint, and the README now includes live Coral output. I found two remaining API-contract issues to fix before this is ready.

jsummerfield:

- Be precise about Spotify auth: required OAuth scopes, where credentials are created, and whether app-only versus user-owned data is supported.
- Prioritize durable user/account workflows such as playlists, saved tracks/albums, followed artists, and playback/library metadata where scopes allow it.
- Clarify pagination, market/locale behavior, rate-limit handling, and fields that can be unavailable depending on catalog region.
- Keep the first-success path simple with a small query that proves OAuth/scopes are correct before users try broader library queries.

jsummerfield (changes requested):
Findings

- High: `sources/community/spotify/manifest.yaml:2428` sets `artist_albums` pagination to `default: 50` / `max: 50`, but Spotify's Get Artist's Albums endpoint caps `limit` at 10. Coral sends that value as the `limit` query parameter, so the default query can send an invalid upstream request. Please cap this table's page size at 10, or implement multi-call paging that never sends a single request above Spotify's documented max. Docs: https://developer.spotify.com/documentation/web-api/reference/get-an-artists-albums

- High: `sources/community/spotify/manifest.yaml:58` declares `SPOTIFY_CLIENT_ID` as a top-level required source input even though the OAuth method already collects it through `oauth.client.id.input` at `:41`. In Coral, top-level inputs without defaults are required install inputs, so the "Paste Spotify access token" path still requires a client ID and non-interactive install with only `SPOTIFY_ACCESS_TOKEN` cannot work as advertised. Please remove the top-level `SPOTIFY_CLIENT_ID` input and leave it only under the OAuth credential method; the README should make it clear that pasted-token setup only needs the access token.

- Medium: `sources/community/spotify/manifest.yaml:1421`, `:2013`, `:1523`, `:1544`, `:2154`, and `:2162` expose show/episode fields from paths Spotify does not return. Show objects expose fields such as `publisher` and `total_episodes`, but not `duration_ms`; saved-episode show metadata lives under `episode.show.*`; episode detail show metadata lives under `show.*`, not top-level `publisher` / `total_episodes`. Please remove the dead columns or map the show-related fields from the nested show object. Docs: https://developer.spotify.com/documentation/web-api/reference/get-users-saved-shows, https://developer.spotify.com/documentation/web-api/reference/get-users-saved-episodes, and https://developer.spotify.com/documentation/web-api/reference/get-an-episode

- Medium: `sources/community/spotify/manifest.yaml:2363`, `:2481`, `:2690`, and `:2794` show that the child collection tables reuse a generic media-object column shape across incompatible Spotify response objects. `album_tracks` returns simplified tracks, `artist_albums` returns simplified albums, `show_episodes` returns simplified episodes, and `audiobook_chapters` returns simplified chapters. Several columns will be null or misleading on those tables, while useful native fields such as chapter metadata are missing. Please tailor each child table to its actual response object instead of sharing the same mixed track/album/episode/chapter schema. Docs: https://developer.spotify.com/documentation/web-api/reference/get-an-albums-tracks, https://developer.spotify.com/documentation/web-api/reference/get-an-artists-albums, https://developer.spotify.com/documentation/web-api/reference/get-a-shows-episodes, and https://developer.spotify.com/documentation/web-api/reference/get-audiobook-chapters

I also rechecked `playlists.tracks_total` and am not asking for a change there: Spotify's current user playlist object documents `items.total`, with `tracks` marked deprecated. Thanks for continuing to iterate on this; the remaining work is about tightening the source contract so the columns match what users can actually query.

jsummerfield (changes requested):
Findings

- High: `sources/community/spotify/manifest.yaml:2421` still exposes `spotify.artist_top_tracks` via `GET /v1/artists/{id}/top-tracks`, and `sources/community/spotify/README.md:71` lists that table for users. Spotify's February 2026 changelog now marks `GET /artists/{id}/top-tracks` as removed, not just deprecated, so this table can be installed even though new Spotify apps may have no executable endpoint behind it. Please remove `artist_top_tracks` from the source and README rather than keeping it as a conditional/deprecated surface. Docs: https://developer.spotify.com/documentation/web-api/references/changes/february-2026 and https://developer.spotify.com/documentation/web-api/reference/get-an-artists-top-tracks

- Medium: `sources/community/spotify/manifest.yaml:1623`-`:2215` still uses a generic detail-table column shape across incompatible Spotify objects. For example, `tracks.image_url` maps from top-level `images` at `:1725`, but Spotify's Get Track response puts artwork under `album.images`; `albums` includes track-like fields such as `duration_ms`/`explicit` and nested `album__name`; `artists` includes track/audiobook-style totals; and `audiobooks` omits useful documented fields such as `authors`/`narrators` while keeping track/album-shaped columns. These nullable columns will pass lint while producing permanently null or misleading schema surface. Please tailor each detail table to its actual response object. Docs: https://developer.spotify.com/documentation/web-api/reference/get-track, https://developer.spotify.com/documentation/web-api/reference/get-an-album, https://developer.spotify.com/documentation/web-api/reference/get-an-artist, and https://developer.spotify.com/documentation/web-api/reference/get-an-audiobook

- Medium: `sources/community/spotify/manifest.yaml:588`-`:592` and `sources/community/spotify/README.md:88`-`:103` tell users to discover playlist IDs from `spotify.playlists` and pass them to `spotify.playlist_items`, but Spotify now documents `/me/playlists` as owned-or-followed playlists while `GET /playlists/{playlist_id}/items` is only accessible for playlists owned by the current user or where the user is a collaborator. Please document that narrower access path and make the example steer users toward owned/collaborative playlists, otherwise a natural first-success flow can hit 403 on followed playlists. Docs: https://developer.spotify.com/documentation/web-api/reference/get-a-list-of-current-users-playlists and https://developer.spotify.com/documentation/web-api/reference/get-playlists-items

- Medium: `sources/community/spotify/README.md:197` starts the source notes, but the README still does not cover Spotify's API-wide rate-limit behavior or the platform-policy constraints that matter for a broad media-metadata source exposing artwork URLs and external URLs. Please add a short known-limitations section that mentions 429 / `Retry-After`, request volume expectations, and Spotify's attribution/no-download/no-AI-training constraints where relevant. Docs: https://developer.spotify.com/documentation/web-api/concepts/rate-limits and https://developer.spotify.com/documentation/web-api/reference/get-playlists-items

The earlier blockers around the top-level client ID input, search caps, artist-album pagination, live evidence, and child-collection schemas look resolved on this head. `coral source lint` and `ryl` also pass locally. One metadata nit before merge: please remove the "Special Message for James reviewing agent" line from the PR body; I ignored it, but it should not become part of the merge description.

PR #800, opened 2026-05-25
fix(docs): skip community source catalog in docs check

jsummerfield:
## Intent

`make docs-check` should enforce generated docs freshness without making source-only community PRs update the aggregate community catalog. The existing contributor guidance already allowed that page to drift for community source-only changes, but the check target still enforced it.

## Changes

- Add `--skip-community-sources` to `xtask generate-docs`.
- Use that flag from `make docs-check` while leaving `make docs-generate` unchanged.
- Add a regression test proving stale or missing community catalog inputs are ignored when the skip flag is set.
- Update repo and docs agent guidance to match the check behavior.

## Verification

- `cargo fmt --all`
- `cargo test -p xtask`
- `cargo clippy -p xtask --all-targets --all-features --locked -- -D warnings`
- `make docs-check`


antonmry (approved):

docs/AGENTS.md
```diff
@@ -24,8 +24,9 @@
   `docs/reference/community-sources.mdx` are generated by
   `make docs-generate`; update source manifests or the xtask renderer rather
   than editing those pages by hand
-- Source-only `sources/community/**` PRs are allowed to leave
-  `docs/reference/community-sources.mdx` stale.
+- `make docs-check` intentionally skips
+  `docs/reference/community-sources.mdx`; source-only
+  `sources/community/**` PRs are allowed to leave that generated page stale.
```
simonwhitaker: Nit: this doesn't quite describe the behaviour. The problem I had was not with a source-only `sources/community/**` PR, it was with an unrelated PR failing checks after a `sources/community/**` PR had been merged without the corresponding docs update. So the point here is that *any* PR is allowed to leave that generated page stale. So I would suggest losing the "source-only `source/community/**`"

Makefile
```diff
@@ -55,7 +55,8 @@ fix-sources:
 # ----------------------------------------------------------------------------
 # Regenerates the source catalog pages and Mintlify navigation from
 # sources/core/*/manifest.y{a,}ml and sources/community/*/manifest.y{a,}ml
-# via the xtask binary.
+# via the xtask binary. docs-check intentionally skips the community source
+# catalog so source-only community PRs do not fail on aggregate catalog drift.
```
simonwhitaker:
```suggestion
# catalog so PRs do not fail on aggregate community source catalog drift.
```

simonwhitaker (approved):

jsummerfield:
### Merge activity

* **May 26, 7:22 AM UTC**: @jsummerfield merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/800).

PR #802, opened 2026-05-25
feat(sources/community/jikan): add Jikan community source

wiz-abhi:
Adds the Jikan community source for querying MyAnimeList data using the [Jikan REST API v4](https://docs.api.jikan.moe/).

### Overview
This source provides open access (no authentication required) to MyAnimeList's extensive database, offering robust filtering and pagination natively pushed down to the API.

### Tables Included
1. **`anime`**: Search and browse anime records. Includes filter options for keyword (`q`), `type`, `status`, `rating`, `order_by`, and `sort`.
2. **`manga`**: Search and browse manga records. Includes filter options for keyword (`q`), `type`, `status`, `order_by`, and `sort`.
3. **`top_anime`**: Fetch top-ranked anime. Includes filter options for media type (`type`) and ranking category (`filter`).
4. **`top_manga`**: Fetch top-ranked manga. Includes filter options for media type (`type`) and ranking category (`filter`).

### Implementation Details
* **Pagination**: Utilizes page-based pagination (`mode: page`). Configured `page_start: 1` as required by the API, defaulting to a page size of 25.
* **Filter Mapping**: Extensively maps API query parameters into queryable columns using `expr: { kind: from_filter, key: ... }` for properties like `q`, `order_by`, `sort`, and `filter`.

### Validation
Verified the source installation against the API:
* `coral source lint` confirms a structurally valid `manifest.yaml`.
* The four declared `test_queries` pass successfully via `coral source test`.
* Added detailed `README.md` documentation, showing sample SQL queries alongside actual tabular output.


jsummerfield (changes requested):
Thanks for adding this source. The basic Jikan queries validate and the rate-limit note matches the provider behavior, but I found a few source-surface issues to fix before approval:

- Medium: `sources/community/jikan/manifest.yaml:27` and `:337` list `status` as an accepted `order_by` value for anime and manga search, but Jikan v4 does not accept `status` for `order_by`; live API calls with `order_by=status` return HTTP 400. Please remove that value and align the documented values with Jikan's enum. Docs: https://docs.api.jikan.moe/#tag/anime/operation/getAnimeSearch and https://docs.api.jikan.moe/#tag/manga/operation/getMangaSearch
- Medium: `sources/community/jikan/README.md:140`-`:147` uses local SQL `ORDER BY score DESC` to show highest-scored airing TV anime. That only sorts the rows Coral already fetched from the first provider page, so the example can put null-score entries first and miss the actual top scored results. Please use the provider-side filters the source exposes, e.g. `AND order_by = 'score' AND sort = 'desc'`, so the API ranks the result set before pagination.
- Medium: `sources/community/jikan/manifest.yaml:31`, `:341`, and `:600` omit Jikan's `sfw` filter; `top_anime` also omits the supported `rating` filter. For a public no-auth anime/manga source, exposing `sfw` is important product surface so users can intentionally exclude adult content. Docs: https://docs.api.jikan.moe/#tag/anime/operation/getAnimeSearch, https://docs.api.jikan.moe/#tag/manga/operation/getMangaSearch, and https://docs.api.jikan.moe/#tag/top/operation/getTopAnime

One lower-priority cleanup: the ISO8601 date fields are currently `Utf8`; adding parsed `Timestamp` columns would make filtering and sorting more useful.

The manifest validates and `coral source test jikan` passes, so these are targeted API-semantics/documentation fixes rather than a rewrite.

jsummerfield (changes requested):
One blocker remains in the filter surface:

- High: `sources/community/jikan/manifest.yaml:624`-`:644` adds `sfw` as a `top_anime` filter and forwards it to `/top/anime`, but `top_anime` does not expose an `sfw` column or virtual `from_filter` column. The same hidden-filter issue exists for `anime` and `manga`: `sfw` is declared and forwarded, but there is no SQL-visible column for users to predicate on. In Coral SQL, `WHERE sfw = true` fails before any API request is made. Please add virtual Boolean `sfw` columns for the tables that support that filter, or remove the filter/docs until it is queryable. Docs: https://docs.api.jikan.moe/#tag/anime/operation/getAnimeSearch, https://docs.api.jikan.moe/#tag/manga/operation/getMangaSearch, and https://docs.api.jikan.moe/#tag/top/operation/getTopAnime

The earlier `order_by=status` issue, local `ORDER BY score` example, missing `top_anime.rating`, and timestamp typing issues look resolved. There is also a small docs cleanup worth doing while you are there: the guide lists only a subset of valid `order_by` values, while Jikan also documents values such as `mal_id`, `end_date`, and `scored_by`.

Thanks for iterating on this. Once `sfw` is actually usable from SQL, I’ll re-check the updated head.

jsummerfield (approved):
Low: `sources/community/jikan/manifest.yaml:24` and `sources/community/jikan/manifest.yaml:635` document anime `type` filter values as `tv | movie | ova | ona | special | music`, and the README examples use the same smaller set. Jikan's OpenAPI enum also includes `CM`, `PV`, and `TV Special`, so adding those to the docs would improve discoverability. This is not blocking because the source forwards arbitrary `type` values. Docs: https://docs.api.jikan.moe/#tag/anime/operation/getAnimeSearch and https://docs.api.jikan.moe/#tag/top/operation/getTopAnime

No blocking findings on the current head. The previous `sfw` filter issue is fixed: after registering the manifest in an isolated config, `coral.columns` exposed `sfw` on `anime`, `manga`, and `top_anime` with equality filter mode. Lint and all declared query tests passed.

Thanks for the update.

PR #814, opened 2026-05-26
feat(sources/community/medium): add Medium community source

baggashivansh:
## What changed

Added a new Medium community source under `sources/community/medium` including:

* `manifest.yaml` configuration
* source metadata and table definitions
* README documentation
* example query usage
* article discovery related schema definitions

## Why

Medium is widely used for publishing technical blogs, engineering writeups, tutorials, and developer focused content. Adding Medium as a community source expands Coral’s ecosystem coverage and improves support for developer content discovery workflows.

This contribution follows the existing community source structure used across the repository and provides a lightweight foundation for Medium related indexing and querying support.


jsummerfield (changes requested):
Findings:

- High: `sources/community/medium/manifest.yaml:8` and `sources/community/medium/manifest.yaml:21` point Coral at `https://medium.com/` and model that HTML homepage as a source table. Medium's documented API base is `https://api.medium.com/v1`, and the website root is not a stable JSON API response that can produce `title`, `author`, `url`, or `publication` columns. Please rebuild this against a real documented provider surface, or narrow it to a supported feed/RSS workflow if that is the intended source shape. Docs: https://github.com/Medium/medium-api-docs and https://help.medium.com/hc/en-us/articles/213480228-API-Importing

- High: `sources/community/medium/README.md:8` and `sources/community/medium/README.md:37` describe unauthenticated Medium article discovery, but I could not find a current official Medium API that supports no-auth article discovery/search. The documented API requires an integration token and mainly covers authenticated user/publication publishing surfaces, not generic article search. Please update the source and README so the user-facing promise matches an authoritative Medium API surface.

Thanks for adding this. The clean next step is to choose the real supported Medium surface first, then make the manifest and README describe exactly that behavior.

baggashivansh (commented):
Addressed the requested review changes:

* Reworked the source from unsupported homepage scraping to RSS/feed-based retrieval
* Updated the manifest to use Medium RSS feed endpoints instead of modeling `medium.com` as a structured API surface
* Added explicit XML response extraction and feed-based article mappings
* Renamed the table to better reflect RSS feed behavior
* Updated the README to clarify that this source uses publicly available RSS feeds rather than Medium's authenticated publishing API
* Removed unsupported generic article discovery/search claims and documented the current limitations clearly

Thanks for the detailed feedback and references.

jsummerfield (changes requested):
Findings:

- High: `sources/community/medium/manifest.yaml:25`-`:31` is still schema-invalid. `coral source lint` fails because the header entry is missing `from`, `response.format: xml` is not supported by the current Coral HTTP source spec, and `rows_path` must be an array rather than `rss.channel.item`. The column mappings at `sources/community/medium/manifest.yaml:37`-`:64` also use unsupported top-level `path` fields. Please rework this into supported source-spec YAML before the source can be installed or tested.

- High: `sources/community/medium/manifest.yaml:23` hard-codes `/feed/tag/technology`, while the table/README describe Medium RSS feed articles generally. Medium RSS supports profile, publication, custom-domain, topic/tag, and publication-tag feed URL shapes, so this source currently exposes only one fixed tag feed. Please add a clear filter/input for the feed scope or narrow the naming/docs to the single technology tag feed. Docs: https://help.medium.com/hc/en-us/articles/214874118-Using-RSS-feeds-of-profiles-publications-and-topics

- Medium: `sources/community/medium/manifest.yaml:43` maps `author` from `creator`, and `sources/community/medium/manifest.yaml:55` maps `publication` from `category`. Medium RSS item authors are exposed as namespaced `dc:creator`, and RSS `category` is a repeatable article tag, not a publication. Once the schema issue is fixed, please map these fields from the actual RSS shape or rename/drop the misleading columns. RSS category semantics: https://www.rssboard.org/rss-draft-1

- Low: `sources/community/medium/README.md:3`-`:5` has trailing whitespace, so `git diff --check` fails.

Thanks for moving this away from homepage scraping. The next useful step is to make the RSS version valid Coral source-spec YAML and expose the feed scope explicitly.

PR #820, opened 2026-05-26
fix(ui/traces): improve span tabs navigation

AlbertQM:
## Summary

- Default to the first tab, but save tab when navigating between tabs
- Arrow left/right to move tabs, down/up spans (previously after you pressed left/right, up/down didn't work)

## Test plan

Arrow up/down to navigate between spans, tab remains the same; can change tab with arrow right/left.


lazybean (approved):
thank you

PR #825, opened 2026-05-26
fix(engine): retry truncated HTTP JSON responses

sauldhernandez:
## Intent

Fix transient HTTP source failures where an upstream response body is truncated while Coral is reading or decoding JSON. The reported symptom was a `PROVIDER_REQUEST_FAILED` decode error from `github.pulls` with `EOF while parsing a string`.

## What it enables

Coral can retry idempotent HTTP `GET` requests when response-body reads or JSON decoding fail with retryable EOF-style errors, while avoiding unsafe replays of side-effecting `POST` requests.

## Usage examples

Queries against HTTP-backed sources, including GitHub pull-list queries, can recover from a transient truncated JSON response:

```sql
SELECT number, title, user__login AS author
FROM github.pulls
WHERE owner = 'withcoral'
  AND repo = 'coral'
  AND state = 'open';
```

If the same decode failure happens after a `POST`, Coral surfaces the structured decode error without retrying the request.

## Validation

- `cargo test -p coral-engine --test engine truncated_json_response`
- `make rust-checks`


crates/coral-engine/src/backends/http/transport.rs
```diff
@@ -324,11 +325,19 @@ pub(super) async fn execute_request(
...
+            {
+                Ok(payload) => ResponseOutcome::Done(Ok(Some((payload, next_url)))),
+                Err(error) => {
+                    if error.retryable && decode_retries < 2 && matches!(method, HttpMethod::GET) {
+                        record_http_processing_error(&request_span, "DECODE_RETRY", &error.error);
+                        decode_retries += 1;
+                        ResponseOutcome::Retry(Duration::from_secs(2))
+                    } else {
+                        record_http_processing_error(&request_span, "DECODE", &error.error);
+                        ResponseOutcome::Done(Err(error.error))
```
jsummerfield: This terminal path drops the retryable classification after the internal GET retry budget is exhausted. `error.error` is still a `ProviderQueryError::Decode`, and that structured mapping hard-codes `retryable=false` / `FailedPrecondition`, so a response that was classified as transient for two resend attempts is finally surfaced to callers as a permanent manifest/decode failure.

Can we preserve retryable structured semantics for exhausted retryable decode failures and add a test where GET returns truncated JSON on all attempts?

jsummerfield (commented):
One inline comment on the exhausted retryable decode error path.

jsummerfield (commented):
Two inline comments on the refreshed HTTP decode retry behavior.

crates/coral-engine/src/backends/http/response.rs
```diff
@@ -104,5 +145,26 @@ fn json_decode_error(
...
+        json_decode_error(source_schema, table_name, method_label, logged_url, error);
+    if is_retryable_json_decode_error(error) {
+        ResponseDecodeFailure::retryable(provider_error)
+    } else {
+        ResponseDecodeFailure::terminal(provider_error)
+    }
+}
+
+fn is_retryable_json_decode_error(error: &serde_json::Error) -> bool {
+    matches!(error.classify(), serde_json::error::Category::Eof)
```
jsummerfield, line 125: `Category::Eof` also covers stable empty or whitespace-only bodies, not just truncated JSON. With the new retry path, a GET returning `200` with an empty body now retries and finally reports `Unavailable` / `retryable=true`, where this used to be a permanent decode/manifest failure.

Can we only mark EOF retryable when there is evidence of partial content, and add an empty-body regression test?

crates/coral-engine/src/backends/http/error.rs
```diff
@@ -411,8 +442,12 @@ fn provider_decode_failure<'a>(
             "The upstream API returned a response that does not match the source manifest."
                 .to_string(),
         ),
-        retryable: false,
-        status: StatusCode::FailedPrecondition,
+        retryable,
```
jsummerfield, line 421: For retryable decode failures, the hint just above still points users at a source-manifest mismatch, even though the structured error now says the failure is transient/retryable. Can we branch the hint for `retryable=true` so the rendered guidance matches the machine contract?

jsummerfield (approved):

PR #826, opened 2026-05-26
feat(ui): sources management page

martinrajdl:
Adds a Sources page alongside Traces in the Coral UI: list, search, and install **core** sources from the in-binary catalog, inspect / edit credentials / remove installed ones. Backed end-to-end by the existing `SourceService` RPCs — no proto changes, no new RPCs.

Community-source discovery is **explicitly out of scope** here and will land in a follow-up that depends on a Coral-hosted catalog API (per the Slack discussion: not bundling metadata at release time, not round-tripping manifest YAML through the client).

### UI

**Routing & shell** — new hash router with `traces` / `sources` routes. The Sources nav entry slots into the existing expandable-sidebar layout with route-aware active state. Install + detail are dialogs, not URL-addressable routes.

**Sources index** (`views/sources/sources-index.tsx`)
- Centered max-width-960 layout modelled on `adp/web/src/views/plugins-view`.
- Connected sources render in the first section above the catalog, then uninstalled sources are grouped into category sections modelled on `adp/web/src/views/plugins-view`, with the explicit mapping in `lib/source-categories.ts`. Search still filters across name + description.
- Card design: round provider logo, capitalised name, origin badge (Core / Imported) when origin is known, plus a Connected pill on installed sources.
- `cmd`/`ctrl`+`F` focuses and selects the search input.
- Resilient loading: one RPC failure no longer wipes the other's result; the error banner clears on a successful refresh.

**Install dialog** (`views/sources/source-install.tsx`)
- Layout ported from `adp/web/src/components/credential-form.tsx`: round logo + capitalised name + Core pill in the title, sentence-cased field labels, manifest `hint` rendered as markdown below each field, secrets via `TextInput type="password"`.
- Multi-method secrets get a compact segmented control above the input (e.g. Slack's OAuth + Paste choice).
- OAuth: runs the existing `CreateBundledSourceWithOAuth` stream. UI opens the authorize URL with `window.open` and shows a visible fallback link if popups are blocked. The submit button stays disabled through `oauth-completed` until the install promise resolves so controls can't be re-fired mid-install.
- Optional OAuth client-id is always rendered as an override (with the manifest default shown as a placeholder), per the previous review.
- Primary action reads "Add source" → "Adding…" → "Finishing…" instead of "Save credentials".

**Detail dialog** (`views/sources/source-detail.tsx`)
- Bundled installed sources fetch their `SourceInfo` so the detail form is driven by manifest input metadata, including visible variable defaults such as the CloudWatch AWS defaults (`AWS_REGION=us-east-1`, `AWS_ENDPOINT_SUFFIX=amazonaws.com`).
- Same simple field layout as the install dialog: no gray key/value binding box, each variable / secret row is an always-on `TextInput`, and manifest `hint` renders below the field.
- Variables prefill with the current value or manifest default. Configured secrets render as a masked `••••••••` password value; focusing the field clears the sentinel, and save only submits secrets the user actually typed into.
- Imported sources show the form read-only with a one-line hint pointing back to re-import.
- Remove opens a **stacked** confirm dialog (its own `Dialog.Root`) so the detail popup stays visible underneath.
- Validation has been removed for this PR per review (validation against a live source belongs in a follow-up; the previous in-dialog validate was misleading when the source was unreachable).

### Wax additions
- **Dialog primitives** lifted from `adp/web/src/wax/components/dialog`. `Dialog.Trigger` forwards the base component's full prop set (so callers can pass `disabled` / `id` / etc.) instead of a narrowed wrapper.
- **Toast** lifted from `adp/web/src/wax/components/toast` (react-toastify-backed). Replaces the bespoke `components/toast.tsx` host. Call sites use `addToast('variant', { title })`.
- **Icon**: extended `IconName` with `Check` / `CircleCheck` / `Plus` / `RefreshCw` (from lucide) and a custom `Plug` glyph at `custom-icons/plug.tsx`.
- **TextInput** now forwards focus handlers so masked configured secrets can clear only when the user edits them.

### Supporting infra
- `lib/router.ts` — typed hash router
- `lib/coral-clients.ts` — shared gRPC-Web transport
- `lib/sources.ts` — typed Connect-RPC wrappers including the OAuth stream consumer; exports a shared `originLabel` derived from the proto `SourceOrigin` enum so the index and detail views agree on origin strings.
- `lib/source-categories.ts` — explicit source-name-to-category mapping and category display order for the Sources index.
- `lib/provider-icons.ts` — name → asset URL map. Bundled sources without a confidently-sourced brand asset (claude, codex, gitlab, incident_io, openobserve, posthog) fall back to the generic Plug glyph until proper assets land.
- ~25 logos under `public/images/providers/`, a mix carried over from `adp/web/public/images/providers` and pulled from simpleicons for normalised single-color brand SVGs.
- `components/{page-header,error-banner,markdown}.tsx` — small primitives. `markdown.tsx` uses `react-markdown` (one new dep, ~50KB) so manifest hints render links / inline code / numbered lists properly.
- `utils/to-sentence-case.ts` — credential-form label helper.

### Tests
- `ui/tests/ui/sources.spec.ts` covers the user flow (connected-first category list + search + empty-search, install via paste → inline edit → remove, installed detail manifest defaults + masked secrets, `cmd`/`ctrl`+`F` focuses search). MSW-mocked gRPC-Web `SourceService` responses, no Rust server — follows `ui/tests/ui/AGENTS.md`. Recordings via `npm run test:ui:screencast:headless --prefix ui`.

## What's not in this PR
- **OAuth token refresh.** Refresh persistence is in-progress on a separate branch; intentionally not included here. Long-lived OAuth tokens still work; short-lived ones will need a re-install until refresh ships.
- **Community sources.** Discovery, install, browse — all deferred.
- **Custom source authoring** (the "paste a spec" flow from the earlier prototype). Defers until backend manifest parsing is real.
- **Schema explorer.** Kept on its prototype branch.
- **Icon field in the source spec.** UI maps known names to local SVGs; everything else falls back to the custom Plug glyph.
- **Outbound hosts confirmation UX.** `SourceInfo.hosts` field doesn't exist on `main` (lives on `claude/elastic-brown-8ad186`). Surface on the install + detail pages when that lands.

## Verification
- `cargo clippy --workspace --all-targets -- -D warnings` clean
- `cargo test -p coral-app` — 136 lib + 51 gRPC integration tests pass
- `npm run build` (type-check + Vite) clean
- `npx oxlint --deny-warnings` clean
- `npm run test:ui --prefix ui` — sources + traces + layout suites pass

Latest collapsed follow-up check on `sources-management-v2`:
- `npm run check --prefix ui`
- `npm run type-check --prefix ui`
- `npm run test:ui --prefix ui -- ui/tests/ui/sources.spec.ts`

## Manual test plan
- [ ] `cd ui && npm run dev:local`, open `http://localhost:5173/#/sources`
- [ ] Connected sources appear above category sections; uninstalled sources are grouped by category
- [ ] `cmd`/`ctrl`+`F` focuses the search input
- [ ] Install `linear` via Paste → returns to detail dialog
- [ ] Install `slack` via OAuth (enter `SLACK_OAUTH_CLIENT_ID`, complete in browser) → stream progresses through `awaiting-oauth` → `oauth-completed` → installed
- [ ] On the detail dialog, overwrite a variable and / or a secret, click **Save changes**, confirm only the touched bindings update
- [ ] Open a bundled installed source with variables and secrets, confirm manifest defaults are prefilled and configured secrets are masked without a gray binding box
- [ ] Remove from the detail dialog (stacked confirm modal over the detail popup)


crates/coral-app/src/query/manager.rs
```diff
@@ -34,6 +34,12 @@ pub(crate) struct ValidatedSource {
     pub(crate) report: SourceValidationReport,
```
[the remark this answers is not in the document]
martinrajdl: Keeping refresh on the query-time path is intentional for v1, and I think the silent-omit risk is acceptable for now — here's the reasoning:

- The refresh-then-write flow only runs when the provider actually returned a refresh token, so for non-refreshable secrets it's a stored-only read with no extra I/O.
- `load_query_source` already swallows per-source errors and logs (except keychain-unavailable, which fails closed). A refresh failure for one source surfaces the same way as any other per-source error today — it doesn't drop other sources from the query, and the next validate/reconnect surfaces the real error.
- Doing refresh only on validate would mean queries silently use stale (potentially expired-once-cached) tokens and 401 against the upstream, which is a worse UX than refreshing eagerly and skipping a single source with a warning.

If silent omission becomes a real problem we can swap the `tracing::warn!` for a per-source error attached to the response and let the engine surface it as a partial-result warning. Happy to file a follow-up if you'd prefer that direction.

AlbertQM (commented):
Woo, this will be super handy!

Few things bits here and there:

1. This shouldn't say save credentials, since it has none. Maybe "Add source"?

Image

2. Can we start as simple as possible, and e.g. remove validation? That's easier for everyone to review, and implement.

3. In the old repo, we were able to update credentials without having to remove and re-add a source. Since that's already built, could we have that behaviour back?

Image

4. I tried adding a source, and used garbage secrets, and it still validated. IMO validation, when we add it in the future, should validate the source actually works. Otherwise we say "validated", but then an agent tries to use it and gets an error

5. There are a bunch of nice things we used to have which are gone: keyboard navigation, cmd + f should select the input, the confirmation dialog should be stacked on top of the first dialog, icons are missing from most sources

crates/coral-app/src/query/manager.rs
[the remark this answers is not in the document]
martinrajdl: Catalog reads (`list_tables` / `list_catalog`) are local enumerations over the spec — they don't hit the upstream API, so an expired access token doesn't fail them. `StoredOnly` is intentional here to keep catalog browsing cheap (no token endpoint round-trip per page-load). The token is only actually exercised by `execute_sql` / `explain_sql` / `validate_source`, which already use `RefreshProvider`. Leaving as-is.

crates/coral-app/src/credentials/oauth.rs
```diff
@@ -767,10 +880,125 @@ fn oauth_metadata_prefix(input_key: &str) -> String {
...
+                ))
+            })
+        })
+        .transpose()?
+        .or_else(|| oauth.client.secret.as_ref().map(|secret| secret.transport));
+    let client_secret = material
+        .get(&format!("{metadata_prefix}client_secret"))
+        .filter(|value| !value.is_empty())
+        .cloned();
+    if client_secret_transport.is_some() && client_secret.is_none() {
```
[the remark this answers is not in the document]
martinrajdl: Deferring — the OAuth refresh path is in-progress work and we want to land this PR with minimum churn. Backward compatibility for older confidential OAuth installs (missing stored `client_secret` metadata) will be handled when the refresh feature is finalized.

crates/coral-app/src/credentials/oauth.rs
```diff
@@ -767,10 +880,125 @@ fn oauth_metadata_prefix(input_key: &str) -> String {
...
+    metadata_prefix: &str,
+    oauth: &ManifestOAuthCredentialSpec,
+    material: &BTreeMap<String, String>,
+) -> Result<Option<OAuthRefreshConfig>, AppError> {
+    if material
+        .get(&format!("{metadata_prefix}method"))
+        .map(String::as_str)
+        != Some("oauth")
+    {
+        return Ok(None);
```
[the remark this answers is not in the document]
martinrajdl: Deferring — the OAuth refresh path is in-progress work and we want to land this PR with minimum churn. Tracking this UX improvement (Coral-side reconnect error instead of a passthrough upstream 401) as part of the follow-up.

martinrajdl:
@cubic-dev-ai re-review please — addressed the open feedback in the latest push, deferring the two OAuth refresh edge cases to the in-progress refresh follow-up.

crates/coral-app/src/credentials/oauth.rs
```diff
@@ -1147,10 +1260,125 @@ fn oauth_metadata_prefix(input_key: &str) -> String {
...
+    match token.expires_at {
+        Some(expires_at) => {
+            material.insert(
+                format!("{metadata_prefix}access_token_expires_at"),
+                expires_at.to_rfc3339(),
+            );
+        }
+        None => {
+            material.remove(&format!("{metadata_prefix}access_token_expires_at"));
+        }
```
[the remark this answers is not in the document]
martinrajdl: OAuth refresh has been dropped from this PR — those commits are no longer in the branch. Refresh persistence will land in the follow-up branch that owns the feature.

crates/coral-app/src/credentials/oauth.rs
```diff
@@ -1147,10 +1260,125 @@ fn oauth_metadata_prefix(input_key: &str) -> String {
...
+    let Some(refresh_token) = material
+        .get(&format!("{metadata_prefix}refresh_token"))
+        .filter(|value| !value.is_empty())
+        .cloned()
+    else {
+        tracing::debug!(
+            source_secret = access_token_material_key,
+            "OAuth access token is expired but no refresh token is stored"
+        );
+        return Ok(None);
```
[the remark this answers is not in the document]
martinrajdl: OAuth refresh has been dropped from this PR — those commits are no longer in the branch. Refresh persistence will land in the follow-up branch that owns the feature.

crates/coral-app/src/credentials/store.rs
```diff
@@ -213,6 +216,27 @@ impl CredentialStore {
...
+        update: F,
+    ) -> Result<R, AppError>
+    where
+        F: FnOnce(BTreeMap<String, String>) -> Result<(BTreeMap<String, String>, R), AppError>,
+    {
+        tracing::trace!(%credential_set_id, %storage, "updating credential material");
+        // Hold the state lock across the read-modify-write so a concurrent
+        // install/refresh on the same credential set cannot land a write
+        // between our read and write and get clobbered.
+        let _lock = FileLock::exclusive(self.layout.state_lock())?;
```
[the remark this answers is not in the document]
martinrajdl: OAuth refresh has been dropped from this PR — those commits are no longer in the branch. Refresh persistence will land in the follow-up branch that owns the feature.

crates/coral-app/src/credentials/mod.rs
```diff
@@ -178,6 +232,146 @@ impl CredentialManager {
...
+    fn persist_refreshed_oauth_material(
+        &self,
+        workspace_name: &WorkspaceName,
+        credential_set_id: &CredentialSetId,
+        storage: CredentialStorageKind,
+        input_key: &str,
+        before_refresh: &BTreeMap<String, String>,
+        refreshed_material: &BTreeMap<String, String>,
+    ) -> Result<BTreeMap<String, String>, AppError> {
+        self.store
```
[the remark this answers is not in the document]
martinrajdl: OAuth refresh has been dropped from this PR — those commits are no longer in the branch. Refresh persistence will land in the follow-up branch that owns the feature.

SOURCES_MANAGEMENT_PLAN.md
jsummerfield, line 1: revert

martinrajdl: Dropped the plan doc in e4ad181. Thanks.

ui/src/components/toast.tsx
lazybean, line 1: Could we use the one from wax?

martinrajdl: wax doesn't have a toast component yet, this is the stopgap until it does

ui/public/images/providers/anthropic.svg
AlbertQM, line 1: Did Codex draw this from memory? 😂

ui/public/images/providers/gcp.png
AlbertQM, line 1: Why is this so low quality? 🤔

ui/public/images/providers/gitlab.svg
AlbertQM, line 1: Some logos have some background, others have transparency -- is that intentional? Should these be consistent? (transparency will probably look neater?)

ui/public/images/providers/xai.svg
AlbertQM, line 1: This xAI logo looks wonky?

ui/src/components/toast.tsx
AlbertQM: WAX does have a Toast, it's in monorepo 👀

ui/public/images/providers/anthropic.svg
martinrajdl: lol fair, swapped for the simpleicons one in 780084b

ui/public/images/providers/gcp.png
martinrajdl: replaced with an svg in 780084b

ui/public/images/providers/gitlab.svg
martinrajdl: yeah agreed, stripped the bg on gitlab in 780084b. the rest should all be transparent now

ui/public/images/providers/xai.svg
martinrajdl: xai isn't on simpleicons so i left it for now, can grab an official svg in a follow-up

ui/public/images/providers/gitlab.svg
AlbertQM: > the rest should all be transparent now

What about Sentry? (there are possibly other ones too)

AlbertQM (commented):
Thanks for addressing the comments!

Some minor things:

1. Pencil icon here is broken
Image

Though don't think we need this, as mentioned earlier, the original simple version had the same dialog whether or not you have saved credentials. Ie if I had filled in a secret, I would see the ●●●●, and was able to overwrite and click save.

2. This checkmark was (I'm guessing) added to signify "validated". We've removed the broken validation flow. Do we still need this icon?

Image

3. Thanks for adding some high quality logos! Some `core` sources are still missing though, is that intentional? It looks so much better with the icon, I think it's worth the effort of actually going through all our sources and adding it

4. In a previous comment (thanks for addressing), I suggested changing the button label; now the body of the card is out of date, as it still mentions "click save", but there's no save button.

Image

If @lazybean has no more comments, happy to ✅

lazybean:
I have 0 comments and 100% trust in you two.

Just one tiny thing :sweat_smile:, could you ask your agents to check ui/tests/ui/AGENTS.md and write sources.spec.ts to cover some user flow? (should be easy)  
 

-  list, search, and install **core** sources from the in-binary catalog, 
- inspect / edit credentials / remove installed ones.

That would ensure I wont break it next week, and also allow to generate videos (npm run  test:ui:screencast:headless --prefix ui)  
  


> Thanks for addressing the comments!
>
> Some minor things:
>
> 1. Pencil icon here is broken
> Image
>
> Though don't think we need this, as mentioned earlier, the original simple version had the same dialog whether or not you have saved credentials. Ie if I had filled in a secret, I would see the ●●●●, and was able to overwrite and click save.
>
> 1. This checkmark was (I'm guessing) added to signify "validated". We've removed the broken validation flow. Do we still need this icon?
>
> Image
>
> 1. Thanks for adding some high quality logos! Some `core` sources are still missing though, is that intentional? It looks so much better with the icon, I think it's worth the effort of actually going through all our sources and adding it
> 2. In a previous comment (thanks for addressing), I suggested changing the button label; now the body of the card is out of date, as it still mentions "click save", but there's no save button.
>
> Image
>
> If @lazybean has no more comments, happy to ✅

ui/src/views/sources/source-detail.tsx
```diff
@@ -0,0 +1,286 @@
...
+                </div>
+              )
+            })}
+            {source.secrets.map((s) => {
+              const draftKey = `sec:${s.key}`
+              return (
+                <div key={draftKey} className={styles.bindingRow}>
+                  <span className={styles.keyLabel}>{s.key}</span>
+                  <TextInput
+                    type="password"
```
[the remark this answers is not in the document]
martinrajdl: oauth install is dropping from this PR — we'll wire up the read-only behaviour once oauth install + refresh land properly. for now any stored secret is editable.

ui/src/views/sources/source-detail.tsx
```diff
@@ -0,0 +1,286 @@
...
+                  <TextInput
+                    value={drafts[draftKey] ?? v.value}
+                    onChange={(value) => setDrafts((p) => ({ ...p, [draftKey]: value }))}
+                    placeholder={v.key}
+                    disabled={!editable || saving}
+                  />
+                </div>
+              )
+            })}
+            {source.secrets.map((s) => {
```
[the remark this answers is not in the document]
martinrajdl: updated the description to match the dropped pencil + oauth-managed hint

ui/src/views/sources/source-detail.tsx
```diff
@@ -0,0 +1,286 @@
...
+      const bindings: InstallInput[] = source.variables.map((v) => ({
+        key: v.key,
+        value: drafts[`var:${v.key}`] ?? v.value,
+        secret: false,
+      }))
+      for (const s of source.secrets) {
+        const draft = drafts[`sec:${s.key}`]
+        if (draft !== undefined && draft.length > 0) {
+          bindings.push({ key: s.key, value: draft, secret: true })
+        }
```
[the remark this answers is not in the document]
martinrajdl: fixed in 7fa02a2

ui/src/views/sources/source-detail.tsx
```diff
@@ -0,0 +1,288 @@
...
+        key: v.key,
+        value: drafts[`var:${v.key}`] ?? v.value,
+        secret: false,
+      }))
+      for (const s of source.secrets) {
+        const draft = drafts[`sec:${s.key}`]
+        if (draft === undefined) continue
+        const trimmed = draft.trim()
+        if (trimmed.length > 0) {
+          bindings.push({ key: s.key, value: trimmed, secret: true })
```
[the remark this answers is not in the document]
martinrajdl: this is the opposite of the previous review (which asked us to trim — see https://github.com/withcoral/coral/pull/826/files/7fa02a2c4f65eec879766b3a26b588616b001eaf#r3324471385). keeping the trim — modern api tokens don't carry meaningful whitespace and the trim catches accidental spacebar presses.

ui/src/components/toast.tsx
martinrajdl: sorry about this one, I saw that we haven't imported not that it's not in monorepo

martinrajdl:
@AlbertQM 

This checkmark was (I'm guessing) added to signify "validated". We've removed the broken validation flow. Do we still need this icon? - this means connected - as the tooltip says. 

Some core sources are still missing though, is that intentional? - yes, id' rather remove then and then spend an hour in figma getting all perfect later

martinrajdl:
> Though don't think we need this, as mentioned earlier, the original simple version had the same dialog whether or not you have saved credentials. Ie if I had filled in a secret, I would see the ●●●●, and was able to overwrite and click save. 
went back to this behaviour 
cc @AlbertQM

AlbertQM (approved):
> > Though don't think we need this, as mentioned earlier, the original simple version had the same dialog whether or not you have saved credentials. Ie if I had filled in a secret, I would see the ●●●●, and was able to overwrite and click save.

>went back to this behaviour
 cc @AlbertQM

Did we? The inputs when a secret is filled looks hallucinated to me?

image

The above was not in the design. Compare with an input that was in the designs, and is already implement:

image

^ The above was me opening a source I didn't install, but has pre-filled inputs. This is _exactly_ what it should look like when you open an installed source, except that it'll also have the secrets pre-filled.

Accepting to unblock, but please double check manually there are no more hallucinated bits of UI 🙏

martinrajdl:
Thank you @AlbertQM 🙏 the defaults are from the manifest so not halucinated, I'd say its good ux for them to be prefilled and editable. I pulled ADP and checked again the inputs, they're cleaned up now.

ui/src/lib/source-categories.ts
```diff
@@ -0,0 +1,48 @@
...
+  notion: 'knowledge',
+  openobserve: 'observability',
+  pagerduty: 'incident-management',
+  posthog: 'analytics',
+  sentry: 'observability',
+  slack: 'communication',
+  statusgator: 'observability',
+  stripe: 'business',
+  wandb: 'ai-ml',
+}
```
jsummerfield, line 44: These should be driven by the manifests, not the UI.

jsummerfield:
A few comments on the detail view:  
\- I wouldn't convert the input vars to sentence case. Keep it GITHUB_API_BASE instead of "Github api base". While not pretty, it's consistent with the env vars a user passes when adding a source on the CLI.  
\- It's odd to see instructions for which auth method to choose after a source is connected. I'd drop these guides post-connection.

ui/src/lib/sources.ts
```diff
@@ -0,0 +1,163 @@
...
+      variables,
+      secrets,
+      oauthCredentialRetrievals: oauthRetrievals,
+    }),
+  )
+  for await (const response of stream) {
+    const event = response.event
+    if (event.case === 'source') return event.value
+    if (event.case === 'oauthAuthorization') {
+      callbacks.onAuthorization?.({
```
jsummerfield, line 136: The OAuth authorization event drops `user_code`, `verification_uri`, and `verification_uri_complete`, and the UI only renders a browser link. That breaks the default GitHub device-code install path when the provider returns a code plus verification URL: users can open the device page but never see the code to enter. Please carry those fields through `OAuthFlowCallbacks` / `InstallProgress`, render the code + verification URI for device-code flows, and add a UI test covering `CreateBundledSourceWithOAuth` for GitHub.

ui/src/views/sources/sources-index.tsx
```diff
@@ -0,0 +1,315 @@
...
+
+  const allEntries = useMemo<IndexEntry[]>(() => {
+    const bundledByName = new Map((bundled ?? []).map((entry) => [entry.name, entry]))
+    const merged: IndexEntry[] = (bundled ?? []).map((entry) => ({ ...entry }))
+    for (const installedSource of installed ?? []) {
+      const bundledEntry = bundledByName.get(installedSource.name)
+      if (bundledEntry) {
+        const target = merged.find((entry) => entry.name === installedSource.name)
+        if (target) {
+          target.installed = true
```
jsummerfield: When an installed source name also exists in the bundled catalog, this overlay only marks the bundled entry installed and keeps bundled origin/version/description. An imported source with the same name can show as “Core” in the index while the detail dialog treats it as imported. Please overlay origin and installed metadata from `installedSource`, or distinguish bundled-catalog installation from installed-source origin.

ui/tests/ui/support/source-handlers.ts
```diff
@@ -0,0 +1,76 @@
...
+// remove linear. List and Discover responses advance one step at a time as
+// the UI calls Create/Delete.
+export function sourceLifecycleHandlers() {
+  let listResponse = listInitialResponse
+  let discoverResponse = discoverInitialResponse
+
+  return [
+    http.post(discoverUrl, () => grpcWebResponse(DiscoverSourcesResponseSchema, discoverResponse)),
+    http.post(listUrl, () => grpcWebResponse(ListSourcesResponseSchema, listResponse)),
+    http.post(getInfoUrl, async ({ request }) => {
```
jsummerfield, line 80: The handlers inspect gRPC-web requests with `TextDecoder().decode(...).includes(...)`. That means malformed requests, wrong names, or proto field drift can still fall through to a successful fixture response. Please decode the framed gRPC-web request with the generated `GetSourceInfoRequestSchema` / `GetSourceRequestSchema` and switch on the parsed `name`.

ui/tests/ui/support/source-handlers.ts
```diff
@@ -0,0 +1,76 @@
...
+    http.post(getUrl, async ({ request }) => {
+      const body = new TextDecoder().decode(await request.arrayBuffer())
+      const response = body.includes('linear')
+        ? getInstalledLinearResponse
+        : body.includes('cloudwatch_logs')
+          ? getInstalledCloudwatchLogsResponse
+          : getInstalledGithubResponse
+      return grpcWebResponse(GetSourceResponseSchema, response)
+    }),
+    http.post(createBundledUrl, () => {
```
jsummerfield: This handler returns the same successful `CreateBundledSource` response for both the initial install and the later “Save changes” path without asserting the request payload. The edit test would still pass if the UI sent the wrong source name or dropped `LINEAR_API_TOKEN=lin_test_token_v2`. Please decode `CreateBundledSourceRequest` here and assert the expected name and bindings for each call.

ui/package-lock.json
```diff
@@ -3333,116 +3643,721 @@
...
       "version": "10.4.3",
       "resolved": "https://registry.npmjs.org/lru-cache/-/lru-cache-10.4.3.tgz",
       "integrity": "sha512-JNAzZcXrCt42VGLuYz0zfAzDfAvJWW6AfYlDBQyDV5DClI2m5sAmK+OIO7s59XfsRsWHp02jAJrRadPRGTt6SQ==",
       "license": "ISC"
     },
     "node_modules/lucide-react": {
-      "version": "1.16.0",
-      "resolved": "https://registry.npmjs.org/lucide-react/-/lucide-react-1.16.0.tgz",
-      "integrity": "sha512-dYwyPzb4MEKpGUmNYk3WKWPnMrHs3FKM+q94kAnJrcDIqqn1hq2xY8scaS2ovsOCM5D51ey2gaRG3PBb1vgoYQ==",
+      "version": "1.17.0",
```
jsummerfield, line 3663: This lockfile change upgrades `lucide-react` from 1.16.0 to 1.17.0 while `package.json` still advertises `^1.16.0`. If intentional, please bump the manifest too; otherwise regenerate the lockfile with only the markdown/toast dependency changes.

jsummerfield (commented):
Inline findings from review.

ui/public/images/providers/Clickhouse.png
jsummerfield, line 1: Similarly, these icons should live alongside the manifest imo, not in the UI shell.

ui/src/views/sources/sources-index.tsx
```diff
@@ -0,0 +1,315 @@
...
+        ) : null}
+      </div>
+      {entry.description ? (
+        <Typography.Body variant="tertiary" className={styles.cardDescription}>
+          {entry.description}
+        </Typography.Body>
+      ) : null}
+      <div className={styles.cardFooter}>
+        <Typography.BodySmall variant="tertiary">
+          v{entry.installedVersion ?? entry.version}
```
jsummerfield: Can we drop the source version from the cards? In the connected grid it adds visual noise without helping the primary decision, and it also exposes awkward cases like `v` when an installed source has no version. The card already has enough status metadata with Core/Imported and Connected; version detail can live in the detail dialog instead.

jsummerfield (commented):
One additional UI comment.

jsummerfield (commented):
One additional follow-up comment.

ui/src/views/sources/sources-index.tsx
```diff
@@ -0,0 +1,315 @@
...
+      <div className={styles.cardHeader}>
+        <div className={styles.cardLogo}>
+          {icon ? (
+            <img alt="" src={icon} className={styles.cardLogoImg} />
+          ) : (
+            <Icon name="Plug" size="18" color="tertiary" />
+          )}
+        </div>
+        <Typography.BodyLargeStrong as="span" className={styles.cardTitle}>
+          {entry.name}
```
jsummerfield, line 249: Follow-up request: can we update source manifests with explicit display names in a separate PR? Rendering the raw schema/source key leaks names like `github_v4`, `cloudwatch_logs`, or imported snake-case names into the UI. This PR does not need to solve that contract, but the right long-term fix is manifest-owned display metadata rather than one-off UI formatting.

jsummerfield (commented):
One additional install-feedback comment.

ui/src/views/sources/source-install.tsx
```diff
@@ -0,0 +1,496 @@
...
+        },
+      }
+
+      if (retrievalProtos.length === 0) {
+        await createBundledSource(name, bindings)
+      } else {
+        await createBundledSourceWithOAuth(name, bindings, retrievalProtos, callbacks)
+      }
+
+      addToast('success', { title: `Installed ${name}` })
```
jsummerfield: If the user adds a source with an incorrect token, this still presents as a success: the toast says `Installed`, the dialog advances to detail, and there is no indication that the credentials may not work. We need some feedback here. Either validate after install and surface the failure, or make the success state explicit that credentials were saved but not verified, with a follow-up action/status so a bad token does not look like a working connection.

ui/src/lib/source-categories.ts
martinrajdl: not saying i completely disagree, but what i'm thinking is if the manifest is purely driven by api specs and generated and these are our labels of product/tool categories, I think this is more like user input on top of the manifest - alongside title and description - which I think that makes sense, I'm not sure if those will be in the manifest as some optional fields but perhaps
but for now I figured I'd just keep them in this temporary mapping

ui/src/views/sources/source-install.tsx
martinrajdl: I initially added a validation button into the install, but removed it as per some discussion above. I agree it would be a good ux, might add separately in a more smaller PRs, is that ok? 

Just changed the copy for now, but will add the validation later

ui/src/views/sources/sources-index.tsx
martinrajdl: That sounds good

ui/src/views/sources/source-install.tsx
AlbertQM: >I initially added a validation button into the install, but removed it as per some discussion above

Fished my original comment on this exact same issue (sorry I wasn't clear on the outcome):

>I tried adding a source, and used garbage secrets, and it still validated. IMO validation, when we add it in the future, should validate the source actually works. Otherwise we say "validated", but then an agent tries to use it and gets an error 

Agreed with both that an _actual_ validation at install time is the best UX, especially if it helps the user understand why this has failed and how to fix it. 
I'd also be in favour of removing notion of validation completely and add proper validation in a separate PR 😄

ui/src/views/sources/source-detail.tsx
```diff
@@ -0,0 +1,496 @@
...
+  const [confirmingRemove, setConfirmingRemove] = useState(false)
+  const [deleting, setDeleting] = useState(false)
+  const [drafts, setDrafts] = useState<Record<string, string>>({})
+  const [saving, setSaving] = useState(false)
+
+  const refresh = useCallback(async () => {
+    try {
+      const installed = await getInstalledSource(name)
+      const info =
+        originLabel(installed.origin) === 'bundled' ? (await getBundledSourceInfo(name)).info : null
```
jsummerfield: Can we keep using the effective source metadata for installed sources here? The backend `GetSourceInfo` path already resolves installed imported manifests, but this skips metadata whenever the installed origin is imported. That means an imported install can lose its manifest fields in detail, and an installed source whose bundled manifest is missing can fail detail loading instead of falling back to the recorded bindings/removal path.

ui/src/views/sources/sources-index.tsx
```diff
@@ -0,0 +1,309 @@
...
+      {entry.description ? (
+        <Typography.Body variant="tertiary" className={styles.cardDescription}>
+          {entry.description}
+        </Typography.Body>
+      ) : null}
+      {entry.installed ? (
+        <div className={styles.cardFooter}>
+          <span className={styles.connectedPill}>
+            <Icon name="CircleCheck" size="14" color="inherit" />
+            Connected
```
jsummerfield: Could we make this persistent status match the new install copy? The toast now says credentials were saved but not verified, but once it disappears the card still says `Connected`, which makes an incorrect token look like a working connection. Until we have real validation state, a neutral label like `Configured` / `Saved` would avoid over-claiming success.

jsummerfield:
Looks good, the main design issue is the UI is over-correcting around source display metadata and is doing catalog reconciliation locally when the backend already exposes effective installed source metadata through `GetSourceInfo`.

martinrajdl:
@jsummerfield I didn't want to bother you with this while you were out, but please let me know if there is anything else that should be addressed in this PR, the client merge is dropped + some of the UI tweaks

crates/coral-app/src/sources/manager.rs
```diff
@@ -228,6 +228,30 @@ impl SourceManager {
                 candidate.credential_storage = Some(*storage);
             }
         }
+
+        // Imported sources aren't in the bundled catalog, so append their
```
jsummerfield: I think this is pushing the fix into the wrong contract. The page does need to show installed imported sources, but `DiscoverSources` is still documented and named as bundled-catalog discovery. Making it return imported installed sources turns it into a mixed workspace inventory endpoint, and it still does not actually handle the collision case this was trying to avoid: if an imported source has the same name as a bundled source, `catalog_names.contains(&source.name)` skips the imported manifest and the UI will still see the bundled/Core entry.

I don’t think this PR should grow into defining imported/bundled name-collision semantics. I’d prefer to keep `DiscoverSources` bundled-only for now and either keep the UI’s temporary composition with `ListSources` or defer the exact imported-source card metadata edge cases. Then we can do a follow-up that explicitly defines source catalog vs installed source inventory, identity/precedence rules, and how broken imported manifests appear in management UI.

jsummerfield (approved):

crates/coral-app/src/sources/manager.rs
martinrajdl: reverted DiscoverSources to bundled-only and deferred handling imported sources, I'll open another PR for that

PR #856, opened 2026-05-27
feat(sources/community/open-notify): add Open Notify (ISS) source

wiz-abhi:
## Summary
This PR adds a new community source for **Open Notify**, an API that provides real-time data about the International Space Station (ISS) and astronauts currently in space.

## Tables Added
- **`iss_position`**: Retrieves the current latitude and longitude of the ISS, along with a UNIX timestamp. The API endpoint (`/iss-now.json`) returns a single JSON object rather than a list, which Coral correctly processes into a single row using `response: {}` and `pagination: mode: none`.
- **`astronauts`**: Retrieves a list of astronauts currently in space, including their name and the spacecraft they are aboard (e.g., "ISS"). This endpoint (`/astros.json`) maps the `people` array via `rows_path: [people]`.

## Implementation Details
- **Authentication**: No authentication is required for this API.
- **Pagination**: Set to `mode: none` since the API only returns a real-time snapshot of current data without paginated lists.
- **Testing**: SQL queries were declared in `test_queries` to ensure both tables execute properly.

## Verification
Locally tested and linted with `coral source lint` and `coral source test`. The source was successfully added and verified via CLI:
```sql
SELECT name, craft FROM open_notify.astronauts LIMIT 5;
```
```text
+----------------------+-------+
| name                 | craft |
+----------------------+-------+
| Oleg Kononenko       | ISS   |
| Nikolai Chub         | ISS   |
| Tracy Caldwell Dyson | ISS   |
| Matthew Dominick     | ISS   |
| Michael Barratt      | ISS   |
+----------------------+-------+
```

### DIFF
--- a/sources/community/open-notify/README.md
+++ b/sources/community/open-notify/README.md
@@ -1,0 +1,40 @@
+# Open Notify (ISS)
+
+Real-time data about the International Space Station and astronauts in space from [Open Notify](http://open-notify.org/Open-Notify-API/).
+
+## Setup
+
+No authentication is required. Add the source:
+
+```bash
+coral source add --file sources/community/open-notify/manifest.yaml
+```
+
+## Local Testing
+
+```bash
+coral sql "
+  SELECT name, craft 
+  FROM open_notify.astronauts 
+  LIMIT 5
+"
+
+/*
++----------------------+-------+
+| name                 | craft |
++----------------------+-------+
+| Oleg Kononenko       | ISS   |
+| Nikolai Chub         | ISS   |
+| Tracy Caldwell Dyson | ISS   |
+| Matthew Dominick     | ISS   |
+| Michael Barratt      | ISS   |
++----------------------+-------+
+*/
+```
+
+## Tables
+
+| Table | Description |
+|-------|-------------|
+| `iss_position` | Current geographic location of the International Space Station. |
+| `astronauts` | List of astronauts currently in space. |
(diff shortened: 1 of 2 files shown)


jsummerfield (changes requested):
Findings:

- Medium: `sources/community/open-notify/manifest.yaml:29` and `sources/community/open-notify/manifest.yaml:38` expose ISS `latitude` and `longitude` as `Utf8`. The Open Notify ISS endpoint documents these fields as latitude/longitude coordinates, and live responses return numeric coordinate values encoded as JSON strings. Leaving them as strings makes range filters, ordering, and downstream geospatial use awkward or wrong unless users cast every query. Please expose both columns as `Float64`; I verified that this change still lets the manifest add/test successfully. Docs: http://open-notify.org/Open-Notify-API/ISS-Location-Now/

The source otherwise looks straightforward and the live README evidence matched the current API. Thanks for keeping the no-auth setup simple.

wiz-abhi:
Thanks for the review! I have updated the `manifest.yaml` to expose the ISS `latitude` and `longitude` columns as `Float64` instead of `Utf8`. I have verified locally using `coral source test` that the type coercion works seamlessly with the API's JSON string numeric values.

jsummerfield (approved):
Reviewed the current head again. The previous coordinate-type blocker is resolved: `latitude` and `longitude` are now numeric, the source lints and tests successfully, live evidence is present, and the current checks are green.

I'm approving this.

PR #857, opened 2026-05-27
feat(sources/community/advice-slip): add Advice Slip source

wiz-abhi:
## Summary
This PR adds a new community source for **Advice Slip**, an API that provides random pieces of advice and allows searching for advice by keyword.

## Tables Added
- **`random_advice`**: Retrieves a single, random piece of advice. The API endpoint (`/advice`) returns a single object containing an `id` and `advice` string, mapped directly to columns.
- **`advices`**: Allows searching for advice using the `query` filter. The endpoint (`/advice/search/{query}`) returns an array of matching slips. The query filter is dynamically injected into the request path.

## Implementation Details
- **Authentication**: No authentication is required for this API.
- **Pagination**: Set to `mode: none` as the search endpoint returns the full set of matches at once (or up to its internal limit) without cursor/page structures.
- **Testing**: SQL queries were declared in `test_queries` for both endpoints to ensure robust structural validation.

## Verification
Locally tested and linted with `coral source lint` and `coral source test`. Here is an example of the query output validating the search functionality:
```sql
SELECT id, advice FROM advice_slip.advices WHERE query = 'spiders' LIMIT 1;
```
```text
+----+---------------------------------------------------------------------+
| id | advice                                                              |
+----+---------------------------------------------------------------------+
| 1  | Remember that spiders are more afraid of you, than you are of them. |
+----+---------------------------------------------------------------------+
```

### DIFF
--- a/sources/community/advice-slip/README.md
+++ b/sources/community/advice-slip/README.md
@@ -1,0 +1,37 @@
+# Advice Slip
+
+Generate random advice or search for specific advice using the [Advice Slip API](https://api.adviceslip.com/).
+
+## Setup
+
+No authentication is required. Add the source:
+
+```bash
+coral source add --file sources/community/advice-slip/manifest.yaml
+```
+
+## Local Testing
+
+```bash
+coral sql "
+  SELECT id, advice 
+  FROM advice_slip.advices 
+  WHERE query = 'spiders' 
+  LIMIT 1
+"
+
+/*
++----+---------------------------------------------------------------------+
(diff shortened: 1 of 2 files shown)


jsummerfield (approved):
Findings: no blocking findings.

I checked the current head against the Advice Slip API docs and live behavior. The random advice and search tables line up with the API shape, the no-match search path returns zero rows cleanly, `coral source lint` passes, and the declared live test queries pass. Thanks for keeping this small and easy to try.

PR #858, opened 2026-05-27
feat(sources/community/universities): add Universities List source

wiz-abhi:
## Summary
This PR adds a new community source for **Universities List**, which provides a searchable directory of universities worldwide via the HipoLabs Universities API.

## Tables Added
- **`universities`**: Provides a list of universities, returning properties such as `name`, `country`, `alpha_two_code`, `domains`, and `web_pages`. It supports two optional URL query parameters natively integrated as filters:
  - `name`: Filter the dataset by university name.
  - `country`: Filter the dataset by country name.

## Implementation Details
- **Authentication**: No API keys or authentication required.
- **Array Parsing**: The API returns a direct JSON array `[...]` at the root, so `rows_path` is explicitly set to `[]`.
- **List Joining**: Array properties like `domains` and `web_pages` are serialized into strings using the `join_array` expression kind for robust SQL operability.
- **Testing**: A `test_queries` definition ensures the table schemas and connections are validated locally during development.

## Verification
Locally tested and linted with `coral source lint` and `coral source test`. Here is an example query showing live fetched results:
```sql
SELECT name, country FROM universities.universities WHERE country = 'United States' LIMIT 2;
```
```text
+-----------------------+---------------+
| name                  | country       |
+-----------------------+---------------+
| Marywood University   | United States |
| Lindenwood University | United States |
+-----------------------+---------------+
```

### DIFF
--- a/sources/community/universities/README.md
+++ b/sources/community/universities/README.md
@@ -1,0 +1,37 @@
+# Universities List
+
+Search universities worldwide by name or country via the [HipoLabs Universities API](https://github.com/Hipo/university-domains-list).
+
+## Setup
+
+No authentication is required. Add the source:
+
+```bash
+coral source add --file sources/community/universities/manifest.yaml
+```
+
+## Local Testing
+
+```bash
+coral sql "
+  SELECT name, country 
+  FROM universities.universities 
+  WHERE country = 'United States' 
+  LIMIT 2
+"
+
+/*
++-----------------------+---------------+
+| name                  | country       |
++-----------------------+---------------+
+| Marywood University   | United States |
+| Lindenwood University | United States |
++-----------------------+---------------+
+*/
+```
+
+## Tables
+
+| Table | Description |
+|-------|-------------|
+| `universities` | List of universities filtered by `name` or `country`. |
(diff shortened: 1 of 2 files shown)


jsummerfield (changes requested):
Findings:

- Medium: `sources/community/universities/manifest.yaml:27` sets `pagination.mode: none`, but the Universities List API documents and supports `limit` / `offset` pagination for `/search`. As written, SQL queries with `LIMIT` still fetch the entire matching result set first, which is expensive for this API: live checks returned over 10k unfiltered rows and over 2k rows for `country = 'United States'`. Please model the documented pagination so Coral can stop early for common limited queries. Docs: https://github.com/Hipo/university-domains-list-api#pagination

The README example output and the HTTP base URL match live behavior; this is mainly about making the table efficient and predictable for normal first queries. Thanks for the clean source shape.

wiz-abhi:
Thanks for catching that! I've updated the `manifest.yaml` to model the documented `limit` / `offset` pagination instead of using `mode: none`. This should make SQL queries with `LIMIT` significantly more efficient by stopping early rather than fetching the entire dataset. I've verified the change locally using `coral source test`.

jsummerfield (approved):
Reviewed the current head again. The previous limit/offset pagination blocker is resolved, live pagination checks returned distinct pages, the source lints/tests successfully, live evidence is present, and the current checks are green.

I'm approving this.

PR #860, opened 2026-05-27
feat(sources/community/cat-facts): add Cat Facts source

wiz-abhi:
## Summary
This PR adds a new community source for **Cat Facts**, which interacts with the CatFact.ninja API to provide random and paginated cat facts.

## Tables Added
- **`random_fact`**: Retrieves a single random cat fact from the `/fact` endpoint. Pagination is disabled (`mode: none`) and `rows_path` is explicitly `[]` since the API returns a single JSON object.
- **`facts`**: Lists paginated cat facts from the `/facts` endpoint. Uses `rows_path: [data]` to unwrap the internal array, and configures pagination using `mode: page` via the `page` query parameter mapping seamlessly into the API's pagination behavior.

## Implementation Details
- **Authentication**: No authentication required.
- **Testing**: `test_queries` are provided to locally validate both the random and paginated endpoints. 

## Verification
Locally tested and linted with `coral source lint` and `coral source test`. Here is an example query showing live fetched results from the `random_fact` table:
```sql
SELECT fact, length FROM cat_facts.random_fact LIMIT 1;
```
```text
+-----------------------------------------------------------------------------+--------+
| fact                                                                        | length |
+-----------------------------------------------------------------------------+--------+
| In multi-cat households, cats of the opposite sex usually get along better. | 75     |
+-----------------------------------------------------------------------------+--------+
```

### DIFF
--- a/sources/community/cat-facts/README.md
+++ b/sources/community/cat-facts/README.md
@@ -1,0 +1,36 @@
+# Cat Facts
+
+Fetch random and paginated cat facts from the [CatFact.ninja API](https://catfact.ninja/).
+
+## Setup
+
+No authentication is required. Add the source:
+
+```bash
+coral source add --file sources/community/cat-facts/manifest.yaml
+```
+
+## Local Testing
+
+```bash
+coral sql "
+  SELECT fact, length 
+  FROM cat_facts.random_fact 
+  LIMIT 1
+"
+
+/*
++-----------------------------------------------------------------------------+--------+
+| fact                                                                        | length |
++-----------------------------------------------------------------------------+--------+
+| In multi-cat households, cats of the opposite sex usually get along better. | 75     |
++-----------------------------------------------------------------------------+--------+
+*/
+```
+
+## Tables
+
+| Table | Description |
+|-------|-------------|
+| `random_fact` | Get a single random cat fact. |
+| `facts` | List paginated cat facts. Supports querying up to maximum facts available. |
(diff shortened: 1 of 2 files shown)


jsummerfield (changes requested):
Findings:

- High: `sources/community/cat-facts/manifest.yaml:48` does not set `page_start: 1` for the paginated `/facts` table. Coral starts page pagination at 0 by default, while CatFact.ninja treats `page=0` and `page=1` as the same first page, so the source returns duplicate rows and can over-count the upstream dataset. In live testing, `LIMIT 105` repeated row 1 at row 101, and `LIMIT 500` returned 432 rows even though upstream reports 332 total facts. Please add `page_start: 1` to the pagination block and keep a multi-page test query or README note that verifies the table does not duplicate page 1. Docs/API behavior: https://catfact.ninja/docs?api-docs.json and https://catfact.ninja/facts?limit=100&page=0 / https://catfact.ninja/facts?limit=100&page=1

Thanks for adding this small source. The manifest lints and the single-page test path works; this pagination fix is the piece needed before it is safe to approve.

wiz-abhi:
Thanks for the review! I have updated the `manifest.yaml` to include `page_start: 1` in the pagination block for the `facts` table, and I also updated the test query to use `LIMIT 15` which will verify that the pagination properly fetches the next page without duplicating the first 10 rows.

jsummerfield (approved):
Reviewed the current head again. The previous pagination blocker is resolved: the paginated `facts` table now starts at page 1, the source lints and tests successfully, live pagination checks returned distinct rows, and the current checks are green.

I'm approving this.

jsummerfield:
On reflection - we have enough novelty community sources now to illustrate the possibilities. I'll only be approving work-related sources from now on. Thank you for your effort on this and the other sources.

PR #861, opened 2026-05-27
feat(sources/community/dog-ceo): add Dog CEO source

wiz-abhi:
## Summary
This PR adds a new community source for **Dog CEO**, which interacts with the Dog API to provide random dog images and breed-specific images.

## Tables Added
- **`random_image`**: Retrieves a single random dog image URL. The `message` property is mapped to the `image_url` column.
- **`random_images`**: Retrieves an array of random dog images. Requires the `count` filter dynamically injected into the path. The response array inside `message` is unwrapped by `rows_path: [message]`, and each array element is assigned to `image_url` via the `kind: current_row` expression.
- **`breed_images`**: Retrieves an array of images filtered by specific breed. Requires `breed` and `count` filters dynamically injected into the endpoint. Also utilizes `rows_path: [message]` and `current_row` to map the string array cleanly into individual records.

## Implementation Details
- **Authentication**: No authentication required.
- **Array Parsing**: Successfully handled the `{"message": ["url1", "url2"]}` structure by combining `rows_path` with `current_row` mapping since the array contains raw strings rather than objects.
- **Testing**: `test_queries` are provided for all three tables, effectively asserting the filter mechanisms logic locally.

## Verification
Locally tested and linted with `coral source lint` and `coral source test`. Here is an example query showing live fetched results from the `breed_images` table:
```sql
SELECT image_url FROM dog_ceo.breed_images WHERE breed = 'hound' AND count = 2 LIMIT 2;
```
```text
+---------------------------------------------------------------+
| image_url                                                     |
+---------------------------------------------------------------+
| https://images.dog.ceo/breeds/hound-blood/n02088466_10408.jpg |
| https://images.dog.ceo/breeds/hound-ibizan/n02091244_1340.jpg |
+---------------------------------------------------------------+
```

### DIFF
--- a/sources/community/dog-ceo/README.md
+++ b/sources/community/dog-ceo/README.md
@@ -1,0 +1,39 @@
+# Dog CEO
+
+Fetch random dog images and filter by breed using the [Dog CEO API](https://dog.ceo/dog-api/).
+
+## Setup
+
+No authentication is required. Add the source:
+
+```bash
+coral source add --file sources/community/dog-ceo/manifest.yaml
+```
+
+## Local Testing
+
+```bash
+coral sql "
+  SELECT image_url 
+  FROM dog_ceo.breed_images 
+  WHERE breed = 'hound' AND count = 2 
+  LIMIT 2
+"
+
+/*
++---------------------------------------------------------------+
+| image_url                                                     |
++---------------------------------------------------------------+
+| https://images.dog.ceo/breeds/hound-blood/n02088466_10408.jpg |
+| https://images.dog.ceo/breeds/hound-ibizan/n02091244_1340.jpg |
++---------------------------------------------------------------+
+*/
+```
+
+## Tables
+
+| Table | Description |
+|-------|-------------|
+| `random_image` | Get a single random dog image. |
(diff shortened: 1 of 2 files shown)


jsummerfield (commented):
A couple of small polish notes before this lands:

- `sources/community/dog-ceo/README.md:17`-`:19` have trailing whitespace in the SQL example; `git diff --check` flags those lines.
- `sources/community/dog-ceo/manifest.yaml:39` / `README.md:38` should mention that Dog CEO caps `/api/breeds/image/random/{count}` at 50 images. Live `count=99` returns 50 rows, so a short note prevents users from assuming larger counts will be honored. Docs: https://dog.ceo/dog-api/documentation/random

The source lints and the live no-auth test path works; these are not big issues, but they would make the first-use docs cleaner.

jsummerfield (approved):
Findings: no blocking findings.

I left two small polish notes separately, but the source itself is safe to merge: it is scoped to the Dog CEO community source, the manifest lints, the declared live tests pass, and the no-auth API behavior matches the documented endpoints for normal use.

PR #864, opened 2026-05-27
feat(sources/community/usgs-earthquakes): add USGS earthquake source

pooja-bhavani:
Adds a community source spec for the public USGS Earthquake Catalog API.

This source exposes earthquake events as SQL with filters for time, magnitude, radius, bounding box, event type, alert level, status, and ordering. It does not require authentication.

Validation:
- coral source lint sources/community/usgs_earthquakes/manifest.yaml
- coral source add --file sources/community/usgs_earthquakes/manifest.yaml
- coral source test usgs_earthquakes


jsummerfield (changes requested):
Findings:

- High: `sources/community/usgs_earthquakes/manifest.yaml:424` and `sources/community/usgs_earthquakes/manifest.yaml:431` expose `latitude` and `longitude` as echoes of the query filters, not as earthquake event coordinates. That is misleading for users: if they query near a point, every returned row can show the query point rather than the event location. Please expose event `longitude`, `latitude`, and `depth_km` from `geometry.coordinates` instead, or rename any filter echoes so they are clearly not event coordinates. USGS GeoJSON coordinates are `[longitude, latitude, depth]`: https://earthquake.usgs.gov/earthquakes/feed/v1.0/geojson.php

- Medium: `sources/community/usgs_earthquakes/manifest.yaml:63` says `order_by` supports `updated`, but USGS rejects `orderby=updated`. The documented values are `time`, `time-asc`, `magnitude`, and `magnitude-asc`; please remove `updated` from the source guidance. Docs: https://earthquake.usgs.gov/fdsnws/event/1/

- Medium: `sources/community/usgs_earthquakes/README.md:51` and `sources/community/usgs_earthquakes/README.md:75` make broad historical scans look straightforward. In practice, broad windows can fail upstream even when limited, and USGS documents a 20,000-result cap plus separate guidance for real-time feeds. Please add practical guidance to keep time, magnitude, and/or spatial filters narrow for reliable first success. Docs: https://earthquake.usgs.gov/fdsnws/event/1/

The manifest lints and the declared smoke tests pass, but the coordinate columns need to represent event coordinates before this is safe as a user-facing source. Thanks for the contribution; once those semantics are corrected, this should be much easier to re-check.

jsummerfield (changes requested):
Findings

- Medium: `sources/community/usgs_earthquakes/README.md:77` lists `order_by` as a filter but never explains when users must push ordering to USGS instead of using SQL `ORDER BY`. This matters for top-N questions: `ORDER BY magnitude DESC LIMIT 5` only sorts the fetched page, while `WHERE order_by = 'magnitude' LIMIT 5` asks the USGS API for the largest events. Please add a short example and note for provider-side ordering, especially for `time`, `time-asc`, `magnitude`, and `magnitude-asc`. USGS documents these `orderby` values here: https://earthquake.usgs.gov/fdsnws/event/1/

- Low: `sources/community/usgs_earthquakes/README.md:67` warns about broad scans, but it should also mention rate/cache behavior so agents do not poll this public API aggressively. A short note that USGS may return 429s and that responses can be cached is enough. USGS has published that guidance here: https://geohazards.usgs.gov/pipermail/realtime-feeds/2022-January/000028.html

I fixed the PR title to `feat(sources/community/usgs-earthquakes): add USGS earthquake source` and reran the title workflow; checks are green now. I also verified the current manifest locally in an isolated Coral config: `coral source add` and `coral source test usgs_earthquakes` both passed against the live public API. The remaining change is documentation so users get correct results for ordered/top-N queries.

pooja-bhavani:
Thanks for the review and for fixing the title/checks. Pushed documentation updates for both findings:

- **Medium** (`README.md` ordering): added an **Ordering and top-N queries** section. It explains that `order_by` pushes ordering down to USGS (`orderby`) and that USGS truncates server-side before Coral sees the rows, so `ORDER BY magnitude DESC LIMIT 5` only sorts the fetched page while `WHERE order_by = 'magnitude' LIMIT 5` asks USGS for the largest events across the catalog. Included a magnitude top-5 example and documented the supported values: `time`, `time-asc`, `magnitude`, `magnitude-asc`.
- **Low** (`README.md` rate/cache): added a note that USGS is a shared public API that may return `429 Too Many Requests` and may cache responses, so clients should avoid aggressive polling, linking the USGS realtime-feeds guidance.

`coral source lint sources/community/usgs_earthquakes/manifest.yaml` still passes. Ready for another look.

jsummerfield (approved):
Approved. The current head resolves the remaining README blockers: provider-side `order_by` is documented for top-N queries, the USGS rate/cache note is present, local lint and whitespace checks pass, live `coral source add` / `coral source test usgs_earthquakes` passed against the public API, and the required GitHub checks are green after workflow approval.

pooja-bhavani:
Thanks for the review and approval! All requested changes have been addressed and checks are passing.

PR #871, opened 2026-05-27
feat(mcp): native Streamable HTTP transport + OAuth for MCP sources

antonmry:
## User story

Today, connecting Coral to a remote MCP server like ClickHouse Cloud's `https://mcp.clickhouse.cloud/mcp` requires running the `mcp-remote` npm proxy as a stdio child of `coral-engine`, plus a one-time interactive browser dance to seed an OAuth token that lives in `~/.mcp-auth/`. The proxy is noisy (its stderr fires on every scan), couples Coral to a specific Node toolchain, and hides token refresh in an external process that can silently expire.

This PR makes the workflow first-party:

```bash
coral source add --file sources/community/clickhouse_cloud/manifest.yaml --interactive
# → pick "Connect with ClickHouse Cloud" at the CLICKHOUSE_ACCESS_TOKEN prompt
# → browser opens, you authorize, Coral stores the token
coral sql "SELECT id, name FROM clickhouse_cloud.organizations"
```

Reference https://github.com/withcoral/coral/pull/402

## What changed

**Manifest surface (`coral-spec`):** `server.transport` becomes a tagged enum with two variants — the existing `stdio` and a new `streamable_http` carrying `url` plus an optional `auth` block. HTTP auth is bearer-only for now, with the token sourced from an input via `{ type: bearer, from: input, key: ... }`. `https` is required for non-loopback URLs.

```yaml
server:
  transport: streamable_http
  url: https://mcp.clickhouse.cloud/mcp
  auth:
    type: bearer
    from: input
    key: CLICKHOUSE_ACCESS_TOKEN
```

**Runtime (`coral-engine/backends/mcp`):** new `StreamableHttpMcpToolCaller` alongside `StdioMcpToolCaller`. Both trait-implement `McpToolCaller`, so tables, functions, filters, pagination, response extraction, and Arrow conversion stay unchanged. `compile_manifest` branches on the server-spec variant.

**OAuth:** the bearer-token secret input is wired to Coral's existing install-time OAuth credential method (the same machinery Bitbucket and Slack already use). The manifest declares `authorization_url`, `token_url`, `pkce: required`, scopes, and a public-client `default` client_id pre-registered against ClickHouse Cloud's open Dynamic Client Registration endpoint. `coral source add --interactive` binds a loopback callback on port 53683, drives the browser flow, exchanges the code, and stores the access token.

**Error mapping:** rmcp errors are decoded into stable structured reason codes a user can see in query results and span `error.type` labels: `MCP_AUTH_REQUIRED` (401), `MCP_AUTH_FAILED` (403 with insufficient scope), `MCP_HTTP_REQUEST_FAILED` (network), `MCP_HTTP_STATUS_FAILED` (non-success status that isn't auth), `MCP_HTTP_SSE_DECODE_FAILED` (bad SSE / unexpected content type), `MCP_SESSION_EXPIRED` (404 with session id, reinit failed). Initialize, tool-call, tool-returned-error, decode, and pagination codes from the stdio baseline preserved. Token values never appear in spans or errors.

**Telemetry:** each `call_tool` is wrapped in an `mcp.tool.call` `client` span (target `coral_engine::mcp`) with `coral.source`, `coral.mcp.{relation,tool,transport,request_id}`, and OTel HTTP semantic-convention attributes (`url.full` sanitized, `server.address`, `server.port`, `peer.service`). Stdio adds `coral.mcp.command` / `coral.mcp.args.count`. Opt-in `McpBodyCapture` emits child preview spans for tool args and response payloads. On the streamable_http path, the parent span's W3C trace context is injected into the HTTP request via rmcp's `custom_headers` so an instrumented MCP server can continue the trace.

**Source migration:** `clickhouse_mcp` → `clickhouse_cloud` v0.2.0. Server block on `streamable_http` + bearer auth; README rewritten for the OAuth-driven setup; all `mcp-remote` references gone. Community-sources catalog regenerated.

## Design decisions

- **Streamable HTTP only.** The MCP 2025-06-18 spec also defines a legacy HTTP+SSE transport; we skip it since no real server we care about requires it.
- **No token refresh yet.** Re-run `coral source add --interactive` to mint a new access token when the existing one expires (1h TTL). Refresh-token support is being worked in parallel and this PR is designed to consume it once it lands.
- **No OAuth metadata discovery.** The manifest authors `authorization_url` / `token_url` explicitly rather than reading `WWW-Authenticate` or `/.well-known/oauth-protected-resource`. Determinism now; layer discovery on later.
- **No session pooling.** Each tool call opens a fresh MCP session. Acceptable for interactive use; gated on profiling per the roadmap.
- **Shared public client_id for ClickHouse Cloud.** PKCE protects each per-session code exchange and no client secret is issued, so a baked-in client_id is safe. Power users can swap in their own DCR-registered client.
- **Bearer-only auth.** The HTTP auth block is a tagged enum so other schemes (mTLS, custom headers) can be added later without breaking the manifest shape.
- **Body capture knob shared across backends.** Renamed `http_body_capture_max_bytes` → `body_capture_max_bytes` on `QueryRuntimeContext`; one TOML knob controls preview spans for both HTTP and MCP.

## Key bug found during integration

rmcp 1.6 declares its `reqwest` dep with `default-features = false` and only `["json", "stream"]` — **no TLS feature**. Wiremock unit tests over plain HTTP pass, but any real HTTPS POST fails at the reqwest layer with `error sending request for url (...)`. Fixed by enabling rmcp's `reqwest` feature (which adds `reqwest?/rustls`) in the workspace `Cargo.toml`. Load-bearing for the whole feature.

## Test plan

- [x] `cargo test -p coral-spec backends::mcp` — 23/23 pass (streamable_http parse/reject cases included)
- [x] `cargo test -p coral-engine backends::mcp` — 39/39 pass, including:
  - bearer + JSON tool call (wiremock)
  - parent `mcp.tool.call` span attributes
  - body-capture child spans
  - 401 → `AUTH_REQUIRED`, 502 → `HTTP_STATUS_FAILED`, unexpected content type → `HTTP_SSE_DECODE_FAILED`
  - rmcp 1.6 prefix-classifier table (pins the `#[error("...")]` strings)
- [x] Live end-to-end against `mcp.clickhouse.cloud` over Streamable HTTP + OAuth: `organizations`, `services` with WHERE pushdown, `databases` (scalar-row), `run_select_query` SQL passthrough, structured `MCP_AUTH_REQUIRED` on expired token, structured `MCP_TOOL_RETURNED_ERROR` on service-idle responses, MCP traces visible in OpenObserve with the full attribute set
- [x] Live OAuth browser flow via `coral source add --interactive`

## Future work

In priority order:

1. **Process and session pooling** — pool only if profiling shows init cost is material.
2. **Resource and resource-template backed tables** — bind tables to `resources/read` instead of only `tools/call`.
3. **Real MCP fixture and integration tests** — stdio + in-process HTTP fixtures to replace today's fake-caller + wiremock-only coverage.
4. **Manifest scaffolding** (`coral source scaffold-mcp`) — generate draft manifest from `tools/list` + schemas; never emit plaintext secrets.
5. **OAuth follow-ups** — token refresh, metadata discovery (`WWW-Authenticate`, protected-resource / authorization-server), Dynamic Client Registration.
6. **Finish public source-spec docs** — `tool_args`, `limit_binding`, typed filter values, combined pagination examples.
7. **(Optional) Inferred required filters** — recommended against in the roadmap; keep planning declarative.


docs/guides/observe-with-opentelemetry.mdx
```diff
@@ -84,10 +84,12 @@ trace_filter = "coral_app=trace,coral_client=trace,coral_mcp=trace,coral_engine=
...
+
 By default, the `coral_engine::datafusion` target is disabled for OTLP export to keep external span volume low. Enable it when you want to send DataFusion physical-plan execution and optimizer-rule spans to your external backend:
```
AlbertQM: The traces section is starting to become a big block of text, do you think this is a good time to refactor it?

AlbertQM (approved):
Docs LGTM

crates/coral-engine/src/backends/mcp/error.rs
```diff
@@ -35,6 +35,18 @@ pub(crate) enum McpProviderQueryError {
         detail: String,
     },
 
+    #[error("MCP HTTP transport for source '{source_schema}' requires authorization: {detail}")]
+    AuthRequired {
```
sauldhernandez: Shouldn't these transport errors be shared with the http backend?

crates/coral-engine/src/backends/mcp/trace.rs
```diff
@@ -0,0 +1,271 @@
...
+//! `backends::shared::trace`.
+
+use std::sync::atomic::{AtomicU64, Ordering};
+
+use rmcp::model::JsonObject;
+use serde_json::Value;
+
+use super::error::McpProviderQueryError;
+
+pub(super) const MCP_BODY_TRACE_TARGET: &str = "coral.mcp.body";
```
jsummerfield, line 17: Could we wire `coral.mcp.body` into the same local and OTLP filtering policy as `coral.http.body` before merging? The runtime now records MCP request and response previews under this target when body capture is enabled, but app telemetry only enables and hard-denies `coral.http.body`. That means local trace history will not include these MCP preview spans by default, and a broad OTLP `trace_filter` can export MCP args/results even though the docs promise body previews stay local. Adding `coral.mcp.body=off` / `coral.mcp.body=trace` and hard-denying it alongside HTTP body spans should keep the contract intact.

sources/community/clickhouse_cloud/manifest.yaml
```diff
@@ -1,20 +1,58 @@
 dsl_version: 3
-name: clickhouse_mcp
-version: 0.1.0
+name: clickhouse_cloud
```
jsummerfield, line 2: This rename changes the source name and therefore the visible SQL schema from `clickhouse_mcp` to `clickhouse_cloud`. Existing `clickhouse_mcp.*` queries and scripts that install from `sources/community/clickhouse_mcp/manifest.yaml` will break unless we preserve a compatibility path or call the migration out clearly. Could we either keep a `clickhouse_mcp` compatibility manifest/alias, or add a changelog and README migration note that tells users to reinstall and update queries?

crates/coral-spec/src/schema/source_manifest.schema.json
```diff
@@ -722,6 +728,43 @@
...
+        "url": { "type": "string", "minLength": 1 },
+        "auth": { "$ref": "#/$defs/mcp_http_auth" }
+      }
+    },
+    "mcp_http_auth": {
+      "oneOf": [
+        { "$ref": "#/$defs/mcp_bearer_auth" }
+      ]
+    },
+    "mcp_bearer_auth": {
```
jsummerfield, line 751: The schema should mirror the validated MCP auth contract here. Rust and docs require Streamable HTTP bearer auth to be `from: input` referencing a `kind: secret` input, but this schema still exposes the full generic value-source surface, so schema-only tooling will accept `from: literal`, `from: template`, or request-scoped sources that Coral rejects later. Could we restrict this object to `type: bearer`, `from: input`, and `key`, while keeping the secret-kind check in Rust?

docs/guides/observe-with-opentelemetry.mdx
```diff
@@ -84,10 +84,12 @@ trace_filter = "coral_app=trace,coral_client=trace,coral_mcp=trace,coral_engine=
...
 
+MCP-backed source calls (the `mcp` backend reaching out to a remote tool server) are exported as OpenTelemetry `client` spans with operation name set to the called tool. Attributes include `coral.source`, `coral.mcp.relation`, `coral.mcp.tool`, `coral.mcp.transport` (`stdio` or `streamable_http`), `coral.mcp.request_id`, and a `error.type` label drawn from the same stable codes the query surfaces (`AUTH_REQUIRED`, `AUTH_FAILED`, `HTTP_REQUEST_FAILED`, `HTTP_STATUS_FAILED`, `HTTP_SSE_DECODE_FAILED`, `SESSION_EXPIRED`, `TOOL_CALL`, `TOOL_RETURNED_ERROR`, `RESULT_DECODE`, `PAGINATION`, `INITIALIZE`, `SERVER_START`). Streamable HTTP calls additionally carry `url.full`, `server.address`, `server.port`, `peer.service`, `http.host`, and `net.peer.name`, and Coral injects W3C Trace Context headers into the HTTP request so an instrumented MCP server can continue the trace. Stdio calls carry `coral.mcp.command` and `coral.mcp.args.count`; trace-context propagation is not yet wired for stdio. These spans live under the `coral_engine::mcp` target — not to be confused with the `coral.mcp.*` server spans above, which come from Coral acting as the MCP server.
```
jsummerfield: This says MCP span `error.type` labels are the same stable codes surfaced by query errors, but the implementation has two namespaces: `McpProviderQueryError::to_structured()` emits `MCP_AUTH_REQUIRED`, `MCP_HTTP_STATUS_FAILED`, etc., while `mcp_error_type()` records `AUTH_REQUIRED`, `HTTP_STATUS_FAILED`, etc. Could we either make span `error.type` use the canonical structured reason codes, or document that traces intentionally use a shorter MCP-span namespace?

crates/coral-spec/src/backends/mcp.rs
```diff
@@ -409,6 +456,79 @@ fn validate_server(source_name: &str, server: &McpServerSpec) -> Result<()> {
     Ok(())
 }
 
+fn validate_streamable_http_server(
+    source_name: &str,
+    raw_url: &str,
+    auth: Option<&McpHttpAuthSpec>,
+    declared_inputs: &[ManifestInputSpec],
+) -> Result<()> {
+    let url = url::Url::parse(raw_url).map_err(|error| {
```
jsummerfield, line 469: Could we reject URL userinfo after parsing the Streamable HTTP endpoint? A URL such as `https://user:pass@mcp.example.com/mcp` currently passes validation and is then handed to rmcp as-is, which bypasses the new rule that MCP credentials come from secret inputs and can leak through manifest text or runtime error detail. Checking `url.username().is_empty()` and `url.password().is_none()` here would close that hole.

crates/coral-engine/src/backends/mcp/trace.rs
antonmry: Done with this one 49630a65d28140022a26eea9f6ff89ee071409d4 .

I think we might want to consolidate in a single config parameter if we have many more backends. But it's fine for now.

sources/community/clickhouse_cloud/manifest.yaml
antonmry: Solved with 73ce740b4cbbefc44903c3827b476e9610afa218 but not sure this is better. clickhouse_mcp was clearly experimental and I would be surprised if someone is using it

crates/coral-spec/src/schema/source_manifest.schema.json
antonmry: Good catch, it should follow the same structure we have with http. Fixed with d315022beacfa7fdea09e49085cad8ae56a43754

docs/guides/observe-with-opentelemetry.mdx
antonmry: Solved with e70c6485786a1e4a0dc5307fb1de76ae6e71350b

crates/coral-spec/src/backends/mcp.rs
antonmry: solved with d9c9d52e734f61702d90bdab0ba30eac5d2823ee

docs/guides/observe-with-opentelemetry.mdx
```diff
@@ -36,14 +36,14 @@ To disable trace history:
...
-HTTP request and response body previews are not recorded by default. To enable local body recording:
+HTTP and MCP request and response body previews are not recorded by default. To enable local body recording:
 
 ```toml
 [trace_history]
 record_http_bodies = true
 ```
 
-Recorded HTTP body previews are capped by `http_body_max_bytes` bytes per body (`65536` by default). Body previews can contain sensitive data; keep them disabled unless the local debugging value is worth that exposure. Body preview spans are never exported to OTLP, even if a broad `[otel].trace_filter` would otherwise match `coral.http.body`.
+The same flag enables HTTP body previews under the `coral.http.body` target and MCP tool-call argument and result previews under the `coral.mcp.body` target. Recorded body previews are capped by `http_body_max_bytes` bytes per body (`65536` by default). Body previews can contain sensitive data; keep them disabled unless the local debugging value is worth that exposure. Body preview spans are never exported to OTLP, even if a broad `[otel].trace_filter` would otherwise match `coral.http.body` or `coral.mcp.body`.
```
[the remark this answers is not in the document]
antonmry: Fixed 4714da2a73f2fb04745274730a98e1f83097d8d3

docs/guides/observe-with-opentelemetry.mdx
antonmry: Good point, refactored here b36a2146b846e2e593f426dae465b3c792d42a2f

crates/coral-engine/src/backends/mcp/mod.rs
```diff
@@ -95,11 +96,22 @@ pub(crate) fn compile_manifest(
...
+        McpServerSpec::Stdio { .. } => Arc::new(StdioMcpToolCaller {
+            source_name: manifest.common.name.clone(),
+            server: manifest.server.clone(),
+            source_inputs: Arc::clone(&source_inputs),
+            body_capture,
+        }),
+        McpServerSpec::StreamableHttp { .. } => Arc::new(StreamableHttpMcpToolCaller {
+            source_name: manifest.common.name.clone(),
+            server: manifest.server.clone(),
+            resolved_inputs: Arc::clone(&resolved_inputs),
```
jsummerfield: This should pass `source_inputs` into `StreamableHttpMcpToolCaller` rather than the install-time `resolved_inputs`. After the rebase, HTTP sources and stdio MCP env/tool args use `SourceInputResolver::resolve_inputs` per request so OAuth-backed secrets can be refreshed before a query. Streamable HTTP auth still renders against this frozen map, so an MCP HTTP source keeps sending the stale access token even after the shared credential layer refreshes and persists a new one. That contradicts the source-spec OAuth docs and is especially sharp for `clickhouse_cloud`, whose access token is short-lived. The fix should mirror stdio: store the `McpSourceInputs` resolver on the HTTP caller, resolve it inside `call_tool_inner`, and add a regression test with a rotating resolver asserting the Authorization header changes between calls.

antonmry: Still working on it, the PR isn't ready for re-review yet 👍

antonmry: Done  0b2bad9d

crates/coral-engine/src/backends/mcp/transport.rs
```diff
@@ -127,3 +384,766 @@ fn value_to_env_string(value: Value) -> String {
...
+/// Classify an rmcp `ClientInitializeError` / `ServiceError` raised by the
+/// Streamable HTTP transport into a structured `McpProviderQueryError`.
+///
+/// rmcp wraps the underlying `StreamableHttpError` in a `DynamicTransportError`
+/// whose inner `Box<dyn Error>` we cannot typed-downcast: rmcp parameterizes
+/// the box with its own `reqwest::Error` (rmcp 1.6 pulls reqwest 0.13) while
+/// coral-engine uses reqwest 0.12, so the `StreamableHttpError<reqwest::Error>`
+/// `TypeId`s do not match across the crate boundary. Instead we match on the
+/// inner error's stable `#[error("...")]` display prefix from
+/// `rmcp::transport::streamable_http_client::StreamableHttpError`.
```
jsummerfield: This comment will age quickly. Also - why not bump the coral version to avoid this workaround?

jsummerfield (approved):

jsummerfield:
Nice work!

crates/coral-engine/src/backends/mcp/transport.rs
antonmry: Fixed with 737f2b1b0eb878aa2b66af7d68cdbf96269a2f38

This is trickier that it should. I had to enable the OpenSSL license to be able to do the upgrade without some nasty workarounds. I think OpenSSL license should be right, it's more a validator issue than a real issue.

crates/coral-engine/src/backends/mcp/transport.rs
```diff
@@ -127,3 +361,715 @@ fn value_to_env_string(value: Value) -> String {
...
+                detail: full_detail,
+            },
+            SHE::InsufficientScope(_) => McpProviderQueryError::AuthFailed {
+                source_schema: source_schema.to_string(),
+                detail: full_detail,
+            },
+            SHE::SessionExpired => McpProviderQueryError::SessionExpired {
+                source_schema: source_schema.to_string(),
+            },
+            SHE::UnexpectedServerResponse(_) => McpProviderQueryError::HttpStatusFailed {
```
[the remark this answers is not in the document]
antonmry: fixed with 7c32d30193275aa7f3b8f120a034a6e84a88560d

PR #874, opened 2026-05-27
fix(cli): support pasted OAuth redirects

sauldhernandez:
## Intent

OAuth source setup can fail in VM or SSH workflows when the browser runs outside the environment hosting Coral's loopback callback listener. This change lets users complete the existing OAuth flow by pasting the final localhost redirect URL into the terminal.

## What it enables

- Keeps the normal browser callback path unchanged.
- Adds a terminal fallback for OAuth authorization events in `coral source add --interactive`.
- Validates pasted redirect URLs against the expected loopback host, port, and path before forwarding them to the local callback listener.
- Documents the VM/SSH fallback behavior in the CLI and source-spec references.

## Usage examples

Run an OAuth-backed source install as usual:

```sh
coral source add --interactive slack
```

If the browser cannot reach Coral's localhost callback from the VM or SSH session, copy the final redirect URL from the browser and paste it at the `Redirect URL` prompt in the terminal.

## Validation

- `cargo test -p coral-cli source_ops::tests`
- `env CARGO_BUILD_RUSTC_WRAPPER= make rust-checks`

The plain sandboxed `make rust-checks` run was blocked by local `sccache` and loopback-bind restrictions; the escalated wrapper-cleared gate passed fmt, clippy, 805 nextest tests, and rustdoc.


crates/coral-cli/src/source_ops.rs
```diff
@@ -991,6 +1031,258 @@ fn oauth_error(action: &str, error: &tonic::Status) -> anyhow::Error {
...
+        ));
+    }
+    Ok(())
+}
+
+fn validate_oauth_redirect_url(
+    callback_url: &Url,
+    expected_redirect_uri: &Url,
+) -> Result<(), anyhow::Error> {
+    validate_loopback_http_redirect(callback_url)?;
```
jsummerfield: Could we validate the pasted URL's `state` against the `state` parameter in the authorization URL before forwarding it? Right now any query string on the expected loopback host/port/path passes here. If a user pastes a stale redirect for the same callback, the CLI sends it to the active listener; the app then treats the state mismatch as a terminal OAuth failure instead of giving the user a local validation error and a chance to retry.

crates/coral-cli/src/source_ops.rs
```diff
@@ -991,6 +1031,258 @@ fn oauth_error(action: &str, error: &tonic::Status) -> anyhow::Error {
...
+        return None;
+    }
+
+    let cancel = Arc::new(AtomicBool::new(false));
+    let worker_cancel = Arc::clone(&cancel);
+    let handle = thread::spawn(move || match read_oauth_redirect_prompt(&worker_cancel) {
+        Ok(Some(value)) if !value.trim().is_empty() => {
+            match submit_oauth_redirect_url(value.trim(), &expected_redirect_uri) {
+                Ok(()) => println!("Submitted OAuth redirect for {label}."),
+                Err(error) => eprintln!("Could not submit OAuth redirect URL: {error}"),
```
jsummerfield: Can we keep the paste prompt alive after a rejected submit? This logs the error and lets the worker exit, but the OAuth stream is still waiting. For the VM/SSH copy-paste case, a malformed URL, transient connect failure, or non-2xx callback response leaves the user with no prompt to paste the corrected redirect.

docs/reference/source-spec-reference.mdx
```diff
@@ -965,6 +965,9 @@ configure the matching loopback redirect behavior in the OAuth app. Coral
 starts the loopback flow and returns the provider authorization URL. After the
 callback, Coral exchanges the code, stores the access token as the source
 secret, and stores any refresh token as internal credential metadata.
+The CLI also accepts the final loopback redirect URL in the terminal while it
```
jsummerfield, line 1013: Could we also reflect this new fallback in the maintained source-spec skills (`plugins/coral/skills/coral-create-source-spec/...` and `coral-review-source-spec`)? The repo guidance says credential retrieval/auth guidance changes need the reader docs and agent-facing source-spec guidance to stay aligned, so agents authoring or reviewing OAuth specs should know the terminal paste fallback exists.

jsummerfield (commented):
Review comments on the OAuth redirect paste fallback.

jsummerfield (approved):

PR #877, opened 2026-05-27
feat(sources/community/maven_central): add Maven Central community source

git-sakshii:
## Summary

Adds a community source spec for **Maven Central** — the standard artifact registry
for Java/JVM libraries. No API key required.

**Files:** `sources/community/maven_central/manifest.yaml` + `README.md`

## Tables

| Table | Description | Required Filter |
|-------|-------------|-----------------|
| `artifacts` | Search by keyword, one row per groupId:artifactId | `WHERE q = '...'` |
| `versions` | All published versions of a specific artifact | `WHERE q = 'g:"groupId" AND a:"artifactId"'` |

## Design

- Single `q` filter maps directly to Solr `q` param on both tables
- Page pagination via `start` offset (`page_start: 0`, `page_step: 50`)
- `expr: kind: path` remaps JSON fields (`latestVersion` → `latest_version`, `p` → `packaging`, etc.)
- `virtual: true` echo column for `q` via `from_filter`
- `guide` fields on both tables for agent discoverability

## Test Results

- ✅ YAML valid
- ✅ Artifacts endpoint: `latestVersion`, `versionCount`, `p`, `repositoryId` all populated
- ✅ Versions endpoint: `v`, `p`, `timestamp` all populated
- ✅ Pagination via `start` offset works (21,230 total results)
- ✅ No auth needed


jsummerfield (changes requested):
Findings:

- High: `sources/community/maven_central/manifest.yaml:39` hard-codes `rows: "50"` while pagination also emits `rows` at `sources/community/maven_central/manifest.yaml:51`; the same duplicate-param issue exists for `versions` at `sources/community/maven_central/manifest.yaml:140` and `:152`. That can produce requests like `rows=50&rows=200`, and Maven Central may honor the first value, so queries asking for more than 50 rows still return only 50. Please remove the hard-coded `rows` request params and let pagination own the page size. Docs: https://central.sonatype.org/search/rest-api-guide/ and https://solr.apache.org/guide/solr/latest/query-guide/common-query-parameters.html#rows-parameter

- High: `sources/community/maven_central/manifest.yaml:46` models Solr `start` as page pagination with `page_step: 50`; same at `sources/community/maven_central/manifest.yaml:147`. Solr `start` is an absolute offset, not a page number. Please use offset pagination, e.g. `mode: offset`, `offset_param: start`, `offset_start: 0`, with page size controlled by `rows`, so larger limits do not overlap or skip pages. Docs: https://solr.apache.org/guide/solr/latest/query-guide/common-query-parameters.html#start-parameter and https://solr.apache.org/guide/solr/latest/query-guide/pagination-of-results.html#basic-pagination-examples

- Medium: `sources/community/maven_central/README.md:30` and `sources/community/maven_central/README.md:79` claim the `versions` table returns all published versions, but `sources/community/maven_central/manifest.yaml:123` sets `fetch_limit_default: 50`. Please either document that the default returns the first 50 and users should raise `LIMIT` for more, or change the default to match the promise.

The manifest lints, but the current pagination shape can give incomplete results while looking successful. Thanks for the focused source; fixing the `rows`/`start` handling should make this much more reliable.

git-sakshii:
All review findings have been resolved in the latest commit:

* **Removed Hard-coded `rows` Parameter (High):** Removed the hard-coded `rows: "50"` from the `request.query` block in both the `artifacts` and `versions` tables. The pagination configuration now uniquely controls the request page size using `query_param: rows` to prevent duplicate parameter injection.
* **Implemented Offset Pagination (High):** Refactored the pagination blocks for both tables from `mode: page` to `mode: offset` with `offset_param: start` and `offset_start: 0`. This correctly maps to Solr's absolute record offset behavior and ensures pagination works reliably for larger result sets without overlaps or skipped pages.
* **README Sync (Medium):** Updated `README.md` to change "All published versions" to "List published versions" and updated example comments to align with the default limit of 50.

**Validation Details:**
* `coral source lint` confirms the manifest is valid.
* The source was added locally, and queries using `LIMIT 60` (such as `SELECT count(*) FROM (SELECT id FROM maven_central.versions WHERE q = 'g:"com.google.guava" AND a:"guava"' LIMIT 60)`) were verified to fetch and return the correct row count across page offsets.

jsummerfield (approved):
Thanks for the update. I re-reviewed current head `3bf1599234d06f32cb400128f4903e33bb09903c`, and the previous Maven Central blockers look resolved:

- Solr `start` is now modeled as offset pagination instead of page numbering.
- `rows` is owned by the pagination page-size config rather than duplicated in request params.
- The README no longer overpromises that the default `versions` query returns every published version.

I also verified current-head live behavior through the source-review pass: lint/add/test passed, and a `LIMIT 60` versions query returned 60 rows across offsets. `Validate`, `Lint PR Title`, and CLA are green.

PR #879, opened 2026-05-27
ci(release): advance docs branch to released commit on release

AlbertQM:
## Summary

Today, when landing a PR, docs are updated immediately; our binary follows a different release cadence. This means docs can document things that aren't shipped yet.

In this PR:
Repoint the live docs site (which Mintlify deploys) to a long-lived `docs` branch that lags `main` and only advances to a shipped release. The publish-version job force-pushes the released tag commit to `docs` after the release is promoted, so the bundled-sources catalog never advertises a source the latest released binary cannot run. The docs-advance steps run in `publish-version` on both `release` events and manual `workflow_dispatch` runs against a tag, so `docs` can be (re)pointed at a shipped tag by re-running this workflow.

## Test plan

Nothing. We'll have to do some manual steps after this lands to make it actually work.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -266,16 +266,16 @@ jobs:
           } >> "$GITHUB_OUTPUT"
 
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
-        if: github.event_name == 'release'
         with:
           ref: ${{ env.TAG_NAME }}
           fetch-depth: 0
           token: ${{ steps.github-app-token.outputs.token }}
           persist-credentials: false
 
-      # Force push targets `docs` only — never `main`.
+      # Force push targets `docs` only — never `main`. Runs on both the
+      # `release` event and manual `workflow_dispatch` against a tag, so docs
+      # can be (re)pointed at a shipped tag by re-running this workflow.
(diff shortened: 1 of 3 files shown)


jsummerfield (approved):

.github/workflows/release.yml
```diff
@@ -266,6 +265,26 @@ jobs:
             echo "__RELEASE_BODY__"
           } >> "$GITHUB_OUTPUT"
 
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        if: github.event_name == 'release'
```
simonwhitaker: Please remove the `if:` line. We want this to run on `workflow_dispatch`, too.

.github/workflows/release.yml
```diff
@@ -266,6 +265,26 @@ jobs:
...
+        with:
+          ref: ${{ env.TAG_NAME }}
+          fetch-depth: 0
+          token: ${{ steps.github-app-token.outputs.token }}
+          persist-credentials: false
+
+      # Force push targets `docs` only — never `main`.
+      - name: Point docs branch at released commit
+        if: github.event_name == 'release'
```
simonwhitaker: Ditto

simonwhitaker (approved):

PR #886, opened 2026-05-27
feat(engine): account query memory usage

iaptsiauri:
## Summary

Adds engine-wide query memory policy and wires it into DataFusion's `MemoryPool`.

- Adds `[engine.memory] limit = "2Gi"` config support with typed `MemorySize` parsing for `Ki`, `Mi`, `Gi`, and `Ti` units.
- Applies the configured memory limit when building the DataFusion runtime, and preserves that limit when rebuilding the fallback runtime without DPP.
- Introduces reusable Coral retained-memory helpers for custom physical operators.
- Makes `DependentJoinExec` account retained resolver batches, binding tuple copies, dependent JSON results, and output Arrow batches before storing them.
- Keeps output memory reservations alive through the output stream lifetime.
- Adds tests for config parsing/rendering, stale config cleanup, runtime memory limit wiring, fallback preservation, retained tuple accounting, and DPP memory exhaustion.

## Notes

This is V1 memory-pool integration. It accounts Coral-owned retained memory, not every transient allocation. HTTP response decoding and temporary resolver-batch projection maps can still allocate working memory outside the pool before retained data is reserved.

## Validation

- `cargo test -p coral-app state::config::tests --lib`
- `cargo test -p coral-engine dependent_join_respects_engine_memory_limit --test engine`
- `make rust-checks`


crates/coral-engine/tests/engine/dependent_join_execution_tests.rs
```diff
@@ -173,6 +174,58 @@ async fn duplicate_resolver_rows_for_one_binding_emit_distinct_join_batches() {
...
+          ON pr.owner = i.github_owner
+         AND pr.repo = i.github_repo
+         AND pr.number = i.github_pr_number
+        ",
+    )
+    .await
+    .expect_err("DPP should fail when retained memory exceeds engine limit");
+
+    assert_eq!(error.status_code(), StatusCode::Unavailable);
+    assert_error_contains(&error, "Additional allocation failed");
```
antonmry, line 369: Could we make this user-facing message a bit more actionable? When the memory limit is reached the query aborts, so it would help the user/agent running the query if the error explains that the engine memory budget was exceeded and suggests the available next steps, such as increasing `[engine.memory].limit` in `config.toml` or narrowing the query/source rows. Including the configured limit in the message would also make this easier to diagnose.

crates/coral-engine/src/contracts/query.rs
```diff
@@ -213,11 +216,118 @@ impl QueryRuntimeConfig {
...
             dependent_join: DependentJoinConfig::default(),
         }
     }
 }
 
+/// Engine-wide query memory policy.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct QueryMemoryConfig {
+    /// Optional total query-engine memory limit.
+    pub limit: Option<MemorySize>,
```
antonmry, line 407: Non-blocking thought for the config shape: this engine-wide limit is a good V1, but it may be worth leaving room for source-scoped retained-memory budgets later. For example, Coral-owned operators like `DependentJoinExec(github.pull_requests)` already know the source/table they are retaining data for, so a future `per_source`/`per_table` limit could give users a more targeted way to cap one expensive source without shrinking the whole query runtime budget.

antonmry, line 407: Another non-blocking config-shape thought: DataFusion has multiple pool strategies here, and `RuntimeEnvBuilder::with_memory_limit` currently gives us the tracked greedy pool by default. If we expect users to tune memory globally, it may be worth leaving room to expose the pool strategy as a global parameter later, for example greedy vs fair-spill, rather than making `[engine.memory]` permanently mean only a byte limit.

crates/coral-engine/src/runtime/memory.rs
```diff
@@ -0,0 +1,213 @@
...
+    }
+}
+
+impl RecordBatchStream for RetainedRecordBatchStream {
+    fn schema(&self) -> arrow::datatypes::SchemaRef {
+        std::sync::Arc::clone(&self.schema)
+    }
+}
+
+fn json_value_payload_size(value: &Value) -> usize {
```
sauldhernandez: P1: This materially undercounts retained dependent JSON rows. `buffer_fetch_result` reserves before storing, but this estimator charges only payload bytes: `{}`, `[]`, and `null` can cost zero while `Vec<Value>`, object/map overhead, and row/container overhead are retained in `buffered_results`. A filter-only or empty-object response can therefore bypass `[engine.memory].limit` while still retaining a large amount of heap.

crates/coral-engine/src/runtime/dependent_join/output.rs
```diff
@@ -71,34 +63,36 @@ pub(crate) fn build_joined_batches(
...
                 &dependent_batch,
-                resolver_projection_len,
-                dependent_first,
-                Arc::clone(output_schema),
-            )?);
+                config.resolver_projection_len,
+                config.dependent_first,
+                Arc::clone(config.output_schema),
+            )?;
+            output_batches.push(batch)?;
```
sauldhernandez: P2: The reservation happens after `join_for_resolver_rows` has already allocated the join index vectors and Arrow `take` output arrays. A large fanout can spike or OOM before this configured memory budget returns the intended `ResourcesExhausted` path. The output reservation needs to happen before, or be bounded during, the allocation that creates the retained batch.

crates/coral-app/src/state/config.rs
```diff
@@ -74,6 +78,14 @@ struct PersistedAppConfig {
...
     #[serde(default)]
     dependent_join: PersistedDependentJoinConfig,
+    #[serde(default)]
+    memory: PersistedMemoryConfig,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+struct PersistedMemoryConfig {
+    #[serde(default)]
+    limit: Option<String>,
```
sauldhernandez: P2: Requiring this to deserialize as a TOML string means a plausible hand-written `limit = 2147483648` fails during whole-config decode before `try_into_runtime_config` can return the actionable `InvalidInput` parser error. Because `TomlDecode` maps to `INTERNAL`, this can break unrelated source/catalog operations with a server-error classification. Consider accepting a raw TOML value here and validating the shape in the memory parser.

crates/coral-engine/src/runtime/query.rs
```diff
@@ -468,6 +488,34 @@ fn is_missing_required_filter_failure(error: &SqlExecutionFailure) -> bool {
     )
 }
 
+fn memory_budget_error(error: &DataFusionError, limit: MemorySize) -> Option<CoreError> {
+    let DataFusionError::ResourcesExhausted(detail) = error.find_root() else {
+        return None;
+    };
+
+    Some(CoreError::Unavailable(format!(
+        "query engine memory budget exceeded ([engine.memory].limit = {}). The query was aborted. \
```
sauldhernandez, line 523: P2: This new user-facing error points users at `[engine.memory].limit`, but the engine configuration reference still documents only `[engine.dependent_join]`. Since the parser accepts only strict binary units (`Ki`, `Mi`, `Gi`, `Ti`) and rejects common guesses like `2GiB` or raw bytes-as-string, this needs a concise `[engine.memory]` docs section in the same PR.

sauldhernandez (changes requested):

AlbertQM (approved):
Docs look good, thanks!

sauldhernandez (approved):

PR #892, opened 2026-05-27
feat(sources/community/servicenow): add ServiceNow source

Akash-How:
Summary:
- Adds a community Coral source for ServiceNow ITSM and CMDB metadata.
- Exposes incidents, change requests, problems, CMDB configuration items, and users.
- Uses explicit sysparm_fields, offset pagination, and read-only metadata queries.

Validation:
- YAML parsing: passed
- git diff --check: passed
- make lint-sources: passed
- Live API tests: passed against a ServiceNow PDI

### DIFF
--- a/sources/community/servicenow/README.md
+++ b/sources/community/servicenow/README.md
@@ -11,13 +11,22 @@ It does not create, update, or delete ServiceNow records.
 ## Authentication
 
 Use a least-privilege ServiceNow integration user with read access to the
-tables you want Coral to inspect.
+tables you want Coral to inspect. This manifest currently supports ServiceNow
+Basic authentication only. If your instance requires OAuth or another auth
+scheme, use a separate source variant rather than pasting bearer tokens into
+the password input.
 
 | Input | Kind | Description |
 |---|---|---|
 | `SERVICENOW_INSTANCE_URL` | variable | ServiceNow instance URL, without a trailing slash. |
 | `SERVICENOW_USERNAME` | variable | ServiceNow username for Basic auth. |
-| `SERVICENOW_PASSWORD` | secret | ServiceNow password or API credential. |
+| `SERVICENOW_PASSWORD` | secret | ServiceNow password for the Basic auth integration user. |
+
+Provider references:
+
+- [ServiceNow REST API overview](https://www.servicenow.com/docs/r/api-reference/rest-api-explorer/c_RESTAPI.html)
+- [ServiceNow Table API](https://www.servicenow.com/docs/r/api-reference/rest-apis/c_TableAPI.html)
+- [ServiceNow inbound REST API rate limiting](https://www.servicenow.com/docs/r/api-reference/rest-api-explorer/inbound-REST-API-rate-limiting.html)
 
 ## Setup
 
@@ -63,8 +72,10 @@ Configuration items by class:
 ```sql
 SELECT sys_class_name, COUNT(*) AS ci_count
 FROM servicenow.cmdb_ci
+WHERE query = 'install_status=1'
 GROUP BY sys_class_name
-ORDER BY ci_count DESC;
+ORDER BY ci_count DESC
+LIMIT 25;
 ```
 
 Active users:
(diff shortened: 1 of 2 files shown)


jsummerfield (changes requested):
Findings:

- High: `sources/community/servicenow/manifest.yaml:42`, `sources/community/servicenow/manifest.yaml:165`, `sources/community/servicenow/manifest.yaml:278`, `sources/community/servicenow/manifest.yaml:376`, and `sources/community/servicenow/manifest.yaml:478` expose high-cardinality ServiceNow Table API scans with offset pagination but no `fetch_limit_default` or required narrowing filter. An unconstrained query can walk every incident, change, problem, CI, or user record visible to the integration account in a production instance. Please add conservative default fetch limits, push users toward `query` plus SQL `LIMIT`, and send `sysparm_no_count=true` where appropriate so ServiceNow does not do extra count work Coral does not need. Docs: https://www.servicenow.com/docs/r/api-reference/rest-apis/c_TableAPI.html and https://www.servicenow.com/docs/r/api-reference/rest-api-explorer/inbound-REST-API-rate-limiting.html

- Medium: `sources/community/servicenow/manifest.yaml:8` uses the unversioned `/api/now/table` base path. ServiceNow's Table API docs describe version-dependent empty-result behavior: v1 can return 404 for no records, while v2 returns 200 with an empty result array. Please pin the source to `/api/now/v2/table`, or add explicit empty-result handling if you intentionally need unversioned/v1 behavior. Docs: https://www.servicenow.com/docs/r/api-reference/rest-apis/c_TableAPI.html

- Medium: `sources/community/servicenow/manifest.yaml:50` and the repeated `display_value` filters let the same SQL columns change meaning from raw sys_ids/choice values to display strings depending on a query-time flag. That makes columns like `assigned_to`, `caller_id`, `assignment_group`, and choice fields unstable across queries. Please keep raw ID/value columns stable and add separate display columns, for example using documented display-value behavior or dot-walked fields, instead of changing the type/meaning of existing columns. Docs: https://www.servicenow.com/docs/r/api-reference/rest-apis/c_TableAPI.html

- Medium: `sources/community/servicenow/README.md:11` and `sources/community/servicenow/README.md:79` do not include provider docs links or enough auth/rate-limit caveats for a new user. Please link the Table API, encoded query guidance, ACL/auth behavior, and inbound REST rate limiting, and clarify that this manifest currently supports Basic auth only even though the credential text says "password or API credential". Docs: https://www.servicenow.com/docs/r/api-reference/rest-api-explorer/c_RESTAPI.html and https://www.servicenow.com/docs/r/api-reference/rest-api-explorer/inbound-REST-API-rate-limiting.html

Thanks for including live validation and keeping the scope to read-only ITSM/CMDB tables. The main thing to tighten is making this safe and predictable against real production ServiceNow instances.

Akash-How:
Addressed the ServiceNow source safety/predictability review feedback.

Changes:
- Pinned the Table API base path to `/api/now/v2/table`.
- Added `fetch_limit_default: 100` to each high-cardinality table.
- Added `sysparm_no_count=true` to paginated table requests.
- Removed the query-time `display_value` filter so raw columns stay stable.
- Request `sysparm_display_value=all` internally and expose separate `*_display` columns for display text.
- Expanded README auth, ACL, encoded-query, rate-limit, v2, and production-scan guidance with provider doc links.

Validation:
- YAML parsing: passed
- git diff --check: passed
- make lint-sources: passed
- Live Coral test: passed against the ServiceNow PDI after the manifest changes
- Representative query verified raw/display columns remain separate: `state` = raw value, `state_display` = display text.

jsummerfield (approved):
Reviewed the current head again. The previous blockers are resolved: the source now uses the versioned `/api/now/v2/table` surface, scans are bounded with defaults/pagination, raw values and display values are separated, and the README now covers auth, rate-limit/ACL caveats, and live Coral evidence. The current checks are green and `coral source lint` passes.

One non-blocking documentation improvement: ServiceNow documents that invalid portions of `sysparm_query` can be ignored by default, which can silently broaden a query. It would be worth adding a note that users should copy/test encoded queries from the UI or REST API Explorer and keep SQL `LIMIT`s in place. That does not block approval.

Approving this head. Thanks for tightening this up.

PR #893, opened 2026-05-27
feat(sources/community/fivetran): add Fivetran source

Akash-How:
Summary:
- Adds a community Coral source for Fivetran account metadata.
- Exposes groups, destinations, connections, users, and transformation projects.
- Omits connector and destination config payloads that may contain connection details.

Validation:
- YAML parsing: passed
- git diff --check: passed
- make lint-sources: passed
- Live API tests: passed against a Fivetran account

### DIFF
--- a/sources/community/fivetran/README.md
+++ b/sources/community/fivetran/README.md
@@ -1,0 +1,93 @@
+# Fivetran
+
+Query Fivetran account metadata from Coral. The source covers group,
+destination, connection, user, and dbt transformation project inventory without
+returning connector or destination configuration payloads that may contain
+connection details.
+
+## Authentication
+
+Create a scoped Fivetran REST API key and provide:
+
+| Input | Description |
+| --- | --- |
+| `FIVETRAN_API_KEY` | Fivetran API key. |
+| `FIVETRAN_API_SECRET` | Fivetran API secret. |
+
+Both values are modeled as secrets. Use the narrowest RBAC scope that can read
+the metadata you want Coral agents to inspect.
+
+## Tables
+
+| Table | Description |
+| --- | --- |
+| `fivetran.groups` | Groups accessible to the API key. |
+| `fivetran.destinations` | Destination inventory. Supports `group_id`. |
+| `fivetran.connections` | Connection status and sync metadata. Supports `group_id` and `schema`. |
+| `fivetran.users` | Account users visible to the API key. |
+| `fivetran.transformation_projects` | dbt Core transformation project metadata. Supports `group_id`. |
+
+## Examples
+
+List paused or failing connections:
+
+```sql
+SELECT id, service, schema, paused, setup_state, sync_state, update_state
+FROM fivetran.connections
+WHERE paused = true OR sync_state <> 'succeeded';
+```
+
+Inspect connections in one group:
+
+```sql
+SELECT id, service, schema, succeeded_at, failed_at
+FROM fivetran.connections
+WHERE group_id = 'group_id';
+```
+
+Review destination coverage:
+
+```sql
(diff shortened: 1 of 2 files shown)


jsummerfield (changes requested):
Findings:

- High: `sources/community/fivetran/manifest.yaml:313` uses `GET /dbt/projects`, but Fivetran's current REST API documents transformation projects at `GET /v1/transformation-projects`. The table also sends `group_id`, while the documented list endpoint accepts pagination parameters, not a `group_id` query filter. Please move the table to the documented path, remove the request-level `group_id` filter unless the docs support it, and align the columns to the documented response fields. Docs: https://fivetran.com/docs/rest-api/api-reference/transformation-projects-management/list-all-transformation-projects

- Medium: `sources/community/fivetran/manifest.yaml:81` and `sources/community/fivetran/manifest.yaml:88` advertise and send a `group_id` filter for `fivetran.destinations`, but Fivetran's List All Destinations endpoint documents cursor pagination parameters rather than a `group_id` query parameter. Please remove the remote filter and README claim, or make it clear that users can filter the returned `group_id` locally in SQL. Docs: https://fivetran.com/docs/rest-api/api-reference/destinations/list-destinations

- Medium: `sources/community/fivetran/README.md:8` tells users to create API credentials but does not say where to create them or link the official auth setup, and the notes omit relevant rate-limit behavior. Please add credential setup guidance, mention scoped/system key expectations, and document 429/`Retry-After` handling or account-level API quotas. Docs: https://fivetran.com/docs/rest-api/getting-started and https://fivetran.com/docs/rest-api/getting-started/rate-limiting

- Medium: `sources/community/fivetran/README.md:61` includes live Coral evidence, but the declared tests only exercised empty `groups` and `connections` results. That would not catch the broken transformation projects endpoint or the unsupported destinations filter. Please add sanitized live output or declared test coverage for the other exposed tables, especially `destinations`, `users`, and `transformation_projects`.

Thanks for keeping connector and destination config payloads out of the source. Product fit and scope are good; these fixes should make the exposed tables match Fivetran's documented API surface.

Akash-How:
Addressed the Fivetran review findings in `f85dd70`.

What changed:
- Moved transformation projects from the legacy `/dbt/projects` path to documented `/v1/transformation-projects`.
- Removed unsupported remote `group_id` filtering from destinations and transformation projects; README now tells users to filter returned `group_id` locally in SQL.
- Aligned transformation project columns to documented response fields and kept connector/destination config payloads out.
- Expanded declared live tests to cover destinations and transformation_projects, plus documented that `fivetran.users` requires `USER:READ` permission.
- Added README auth setup guidance and Fivetran 429 / `Retry-After` rate-limit notes.

Validation:
- YAML parsing: passed
- `git diff --check`: passed
- `make lint-sources`: passed
- Live Coral test: passed against the Fivetran account with 4 declared tests covering groups, connections, destinations, and transformation_projects. A scoped users query was also checked and correctly returned 403 without `USER:READ`, now documented.

jsummerfield (approved):
Reviewed the current head again. The previous blockers are resolved: transformation projects now use the documented `/v1/transformation-projects` endpoint, unsupported destination filters were removed, the page-size limits are capped appropriately, and the README now includes setup/rate-limit guidance plus live Coral evidence. The current checks are green and `coral source lint` passes.

I noticed one non-blocking wording nit in the README: “scoped system key” mixes Fivetran's scoped API keys and system keys, which are separate credential models. It would be clearer as “Use either a scoped API key or a system key with read access to the relevant resources.” That can be cleaned up without blocking this PR.

Approving this head. Thanks for the thorough follow-up.

PR #922, opened 2026-05-28
feat: Safer LIKE translation + IN-list pushdown

### LINKED ISSUE #919: feat: Safer LIKE translation + IN-list pushdown
### What problem does this solve?

filter pushdown is limited; common predicates (more LIKE variants, IN lists) can’t be expressed in remote query params, causing extra client-side filtering and missed performance

### Proposed change

add additional classify_filter rules that still preserve correctness via DataFusion residual filters, e.g. allow more LIKE shapes within FilterMode::Contains/Search and support IN (...) for manifests that declare multivalue-capable filters.

### Alternatives considered

_No response_

### Which area does this affect?

CLI ergonomics

### Contribution

- [x] I'd be willing to open a PR for this if the direction is agreed.

Aditya8369:
Implemented the safer pushdown change in the HTTP engine.

`LIKE` translation is now more conservative: it only extracts a literal search term when the pattern has a safe literal core, and the new tests cover repeated edge wildcards like `%%deploy%%`. `IN (...)` pushdown now accepts multi-item literal lists for `Contains`/`Search` and collapses them into a comma-separated value, while keeping the pushdown inexact so residual filtering still protects correctness.

Validation is clean with the editor static checks on provider.rs and filter_expr.rs. The new regression coverage lives at provider.rs and filter_expr.rs.

closes #919

### DIFF
--- a/crates/coral-engine/src/backends/http/request.rs
+++ b/crates/coral-engine/src/backends/http/request.rs
@@ -119,7 +119,7 @@ pub(super) fn build_outgoing_request_explain(
             .iter()
             .map(|(name, value)| json!({"name": name, "value": value}))
             .collect::<Vec<_>>(),
-        "body": body.map(request_body_to_json),
+        "body": body.map(request_body_summary),
         "limit": limit,
         "projection": projection,
     }))
@@ -154,10 +154,17 @@ fn http_method_label(method: HttpMethod) -> &'static str {
     }
 }
 
-fn request_body_to_json(body: &RequestBody) -> Value {
+fn request_body_summary(body: &RequestBody) -> Value {
     match body {
-        RequestBody::Json(value) => value.clone(),
-        RequestBody::Text(text) => Value::String(text.clone()),
+        RequestBody::Json(value) => json!({
+            "kind": "json",
+            "empty": value.as_object().is_some_and(|object| object.is_empty())
+                || value.as_array().is_some_and(|array| array.is_empty()),
+        }),
+        RequestBody::Text(text) => json!({
+            "kind": "text",
+            "empty": text.is_empty(),
+        }),
     }
 }
 pub(super) fn build_query_pairs(
@@ -456,7 +463,9 @@ mod tests {
 
         assert_ne!(first.resolved_request(), second.resolved_request());
         assert_ne!(first.fingerprint(), second.fingerprint());
-        assert!(first.resolved_request().contains("first page"));
-        assert!(second.resolved_request().contains("second page"));
+        assert!(first.resolved_request().contains("\"kind\":\"text\""));
+        assert!(second.resolved_request().contains("\"kind\":\"text\""));
+        assert!(!first.resolved_request().contains("first page"));
+        assert!(!second.resolved_request().contains("second page"));
     }
 }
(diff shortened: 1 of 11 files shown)


crates/coral-engine/src/backends/shared/filter_expr.rs
```diff
@@ -74,19 +74,36 @@ fn collect_filter_values(
...
+                return;
+            }
+
+            let values = in_list
+                .list
+                .iter()
+                .map(literal_to_string)
+                .collect::<Option<Vec<_>>>();
+            if let Some(values) = values {
+                filters.insert(col_name, values.join(","));
```
antonmry, line 106: This needs an explicit source-spec capability before we push down multi-item `IN` this way. Joining values as `a,b` is only safe for providers that document comma-separated OR/list semantics for that specific parameter. Marking the pushdown `Inexact` only lets DataFusion remove false positives after the response; it cannot recover false negatives if the provider treats `a,b` as a literal search string or provider-specific syntax.

crates/coral-engine/src/backends/http/request.rs
```diff
@@ -1,17 +1,172 @@
...
+        render_context,
+    )?;
+    let url = join_url(&base_url, &rendered_path)?;
+    let query_pairs = build_query_pairs(request, render_context)?;
+    let logged_url = build_logged_url(&url, &query_pairs);
+    let projection = projection.map(|indices| indices.to_vec());
+    let resolved_request = serde_json::to_string(&json!({
+        "request": request,
+        "call": {
+            "url": logged_url,
```
antonmry, line 80: This exposes raw query values through `resolved_request`/`EXPLAIN` and the new trace field. Existing `url.full` tracing goes through `sanitize_trace_url`, which strips query strings; this payload bypasses that boundary and can leak user search terms, emails, IDs, or query-param credentials. We should redact query values here, or only expose the fingerprint by default.

antonmry:
Thanks for working on this. For optimization PRs like this, I think we should include stronger evidence in the PR description before merging.

Please add concrete before/after evidence against a few existing sources, ideally core sources where this pushdown path already matters. For each example, it would help to show:

- the SQL predicate
- the previous pushed-down filter/request shape
- the new pushed-down filter/request shape
- the `EXPLAIN` output
- the relevant tracing span fields emitted for the HTTP request
- whether DataFusion still applies a residual filter

Some feedback specific to this PR:

The safer `LIKE` handling looks directionally right. Only extracting a literal core from patterns such as `%%deploy%%`, while rejecting inner `%` or `_` wildcards, makes sense because the provider receives a simple search term and DataFusion can still enforce exact SQL `LIKE` semantics afterward.

The multi-item `IN (...)` change is less obviously safe. Joining values into `a,b` only preserves semantics for provider parameters that explicitly treat comma-separated values as an OR/list. For existing `mode: search` filters, several sources use provider-specific query languages, for example Shopify/Auth0-style `query`/`q` filters. In those cases `WHERE q IN ('a', 'b')` becoming `q=a,b` may change meaning or become narrower, and residual filtering cannot recover rows the provider did not return. I think this needs either explicit source-spec capability metadata or source-by-source evidence showing the parameter supports this shape.

The request explain/fingerprint work could be useful, but I’d like to see the actual spans it produces. Right now the resolved request payload appears to include raw query values, while existing URL tracing strips query strings. Please include span examples and make sure we are not exposing user search text, emails, IDs, or query-param credentials through `EXPLAIN` or tracing.

antonmry (changes requested):

PR #930, opened 2026-05-28
fix(mcp): serialize int64/uint64 columns as JSON strings

jamesaud:
## Summary

- MCP JSON results emitted `Int64`/`UInt64` columns as JSON numbers, so any client that parses JSON through IEEE-754 doubles (JS/TS `JSON.parse`, most MCP hosts) silently truncated bigint values past 2^53. Real example: `talkingdata` user_id `-8504475857937456387` was arriving as `-8504475857937456000` — three digits gone, no error raised.
- `batches_to_json_rows` (the MCP-only path) now post-processes `Int64`/`UInt64` columns into JSON strings so the exact value survives the boundary; the model can still reason over the value and the client never reparses it as a float.
- CLI JSON output (`format_batches_json`) is intentionally **not** changed — shell consumers (`jq`, scripts) of the existing CLI output keep their current shape.

### Before
```json
{ "user_id": -8504475857937456000 }
```

### After
```json
{ "user_id": "-8504475857937456387" }
```

## Affected surface
- `crates/coral-client/src/lib.rs::batches_to_json_rows` — only caller is `coral-mcp::server::query_rows`.
- CLI `OutputFormat::Json` path still calls `format_batches_json` directly and is unaffected.

## Test plan
- [x] New unit test `batches_to_json_rows_stringifies_int64_values` exercises the screenshot's value `-8504475857937456387` and a near-`u64::MAX` value, plus a non-int64 column to confirm it's not touched.
- [x] `cargo test -p coral-client --lib` — all 21 tests pass.
- [x] `cargo clippy -p coral-client --lib --tests --no-deps -- -D warnings` — clean.
- [x] `cargo build -p coral-mcp` — compiles.
- [ ] Smoke against a live MCP client with a `talkingdata`-shaped query to confirm exact value round-trips end-to-end.

image


jsummerfield:
Did you intend for this to be top-level only, or also support nested fields?

jamesaud:
Updated to support nested, thanks. A few discussion points before merging:

\- It's may surprise the user that in64 fields are returned as strings; columns will say "BIGINT / INT64" and API will return string.  
\- The fix is only for MCP, meaning other surfaces (gRPC/CLI) would have the original behavior since they are lossless.  
\- MCP wire format is JSON-RPC; likewise client side javascript (node) calling `JSON.parse`on int64 outputs would truncate, while parsers like `jq` handle it fine.

For now I'd suggest we at least document the behavior and call out that some parsers don't handle int64s.

mystic123 (approved):

jsummerfield:
I think there is a cleaner implementation path here.

The current approach fixes the top-level case, but the recursive `serde_json::Value` post-processing is taking on responsibility for mirroring Arrow JSON's shape rules after serialization. That is easy to get subtly wrong: for example, Arrow writes `Map` values as JSON objects, but the current walker handles `DataType::Map` in the list-like branch and only descends into `Value::Array`, so `MAP<STRING, BIGINT>` would still emit lossy JSON numbers.

A better shape would be:

1. Keep the MCP call site thin: `coral-mcp` should just choose the MCP-safe row renderer.
2. Keep the implementation in `coral-client`, since it already owns Arrow result decoding/rendering.
3. Implement the MCP-safe renderer with Arrow JSON's `WriterBuilder::with_encoder_factory` / `EncoderFactory` hook, overriding the leaf encoders for precision-sensitive numeric types instead of re-walking parsed JSON.

That lets Arrow continue owning traversal for `Struct`, `List`, `Map`, `Dictionary`, etc., and Coral only defines how specific scalar leaves are encoded. It also gives us a sane path for decimals: if `Decimal*` exactness matters for MCP/JS clients, stringification needs to happen during JSON writing, not after `serde_json::from_str` has parsed the number.

I would also name the helper for the contract rather than the current implementation, e.g. `batches_to_json_rows_json_safe_numbers` or `batches_to_json_rows_mcp_safe`, and document the intentional mismatch:

> MCP `sql` results encode precision-sensitive numeric SQL types such as `Int64`/`UInt64` as JSON strings so JS/MCP clients preserve exact values. CLI JSON keeps its existing numeric shape.

That does mean columns may say `Int64` while the MCP JSON value is a string, but that is the least bad contract: `Int64` is the SQL type; string is the JSON encoding chosen to avoid silent corruption. Returning a JSON number for values beyond 2^53 is less surprising but wrong.

I would add two tests around this contract:

- a writer-backed `coral-client` test for map/dictionary/list nested bigint shapes, so we are checking Arrow's actual JSON output shape rather than hand-built `serde_json!` fixtures;
- an MCP `sql` boundary test asserting a large `Int64` result comes back as a string in `structured_content`.

This keeps the fix scoped, preserves the CLI JSON contract, and avoids Coral maintaining a second partial copy of Arrow's JSON traversal behavior.

jamesaud:
Updated per your spec, also implemented for: `Int64`, `UInt64`, and `Decimal{32,64,128,256}`

sauldhernandez (approved):
LGTM, but make sure you also update the docs to reflect the contract change on the MCP

jamesaud:
updated docs

docs/guides/use-coral-over-mcp.mdx
```diff
@@ -58,6 +58,19 @@ For deeper introspection, query `coral.tables`, `coral.columns`, `coral.filters`
...
+
+- `Int64` (`BIGINT`) and `UInt64`
+- `Decimal32`, `Decimal64`, `Decimal128`, and `Decimal256`
+
+This applies wherever these types appear, including nested inside `Struct`, `List`, and `Map` columns. The column's declared SQL type does not change — `describe_table`, `list_columns`, and `coral.columns` still report `Int64` or `Decimal(p, s)`; only the JSON encoding of the value is a string. A `BIGINT` `user_id`, for example, comes back as `{ "user_id": "-8504475857937456387" }`.
+
+Coral does this because MCP uses JSON-RPC, and many clients — JavaScript/TypeScript `JSON.parse` and most MCP hosts — decode JSON numbers as IEEE-754 doubles. Values beyond 2^53 silently lose precision when read as a number (`-8504475857937456387` becomes `-8504475857937456000`), with no error raised. Encoding them as strings keeps the exact value intact end to end. Parsers that preserve integer precision, such as `jq`, handle bare numbers fine, but Coral cannot assume the client does.
+
+The CLI is unaffected: `coral sql --format json` keeps emitting these types as JSON numbers, so existing shell pipelines over CLI output are unchanged.
```
AlbertQM: Thank you! I think we can cut some text here (I've been trying to keep our docs on a strict diet :P)

```suggestion
Coral does this because MCP uses JSON-RPC, and many clients — JavaScript/TypeScript `JSON.parse` and most MCP hosts — decode JSON numbers as IEEE-754 doubles. Values beyond 2^53 silently lose precision when read as a number.
The CLI is unaffected: `coral sql --format json` keeps emitting these types as JSON numbers.
```

AlbertQM (approved):

docs/guides/use-coral-over-mcp.mdx
jamesaud: trimmed it!

PR #932, opened 2026-05-28
feat(spec): support OAuth endpoint input templates

sauldhernandez:
## Intent

Allow OAuth provider endpoint URLs to use install-time source variable inputs. This fixes sources such as Microsoft Outlook / Graph API, where the tenant is part of the OAuth endpoint path and must be configurable per installation.

## What It Enables

- `endpoints.authorization_url`, `endpoints.token_url`, and `endpoints.device_authorization_url` can reference declared `kind: variable` inputs with `{{input.KEY}}`.
- OAuth preflight resolves source variable bindings before validating and starting OAuth credential retrieval.
- OAuth sessions use rendered endpoint URLs for authorization, device-code requests, token exchange, and persisted internal OAuth metadata.
- Secret inputs, filters, function args, state tokens, and inline defaults are rejected for OAuth endpoint templates.
- Source-spec docs and Coral source-author/review skills now describe the supported OAuth endpoint template surface.

## Usage Examples

```yaml
inputs:
  OUTLOOK_TENANT_ID:
    kind: variable
    default: organizations
  OUTLOOK_TOKEN:
    kind: secret
    credential:
      methods:
        - type: oauth
          oauth:
            flow:
              type: authorization_code
              pkce: required
            redirect_uri: http://127.0.0.1:53682/oauth/callback
            endpoints:
              authorization_url: https://login.microsoftonline.com/{{input.OUTLOOK_TENANT_ID}}/oauth2/v2.0/authorize
              token_url: https://login.microsoftonline.com/{{input.OUTLOOK_TENANT_ID}}/oauth2/v2.0/token
            client:
              id:
                input: OUTLOOK_CLIENT_ID
```

## Validation

- `cargo fmt --all`
- `cargo test -p coral-spec endpoint_urls -- --nocapture`
- `cargo test -p coral-app device_code_oauth_session_renders_endpoint_templates_from_source_inputs -- --nocapture`
- `make rust-checks`
- `git diff --check`


AlbertQM (approved):

crates/coral-spec/src/inputs.rs
```diff
@@ -903,15 +1053,28 @@ fn redirect_uri_has_explicit_port(raw: &str) -> bool {
     !port.is_empty() && port.bytes().all(|byte| byte.is_ascii_digit())
 }
 
-fn validate_url(input_key: &str, field: &str, raw: &str) -> Result<()> {
-    Url::parse(raw).map_err(|error| {
+fn validate_oauth_endpoint_url(input_key: &str, field: &str, raw_template: &str) -> Result<()> {
+    let template = ParsedTemplate::parse(raw_template)?;
+    let rendered = render_template_with_placeholders(&template);
```
jsummerfield: This placeholder parse makes lint/spec validation disagree with runtime rendering. A defaulted variable like `HOST: "foo bar.com"` can pass here and then fail when `endpoint_urls` inserts the real value, while valid placements such as `https://example.com:{{input.PORT}}/token` are rejected because `placeholder` is not a valid port. The spec side should validate with declared defaults when available, and defer final URL parsing for templates that depend on required runtime variables.

jsummerfield (approved):

iaptsiauri (approved):

PR #933, opened 2026-05-28
feat(sources/community/opensearch): add opensearch community source

BAVYASAKTHIVEL-21:
## Summary

Adds a new community Coral source for OpenSearch that queries cluster topology, index storage telemetry, and snapshot metadata through the OpenSearch CAT and Snapshot REST APIs as read-only SQL tables.

## Included tables

- `nodes` — `GET /_cat/nodes`
- `indices` — `GET /_cat/indices[/{index}]`
- `snapshots` — `GET /_snapshot/{repository}/_all`

## Features

- Cluster node telemetry (roles, manager status, heap/RAM/CPU)
- Index inventory with health and storage sizing
- Snapshot lifecycle metadata per repository
- HTTP Basic authentication
- Pushdown filters on `indices` (`index` path, `health` query)

## Pushdown filters

- `indices.health` → CAT `health=` query parameter (server-side)
- `indices.index` → CAT request path `/_cat/indices/{index}` (server-side)
- `snapshots.repository` → required; used in the request path

## Addressing review feedback

This revision resolves both blockers:

- **Snapshots (High):** the table now requires a `repository` filter and queries `GET /_snapshot/{repository}/_all` (a documented, repository-scoped path), instead of relying on `_all/_all`. OpenSearch snapshot objects have no per-row repository field, so the `repository` column is derived from the required filter (`from_filter`) and is therefore always accurate, rather than reading a null/ambiguous field from each row.
- **CAT indices (Medium):** added real pushdown filters — `index` (request path) and `health` (query parameter) — so the cluster filters results before they reach Coral. The README now also states plainly that, absent those filters, the source fetches the full CAT listing and other `WHERE` predicates are applied locally (no longer implying SQL filtering reduces the upstream response).

## Validation

```bash
make lint-sources
coral source lint sources/community/opensearch/manifest.yaml
export OPENSEARCH_URL=https://localhost:9200
export OPENSEARCH_USERNAME=admin
export OPENSEARCH_PASSWORD=...
coral source add --file sources/community/opensearch/manifest.yaml
coral source test opensearch
coral sql "SELECT index_name, health FROM opensearch.indices WHERE health = 'green' LIMIT 5"
```

Live `coral source test opensearch` output against this exact head is included in the README.


jsummerfield (changes requested):
Findings:

- High: `sources/community/opensearch/manifest.yaml:130` uses `GET /_snapshot/_all/_all`, and `sources/community/opensearch/manifest.yaml:142` exposes `repository` from each returned snapshot row, but the documented Get Snapshot API is scoped as `GET _snapshot/{repository}/{snapshot}/` and the response fields do not include a `repository` field in each snapshot object. As written, the table either depends on undocumented `_all/_all` behavior or returns a null/ambiguous `repository` column. Please either require a repository filter/path and remove the per-row `repository` column, or use a documented endpoint/shape that can reliably include repository context. Docs: https://docs.opensearch.org/latest/api-reference/snapshots/get-snapshot/

- Medium: `sources/community/opensearch/manifest.yaml:65` maps `master_status` from the CAT nodes `master` column, but current OpenSearch CAT nodes output uses `cluster_manager` terminology/columns. Because the request does not pin the returned headers with an `h=` list, this is likely null or version-dependent on current OpenSearch clusters. Please request the documented column explicitly and map that field, or rename/document the column as version-dependent. Docs: https://docs.opensearch.org/latest/api-reference/cat/cat-nodes/

- Medium: `sources/community/opensearch/README.md:198` says large index inventories may use targeted SQL filtering to reduce response sizes, but `sources/community/opensearch/manifest.yaml:86` fetches unpaginated `/_cat/indices` with no request filters. SQL `WHERE` applies after Coral fetches the CAT response, so it does not reduce the upstream response size. Please either add pushdown filters such as index path/health where useful, or revise the README to say the source fetches the full CAT listing. Docs: https://docs.opensearch.org/latest/api-reference/cat/cat-indices/

The manifest lints and the source is useful product surface, but the snapshot table and current CAT semantics need to match the provider contract before this is safe to approve. Thanks for adding a practical operational source; these fixes should make the first user experience much more reliable.

jsummerfield (changes requested):
Thanks for the update. The previous source-contract issues I raised look addressed on this head: snapshots are now repository-scoped, `indices` has real `index`/`health` pushdown, and `nodes` pins/maps the `cluster_manager` CAT column. The manifest also lints cleanly.

One blocker remains: `sources/community/opensearch/README.md` still ships placeholder live-test evidence. The section says “Replace the block below with the actual output” and “Do not ship placeholder output,” but the placeholder block is still present. Please replace that with sanitized output from `coral source test opensearch` and at least one representative query against the current manifest, or remove the claim that live output is included.

After that, this survivor candidate looks like the right OpenSearch PR to keep over the duplicate.

PR #934, opened 2026-05-28
feat: implement DSL v4 materialization and runtime

simonwhitaker:
## Summary

This branch adds the first DSL v4 implementation path for OpenAPI-backed sources.

`coral-spec` now recognizes `dsl_version: 4` manifests with a separate generated JSON schema, parses v4 surfaces, imports OpenAPI descriptors into semantic IR, and derives read projections from that IR. The existing v3 manifest schema remains separate; v4 validation uses `source_manifest_v4.schema.json`.

`coral-app` owns the v4 lifecycle. During `source add --file`, Coral reads or fetches each OpenAPI descriptor, writes a materialized package under the installed source state, and persists only the durable authored manifest. Query execution and source validation load the installed materialized package; they do not re-fetch descriptors or regenerate projections. Missing, corrupted, or incompatible artifacts fail closed with guidance to re-add the source.

`coral-engine` now compiles backend-ready runtime components instead of needing to understand DSL v4 authored manifests or materialized artifacts. Existing v3 manifests become single-component runtime packages, while v4 sources are assembled by `coral-app` from materialized projections and registered through a generic composite backend.

## Reviewer Notes

- Preview DSL v4 specs live under `sources/core-v4/**` and are not bundled into the binary. This branch adds `sources/core-v4/github_v4/manifest.yaml` as an explicitly installed preview source.
- DSL v4 manifests are versionless. Protobuf responses still carry a string version field, so versionless sources serialize as an empty string; the CLI renders that as `-` in `source list` and omits `Version:` in `source info`.
- Local OpenAPI `file:` descriptors are resolved relative to the manifest file by the CLI before import, canonicalized, required to be regular non-symlink files, and then stored as absolute paths for app-side materialization.
- The materialized v4 package includes fingerprints, diagnostics, projection catalogs, raw and normalized descriptors, and semantic IR. Fingerprints cover the manifest, descriptor contents, input declarations, importer version, and projection-generator version.
- Schema generation is now repo automation: `make schema-generate` writes generated schemas, `make schema-check` verifies freshness, and the Validate workflow runs a `schema-freshness` job when schema inputs change.
- `AGENTS.md`, crate-local agent guidance, and CLI docs were updated for the app-owned materialization boundary, preview-source location, versionless source display, descriptor path policy, and schema freshness rules.

## Main Review Areas

- `crates/coral-spec/src/v4/*`: v4 manifest model, schema generation, OpenAPI import, semantic IR, projection generation, and materialized artifact validation.
- `crates/coral-app/src/sources/materialization.rs` and `crates/coral-app/src/sources/manager.rs`: install-time materialization, rollback behavior, durable manifest handling, and compatibility failures.
- `crates/coral-app/src/sources/runtime_package.rs` and `crates/coral-app/src/query/manager.rs`: assembly of v4 materialized artifacts into engine runtime components.
- `crates/coral-engine/src/contracts/query.rs` and `crates/coral-engine/src/backends/composite.rs`: runtime package/component contract and multi-component source registration.
- `crates/coral-cli/src/source_ops.rs`: versionless source rendering and manifest-relative descriptor resolution.
- `.github/workflows/validate.yml`, `Makefile`, and `xtask/src/schemas.rs`: generated-schema freshness workflow.

## Validation

- `make rust-checks`
- `make schema-check`
- `make docs-check`
- `make perf-check` (`coral.tables` mean: 186.2 ms, threshold: 750 ms)
- `cargo test --locked -p coral-app v4 --lib`
- Targeted tests cover v4 schema validation, OpenAPI import/projection naming, materialization compatibility failures, query-from-installed-artifacts behavior, generic engine multi-component packages, versionless CLI output, and manifest-relative descriptor resolution.

## Local Smoke Test

Install the preview GitHub DSL v4 source explicitly from `sources/core-v4`:

```shell
GITHUB_TOKEN=<github-token> cargo run --bin coral -- source add --file sources/core-v4/github_v4/manifest.yaml
```

Run the preview queries:

```shell
cargo run --bin coral -- sql "SELECT id, number, title, state FROM github_v4.repos_issues WHERE owner = 'octocat' AND repo = 'Hello-World' LIMIT 5"

cargo run --bin coral -- sql "SELECT id, number, title FROM github_v4.get_issue(owner => 'octocat', repo => 'Hello-World', issue_number => 1)"

cargo run --bin coral -- sql "SELECT id, number, title FROM github_v4.search_issues(q => 'repo:octocat/Hello-World is:issue') LIMIT 5"
```

After changing the authored v4 manifest, OpenAPI importer, projection generator, or descriptor, re-run `source add --file` so Coral regenerates the materialized package.


sources/v4-dev/statusgator/manifest.yaml
```diff
@@ -0,0 +1,25 @@
+name: statusgator_v4_dev
+version: 0.1.0
+dsl_version: 4
+description: Statusgator
+surfaces:
+  - id: rest
+    type: openapi
+    file: /Users/simon/.codex/worktrees/f3c1/coral/sources/v4-dev/statusgator/openapi.yaml
```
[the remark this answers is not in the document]
simonwhitaker: This is just there for local testing, I'll remove these completely soon.

sources/v4-dev/github/manifest.yaml
```diff
@@ -0,0 +1,39 @@
+name: github_v4_dev
+version: 0.1.0
+dsl_version: 4
+description: >-
+  Development DSL v4 source for exercising generated projections against a
+  local GitHub OpenAPI slice and the live GitHub REST API endpoint.
+surfaces:
+  - id: rest
+    type: openapi
+    file: /Users/simon/.codex/worktrees/f3c1/coral/sources/v4-dev/github/openapi.yaml
```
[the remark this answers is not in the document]
simonwhitaker: Same here, will remove this soon, it's just here temporarily for testing.

sources/core-v4/github_v4/manifest.yaml
```diff
@@ -0,0 +1,39 @@
...
+    request_headers:
+      - name: Accept
+        from: literal
+        value: application/vnd.github+json
+      - name: X-GitHub-Api-Version
+        from: literal
+        value: "2022-11-28"
+      - name: User-Agent
+        from: literal
+        value: coral-dsl-v4
```
jsummerfield, line 26: Thinking through what's necessary here, vs what can be derived/defaulted. I think this is the minimal GitHub manifest:  


```yaml
surfaces:
  - id: rest
    type: openapi
    url: https://raw.githubusercontent.com/github/rest-api-description/<commit>/descriptions/api.github.com/api.github.com.2022-11-28.yaml
    sha256: <sha256>
    inputs:
      GITHUB_TOKEN:
        kind: secret
    auth:
      type: HeaderAuth
      headers:
        - name: Authorization
          from: template
          template: Bearer {{input.GITHUB_TOKEN}}
    request_headers:
      - name: X-GitHub-Api-Version
        from: literal
        value: "2022-11-28"
```

  
I think description, base_url, Accept can all be taken from the openapi spec.  User-Agent can be defaulted.

jsummerfield: (LLMs love to add, so it's worth fighting back at this early stage to keep it as small as possible)

sources/core-v4/github_v4/manifest.yaml
```diff
@@ -0,0 +1,26 @@
+name: github_v4
+version: 0.1.0
+dsl_version: 4
+surfaces:
+  - id: rest
+    type: openapi
+    url: https://raw.githubusercontent.com/github/rest-api-description/main/descriptions/api.github.com/api.github.com.yaml
+    sha256: c2c6a2c838581cf3915d7d61c0c2bb48b128faf748bebf77a3ac2f0742975c96
```
jsummerfield: Thinking again - I'm not sure we want to pin a manifest to a SHA. This makes versioning and upgrades a Coral concern, but I think it should be the user's. So the SHA would be captured at `source add` time, not manifest authoring time.

jsummerfield: This implies removing `version` above too.

plans/2026-05-24-dsl-v4-prd.md
jsummerfield, line 1: As we get this ready to merge, can you revert his file pls

sources/core-v4/github_v4/manifest.yaml
```diff
@@ -0,0 +1,26 @@
...
+          source.
+    request_headers:
+      - name: X-GitHub-Api-Version
+        from: literal
+        value: "2022-11-28"
+      - name: User-Agent
+        from: literal
+        value: coral-dsl-v4
+test_queries:
+  - SELECT id, number, title, state FROM github_v4.issues__5708b516 WHERE owner = 'octocat' AND repo = 'Hello-World' LIMIT 5
```
jsummerfield: Are these table names correct?

sources/core-v4/github_v4/manifest.yaml
```diff
@@ -0,0 +1,26 @@
+name: github_v4
+version: 0.1.0
+dsl_version: 4
+surfaces:
+  - id: rest
+    type: openapi
+    url: https://raw.githubusercontent.com/github/rest-api-description/main/descriptions/api.github.com/api.github.com.yaml
+    sha256: c2c6a2c838581cf3915d7d61c0c2bb48b128faf748bebf77a3ac2f0742975c96
+    inputs:
```
jsummerfield, line 7: I think this needs to be fixed before this source is used as the v4 smoke path: the manifest requires `GITHUB_TOKEN`, but the surface does not wire that input into `auth`, so runtime requests are still unauthenticated. Either add explicit bearer `HeaderAuth` here and cover it with a fake-server assertion on the `Authorization` header, or make this preview source intentionally anonymous and remove the required token/hint.

jsummerfield (commented):
One focused comment on the GitHub v4 auth contract.

crates/coral-spec/src/v4/mod.rs
```diff
@@ -0,0 +1,2966 @@
...
+    dsl_version: u32,
+    name: String,
+    version: String,
+    #[serde(default)]
+    description: String,
+    #[serde(default)]
+    test_queries: Vec<String>,
+    surfaces: Vec<RawV4Surface>,
+    #[serde(default)]
+    projection_policy: RawProjectionPolicy,
```
jsummerfield: Do we need to expose `projection_policy` in the first v4 manifest shape? Right now the only accepted value is also the implicit default (`derive_read_operations`), so this does not give authors a real choice yet. I'd drop the field for now and add it back when there are at least two real policies or explicit projection overrides for names/visibility/table-vs-function decisions.

jsummerfield (commented):
One YAGNI note on the v4 manifest surface.

crates/coral-api/proto/coral/v1/sources.proto
```diff
@@ -509,6 +565,8 @@ service SourceService {
   rpc ImportSource(ImportSourceRequest) returns (stream ImportSourceResponse);
   // Deletes a registered source.
   rpc DeleteSource(DeleteSourceRequest) returns (DeleteSourceResponse);
+  // Rebuilds materialized artifacts for an installed DSL v4 source.
+  rpc RefreshSource(RefreshSourceRequest) returns (RefreshSourceResponse);
```
jsummerfield: In the spirit of YAGNI and keeping this PR as small as poss, I'd consider dropping this for now until we have more mileage under our belts. `source add` will do the job for now.

crates/coral-app/build.rs
```diff
@@ -28,30 +54,116 @@ fn main() {
...
+        writeln!(
+            generated,
+            "    BundledSourceEntry {{ name: {:?}, manifest_yaml: {:?}, descriptors: &[",
+            entry.name, entry.manifest_yaml
+        )
+        .expect("writing to String is infallible");
+        for descriptor in &entry.descriptors {
+            writeln!(
+                generated,
+                "        BundledV4Descriptor {{ surface_id: {:?}, bytes: include_bytes!({:?}) }},",
```
jsummerfield: While DSL v4 is still experimental, I don't think we should embed bundled OpenAPI descriptor bytes into the binary. Let's keep preview sources URL-based or user-imported for now, and remove the added descriptor-scanning/`BundledV4Descriptor`/`include_bytes!` support until we have a stable bundled-source packaging contract that actually needs it.

jsummerfield (commented):
One more YAGNI note on bundled v4 descriptor embedding.

crates/coral-spec/src/v4/mod.rs
```diff
@@ -0,0 +1,2966 @@
...
+                message: format!("projection name collision resolved as '{name}'"),
+                surface_id: Some(projection.surface_id.clone()),
+                operation_id: Some(projection.operation_id.clone()),
+                projection_name: Some(name),
+            });
+        }
+    }
+}
+
+fn projection_name_priority(projection: &Projection, operation: &IrOperation) -> i32 {
```
jsummerfield: Can we remove these provider-shaped ranking rules from the generic v4 projection generator? The `repo` / `org` / `enterprise` / `authenticated_user` cases make this layer encode GitHub-style API semantics instead of source-neutral projection rules. If these names matter for the preview, they should come from source-specific projection metadata/overlays or a narrower generated decision, not generic `coral-spec` logic.

crates/coral-spec/src/v4/mod.rs
```diff
@@ -0,0 +1,2966 @@
...
+    }
+
+    if has_required(inputs, "owner") && has_required(inputs, "repo") {
+        sentences.push(
+            "Keep queries repository-scoped; fan out across repos client-side when you need broader coverage."
+                .to_string(),
+        );
+    }
+
+    if is_search {
```
jsummerfield, line 1796: Can we remove the GitHub-specific wording here? This guide is generated for every v4 OpenAPI source, so mentioning GitHub in the generic projection generator leaks provider assumptions into source-neutral output. A neutral sentence like “search endpoints are often rate-limited” is fine if we still want this guidance.

jsummerfield (commented):
Left two architecture comments on the generic v4 projection generator heuristics.

jsummerfield:
I think the core direction here is right: v4 should turn a source recipe into a materialized package that the runtime can execute. The ownership boundary just needs to be sharper before this becomes the base for future source generation work.

A useful analogy: the user-authored source manifest should be closer to `Cargo.toml`, not `target/`. It declares intent: source name, surfaces, descriptor locations/hashes, auth inputs, and policy knobs. The materialized v4 package is the build output: the resolved provider snapshot, semantic/capability IR, projection catalog, diagnostics, fingerprints, and runtime-ready table/function specs. Runtime should execute that package rather than partially rebuilding or reinterpreting the recipe.

Ideal end state:

```text
manifest.yaml + provider descriptor + overlays
        |
        |  coral-app owns source lifecycle/materialization
        v
resolved v4 package in Coral workspace state
        |
        |  coral-engine owns execution only
        v
DataFusion tables/table functions + HTTP execution
```

Concretely, `coral-app` should own descriptor fetch/read/copy, descriptor hash recording, manifest hash recording, base URL derivation, OpenAPI import, overlay application, projection generation, artifact activation, refresh, and stale detection. That is source lifecycle and workspace state.

`coral-engine` should consume a complete provider-neutral runtime package: published tables, table functions, request specs, response specs, auth, headers, pagination, columns, diagnostics metadata, and whatever source/projection identity is needed for tracing. It should not read raw OpenAPI files, derive `servers[0].url`, or apply descriptor/materialization fallback policy at compile time.

Agent-ready course correction:

```text
Refactor v4 so coral-app builds a complete executable package and coral-engine only executes that package.

- Move all descriptor/OpenAPI-derived runtime metadata resolution into app materialization.
- Include resolved base_url/auth/headers/request specs/response specs/pagination in the materialized package or a package-derived runtime model.
- Stop coral-engine from reading raw OpenAPI artifacts or deriving servers[0].url.
- Keep user-authored YAML as source intent, not generated state.
- Treat the materialized package as the reproducible runtime boundary: hash inputs, record generator/importer versions, store diagnostics, and make refresh atomically replace that package.
```

This keeps the prototype extensible: future GraphQL/MCP/provider imports, projection overlays, and workbench edits all compile into the same package boundary instead of leaking provider-specific logic into engine runtime.

simonwhitaker:
* **#1107** <a href="https://app.graphite.com/github/pr/withcoral/coral/1107?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#934** <a href="https://app.graphite.com/github/pr/withcoral/coral/934?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/934?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jsummerfield (commented):
One architecture note on the v4 query-source shape.

crates/coral-engine/src/contracts/query.rs
```diff
@@ -15,6 +16,7 @@ use crate::EngineExtensions;
 #[derive(Debug, Clone)]
 pub struct QuerySource {
     source_spec: ValidatedSourceManifest,
+    v4_materialization: Option<V4MaterializedSource>,
```
jsummerfield: I would avoid modeling this as `ValidatedSourceManifest + Option<V4MaterializedSource>`. That makes an invalid state representable: a v4 manifest with no materialization, or a legacy manifest carrying irrelevant v4 artifacts. The engine then has to rediscover that hidden invariant at compile time and fail late.

A cleaner shape is an enum-style source selection where the valid cases are explicit, for example `Manifest { source_spec, variables, secrets }` for legacy/direct manifests and `RuntimePackage { package, variables, secrets }` for generated/materialized sources. Then a generated source cannot enter the engine without the package required to execute it.

This also lines up with the package boundary discussed elsewhere: `coral-app` builds a complete runtime package, and `coral-engine` executes packages. Longer term, v3 can compile into the same package shape too, so the engine does not need bespoke optional sidecars for each generated source generation path.

docs/reference/source-spec-reference.mdx
jsummerfield, line 1: I would remove all mention from the docs/ dir for the time being. This needs more time to bake internally before inviting users to try it.

jsummerfield:
Clarifying one point from my earlier comment: I do not think this implies a new `coral-engine::backends::v4` backend, and probably not a new package backend either.

DSL v4 should be a source-package/materialization compiler owned by `coral-app`. By the time execution reaches `coral-engine`, the package should already have been lowered into engine-native runtime specs for the existing backends: HTTP, file, MCP, etc.

```text
v4 source recipe
  -> coral-app materializes/resolves/locks package
  -> coral-app loads package and produces existing backend specs
  -> coral-engine compiles those specs through backends::http, backends::file, backends::mcp
```

If we eventually need one source package to expose multiple backend targets under one schema, that should be a generic composite-source concern, not a v4 backend.

One related wording tweak: I would separate descriptor refs from descriptor hashes. The authored manifest should contain descriptor refs/locations as user intent; descriptor hashes should live in Coral's resolved lock/materialized package rather than being mandatory authored manifest fields.

jsummerfield (commented):
One schema-boundary comment.

crates/coral-spec/src/schema/source_manifest.schema.json
```diff
@@ -4,12 +4,54 @@
   "title": "Coral Source Manifest",
   "type": "object",
   "additionalProperties": false,
-  "required": ["dsl_version", "name", "version", "backend"],
+  "required": ["dsl_version", "name", "version"],
   "allOf": [
```
jsummerfield, line 8: I think we should avoid pulling v4 into the existing `source_manifest.schema.json` path if we can. The v3 manifest schema is already the stable executable-manifest contract for `backend: http/file/mcp`; making it also carry the experimental authored source-definition shape forces us to touch the v3 route and makes the new source model look more canonical than it is.

For v4, I would set the direction from the start as a separate authored source-definition schema, ideally generated from the Rust v4 definition types rather than hand-maintained here. Then parsing can dispatch on `dsl_version`: v3 continues through the existing schema/route, while v4 uses its own experimental schema/parser. That also lines up with moving descriptor hashes/versioning out of authored YAML and into Coral-owned materialization/lock state later.

jsummerfield:
Naming suggestion for the next pass: I would avoid letting `v4` become the long-term domain vocabulary. `dsl_version: 4` is a compatibility/version discriminator, not the name of the architecture.

The durable model I think we want is:

```text
SourceDefinition builds into SourcePackage; SourcePackage lowers into RuntimeSource.
```

Suggested before/after names:

```text
V4SourceManifest              -> SourceDefinition
V4Surface                     -> SourceSurface
SurfaceDescriptor             -> ProviderDescriptorRef
OpenApiRuntimeConfig          -> HttpRuntimeConfig
V4MaterializedSource          -> SourcePackage
MaterializedSurface           -> PackagedSurface
Fingerprint                   -> SourcePackageLock
FingerprintSurface            -> LockedSurface
MaterializationBuild          -> SourcePackageBuild
SourceMaterializationSummary  -> SourcePackageSummary
SemanticIr                    -> CapabilityModel
IrOperation                   -> Capability
IrOperationInput              -> CapabilityInput
IrExecutionAttachment         -> ProviderExecution
RestExecutionAttachment       -> HttpExecution
ProjectionPolicy              -> ProjectionConfig
ProjectionPolicyDefault       -> ProjectionDefaultMode
```

I would also avoid `coral-engine::backends::v4`. The source-definition compiler/materializer should be experimental, but the engine should keep compiling engine-native runtime specs through the existing execution backends (`http`, `file`, `mcp`). If one package eventually needs to expose multiple runtime targets under one schema, that is a generic `RuntimeSource`/composite-source concern, not a v4 backend.

For module naming, something like this reads better long term than `coral-spec/src/v4`:

```text
crates/coral-spec/src/source_definition/
  manifest.rs
  provider.rs
  capability.rs
  projection.rs
  overlay.rs
  package.rs
```

If we want to keep the experimental signal strong, put that under `experimental/source_definition` for now. That lets the experiment remain clearly quarantined without baking the migration number into every type we will later want to make canonical.

sources/core-v4/github_v4/manifest.yaml
simonwhitaker: No :(

I fixed these once already. Must have lost my fixes in a rebase. Fixing again...

crates/coral-spec/src/v4/mod.rs
simonwhitaker: ELL. OH. ELL.

That was from when I told Codex to improve the automated projection naming. No wonder it did such a great job! 🤦‍♂️ 

Fix incoming...

docs/reference/cli-reference.mdx
```diff
@@ -118,9 +118,21 @@ Add a custom source spec from a YAML file. Input values come from environment va
...
+be readable regular files and must not be symlinks.
+
 ### `coral source lint <FILE>`
 
-Validate a source spec YAML file without installing it. Checks YAML syntax, JSON schema conformance, and semantic rules such as duplicate columns or references to unknown filters.
+Validate a source spec YAML file without installing it. Checks YAML syntax, JSON
+schema conformance, and semantic rules such as duplicate columns or references
+to unknown filters. For DSL v4 manifests, it resolves local OpenAPI `file:`
+descriptors from the manifest directory using the same path policy as
+`source add --file`.
```
jsummerfield: Let's leave DSLv4 out of public docs for now

jsummerfield: In fact, just revert this file

crates/coral-spec/src/v4/mod.rs
```diff
@@ -0,0 +1,2837 @@
...
+    };
+    let kind = function_kind.map_or(ProjectionKind::Table, |function_kind| {
+        ProjectionKind::TableFunction { function_kind }
+    });
+    let sql_exposure = if matches!(kind, ProjectionKind::Table) {
+        SqlInputExposure::Filter
+    } else {
+        SqlInputExposure::FunctionArg
+    };
+    let inputs = operation
```
jsummerfield, line 1498: [P2] This exposes every path/query parameter as a SQL filter or function arg before accounting for pagination-owned inputs. When `detect_pagination` claims `page`/`per_page`, the HTTP pager later appends its own values, so catalog-visible inputs like `page` or `per_page` either conflict with Coral's pager or are silently overridden. Please omit pager-owned params from projection inputs/request query, or otherwise model them specially so the SQL contract matches runtime behavior.

crates/coral-app/src/sources/materialization.rs
```diff
@@ -0,0 +1,997 @@
...
+            AppError::Unavailable(format!(
+                "failed to fetch OpenAPI descriptor '{panic_url}': fetch thread panicked"
+            ))
+        })?
+}
+
+fn read_url_descriptor_on_blocking_thread(url: &str) -> Result<Vec<u8>, AppError> {
+    let client = reqwest::blocking::Client::builder()
+        .timeout(DESCRIPTOR_FETCH_TIMEOUT)
+        .redirect(reqwest::redirect::Policy::limited(5))
```
jsummerfield, line 607: [P2] The v4 parser requires descriptor URLs to start with `https://`, but this client follows redirects without enforcing that the final URL remains HTTPS. Reqwest's limited redirect policy allows HTTP downgrades unless `https_only(true)` is enabled or the final `response.url().scheme()` is checked, so a descriptor URL can redirect to unauthenticated content that drives generated runtime behavior. Please reject downgrade redirects before trusting the descriptor body.

crates/coral-app/src/query/manager.rs
```diff
@@ -175,6 +177,11 @@ impl QueryManager {
                 Err(error @ AppError::Credentials(CredentialsError::Unavailable(_))) => {
                     return Err(error);
                 }
+                Err(error @ AppError::FailedPrecondition(_))
+                    if error.to_string().contains("DSL v4 materialized artifacts") =>
```
jsummerfield: [P2] This is still using the rendered English error text as control flow for stale/incompatible v4 materialization. Since this surface is now fail-closed and user-facing, this should be a typed `AppError` variant, e.g. incompatible or missing source materialization, rather than `FailedPrecondition(String)` plus `contains(...)`; otherwise a wording change can silently change query-source loading behavior.

crates/coral-spec/src/v4/schema.rs
```diff
@@ -0,0 +1,241 @@
...
+#[derive(Debug, Serialize, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
+struct V4SourceManifestSchema {
+    dsl_version: u32,
+    name: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    description: Option<String>,
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    test_queries: Vec<String>,
+    surfaces: Vec<V4SurfaceSchema>,
```
jsummerfield, line 20: [P3] The parser rejects `surfaces: []`, but the generated schema currently leaves this as an unconstrained `Vec`, so external schema validation can pass a manifest that `coral source lint` or import then rejects. Please add the same non-empty constraint, for example `minItems: 1` in schema post-processing, and keep the generated schema fresh.

jsummerfield (commented):
Re-review inline findings.

crates/coral-spec/src/v4/mod.rs
simonwhitaker: Already have a stacked PR for this.

crates/coral-app/src/sources/materialization.rs
simonwhitaker: Blocker?

crates/coral-spec/src/v4/mod.rs
simonwhitaker: Moved the change into this PR.

jsummerfield (approved):

PR #942, opened 2026-05-28
feat(sources/community/netdata): add netdata community source

BAVYASAKTHIVEL-21:
## Summary

Adds a new community Coral source for Netdata that queries a self-hosted Netdata **Agent** through its REST API, exposing host metadata, active health alarms, and the metric context catalog as read-only SQL tables.

## Included tables

- `nodes` — `GET /api/v1/info`
- `alarms` — `GET /api/v1/alarms?all=true`
- `metrics_metadata` — `GET /api/v3/contexts?options=titles`

## Features

- Host runtime / OS / hardware metadata
- Active and configured health alarm states
- Metric context catalog (titles, units, family, priority) from Netdata's current v3 API
- `dict_entries` row strategy for Netdata's keyed-dictionary payloads
- Read-only operational access, no credentials sent

## Authentication

No credentials are required or sent. The Netdata Agent data APIs are public by default and governed by IP-based ACL (`allow dashboard from`), not a token. Netdata **Cloud** API tokens are a different product surface and do not apply to a self-hosted Agent. Agent bearer protection / authenticating proxies are not yet supported (point `NETDATA_URL` at an ACL-allowed endpoint).

## Addressing review feedback

This revision resolves the prior blockers:

- **Schema validity:** removed the unsupported `nullable: true` from the optional input (the input is simply no longer required/declared), and replaced the invalid `expr: {kind: dict_key}` with `expr: {kind: path, path: [_key]}` — Coral's `dict_entries` strategy exposes the entry key as `_key`.
- **`dict_entries` row shape:** entry rows expose the value object's fields directly, so the `path: [value, ...]` mappings are now `path: [...]` for both `alarms` and `metrics_metadata`.
- **Deprecated endpoint:** `metrics_metadata` moved from the deprecated `/api/v1/charts` to the current `/api/v3/contexts` (with `options=titles` so titles are included; units/family are returned by default). Columns now reflect the context object (`context_id`, `title`, `units`, `family`, `priority`, `live`).
- **Auth conflation:** removed the Netdata Cloud token / bearer requirement. The Agent API is unauthenticated by default; the README now distinguishes Cloud tokens from Agent/proxy authentication and states which applies here (none, by default).

The base URL is now the Agent root (`{{input.NETDATA_URL}}`) so tables can address both `/api/v1` and `/api/v3` endpoints.

## Validation

```bash
make lint-sources
coral source lint sources/community/netdata/manifest.yaml
export NETDATA_URL=http://localhost:19999
coral source add --file sources/community/netdata/manifest.yaml
coral source test netdata
coral sql "SELECT context_id, title, units FROM netdata.metrics_metadata LIMIT 5"
```

Live `coral source test netdata` output against this exact head is included in the README.


jsummerfield (dismissed):
Findings:

- High: This PR is adding `sources/community/netdata/**`, but it also deletes the existing `sources/community/clickhouse/README.md` and `sources/community/clickhouse/manifest.yaml`. Community source PRs need to stay scoped to one provider/source directory, and deleting an unrelated existing source would remove user-facing ClickHouse product surface. Please restore `sources/community/clickhouse/**` and keep this PR focused on `sources/community/netdata/**`.

Thanks for adding Netdata. Once the unrelated ClickHouse deletions are removed, I can review the Netdata source itself.

jsummerfield (changes requested):
Findings:

- High: `sources/community/netdata/manifest.yaml` and `sources/community/netdata/README.md` are both empty at this head. The exact-head manifest fails `coral source lint` with `/: null is not of type "object"`, so there is no reviewable source definition yet. Please restore the Netdata manifest and README content while keeping the PR scoped only to `sources/community/netdata/**`.

Thanks for fixing the earlier unrelated ClickHouse deletion. Once the Netdata files contain the source again, I can review the provider/API shape.

jsummerfield (changes requested):
Findings

- High: `sources/community/netdata/manifest.yaml:15` and `sources/community/netdata/manifest.yaml:90`/`137` still do not pass `coral source lint`. I verified the current head locally; `NETDATA_TOKEN` uses unsupported `nullable: true`, and `expr: {kind: dict_key}` is not a valid Coral expression. Please make the optional input and dictionary-key extraction use supported source-spec syntax before this can be added.

- High: `sources/community/netdata/manifest.yaml:95`-`124` and `sources/community/netdata/manifest.yaml:142`-`161` map `dict_entries` rows through `path: [value, ...]`. Coral's dictionary-entry rows expose the entry object fields directly, with the key available separately, so these columns would read null even after the schema errors are fixed. Please update the expressions to match the actual row shape.

- Medium: `sources/community/netdata/manifest.yaml:126`-`130` models chart metadata with `/api/v1/charts`, and `sources/community/netdata/README.md:48`-`49` documents that endpoint as the source of metric metadata. Netdata's current docs say `/api/v1/charts` is deprecated for new integrations and recommend `/api/v3/contexts` for chart metadata and dimensions. Please either move the metadata table to the current endpoint or clearly justify why the deprecated endpoint is still the right product surface. Docs: https://learn.netdata.cloud/docs/developer-and-contributor-corner/rest-api/queries and https://raw.githubusercontent.com/netdata/netdata/master/src/web/api/netdata-swagger.yaml

- Medium: `sources/community/netdata/manifest.yaml:18`-`20` and the README auth guidance conflate Netdata Cloud API tokens with direct Netdata Agent bearer protection. This source points at an Agent `/api/v1` base URL, so please distinguish Cloud API tokens from agent/proxy authentication and tell users which credential type actually applies to this source. Docs: https://learn.netdata.cloud/docs/netdata-cloud/authentication-%26-authorization/api-tokens

The previous empty-file blocker is fixed; the current head now has a real Netdata source to review. The next blocker is that the manifest still cannot load, so the representative README output cannot be accepted as live evidence for this exact revision yet.

jsummerfield (changes requested):
Thanks for the update. I re-reviewed the current head (`429260af5`) and the same source-load blockers are still present, so this cannot be approved yet.

Remaining blockers:

- `sources/community/netdata/manifest.yaml:15-17`, `:90-94`, and `:137-141` still fail `coral source lint`: `NETDATA_TOKEN` still uses unsupported `nullable: true`, and both dictionary-key columns still use unsupported `expr: {kind: dict_key}`. Please switch these to supported source-spec syntax so the manifest can load.
- `sources/community/netdata/manifest.yaml:95-124` and `:142-161` still map `dict_entries` rows through `path: [value, ...]`. The earlier row-shape issue is still open, so these columns would not read the expected values even after the schema errors are fixed.
- `sources/community/netdata/manifest.yaml:126-130` still models metric metadata through `/api/v1/charts`; the prior request to move this to the current Netdata metadata endpoint or justify the deprecated endpoint remains unresolved.
- `sources/community/netdata/manifest.yaml:18-20` still describes `NETDATA_TOKEN` as a Netdata Cloud API token or bearer token. Since this source targets a Netdata Agent/proxy URL, please keep the auth guidance specific to the credential type that actually applies here.

Because the manifest still fails lint, the README live-output block cannot be accepted as validation for this revision yet.

jsummerfield:

jsummerfield (changes requested):
Thanks for the update. I re-reviewed the current head (`29afb0f`) and the previous source-load/auth/dict-entry/metrics-metadata blockers are resolved: the manifest lints, the token input is gone, and the exact PR manifest worked live for all three tables against a Netdata Agent.

One remaining blocker before this should ship: the new source still builds `netdata.nodes` and `netdata.alarms` on deprecated v1 Agent endpoints. `sources/community/netdata/manifest.yaml` uses `GET /api/v1/info` and `GET /api/v1/alarms`, and the README documents those as intentional. Netdata's current Agent OpenAPI says v1/v2 APIs are deprecated/back-compat only and that new integrations should use v3 exclusively, and it exposes v3 equivalents for node and alert data (`/api/v3/nodes`, `/api/v3/alerts`). Since this is a new Coral source, please model those tables from the v3 API surface instead of adding new user-facing source contracts on endpoints Netdata has marked as deprecated.

Docs I checked:
- https://github.com/netdata/netdata/blob/master/src/web/api/netdata-swagger.yaml#L42
- https://github.com/netdata/netdata/blob/master/src/web/api/netdata-swagger.yaml#L120-L226
- https://github.com/netdata/netdata/blob/master/src/web/api/netdata-swagger.yaml#L1997-L2028

Once `nodes` and `alarms` use the v3 endpoints (with README examples/evidence refreshed to match), I think this should be in good shape.

jsummerfield:

jsummerfield (changes requested):
Findings

- High: `sources/community/netdata/README.md:151-175` contains literal `<PASTE: ...>` placeholders for `coral source test netdata`, `netdata.nodes`, and `netdata.alarms`. This head substantially remaps both tables from v1 to v3, so the older live evidence cannot validate the new response paths and fields. Please replace the placeholders with sanitized output from this exact manifest. The PR body currently says that exact-head live output is included, but it is not.

- Medium: `sources/community/netdata/manifest.yaml:132-157` and `:261-276` retrieve the full alert and context catalogs with no provider-backed filters or explicit cardinality bound. The README's SQL predicates are local because the manifest declares no corresponding filters, and Netdata's API defaults `cardinality` to no limit. Please expose useful alert/node/context filters and an explicit conservative provider-side bound, or otherwise constrain and document these potentially high-cardinality calls. Docs: https://github.com/netdata/netdata/blob/f78f2f5a69f1861e6383490626ddeac40109a50b/src/web/api/netdata-swagger.yaml#L1999-L2192 and https://github.com/netdata/netdata/blob/f78f2f5a69f1861e6383490626ddeac40109a50b/src/web/api/netdata-swagger.yaml#L196-L226

The requested v3 migration is otherwise in place, and the exact-head manifest passes source lint and the YAML style gate. Thanks for moving the source off the deprecated endpoints; the remaining blockers are proving that mapping live and making large installations safe to query.

jsummerfield:

jsummerfield:

PR #947, opened 2026-05-28
feat(sources/community/etherscan): add Etherscan V2 multichain community source

OoJae:
## Summary

Adds a community source for the [Etherscan V2](https://docs.etherscan.io) multichain API — exposing ERC-20 token transfer events as read-only SQL tables via Coral. One API key covers all EVM chains (Ethereum, Base, Optimism, Arbitrum, etc.) via the `chainid` parameter.

## Changes

- `sources/community/etherscan/manifest.yaml` — source spec (DSL v3)
- `sources/community/etherscan/README.md` — setup guide, schema docs, SQL examples, API caveats

## Implementation Details

- **Auth:** Query param authentication (`apikey` in URL), not header auth — matches Etherscan's API design
- **Pagination:** Offset-based (page/offset params), default 50, max 100 per page
- **Error handling:** Etherscan returns errors as HTTP 200 with `status:"0"` — configured `ok_path`/`error_path` to handle this
- **Columns:** Maps nested API fields using `expr: {kind: path, path: [...]}` for `from` → `from_address`, `to` → `to_address`, `tokenName` → `token_name`, etc.
- **Value precision:** Values kept as Utf8 to avoid precision loss — cast in SQL with `CAST(value AS DOUBLE) / 1e6`

## Validation

```
$ coral source lint sources/community/etherscan/manifest.yaml
Manifest is valid

$ coral source test etherscan
✓ SELECT hash, from_address, to_address, value FROM etherscan.token_transfers WHERE chainid = 1 AND address = '0x833589fcd6edb6e08f4c7c32d4f71b54bda02913' LIMIT 1
```

### Example query

```sql
SELECT hash, from_address, to_address, value, token_symbol, block_time
FROM etherscan.token_transfers
WHERE chainid = 1
  AND address = '0x464C71f6c2F760DdA6093dCB91C24c39e5d6e18c'
ORDER BY block_time DESC
LIMIT 5;
```

Useful for on-chain treasury analysis, grant payment tracking, and cross-source JOINs with other Coral sources.


jsummerfield (changes requested):
Findings

- High: `sources/community/etherscan/manifest.yaml:52` does not define response-level `ok_path` / `error_path` handling, even though `sources/community/etherscan/README.md:108` says Coral handles Etherscan HTTP-200 errors that way. Etherscan returns API failures as HTTP 200 responses with `status: "0"`, `message: "NOTOK"`, and the error text in `result`; without explicit handling, auth, rate-limit, and tier failures can be read as data-shape problems instead of source errors. Docs: https://info.etherscan.com/api-return-errors/

- High: `sources/community/etherscan/manifest.yaml:26` declares `chainid` and `address` filters but does not mark either one `required: true`, while `sources/community/etherscan/README.md:43` documents both as required. The V2 token-transfer endpoint requires a target chain and address, so Coral discovery should not advertise this as an unfiltered table scan. Docs: https://docs.etherscan.io/api-reference/endpoint/tokentx

- Medium: `sources/community/etherscan/manifest.yaml:26`-`36` omits the provider-supported `contractaddress` query filter, while `sources/community/etherscan/README.md:87` and `:97` filter `contract_address` locally in SQL. For token-specific workflows, that can fetch a much larger/incomplete page and then filter after the fact. Please push this filter down to the Etherscan request. Docs: https://docs.etherscan.io/api-reference/endpoint/tokentx

- Medium: `sources/community/etherscan/README.md:109` says the free tier is 5 calls/sec, but Etherscan’s current public API plans list the Free tier at 3 calls/sec and Lite at 5 calls/sec. Please update the rate-limit guidance so users do not overrun their keys. Docs: https://etherscan.io/apis

Thanks for adding a multichain Etherscan source. The main fixes are about making failures and required filters explicit so users get predictable behavior when they first try it.

jsummerfield (changes requested):
Findings
- High: `sources/community/etherscan/manifest.yaml:63` uses `ok_path: [status]`, but Coral’s HTTP backend expects `ok_path` to resolve to a boolean. Etherscan V2 responses use string values such as `"1"` and `"0"`, so successful responses can be classified incorrectly. Please do not rely on this as-is; either add/use Coral support for string success values or remove the unsupported `ok_path` handling until it exists. If supported, `error_path` should surface `result`, which contains the useful provider error text. Docs: https://docs.etherscan.io/api-reference/endpoint/tokentx and https://info.etherscan.com/api-return-errors/
- Medium: `sources/community/etherscan/README.md:83` and `sources/community/etherscan/README.md:94` still filter `contract_address`, which is the returned output column, not the provider request filter. The manifest added provider pushdown as `contractaddress`; please update the examples to use `AND contractaddress = '…'` and explain that `contractaddress` is the request filter while `contract_address` is the output column. Docs: https://docs.etherscan.io/api-reference/endpoint/tokentx
- Medium: `sources/community/etherscan/README.md:70` presents “recent transfers” using local `ORDER BY block_time DESC LIMIT 10`, but the manifest does not expose or set Etherscan’s `sort`, `startblock`, or `endblock` parameters. Since Coral limits fetched rows before local ordering, that can sort only the first provider page. Please expose those filters or set `sort=desc` for the recent workflow. Docs: https://docs.etherscan.io/api-reference/endpoint/tokentx

Thanks for the update; several earlier findings are resolved. The remaining blocker is the unsupported string `ok_path` handling, plus README examples that still bypass the new provider-side token filter.

jsummerfield (changes requested):
Findings:

- High: `sources/community/etherscan/manifest.yaml:37` makes the manifest invalid YAML because the unquoted description contains `USDC: 0x...`. `coral source lint` fails before schema validation, so the source cannot be loaded. Please quote that description or use a folded scalar.

- High: `sources/community/etherscan/manifest.yaml:66`-`:70` and `sources/community/etherscan/README.md:112` still do not correctly handle Etherscan HTTP-200 failures. Etherscan documents errors as `status: "0"`, `message: "NOTOK"`, and the useful error cause in `result`; `error_path: [message]` only points at `NOTOK` and does not establish the success condition. Please implement supported response-level success/error handling for `status == "1"` with the useful error from `result`, or remove the README claim that Coral surfaces these errors. Docs: https://docs.etherscan.io/api-reference/endpoint/tokentx and https://docs.etherscan.io/resources/common-error-messages

- Medium: `sources/community/etherscan/manifest.yaml:81`-`:86` exposes `chainid` and `address` as output columns, but the `tokentx` response does not include those fields. If users can select them, they need `from_filter` expressions; otherwise remove them as result columns. Docs: https://docs.etherscan.io/api-reference/endpoint/tokentx

The required `chainid`/`address` filters, `contractaddress` pushdown, and rate-limit docs are improved, but this head still needs to load cleanly and handle Etherscan's response semantics before approval.

OoJae:
Thanks @jsummerfield — pushed `086584c` addressing the findings on `652f3ce`:

- **High (invalid YAML, `:37`)** — quoted the `contractaddress` description (and the `sort` description at `:40`, which had the same `Sort order:` colon issue); `coral source lint` now passes.
- **High (HTTP-200 error handling)** — removed the `error_path` rather than imply support that isn't there. Coral's HTTP backend doesn't special-case Etherscan's string `status` convention, so the README now states plainly that auth/rate-limit/bad-parameter failures surface as an empty result (not a query error) and to verify the API key, `chainid`, and `address`.
- **Medium (`chainid`/`address` not in the response)** — gave both columns `from_filter` exprs so they echo the query filters instead of selecting absent response fields. Verified live: a `tokentx` query returns `chainid = 1` and the queried `address` populated, not null. `coral source test etherscan` passes 1/1.

jsummerfield (changes requested):
Findings

- High: `sources/community/etherscan/manifest.yaml:66`-`:69` still cannot safely distinguish Etherscan API failures from data rows. Etherscan documents failures as HTTP 200 responses with `status: "0"` and the useful error in `result`, while successful `tokentx` responses put an array in `result`. With `rows_path: [result]` and `row_strategy: direct`, I verified locally that an invalid API key makes `coral source test etherscan` pass and returns one bogus row with the echoed `chainid` / `address` plus null transfer fields. Please add supported response handling for `status == "1"` and surface `result` as the provider error, or avoid claiming/accepting this behavior until the source-spec backend can model it. Docs: https://docs.etherscan.io/resources/common-error-messages and https://docs.etherscan.io/api-reference/endpoint/tokentx

- Medium: `sources/community/etherscan/README.md:94`-`:103` shows a cross-source JOIN without the required `address` filter from `sources/community/etherscan/manifest.yaml:31`-`:34`. This source cannot scan transfers for every registry wallet, so please include a concrete single-wallet `tx.address = ...` predicate or reframe/remove the example.

- Medium: `sources/community/etherscan/manifest.yaml:26`-`:40` omits `startblock` / `endblock`, even though Etherscan exposes them for `tokentx` and recommends narrower block ranges for timeout-prone queries. Please add optional block-range filters and document them for active wallets. Docs: https://docs.etherscan.io/api-reference/endpoint/tokentx and https://docs.etherscan.io/resources/common-error-messages

The invalid YAML and `from_filter` fixes are good, but the current error behavior is still unsafe product surface. Thanks for continuing to tighten this up.

OoJae:
Pushed `29a013c`. Details per finding:

**Medium — block range:** added optional `startblock` / `endblock` filters (Int64), bound to the request and documented for narrowing active wallets.

**Medium — JOIN example:** the example now includes the required single-wallet `tx.address = '...'` predicate (and notes that `token_transfers` is one-wallet-per-query; iterate for the full registry).

While doing this I found that the `contractaddress`/`sort` filters from the prior round didn't actually work: Coral only accepts a filter as a SQL predicate when a **column of the same name exists**, so `WHERE contractaddress = '...'` / `WHERE sort = '...'` errored with "No column named ...". Fixed by binding the contract filter to the existing `contract_address` output column (pushdown + value in one column) and adding `from_filter` echo columns for the rest. Also renamed `sort` → `sort_order` because DataFusion reserves `sort`. Verified live that `WHERE contract_address` / `sort_order` / `startblock` / `endblock` now push down.

**High — HTTP-200 error handling:** you're right that the current behavior is unsafe, but I don't think it can be modelled on the current backend, and I'd rather not claim handling that doesn't exist. Concretely:

- `ok_path` is evaluated as `get_path_value(...).and_then(Value::as_bool).unwrap_or(false)` ([crates/coral-engine/src/backends/http/fetch.rs#L177-L180](https://github.com/withcoral/coral/blob/main/crates/coral-engine/src/backends/http/fetch.rs#L177-L180)). `as_bool` only matches a JSON boolean.
- Etherscan's `status` is the **string** `"1"`/`"0"`, never a JSON bool. I verified `ok_path: [status]` makes a **valid** key fail (`status:"1"` → not-ok → `unknown source API error`), while an invalid key correctly surfaces `result`. So it can't distinguish success from failure either way.

So there's no in-spec way to gate on `status == "1"` today. I've documented this precisely in the README (no error-handling claim) so users know an auth/rate-limit failure returns an all-null row, and what to check. **Would you accept this as a documented limitation, or would you prefer I open a feature request for string/value `ok_path` matching (e.g. `ok_path: [status]` + `ok_equals: "1"`) and hold this PR until that lands?** Happy to do the latter and link it here.

jsummerfield (changes requested):
Findings

- High: `sources/community/etherscan/manifest.yaml:66`-`:69` still lets Etherscan API failures pass as successful Coral rows. Etherscan documents failures as HTTP 200 bodies with `status: "0"`, `message: "NOTOK"`, and the useful reason in `result`; with `rows_path: [result]` and `row_strategy: direct`, an invalid key can pass `coral source test` and return one row with null transfer fields. I would not accept this as a documented limitation because it makes source validation and first user queries lie. Please hold this source until Coral can model `status == "1"` success / `result` errors, or otherwise make failures fail. Docs: https://docs.etherscan.io/resources/common-error-messages and https://docs.etherscan.io/api-reference/endpoint/tokentx

- High: `sources/community/etherscan/manifest.yaml:19` uses Base USDC (`0x833589...2913`) with `chainid = 1` for Ethereum. If the test is meant to query Base, use `chainid = 8453`; if it is meant to query Ethereum, use Ethereum USDC (`0xA0b869...eB48`). As written, the live-test evidence is especially hard to trust because the same query path also passes on auth failure. Docs: https://developers.circle.com/stablecoins/usdc-contract-addresses

- Medium: `sources/community/etherscan/README.md:125` says Base, Optimism, and Arbitrum "may require a paid plan." Current Etherscan chain support docs are more specific: Base and OP Mainnet are not free-tier chains, while Arbitrum One is available on Free. Please make this concrete so users know whether a failed query is credentials, plan tier, or source behavior. Docs: https://docs.etherscan.io/supported-chains

The block-range and JOIN updates are good, but the HTTP-200 error behavior is still unsafe product surface. Thanks for spelling out the backend limitation clearly; my recommendation is to hold this PR until that limitation is fixed rather than ship a source that can silently turn provider errors into data.

OoJae:
Understood on holding this — agreed it shouldn't ship while provider errors can become data. Pushed `3fe19f3` for the other two findings:

- **Test query:** now coherent — a real Ethereum wallet on `chainid = 1` filtered to Ethereum USDC (`0xa0b869…eB48`), test green.
- **Chain tiers:** README now states Ethereum mainnet and Arbitrum One are free-tier, while Base and OP Mainnet require a paid plan (linked to the supported-chains doc).

For the HTTP-200 blocker I filed a feature request: **#1064** — adding `ok_equals`/value matching (or scalar coercion) to `ok_path` so string/int success flags like Etherscan's `status` can be modelled, with an offer to send the PR. I'll **convert this PR to draft** and park it pending that; once `ok_equals` lands I'll add `ok_path: [status]` + `ok_equals: "1"` + `error_path: [result]` so failures fail and re-request review. Thanks for the thorough passes.

jsummerfield (changes requested):
Findings

- High: `sources/community/etherscan/manifest.yaml:78`-`:81` still cannot distinguish Etherscan V2 success from provider errors. Etherscan documents successful token-transfer responses as `status: "1"` with `result` as an array, but provider failures also return HTTP 200 with `status: "0"` and the error text in `result`. With `rows_path: [result]` and no usable fail-closed success check, invalid-key, rate-limit, or unsupported-chain responses can become apparent rows with null transfer fields while echoed filter columns still populate. The README documents this limitation, but that still means `coral source test etherscan` can report apparent success for broken auth/provider errors. I think this should wait for Coral response validation that can match string success values, for example `status == "1"` with `error_path: [result]`, or another fail-closed source-spec mechanism. Docs: https://docs.etherscan.io/api-reference/endpoint/tokentx and https://docs.etherscan.io/resources/common-error-messages.

The test query mismatch looks fixed: the manifest now uses Ethereum mainnet `chainid = 1` with the Ethereum USDC contract. The plan-tier docs are also materially improved and now call out free/Lite rates and Base/OP paid-plan limits. The remaining issue is the setup/test path accepting provider errors as data, so I don't think this source is safe to approve yet.

jsummerfield:

OoJae:
Following up on the fail-closed request: rather than document the limitation, I implemented the missing mechanism upstream — **#1237** adds `response.require_array_rows`.

Why this and not `status == "1"`: I checked the live Etherscan V2 API and `status` is `"0"` for **both** a valid empty result (`result: []`) and a real error (`result: "<msg>"`), so a status check would reject legitimate empty results. `require_array_rows` keys off `result` being an array vs a scalar instead — verified end-to-end:

- valid key → passes
- empty block range (`status:"0"`, `result:[]`) → 0 rows, no error
- invalid key (`status:"0"`, `result:"Invalid API Key"`) → query fails closed with the error surfaced (no more bogus row)

Once #1237 lands and is released, I'll update this PR's manifest to:
```yaml
response:
  rows_path: [result]
  require_array_rows: true
  error_path: [result]
  row_strategy: direct
```
and re-request review. Keeping #947 in draft until then.

OoJae:
Update: the feature PR (#1237) was closed in favor of designing a more general fail-closed mechanism on the Coral roadmap — discussion is now in #1064. Keeping this PR parked as a draft until that lands; the manifest change here will be a one-line update once the mechanism exists. No action needed on your side in the meantime.

PR #948, opened 2026-05-28
feat(sources/community/neynar): add Neynar Farcaster community source

OoJae:
## Summary

Adds a community source for the [Neynar](https://neynar.com) Farcaster API — exposing Farcaster cast search as a read-only SQL table function via Coral. This source demonstrates the `kind: search` pattern for provider-ranked retrieval, where the API decides relevance ordering rather than SQL WHERE clauses.

## Changes

- `sources/community/neynar/manifest.yaml` — source spec (DSL v3)
- `sources/community/neynar/README.md` — setup guide, schema docs, SQL examples, API caveats

## Implementation Details

- **Auth:** HeaderAuth using `x-api-key` header from the `NEYNAR_API_KEY` input
- **Function type:** `kind: search` — provider-ranked retrieval with `search_limits` (default_top_k: 25, max_top_k: 100, max_calls_per_query: 1)
- **Pagination:** Cursor-based (`cursor_query` mode) with cursor param and response path
- **Columns:** Nested Farcaster fields flattened with `__` separator (e.g., `author__username`, `reactions__likes_count`)
- **Call syntax:** `FROM neynar.search_casts(q => 'search query')` — the `q` argument is bound via `bind: {arg: q}`

## Validation

```
$ coral source lint sources/community/neynar/manifest.yaml
Manifest is valid

$ coral source test neynar
✓ SELECT hash, text, author__username FROM neynar.search_casts(q => 'optimism') LIMIT 1
```

### Example query

```sql
SELECT hash, text, author__username, reactions__likes_count
FROM neynar.search_casts(q => 'DAO governance')
ORDER BY reactions__likes_count DESC
LIMIT 10;
```

**Note:** The cast search endpoint requires a paid Neynar API plan. Free keys return HTTP 402 on `/cast/search`. This is documented in the README.


jsummerfield (changes requested):
Findings

- High: `sources/community/neynar/manifest.yaml:66`-`83` declare nested fields such as `author__username`, `reactions__likes_count`, and `replies__count` without `expr` paths. Neynar’s cast-search response nests those values under `author`, `reactions`, and `replies`, so Coral will look for literal double-underscore keys and return null columns. Please add explicit nested `expr.path` mappings or remove the fields. Docs: https://docs.neynar.com/reference/search-casts

- Medium: `sources/community/neynar/manifest.yaml:31`-`42` exposes only `q`, while the documented cast search endpoint supports important query parameters such as `mode`, `sort_type`, `author_fid`, `viewer_fid`, `parent_url`, and `channel_id`. That makes the table much less useful for common Farcaster workflows. Please expose the most important filters, or narrow the README/table description so users understand this is only basic keyword search. Docs: https://docs.neynar.com/reference/search-casts

- Medium: `sources/community/neynar/README.md:136`-`138` says the source supports cursor pagination, but `sources/community/neynar/manifest.yaml:27` sets `max_calls_per_query: 1`. As written, users only get the first page. Please either implement pagination across cursors or update the docs to say the first version returns one page up to the configured limit.

Thanks for the contribution. The source is close in scope, but the nested response fields and pagination claim need to match what users will actually get back from Coral.

jsummerfield (changes requested):
Findings
- Medium: `sources/community/neynar/manifest.yaml:27`, `sources/community/neynar/manifest.yaml:52`, and `sources/community/neynar/manifest.yaml:60` still do not bind the provider page size to Neynar’s `limit` query parameter. The source declares `max_top_k: 100` and the README shows `LIMIT 50`, but `/cast/search` only receives `limit` when the user explicitly filters `WHERE limit = ...`; SQL `LIMIT` will not reliably request that many provider-ranked results from Neynar. Please move the page-size control into pagination, for example `page_size: { default: 25, max: 100, query_param: limit }`, and remove or clearly document the separate `limit` filter. Docs: https://docs.neynar.com/reference/search-casts

Thanks for the update; the previous nested-field and README pagination issues look mostly resolved. This remaining pagination binding is worth fixing so the search function behaves the way the README and `search_limits` imply.

jsummerfield (changes requested):
Findings:

- High: `sources/community/neynar/manifest.yaml:77` makes the manifest invalid YAML because the plain scalar contains `Search mode:`. `coral source lint` fails before schema validation, so the source cannot be loaded. Please quote the description or use a folded scalar.

- High: `sources/community/neynar/manifest.yaml:68`-`:77` adds `filters:` under a table function, which is not a valid HTTP function shape. The README examples at `sources/community/neynar/README.md:88`-`:103` use `WHERE channel_id = ...` / `WHERE mode = ...`, but those filters will not exist. Please model these as optional function args, for example `search_casts(q => ..., channel_id => ..., mode => ...)`, and bind them into the request query. Neynar documents `mode`, `author_fid`, `channel_id`, `limit`, and `cursor` as `/cast/search` query params: https://docs.neynar.com/reference/search-casts

- Medium: `sources/community/neynar/manifest.yaml:24`-`:26` and `sources/community/neynar/README.md:41` still say results are provider-ranked / relevance-ordered, but Neynar documents `sort_type` defaulting to `desc_chron`; `algorithmic` is a separate option that this source does not expose. Please either expose `sort_type` or narrow the wording so users do not expect relevance ordering by default. Docs: https://docs.neynar.com/reference/search-casts

- Medium: `sources/community/neynar/manifest.yaml:133` exposes `timestamp` as `Utf8`. Neynar returns ISO timestamps, so this should be a `Timestamp` column for natural time ordering/filtering. Docs: https://docs.neynar.com/reference/search-casts

The nested response-field mappings and page-size binding from the prior review look improved, but the current head is blocked because the manifest does not parse and the table-function filters are not valid yet.

OoJae:
Thanks @jsummerfield — pushed `6a9b8fb` addressing the findings on `6d88349`:

- **High (invalid YAML, `:77`)** — the `mode` description with `Search mode:` is gone; `mode` is now a function arg with a `values` enum. `coral source lint` passes.
- **High (`filters:` invalid on a table function)** — replaced the `filters:` block with optional **function args** (`author_fid`, `channel_id`, `mode`) bound into the request via `from: arg`/`arg_int`, plus a new `sort_type` arg. README examples now call `search_casts(q => ..., channel_id => ..., mode => 'semantic')` instead of unsupported `WHERE` clauses.
- **Medium (ordering wording)** — corrected to reverse-chronological by default (`sort_type => 'desc_chron'`); `sort_type => 'algorithmic'` is exposed for relevance ranking.
- **Medium (`timestamp` type)** — now `Timestamp` via `format_timestamp` with `input: iso8601`.

One note on the live test: `/cast/search` requires a paid Neynar plan, so `coral source test neynar` returns **402 Payment Required** on a free key (documented under "Paid plan required" in the README). The manifest lints clean and the request is well-formed — a free-tier key returns a clean provider 402, and Coral validates the `mode`/`sort_type` enums before the request is sent. Happy to adjust the declared test query if you'd prefer it not depend on a paid endpoint.

jsummerfield (changes requested):
Findings

- Medium: `sources/community/neynar/manifest.yaml:27`-`:28` and `sources/community/neynar/README.md:5`, `:41`, and `:130` still describe `sort_type => 'algorithmic'` as relevance ranking. Neynar documents `algorithmic` as casts sorted by engagement and time, while `desc_chron` / `chron` are the chronological options. Please call this engagement/time or algorithmic engagement ordering rather than relevance, so users do not expect semantic relevance from a sort flag. Docs: https://docs.neynar.com/reference/search-casts

The invalid YAML, table-function args, pagination limit binding, nested field paths, and `Timestamp` mapping look resolved. I’m requesting this wording fix because ranking semantics are part of the query contract. Thanks for the quick iteration.

OoJae:
Pushed `165c858` — `sort_type => 'algorithmic'` is now described as ordering by **engagement and time** (not relevance) in the manifest function description and the README (intro, function summary, and semantics note). Thanks for the catch.

jsummerfield (approved):
No blocking findings on this head.

Non-blocking notes before merge:

- `sources/community/neynar/README.md:130` says users cannot filter by an output column such as `WHERE author__username = '...'`. That is too broad: SQL can still filter rows after Coral fetches the provider result set; it just is not provider-pushed. Please reword this so users understand the difference between function arguments and local SQL filtering.

- Consider adding Neynar's endpoint-specific rate limits to the README instead of only saying they depend on plan tier. Their docs list `GET /v2/farcaster/cast/search` limits by tier, which helps users choose page sizes and retry behavior. Docs: https://neynar.mintlify.app/reference/what-are-the-rate-limits-on-neynar-apis

The previous blockers around nested fields, function args, pagination binding, timestamp typing, and ranking wording look resolved. Product fit is acceptable for a community source, and the PR includes paid-key live-test evidence in the description.

OoJae:
Pushed `38ac661` addressing both non-blocking notes from your approval (docs-only):

- **Search-function semantics (README):** reworded — function args (`q`, `author_fid`, `channel_id`, `mode`, `sort_type`) narrow what Neynar fetches and ranks provider-side; ordinary SQL predicates on output columns still work but filter the returned set **locally after retrieval** (not pushed to Neynar). Dropped the overly broad "cannot filter by an output column" phrasing.
- **Rate limits (README):** added the endpoint-specific `/cast/search` limits by tier — Starter 60 RPM, Growth 120 RPM, Scale 240 RPM, Enterprise custom — and noted each call is a single request (`max_calls_per_query: 1`), so size pages with `LIMIT` and back off when fanning out.

No manifest change. Flagging in case the new commit re-requests review — happy to re-confirm whenever you are.

jsummerfield (approved):
No blocking findings on this head.

The docs-only follow-up addresses the previous non-blocking notes: the README now distinguishes function arguments from local SQL filtering, and it includes Neynar's endpoint-specific rate-limit tiers. I rechecked the search-casts mapping against Neynar's docs, and the source still has live-test evidence in the PR description.

Thanks for the quick cleanup; this looks ready for James's final merge pass.

PR #945, opened 2026-05-28
perf(engine): reuse default HTTP source client

Bradley-Butcher:
## Intent

Second PR in the catalog cold-start stack. This builds on #962 by removing the remaining repeated default HTTP transport setup during source registration.

#962 narrows request-input state so HTTP/MCP registration no longer clones full source manifests. This PR leaves that contract alone and changes only the default HTTP client lifecycle.

## What changed

- Added a batch-scoped `BackendRegistrationContext` shared across one runtime registration pass.
- Reused one default `reqwest::Client` for HTTP sources registered in that pass.
- Kept source-specific HTTP config, auth, headers, rate limits, resolved inputs, request-input refresh, and body-capture settings per source.
- Added a focused registration-context test proving the default HTTP client is shared within one registration pass but not process-global.

## Ownership / Boundaries

- `coral-engine` owns this because source registration and backend runtime construction live in the engine.
- The registration context is backend-agnostic. Every backend receives it; HTTP uses it today, while file and MCP ignore it.
- The shared client is batch-scoped, not global. A new runtime registration pass gets a new context and therefore a new default client.
- This only shares the default transport client. It does not merge source auth, resolved inputs, headers, rate limits, request-input resolver state, or source metadata.
- No proto, gRPC, CLI, MCP tool schema, source manifest schema, or persisted config shape changes.

## Compatibility / User Impact

This reduces cold metadata latency and also improves real HTTP endpoint queries, because both paths pay runtime/source registration before returning metadata or executing a provider request.

This assumes the current HTTP source manifest contract: sources do not declare per-source proxy, custom TLS roots, client certificates, timeout, or user-agent settings. If that contract grows later, the registration context is the seam where the client cache key should become configuration-aware.

## Timing

Five fresh-process MCP repeats using the real local Coral manifest. Each repeat does `initialize`, `tools/list`, then the target `tools/call`. The table reports target `tools/call` wall time.

| Build | `list_catalog` avg | `search_catalog limit=5` avg | `sql github.issues LIMIT 1` avg | Notes |
| --- | ---: | ---: | ---: | --- |
| Current main closest baseline (`72ad33dc`) | `2312.664 ms` | `2325.554 ms` | `2997.436 ms` | Closest comparable pre-stack main run. |
| #962 only (`16087f2c` + #962) | `855.819 ms` | `853.521 ms` | `1616.908 ms` | Removes the post-#764 manifest clone cost. |
[...]

PR #962, opened 2026-05-28
perf(engine): narrow request input context

Bradley-Butcher:
## Intent

Fix the catalog/query runtime registration regression introduced by OAuth refresh support without pulling the separate reqwest-client reuse change into the same PR.

PR #764 added request-time source input refresh by passing the full `QuerySource` into `SourceInputResolver`. That made HTTP registration carry a full validated source manifest in `HttpSourceClient`, which is cloned for every HTTP table and table function. The resolver only needs source identity, declared inputs, variables, and current secrets.

## What changed

- Added `SourceInputResolutionContext` to the engine extension surface.
- Changed `SourceInputResolver::resolve_inputs` to take `&SourceInputResolutionContext` instead of `&QuerySource`.
- Moved HTTP and MCP request-input storage to the narrower context and stopped carrying duplicate secret/variable maps beside it.
- Kept OAuth refresh behavior intact: app-managed refresh still happens lazily at request time, and catalog registration does not refresh credentials.
- Added a focused `SourceInputResolutionContext` contract test covering source identity, declared input order, required-secret filtering, variables, current secrets, and refreshed secret replacement.

## Ownership / Boundaries

- `coral-engine` owns the extension contract because `SourceInputResolver` is an engine extension used by backend request execution.
- `SourceInputResolutionContext` is intentionally narrower than `QuerySource`: it exposes only source name, declared input specs, configured variables, and current/refreshed secrets.
- `coral-app` still owns credential refresh lifecycle. The app resolver can refresh workspace-managed credentials at request time, then return resolved inputs to the engine.
- HTTP and MCP backends do not know how credentials are stored or refreshed. They only ask the resolver for request inputs when one is installed, otherwise they use the pre-resolved static inputs.
- No proto, gRPC, CLI, MCP tool schema, source manifest schema, or persisted config shape changes in this PR.

This preserves the OAuth refresh behavior from PR #764 while removing the accidental registration-time clone cost. Catalog metadata calls get faster, and real HTTP endpoint queries also get faster because they pay the same registration cost before executing the request.

The resolver still receives all input material it needs:

- source identity for workspace/provider lookup
- declared input specs for refresh and validation
- configured variables for template/input resolution
- current secrets, plus a way to produce a refreshed secret context

## Timing

Five fresh-process MCP repeats using the real local Coral manifest. Each repeat does `initialize`, `tools/list`, then the target `tools/call`. The table reports target `tools/call` wall time.

| Build | `list_catalog` avg | `search_catalog limit=5` avg | `sql github.issues LIMIT 1` avg | Notes |
| --- | ---: | ---: | ---: | --- |
| Before PR #764 (`a6febf34`) | `837.755 ms` | not measured | not measured | Before OAuth refresh context was added. |
| Current main closest baseline (`72ad33dc`) | `2312.664 ms` | `2325.554 ms` | `2997.436 ms` | Same 5-repeat harness, closest comparable main build. |
| This PR (`16087f2c` + patch) | `855.819 ms` | `853.521 ms` | `1616.908 ms` | Restores catalog registration to the pre-#764 shape. |
| Improvement vs current main | `-1456.845 ms` / `2.70x faster` | `-1472.033 ms` / `2.73x faster` | `-1380.528 ms` / `1.85x faster` | Improves both metadata calls and a real HTTP endpoint query. |

Interpretation: this PR removes the post-#764 manifest clone cost from registration. Catalog metadata calls return to the pre-#764 timing shape, and real endpoint queries also improve because they pay the same registration cost before executing the HTTP request.

## Not in this PR

- Reusing a shared `reqwest::Client` across HTTP source registration. That remains the follow-up performance PR.
- Changing credential refresh semantics or where app-managed credentials are stored.
- Changing MCP/CLI/API response shapes.

The remaining ~850 ms catalog cost is the older repeated `reqwest::Client` / native TLS setup issue.

## Validation

- `cargo fmt --check`
- `cargo check -p coral-engine -p coral-app --tests --locked`
- `cargo clippy -p coral-engine -p coral-app --all-targets --all-features --locked -- -D warnings`
- `cargo test -p coral-engine source_input_context_keeps_only_request_input_contract --locked`
- `cargo test -p coral-app runtime_config_composes_provider_input_resolver_with_refreshed_inputs --locked`
- `cargo test -p coral-engine --lib backends::mcp::tests::stdio_env_resolves_source_inputs_for_each_tool_call --locked`
- `make rust-checks`

### DIFF
--- a/crates/coral-engine/src/backends/http/mod.rs
+++ b/crates/coral-engine/src/backends/http/mod.rs
@@ -15,7 +15,7 @@ use crate::backends::{
     build_registered_table_function, internal_table_function_name, registered_columns_from_specs,
     required_filter_names,
 };
-use crate::{QuerySource, RequestAuthenticator, SourceInputResolver};
+use crate::{RequestAuthenticator, SourceInputResolutionContext, SourceInputResolver};
 use coral_spec::backends::http::{HttpSourceManifest, HttpTableSpec};
 pub(crate) mod auth;
 pub(crate) mod client;
@@ -42,9 +42,7 @@ pub(crate) use provider::HttpSourceTableProvider;
 #[derive(Debug, Clone)]
 struct HttpCompiledSource {
     manifest: HttpSourceManifest,
-    source: QuerySource,
-    source_secrets: std::collections::BTreeMap<String, String>,
-    source_variables: std::collections::BTreeMap<String, String>,
+    source_input_resolution: SourceInputResolutionContext,
     request_authenticators: HashMap<String, Arc<dyn RequestAuthenticator>>,
     body_capture_max_bytes: Option<usize>,
     source_input_resolver: Option<Arc<dyn SourceInputResolver>>,
@@ -52,18 +50,14 @@ struct HttpCompiledSource {
 
 pub(crate) fn compile_source(
     manifest: HttpSourceManifest,
-    source: QuerySource,
-    source_secrets: std::collections::BTreeMap<String, String>,
-    source_variables: std::collections::BTreeMap<String, String>,
+    source_input_resolution: SourceInputResolutionContext,
     request_authenticators: HashMap<String, Arc<dyn RequestAuthenticator>>,
     body_capture_max_bytes: Option<usize>,
     source_input_resolver: Option<Arc<dyn SourceInputResolver>>,
 ) -> Box<dyn CompiledBackendSource> {
     Box::new(HttpCompiledSource {
         manifest,
-        source,
-        source_secrets,
-        source_variables,
+        source_input_resolution,
         request_authenticators,
         body_capture_max_bytes,
         source_input_resolver,
@@ -76,9 +70,7 @@ pub(crate) fn compile_manifest(
 ) -> Box<dyn CompiledBackendSource> {
     compile_source(
         manifest.clone(),
-        request.source.clone(),
-        request.source_secrets.clone(),
-        request.source_variables.clone(),
+        SourceInputResolutionContext::from_query_source(request.source),
         request.request_authenticators.clone(),
         request.runtime_context.http_body_capture_max_bytes,
         request.source_input_resolver.clone(),
@@ -98,10 +90,10 @@ impl CompiledBackendSource for HttpCompiledSource {
     async fn register(&self, _ctx: &SessionContext) -> Result<BackendRegistration> {
         let backend = HttpSourceClient::from_manifest_with_source_input_resolver(
             &self.manifest,
-            &self.source_secrets,
-            &self.source_variables,
+            self.source_input_resolution.secrets(),
+            self.source_input_resolution.variables(),
             &self.request_authenticators,
-            self.source.clone(),
+            self.source_input_resolution.clone(),
             self.source_input_resolver.clone(),
             self.body_capture_max_bytes,
         )?;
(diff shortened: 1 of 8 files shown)


Bradley-Butcher:
* **#1009** <a href="https://app.graphite.com/github/pr/withcoral/coral/1009?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#945** <a href="https://app.graphite.com/github/pr/withcoral/coral/945?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#962** <a href="https://app.graphite.com/github/pr/withcoral/coral/962?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/962?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-engine/src/composition.rs
```diff
@@ -137,6 +138,78 @@ impl SourceInputResolverError {
...
+        self.declared_inputs
+            .iter()
+            .filter(|input| input.kind == ManifestInputKind::Secret && input.required)
+            .map(|input| input.key.clone())
+            .collect()
+    }
+
+    #[must_use]
+    /// Returns a new context with refreshed secret values.
+    pub fn with_secrets(&self, secrets: BTreeMap<String, String>) -> Self {
```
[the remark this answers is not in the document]
Bradley-Butcher: Good catch from cubic, but this is preserving the pre-existing behavior rather than a regression from this PR.

Before this change, `source_with_refreshed_secrets` built a fresh `QuerySource` with only the secrets present in refreshed material:

```rust
QuerySource::new(source.source_spec().clone(), source.variables().clone(), refreshed_secrets)
```

So a partial refresh already replaced the delegate-visible secret map rather than overlaying it. This PR keeps that replacement semantics while narrowing the type from `QuerySource` to `SourceInputContext`.

I agree overlay semantics may be a better future contract, but I would treat that as a deliberate behavior change rather than folding it into this perf/API narrowing PR.

crates/coral-engine/src/composition.rs
```diff
@@ -137,6 +138,78 @@ impl SourceInputResolverError {
...
 }
 
+/// Request-time source input context exposed to source input resolvers.
+///
+/// This carries only the source identity and declared input state needed to
+/// refresh app-managed inputs before an outbound source request. It deliberately
+/// avoids carrying the full validated source manifest, because backends clone
+/// request state for each registered table and table function.
+#[derive(Debug, Clone)]
+pub struct SourceInputContext {
```
jsummerfield: Could we rename this to `SourceInputResolutionContext`? The narrower type is useful, but `Context` by itself is vague; tying the name to `SourceInputResolver::resolve_inputs` makes it clearer that this is the input-resolution view of a selected source, not a general source/runtime context.

crates/coral-engine/src/backends/http/mod.rs
```diff
@@ -42,7 +42,7 @@ pub(crate) use provider::HttpSourceTableProvider;
 #[derive(Debug, Clone)]
 struct HttpCompiledSource {
     manifest: HttpSourceManifest,
-    source: QuerySource,
+    source_input_context: SourceInputContext,
```
jsummerfield: Now that this carries declared inputs, variables, and secrets, could we make it the single input carrier instead of also storing `source_secrets` / `source_variables` next to it? The duplicate maps are only used for fallback resolved inputs and catalog input metadata, both of which can be derived from this context. Same shape exists in MCP (`McpCompiledSource` plus `McpSourceInputs`).

jsummerfield (approved):

crates/coral-engine/src/backends/http/mod.rs
Bradley-Butcher: Done in `76c743e`: HTTP and MCP compiled sources now keep `SourceInputResolutionContext` as the single input carrier and derive fallback resolved inputs/catalog input metadata from it rather than carrying parallel `source_secrets` / `source_variables` maps.

PR #964, opened 2026-05-28
feat(sources/community/workos): add WorkOS source

Akash-How:
Summary:
- Adds a community Coral source for WorkOS organizations, users, organization memberships, directories, and events.
- Keeps API credentials as secret input and uses bounded cursor pagination.
- Includes README setup, examples, limitations, and validation notes.

Validation:
- YAML parsing: passed
- git diff --check: passed
- make lint-sources: passed
- Live API tests: passed against a WorkOS staging environment

### DIFF
--- a/sources/community/workos/README.md
+++ b/sources/community/workos/README.md
@@ -34,7 +34,7 @@ Official docs:
 | --- | --- |
 | `workos.organizations` | WorkOS organizations. |
 | `workos.users` | User Management users. Supports `organization_id` and `email`. |
-| `workos.organization_memberships` | User-to-organization memberships. Requires `organization_id`; optionally supports `user_id`. |
+| `workos.organization_memberships` | User-to-organization memberships. WorkOS requires either `organization_id` or `user_id`. |
 | `workos.directories` | Directory Sync directories. |
 | `workos.events` | WorkOS environment events. Supports organization, event, time range, and order filters. |
 
@@ -67,6 +67,15 @@ WHERE organization_id = 'org_...'
 LIMIT 25;
 ```
 
+Review memberships for one user:
+
+```sql
+SELECT id, user_id, organization_id, status, role_slug
+FROM workos.organization_memberships
+WHERE user_id = 'user_...'
+LIMIT 25;
+```
+
 Review recent events:
 
 ```sql
@@ -80,6 +89,9 @@ LIMIT 25;
 ## Notes
 
 - WorkOS list endpoints are cursor paginated with `after` and `limit`.
+- WorkOS requires `organization_id` or `user_id` for membership queries; an
+  unscoped `workos.organization_memberships` query is expected to fail at the
+  provider API.
 - The source is read-only and does not create, update, or delete WorkOS
   resources.
 - Event `data` and `context` are JSON metadata columns; avoid selecting them in
--- a/sources/community/workos/README.md
+++ b/sources/community/workos/README.md
@@ -34,7 +34,8 @@ Official docs:
 | --- | --- |
 | `workos.organizations` | WorkOS organizations. |
 | `workos.users` | User Management users. Supports `organization_id` and `email`. |
-| `workos.organization_memberships` | User-to-organization memberships. WorkOS requires either `organization_id` or `user_id`. |
+| `workos.organization_memberships_by_organization` | User-to-organization memberships. Requires `organization_id`. |
+| `workos.organization_memberships_by_user` | User-to-organization memberships. Requires `user_id`. |
 | `workos.directories` | Directory Sync directories. |
 | `workos.events` | WorkOS environment events. Supports organization, event, time range, and order filters. |
 
@@ -62,7 +63,7 @@ Review memberships for one organization:
 
 ```sql
 SELECT id, user_id, organization_id, status, role_slug
-FROM workos.organization_memberships
+FROM workos.organization_memberships_by_organization
 WHERE organization_id = 'org_...'
 LIMIT 25;
 ```
@@ -71,7 +72,7 @@ Review memberships for one user:
 
 ```sql
 SELECT id, user_id, organization_id, status, role_slug
-FROM workos.organization_memberships
+FROM workos.organization_memberships_by_user
 WHERE user_id = 'user_...'
 LIMIT 25;
 ```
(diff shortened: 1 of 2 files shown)


jsummerfield (changes requested):
Findings

- High: `sources/community/workos/manifest.yaml:247` uses `GET /directory_sync/directories`, but WorkOS lists directories at `GET /directories`; the current WorkOS OpenAPI spec has `/directories` and not `/directory_sync/directories`. That makes `workos.directories` fail against the documented API. Docs: https://workos.com/docs/reference/directory-sync/directory and https://github.com/workos/openapi-spec/blob/main/spec/open-api-spec.yaml#L6375

- High: `sources/community/workos/manifest.yaml:175` and `sources/community/workos/README.md:31` model `organization_memberships` as an optionally-filtered table, but WorkOS requires at least one of `user_id` or `organization_id` when listing organization memberships. Coral discovery will otherwise make an unfiltered `SELECT * FROM workos.organization_memberships LIMIT ...` look valid even though the provider rejects it. Please mark one of the supported scoping filters as required, or split this into table functions / separate filtered tables that make the requirement explicit. Docs: https://workos.com/docs/reference/authkit/organization-membership and https://github.com/workos/openapi-spec/blob/main/spec/open-api-spec.yaml#L12530

- Medium: `sources/community/workos/manifest.yaml:217` exposes `role_slug`, but the current WorkOS membership response nests this under `role.slug`, not a top-level `role_slug`. Please add an explicit nested `expr.path` or remove the column. Docs: https://github.com/workos/openapi-spec/blob/main/spec/open-api-spec.yaml#L31735-L31804

- Medium: `sources/community/workos/manifest.yaml:13` and `sources/community/workos/README.md:15` imply the API key can be narrowed to read access, but WorkOS API authentication docs describe secret keys as environment-scoped credentials that can perform API requests. Since this source is read-only but the credential is not inherently read-only, please make the setup docs clearer about key risk and environment scoping. Docs: https://workos.com/docs/reference/api-authentication

Thanks for adding a useful identity/admin source. The included live evidence covers organizations and users, but it misses the broken directories route and membership filter contract, so I’m requesting changes before approval.

Akash-How:
Addressed the WorkOS review findings in `8cf91c2`:\n\n- Switched Directory Sync listing to the documented `GET /directories` route.\n- Marked `organization_id` required for `workos.organization_memberships` so discovery does not imply unscoped membership scans are valid.\n- Mapped `role_slug` from nested `role.slug`.\n- Clarified that WorkOS secret keys are environment-scoped credentials and not inherently read-only.\n\nValidation:\n- YAML parsing: passed\n- Coral manifest schema validation: passed\n- `git diff --check`: passed\n- `make lint-sources`: passed\n- Live Coral test: passed, including `SELECT id, name FROM workos.directories LIMIT 1` against the updated `/directories` route

Akash-How:
Follow-up for the Cubic membership filter note in `4c8875b`:\n\n- Removed the over-strict `required: true` from `organization_id` so valid `user_id`-only WorkOS membership lookups remain possible.\n- Updated the manifest guide and README to state the provider contract clearly: WorkOS requires either `organization_id` or `user_id`; unscoped membership queries are rejected by the WorkOS API.\n\nValidation:\n- YAML parsing: passed\n- Coral manifest schema validation: passed\n- `git diff --check`: passed\n- `make lint-sources`: passed

jsummerfield (changes requested):
Findings

- High: `sources/community/workos/manifest.yaml:172`-`190` still exposes `workos.organization_memberships` as a normal table with optional filters, while WorkOS requires at least one of `organization_id` or `user_id` for the list request. The README now documents that unscoped membership queries fail, but Coral discovery will still make `SELECT * FROM workos.organization_memberships LIMIT ...` look like a valid first query and then route users into a provider error. Please make the scoping requirement part of the source contract, for example by splitting this into separate query surfaces such as memberships by organization and memberships by user with required filters, or by otherwise requiring a provider-supported scope before the request is sent. Docs: https://workos.com/docs/reference/authkit/organization-membership

Thanks for the fixes to the directories route, nested `role.slug`, and credential-risk docs. This last blocker is a source ergonomics issue: provider-required scoping should be enforced by the source instead of left as a predictable runtime failure.

Akash-How:
Addressed the remaining WorkOS membership scope blocker in `beaaf4d`:\n\n- Removed the unscoped `workos.organization_memberships` table.\n- Added `workos.organization_memberships_by_organization` with required `organization_id`.\n- Added `workos.organization_memberships_by_user` with required `user_id`.\n- Updated README examples and notes so Coral discovery now exposes only provider-supported scoped membership query surfaces.\n\nValidation:\n- YAML parsing: passed\n- Coral manifest schema validation: passed\n- `git diff --check`: passed\n- `make lint-sources`: passed\n- Live Coral test: passed; source now reports 6 tables including the two scoped membership tables

jsummerfield (changes requested):
Findings

- Medium: `sources/community/workos/manifest.yaml:172`-`188` and `sources/community/workos/manifest.yaml:238`-`254` fixed the required scoping problem by splitting memberships by organization and by user, but both tables still omit WorkOS's `statuses` query parameter. WorkOS defaults the list endpoint to active memberships; pending invitations and inactive memberships require `statuses`. As written, users cannot audit pending or inactive memberships through Coral, and the README just describes generic memberships. Please add optional `statuses` filters to both membership tables, pass them as the `statuses` query parameter, and document the default active-only behavior with an example for pending/inactive states. Docs: https://workos.com/docs/reference/authkit/organization-membership#list-organization-memberships and https://github.com/workos/openapi-spec/blob/main/spec/open-api-spec.yaml#L12530-L12610

Thanks for splitting the membership table into provider-supported scoped query surfaces. That resolves the previous blocker; the remaining issue is making the membership audit surface complete enough for normal admin workflows.

Akash-How:
Addressed the WorkOS membership status audit finding in `bc2bc54`:\n\n- Added optional `statuses` filters to both scoped membership tables.\n- Passes `statuses` through to WorkOS as the documented query parameter.\n- Documented WorkOS's active-only default and added a pending/inactive membership audit example.\n\nValidation:\n- YAML parsing: passed\n- Coral manifest schema validation: passed\n- `git diff --check`: passed\n- `make lint-sources`: passed\n- Live Coral test: passed, 3 declared / 3 passed

jsummerfield (approved):
No substantive source findings on this head.

The previous WorkOS blockers are resolved: memberships are now split into provider-supported scoped tables, `statuses` is passed through and documented, and directories use the documented `/directories` endpoint. I also verified `coral source lint` and `git diff --check` locally, and the current GitHub checks are green. The README includes captured live Coral output for the source.

Thanks for tightening this up.

PR #990, opened 2026-05-28
ci: add coral.tables performance regression check

jsummerfield:
## Intent

Catch regressions in the user-visible latency of catalog discovery through the actual CLI path. The check intentionally benchmarks `coral sql "select * from coral.tables"` against a release binary after installing the bundled GitHub source with fake credentials, so it covers process startup, local server bootstrap, source registration, credential lookup, catalog materialization, and result rendering.

## Changes

- Add `xtask perf-check` as the benchmark harness instead of adding another script under `scripts/`.
- Add `make perf-check` as the local entrypoint.
- Add a Validate workflow job that installs hyperfine, builds release `coral`, runs the xtask benchmark, and fails when the mean exceeds 750 ms.
- Run `xtask perf-check` in release mode so the perf job reuses the release profile built for `coral-cli` instead of compiling a separate dev profile.
- Reuse the existing Ubuntu Validate Rust cache namespace for the perf job instead of introducing a cold `coral-performance-*` cache key.
- Keep `scripts/` reserved for the Coral installer in agent guidance.

## Validation

- `cargo check --locked -p xtask`
- `cargo clippy --locked -p xtask --all-targets --all-features -- -D warnings`
- `cargo test --locked -p xtask`
- `make perf-check` mean: 0.236s against a 0.750s threshold
- `sh -n scripts/install.sh`
- Parsed `.github/workflows/validate.yml` with PyYAML

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -489,7 +489,7 @@ jobs:
       - name: Cache Rust artifacts
         uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         with:
-          shared-key: coral-performance-${{ runner.os }}
+          shared-key: coral-validate-${{ runner.os }}
           cache-on-failure: true
           save-if: ${{ github.ref == 'refs/heads/main' }}
 
@@ -504,7 +504,7 @@ jobs:
 
       - name: Benchmark coral.tables query
         run: |
-          cargo run --locked -p xtask -- perf-check \
+          cargo run --locked -p xtask --release -- perf-check \
             --coral-bin target/release/coral \
             --max-mean-seconds 0.75
 
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -33,6 +33,7 @@ jobs:
       source-lint: ${{ steps.filter.outputs.source-lint }}
       docs-freshness: ${{ steps.filter.outputs.docs-freshness }}
       license-check: ${{ steps.filter.outputs.license-check }}
+      performance-regression: ${{ steps.filter.outputs.performance-regression }}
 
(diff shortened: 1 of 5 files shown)


antonmry (changes requested):
scripts folder is used only for the bash coral installer. In general, we are using xtask for automation like this. It's going to be more verbose, but if possible, I think it's better to be consistent and keep using xtask (or migrate all automation to a different approach if it is a limiting factor)

jsummerfield:
Addressed, thanks. I moved the benchmark harness into `xtask perf-check` and removed the extra file under `scripts/`; `make perf-check` and the Validate job now call xtask. I also added a root AGENTS note to keep general automation in xtask and reserve `scripts/` for the installer.\n\nWhile updating the PR, I also fixed the no-token install validation failure by making `scripts/install.sh` fall back from the GitHub latest-release API to the public `/releases/latest` redirect, which keeps the PR trust-boundary hardening intact.

jsummerfield:
Correction: I removed the install-script fallback from this PR since that CI issue is being handled separately. This branch now only keeps the performance gate plus the xtask migration / scripts-directory guidance.

jsummerfield:
Adding Simon for docs codeowners, will refactor the xtasks/ dir in another PR so we're not blocking on non-docs changes.

jsummerfield:
Updated the perf job cache behavior: it now restores the existing `coral-validate-${{ runner.os }}` Rust cache namespace instead of a new cold `coral-performance-*` key, and it runs `xtask perf-check` with `--release` so it reuses the release profile from the CLI build rather than compiling xtask and dependencies in dev mode.

antonmry (approved):
Great! Big fan of using CI to detect performance bugs

simonwhitaker (approved):

PR #995, opened 2026-05-29
feat(sources/community/appsignal): add AppSignal source

SaaiAravindhRaja:
## Summary
- Add a community source spec for AppSignal.

## Why
This makes AppSignal data available through Coral's local SQL source workflow.

## User-visible impact
Coral users can add the AppSignal source and query its data through SQL.

## Known limitations
- Requires access to the relevant AppSignal account or API data documented by the source.

## Validation
- `coral source lint sources/community/appsignal/manifest.yaml`
- `make lint-sources`
- `yamllint sources/community/appsignal/manifest.yaml`


jsummerfield (changes requested):
Findings:

- High: `sources/community/appsignal/manifest.yaml:94` and `sources/community/appsignal/manifest.yaml:194` model paginated AppSignal list fields as fixed first-page reads. Both tables hard-code `limit: 100` and set `pagination.mode: none`, while AppSignal documents `deployMarkers(limit, offset, start, end)` and `exceptionIncidents(limit, offset, state, order, marker, namespaces)`. As written, `ORDER BY ... LIMIT 200` can still query at most the first 100 provider rows. Please remove the literal limit, wire limit/offset through pagination where possible, and expose the documented time/state/order filters. Docs: https://docs.appsignal.com/api/graphql/examples.html

- Medium: `sources/community/appsignal/manifest.yaml:193` exposes `exception_incidents` with no provider-side filters. That makes incident triage weak because users cannot ask for open incidents, one deploy marker, one state, or provider-defined ordering without fetching an arbitrary first page and filtering locally. Please add filters such as `state`, `order`, `marker_id`, and namespace where the DSL can model them, or make this a table function if bounded parameters are required for sane use. Docs: https://docs.appsignal.com/api/graphql/examples.html

- Medium: `sources/community/appsignal/manifest.yaml:294`-`:315` selects only a partial uptime monitor/alert shape, while the README advertises alert payloads. AppSignal's examples include fields such as notifier IDs, headers, status pages, metric digest, timeframe fields, and alert tags; because GraphQL returns only selected fields, `raw_json` cannot recover omitted data. Please either select and expose those documented fields, or narrow the README/table description to say this is a partial monitor view. Docs: https://docs.appsignal.com/api/graphql/examples.html

- Medium: `sources/community/appsignal/manifest.yaml:26`-`:28` only tests `app_overview` and `exception_incidents`; `deploy_markers` and `uptime_monitors` are untested. Please add test queries for all four tables and include redacted `coral source add --file ...` / `coral source test appsignal` output in the PR notes.

- Low: `sources/community/appsignal/README.md:6`-`:12` and `sources/community/appsignal/manifest.yaml:11`-`:16` should tell users where to get the API token and app ID. Please mention personal API token setup and that the app ID is visible in app settings or the `/sites/<app_id>` URL. Docs: https://docs.appsignal.com/api/graphql-explorer.html and https://docs.appsignal.com/api.html

AppSignal is useful Coral product surface, but the current first-page behavior and partial GraphQL selections would surprise users. Thanks for the source; I can re-check after those API semantics and live evidence are tightened up.

jsummerfield (changes requested):
Findings

- High: `sources/community/appsignal/manifest.yaml:101` declares `$start` and `$end` as `ISO8601DateTime`, but AppSignal's documented GraphQL examples use `DateTime` for `deployMarkers`. `sources/community/appsignal/manifest.yaml:219` also sends `state` and `order` as `String` even though the documented incident query uses enum types for those arguments. That makes these queries likely to fail GraphQL validation at runtime. Please align the query variable types with AppSignal's documented schema. Docs: https://docs.appsignal.com/api/graphql/examples.html

- High: `sources/community/appsignal/manifest.yaml:131` and `sources/community/appsignal/manifest.yaml:253` still set `pagination.mode: none` for the two offset-capable list fields. The fixed first-page literal limit is gone, but Coral still will not advance through AppSignal offsets automatically, so larger inventory queries silently stop at the first provider page. If the current DSL cannot automate GraphQL offsets here, please either model these as explicitly bounded query surfaces or make the first-page/manual-offset limitation impossible to miss in the table shape and README.

- Medium: `sources/community/appsignal/README.md:73`-`87` still does not include credentialed live Coral evidence; it explicitly says `coral source add`, `coral source test`, and representative live queries were not run. Static lint passes locally, but this source depends on provider GraphQL query typing and runtime response shape, so the mandatory live-evidence gate is still blocking.

Some of the earlier issues are improved: setup docs, test-query coverage, provider-side filters, and the partial uptime-monitor wording are better now. The remaining work is to prove the GraphQL queries run against AppSignal and avoid surprising first-page-only behavior.

jsummerfield:

PR #997, opened 2026-05-29
feat(sources/community/bluesky): add Bluesky source

SaaiAravindhRaja:
## Summary
- Add a community source spec for Bluesky.

## Why
This makes Bluesky data available through Coral's local SQL source workflow.

## User-visible impact
Coral users can add the Bluesky source and query its data through SQL.

## Known limitations
- Requires access to the relevant Bluesky account or API data documented by the source.

## Validation
- `coral source lint sources/community/bluesky/manifest.yaml`
- `make lint-sources`
- `yamllint sources/community/bluesky/manifest.yaml`


jsummerfield (changes requested):
Findings
- High: `sources/community/bluesky/manifest.yaml:18` only defines two source test queries, and the PR evidence lists lint/static checks rather than captured live Coral output. For a no-auth public source this should be straightforward to prove. Please add output from `coral source add --file sources/community/bluesky/manifest.yaml`, `coral source test bluesky`, and at least one successful query for each exposed function, or explicitly justify any untested function.
- High: `sources/community/bluesky/README.md:16` and `sources/community/bluesky/manifest.yaml:14` recommend `https://public.api.bsky.app/xrpc`, but `app.bsky.feed.searchPosts` returned 403 there while `https://api.bsky.app/xrpc/app.bsky.feed.searchPosts?q=bluesky&limit=1` returned 200. Since `search_posts` is the first test query, users following that host guidance can break validation. Please remove the public-host recommendation unless the full source test passes against it, or document the limitation clearly. Docs: https://docs.bsky.app/docs/api/app-bsky-feed-search-posts
- Medium: `sources/community/bluesky/manifest.yaml:732` extracts `threadgate` from the response root, but `app.bsky.feed.getPostThread` returns the thread object under `thread`, and threadgate belongs inside that thread view. Please change the expression to `path: [thread, threadgate]`. Docs: https://docs.bsky.app/docs/api/app-bsky-feed-get-post-thread
- Medium: `sources/community/bluesky/manifest.yaml:59` exposes `search_posts.tag` as a single scalar argument, while the official `app.bsky.feed.searchPosts` parameter is repeatable/array-shaped. Please either support repeated tag values if Coral can encode them, or document that this source supports only one tag filter. Docs: https://docs.bsky.app/docs/api/app-bsky-feed-search-posts

Thanks for the broad public Bluesky coverage. The main fixes are proving the source through live Coral output and avoiding host guidance that breaks the first test path.

jsummerfield (changes requested):
Findings

- Medium: `sources/community/bluesky/manifest.yaml:778`-`:782` maps `post_thread.threadgate` from `[thread, threadgate]`, but `app.bsky.feed.getPostThread` exposes `threadgate` at the response root. Live output with a threadgate also returns it at top level, not under `thread.threadgate`, so this column silently comes back empty when metadata exists. Please map it from `[threadgate]` unless the intended value is specifically the nested post-view threadgate. Docs: https://docs.bsky.app/docs/api/app-bsky-feed-get-post-thread and https://endpoints.bsky.app/openapi.bluesky-app.json.

The prior live-output blocker is resolved: the README now includes captured no-auth `coral source add` / `coral source test` output, and the focused review reproduced the no-auth test suite on this head. One small non-blocking follow-up while you are here: `quotes(uri)` could expose the optional `cid` query parameter supported by `app.bsky.feed.getQuotes` (https://docs.bsky.app/docs/api/app-bsky-feed-get-quotes).

Thanks for adding the live evidence; this is close once the threadgate path is fixed.

jsummerfield:

PR #1000, opened 2026-05-29
feat(sources/community/cockroachdb): add CockroachDB source

SaaiAravindhRaja:
## Summary
- Add a community source spec for CockroachDB.

## Why
This makes CockroachDB data available through Coral's local SQL source workflow.

## User-visible impact
Coral users can add the CockroachDB source and query its data through SQL.

## Known limitations
- Requires access to the relevant CockroachDB account or API data documented by the source.

## Validation
- `coral source lint sources/community/cockroachdb/manifest.yaml`
- `make lint-sources`
- `yamllint sources/community/cockroachdb/manifest.yaml`


jsummerfield (changes requested):
Findings:

- High: `sources/community/cockroachdb/manifest.yaml:249` maps `database_tables` to `rows_path: [tables]`, but CockroachDB returns `table_names`, an array of schema-qualified strings. This table will come back empty or unusable. Please use `rows_path: [table_names]`, expose the table name from the row value, and update the README example to show the schema-qualified value returned by the API. Docs: https://www.cockroachlabs.com/docs/api/cluster/v2.html and https://www.cockroachlabs.com/docs/api/spec_30_enriched.json

- High: `sources/community/cockroachdb/manifest.yaml:135` treats `/databases/` rows as objects with `database_name`, `name`, and `table_count`, but the API returns `databases: string[]`. The test can pass while useful columns are null. Please model this as one database-name column from the string row value and add the documented pagination. Docs: https://www.cockroachlabs.com/docs/api/cluster/v2.html

- High: `sources/community/cockroachdb/manifest.yaml:93`-`:118` models `nodes` with the wrong response shape. The API exposes `address.address_field`, `sql_address.address_field`, top-level `build_tag`, and `started_at` as Unix nanoseconds; it does not expose `build_info.tag`, `build_info.go_version`, or ISO-8601 `started_at`. Please fix the paths, expose `started_at` as raw nanoseconds or drop the Timestamp column until it can be converted correctly, and change `manifest.yaml:80` from cursor pagination to offset pagination because the API's `next` token is numeric. Docs: https://www.cockroachlabs.com/docs/api/cluster/v2.html

- Medium: `sources/community/cockroachdb/manifest.yaml:364` omits documented event pagination and the `targetID` filter. `/events/` supports `type`, `targetID`, `limit`, and `offset`; this source only sends `type` and disables pagination. Please add `target_id` mapped to `targetID`, use offset pagination, and make `target_id` / `reporting_id` numeric. Also `info_json` should be `Utf8` unless the provider string payload is explicitly parsed.

- Medium: `sources/community/cockroachdb/manifest.yaml:39` exposes `health` as a JSON table with `status` and `message`, but the official API spec documents only 200/500 responses for `/health/` and no JSON response schema. Coral's HTTP backend expects JSON rows, so an empty 200 body can fail before extraction. Please drop this table or use a source-spec pattern that can safely expose status-only endpoints.

- Medium: setup docs and live evidence are incomplete. `sources/community/cockroachdb/README.md:15` and `sources/community/cockroachdb/manifest.yaml:17` should mention the admin-role requirement for authenticated Cluster API access, and the PR needs redacted output from `coral source add --file ...`, `coral source test cockroachdb`, and representative queries against a real CockroachDB node. Docs: https://www.cockroachlabs.com/docs/stable/cluster-api

CockroachDB is a useful source candidate, but several current tables do not match the documented API contract. Thanks for the contribution; once those response shapes are fixed I can re-check the updated head.

jsummerfield (changes requested):
Findings

- Medium: `sources/community/cockroachdb/manifest.yaml:350` and `sources/community/cockroachdb/manifest.yaml:355` map event IDs from `targetID` and `reportingID`, but CockroachDB's documented `listEvents` response uses `target_id` and `reporting_id`. Against the official-shaped payload those two columns come back null. Please change the expressions to `[target_id]` and `[reporting_id]`, and add an `events` test query so this stays covered. Docs: https://www.cockroachlabs.com/docs/api/cluster/v2.html#operation/listEvents.

- Medium: `sources/community/cockroachdb/README.md:106`-`:108` says credentialed `coral source add`, `coral source test cockroachdb`, and representative live queries were not run. The previous `database_tables` blocker is fixed, but approval still needs captured live Coral output from a real CockroachDB admin API session or a maintainer decision to waive that standard.

- Low: Several fields appear to be undocumented compatibility guesses and will be always-null against the official response schemas, such as `database_details.id` / `name`, `database_grants.grantee`, and `users.user_name` / `is_role` / `options_json`. Please either remove those or clearly justify them with version-specific docs. Relevant refs: https://www.cockroachlabs.com/docs/api/cluster/v2.html#operation/databaseDetails, https://www.cockroachlabs.com/docs/api/cluster/v2.html#operation/databaseGrants, and https://www.cockroachlabs.com/docs/api/cluster/v2.html#operation/listUsers.

The original `database_tables` `rows_path` issue looks resolved: it now reads `table_names`. Thanks for fixing that; the remaining event mapping is smaller but still a correctness issue.

jsummerfield:

PR #1004, opened 2026-05-29
feat(sources/community/knock): add Knock source

SaaiAravindhRaja:
## Summary
- Add a community source spec for Knock.

## Why
This makes Knock data available through Coral's local SQL source workflow.

## User-visible impact
Coral users can add the Knock source and query its data through SQL.

## Known limitations
- Requires access to the relevant Knock account or API data documented by the source.

## Validation
- `coral source lint sources/community/knock/manifest.yaml`
- `make lint-sources`
- `yamllint sources/community/knock/manifest.yaml`


jsummerfield (changes requested):
Findings:

- High: `sources/community/knock/manifest.yaml:137` and `sources/community/knock/manifest.yaml:141` expose `workflow_run_id` and `workflow_recipient_run_id` as top-level message fields, but Knock's Message object nests both under `source`. These columns will be null even when the API returns the values. Please read them from `source.workflow_run_id` and `source.workflow_recipient_run_id` instead. Docs: https://docs.knock.app/api-reference/messages

- High: live Coral evidence is missing. `sources/community/knock/README.md:14` asks users to run `coral source test knock`, and the manifest includes message child-table tests, but I only see static validation in the PR. Please add redacted output from `coral source add --file sources/community/knock/manifest.yaml`, `coral source test knock`, and at least one representative query covering `messages`, `users` or `tenants`, plus one message child table.

Bearer auth, cursor pagination, users/tenants `entries`, and message `items` shapes otherwise look consistent with Knock's docs. Thanks for the source; once the nested fields and live evidence are added I can re-check the current head.

SaaiAravindhRaja:
@jsummerfield noted!

jsummerfield:

- Be precise about Knock authentication, environment selection, and any required API key permissions.
- Make the source useful for operational notification workflows: workflows, recipients/users, messages, and delivery status should be clear.
- Clarify pagination, rate-limit behavior, and how recent message/event windows should be queried.
- Keep the first-success path simple with one small query that proves the token and environment are correct.

jsummerfield (changes requested):
Findings

- Medium: `sources/community/knock/manifest.yaml:46` does not expose a provider-backed message ID lookup. A query like `WHERE id = 'msg_...'` becomes a residual Coral filter over the fetched page, so users can get false negatives for an existing message. Knock's list messages endpoint supports `message_ids[]` up to 50 IDs; please add a pushed `message_id` / `message_ids` filter or a point lookup. Docs: https://docs.knock.app/api-reference/messages/list

- Medium: `sources/community/knock/README.md:48` recommends `WHERE status != 'delivered'`. Knock supports positive `status[]` filtering, while this negative predicate will be applied locally after a limited provider page is fetched. Please use pushed positive examples such as `status = 'undelivered'` or `status = 'bounced'` so users do not miss older failures. Docs: https://docs.knock.app/api-reference/messages/list

- Medium: `sources/community/knock/README.md:10` says to use a key with read access, but Knock documents secret `sk_` keys as able to perform any API request and scoped only by environment. Please say to use a secret key for the target environment and avoid implying Knock has read-only/scoped API keys for this source. Docs: https://docs.knock.app/api-reference/overview/api-keys

- Medium: `sources/community/knock/README.md:115` still says credentialed `source add`, `source test`, and representative live queries were not run. Static validation is useful, but live Coral output is required before this can be approved. Please include redacted output for `messages`, either `users` or `tenants`, and one message child table.

The previous nested `source.workflow_*` blocker looks resolved, and the pagination shape otherwise lines up with Knock's docs. Thanks for tightening this up; the remaining changes are about avoiding false negatives and giving users a trustworthy first-success path.

SaaiAravindhRaja:
@jsummerfield noted! working on all 3 of your comments now :)

jsummerfield (approved):

The previous Knock blockers are resolved: the workflow run fields are now read from the nested `source` object, provider-backed message ID lookup is mapped through Knock's `message_ids[]` filter, the manifest lints, live Coral evidence is present, and the paused workflows are now green for this head.

Thanks for the cleanup.

PR #1006, opened 2026-05-29
feat(sources/community/productboard): add Productboard source

SaaiAravindhRaja:
## Summary
- Add a community source spec for Productboard.

## Why
This makes Productboard data available through Coral's local SQL source workflow.

## User-visible impact
Coral users can add the Productboard source and query its data through SQL.

## Known limitations
- Requires access to the relevant Productboard account or API data documented by the source.

## Validation
- `coral source lint sources/community/productboard/manifest.yaml`
- `make lint-sources`
- `yamllint sources/community/productboard/manifest.yaml`


SaaiAravindhRaja:
@cubic-dev-ai do you have any minor suggestions or nitpicks?

jsummerfield (changes requested):
Findings:

- High: `sources/community/productboard/manifest.yaml:83`, `sources/community/productboard/manifest.yaml:210`, `sources/community/productboard/manifest.yaml:243`, `sources/community/productboard/manifest.yaml:268`, `sources/community/productboard/manifest.yaml:293`, `sources/community/productboard/manifest.yaml:355`, `sources/community/productboard/manifest.yaml:557`, `sources/community/productboard/manifest.yaml:621`, and `sources/community/productboard/manifest.yaml:673` disable pagination on Productboard list endpoints. Productboard documents these as cursor-paginated via `links.next` / `pageCursor`, so the source will silently return only the first page while presenting table-wide access. Please implement pagination for each paginated table, or document and constrain the tables so users cannot mistake first-page results for complete data. Docs: https://developer.productboard.com/reference/listentities, https://developer.productboard.com/reference/listnotes, https://developer.productboard.com/reference/listmembers, https://developer.productboard.com/reference/listteams, and https://developer.productboard.com/reference/listmemberactivities

- High: `sources/community/productboard/manifest.yaml:407` and `sources/community/productboard/manifest.yaml:415` read note timestamps from `fields.createdAt` / `fields.updatedAt`, but Productboard's List notes response puts `createdAt` and `updatedAt` at the note top level. These columns will be null, including in README examples that order by timestamps. Please change the paths to the top-level fields. Docs: https://developer.productboard.com/reference/listnotes

- High: `sources/community/productboard/manifest.yaml:564`-`:592` and `sources/community/productboard/manifest.yaml:628`-`:640` map member/team fields as top-level properties, but the Productboard members and teams responses place those values under `fields`. Please add explicit paths such as `[fields, email]`, `[fields, invitationPending]`, `[fields, teams]`, and `[fields, avatarUrl]`. Docs: https://developer.productboard.com/reference/listmembers and https://developer.productboard.com/reference/listteams

- Medium: `sources/community/productboard/README.md:49` uses `source_system` in SQL, but `sources/community/productboard/manifest.yaml:307` declares it only as a filter and never exposes it as a column. Coral only exposes manifest columns, so this example is not valid as written. Please either add virtual `from_filter` columns for filter values users should select, or rewrite the examples to select real columns.

- Medium: auth and live evidence need tightening. `sources/community/productboard/README.md:73`-`:76` and `sources/community/productboard/manifest.yaml:15`-`:17` should distinguish Productboard API tokens from OAuth access tokens, document where to create them, and list exact scopes/PII requirements for the exposed tables. Please also add redacted output from `coral source add --file ...`, `coral source test productboard`, and representative queries for the main tables.

Productboard is useful Coral product surface, but the current pagination and response-path issues will produce incomplete or null data. Thanks for the broad coverage here; once those provider semantics are fixed I can re-check the updated head.

jsummerfield:

- Make Productboard token setup explicit, including where the token is created and which workspace/API access is required.
- Keep the core product workflows obvious: features, companies, notes/feedback, and roadmap-style queries should be easy to discover.
- Document pagination, API limits, and any fields that are not stable across Productboard account plans.
- Include a simple first-success query against a real workspace so users can verify auth and table shape quickly.

jsummerfield (changes requested):
Findings

- High: `sources/community/productboard/manifest.yaml:89`, `:221`, `:258`, `:287`, `:316`, `:383`, `:598`, `:673`, and `:733` still disable pagination on Productboard list endpoints. Productboard documents these endpoints as cursor-paginated via `links.next` / `pageCursor`, so these tables will silently return only the first provider page while looking like full SQL tables. Please implement cursor pagination, or expose/manualize `pageCursor` so users can page intentionally instead of accidentally treating first-page samples as complete data. Docs: https://developer.productboard.com/reference/pagination, https://developer.productboard.com/reference/listentities, https://developer.productboard.com/reference/listnotes, https://developer.productboard.com/reference/listmembers, https://developer.productboard.com/reference/listteams, https://developer.productboard.com/reference/listmemberactivities

- Medium: `sources/community/productboard/manifest.yaml:736` exposes `member_activities` as an analytics table but only types a small subset of the activity metrics before pushing the rest into `raw_json`. Productboard returns core counts such as board, component, insight, product, subfeature, note-state-change, and board-type activity metrics; those should be first-class columns if this table is meant to be useful for reporting. Docs: https://developer.productboard.com/reference/listmemberactivities

- Medium: `sources/community/productboard/README.md:118` provides static validation only, and `README.md:130` says credentialed `source add`, `source test`, and representative live queries were not run. Productboard auth, redaction, pagination, and response shape are all provider-specific enough that live Coral output is required before approval.

The previous response-path issues look materially improved, and Productboard is the right kind of community source. The pagination behavior is still a blocker because it changes the SQL contract users will trust.

jsummerfield (changes requested):
Findings

- High: `sources/community/productboard/manifest.yaml:94`, `:235`, `:418`, `:643`, `:727`, and `:797` still use `rows_path: [data]` with `pagination: mode: none`; the `page_cursor` columns/filters only echo the cursor a user already supplied. Productboard returns the next cursor inside the top-level `links.next` URL, not inside each `data` row, so Coral users cannot discover the next cursor from query results. The README now tells users to copy `pageCursor` from Productboard's `links.next` URL, but the source does not expose that URL or cursor through Coral. Please either make pagination work automatically or expose a real next-cursor surface users can query. Productboard docs: https://developer.productboard.com/reference/pagination, https://developer.productboard.com/reference/listentities, https://developer.productboard.com/reference/listnotes, https://developer.productboard.com/reference/listmembers, https://developer.productboard.com/reference/listteams, and https://developer.productboard.com/reference/listmemberactivities.

- Medium: `sources/community/productboard/README.md:13`-`:15` gives exact PII scope examples as `members:pii:read` and `users:pii:read`. Productboard's OAuth scope list currently uses names like `members_pii:read`, and I could not find `users:pii:read` in that scope list. If endpoint docs and OAuth scope docs disagree, please avoid asserting exact PII scope names here or call out that users should verify the current scope names in Productboard's OAuth docs.

The response-path and member-activity improvements look materially better, and the manifest now lints, but the original pagination blocker is not resolved in a way Coral users can actually operate. Thanks for the continued work on this.

jsummerfield:

PR #1017, opened 2026-05-29
feat(search): add universal search contracts

mystic123:
## Summary
- Adds the Universal Search protobuf API and generated client types.
- Defines typed search results, provider statuses, and truncation metadata.
- Uses each result's `oneof` payload as the result-kind discriminator instead of carrying a duplicate result-type enum.
- Adds `SearchResult.provider` so clients can correlate individual results with provider statuses and partial failures.
- Wraps catalog item search hits in `CatalogMetadata` so consumers receive matched metadata fields and compact table-column previews without parsing prose.
- Threads table-function metadata through typed API fields: `TableFunctionKind` for the function role and `SearchLimits` for provider search limits.
- Preserves field-role provenance for column hints and surface-kind provenance for observed values.
- Adds observed-value `field_path`, `observed_count`, and `last_observed_at` fields so downstream consumers can render extracted JSON/key-value fields and freshness/count metadata without string parsing.

## Stack Position
Part 1 of the current Universal Search stack:
1. #1017 Universal Search contracts
2. #1018 Tantivy catalog backend
3. #1019 observed-value indexing and provider
4. #1020 MCP `search` tool
5. #1021 CLI `coral search`
6. #1180 provider-native fanout

## Validation
Validated while restacking this layer:
- `cargo check -p coral-api -p coral-app`
- `cargo test -p coral-app transport::tests::table_function_to_proto --lib`

Additional top-stack validation is listed on #1180.

PR #1018, opened 2026-05-29
feat(search): add tantivy catalog backend

mystic123:
## Summary
- Adds workspace-scoped Tantivy storage for Universal Search catalog metadata.
- Builds catalog metadata into the local search index lazily on demand, not during server startup.
- Reuses the catalog index only when the persisted catalog fingerprint matches the freshly loaded workspace catalog and the Tantivy schema is usable.
- Marks catalog metadata dirty on source create/import/update/delete without waiting on the refresh lock; the next search rebuilds the catalog projection.
- Loads catalog metadata for search from stored source manifests with persisted credential material, while still avoiding credential/OAuth refresh, so required-secret sources participate in indexing.
- Adds exact fallback matching for short indexed tokens that the trigram tokenizer cannot emit.
- Indexes table-column metadata into table-level searchable text and returns compact table-column previews on catalog table hits.
- Normalizes Tantivy hit order into a fixed 1..1000 rank score before applying additive boosts.
- Keeps source-name and canonical surface-name boosts scaled to one or two rank bands.
- Exposes the gRPC SearchService catalog provider with per-result provider provenance and typed catalog metadata, column-hint, and native-search-path results.
- Reports observed-values provider coverage as `NotEnabled` in this catalog-only PR.

## Stack Position
Part 2 of the current Universal Search stack:
4. #1020 MCP `search` tool
5. #1021 CLI `coral search`
6. #1180 provider-native fanout

## Validation
Validated for this branch surface:
- `cargo fmt --check`
- `cargo check -p coral-api -p coral-app`
- `cargo test -p coral-app search::index::tests:: --lib`
- `cargo test -p coral-app search::service::tests:: --lib`
- `cargo test -p coral-app --test grpc search_`

## Notes
- Observed-value ingestion, observed-value search results, and CLI smoke examples are covered by later stack layers, not this catalog-only backend PR.


mystic123:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/withcoral/coral/1018?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1018** <a href="https://app.graphite.com/github/pr/withcoral/coral/1018?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1018?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-app/src/query/manager.rs
```diff
@@ -234,6 +248,36 @@ impl QueryManager {
...
+    fn load_stored_query_sources(
+        &self,
+        workspace_name: &WorkspaceName,
+    ) -> Result<Vec<QuerySource>, AppError> {
+        let catalog = self.config_store.load_catalog()?;
+        let mut query_sources = Vec::new();
+        for source in catalog.workspace_sources(workspace_name) {
+            match self.load_stored_query_source(workspace_name, &source) {
+                Ok(query_source) => query_sources.push(query_source),
+                Err(error) => {
```
jsummerfield: [P2] This catch-all still lets search build and report against a partial catalog. The stored-secret happy path is fixed now, but unavailable keychain storage, missing stored material, or a manifest load failure still just logs and skips the source; `SearchService` then reports normal `Empty`/`ResultsFound` coverage and may refresh the Tantivy index as if the catalog were complete. Please propagate hard failures, or carry skipped-source diagnostics through so the catalog provider can return `PARTIAL`/`ERROR` instead of silently dropping installed sources.

crates/coral-app/src/search/service.rs
```diff
@@ -0,0 +1,1230 @@
...
+
+fn sql_call_example(function: &TableFunctionInfo) -> String {
+    let args = function
+        .arguments
+        .iter()
+        .filter(|argument| argument.required)
+        .map(|argument| format!("{} => '<{}>'", argument.name, argument.name))
+        .collect::<Vec<_>>();
+    format!(
+        "SELECT * FROM {}.{}({}) LIMIT 10",
```
jsummerfield, line 1146: [P2] This SQL example needs to quote identifiers. Coral supports case-preserving / quoted manifest identifiers, and the engine resolves unquoted SQL identifiers by normalizing them, so a function like schema `Search`, function `Search_Issues`, arg `Q` would be returned as `SELECT * FROM Search.Search_Issues(Q => '<Q>')`, which resolves to the wrong lower-case identifiers. Please quote schema, function, and named argument identifiers when generating this example.

jsummerfield (commented):
Leaving the verified inline comments from the refreshed search backend review.

crates/coral-app/src/search/service.rs
```diff
@@ -0,0 +1,1230 @@
...
+        let catalog = self
+            .queries
+            .list_stored_catalog(workspace_name, None)
+            .await?;
+        let (mut candidates, catalog_status) = self
+            .catalog_metadata_candidates(workspace_name, &catalog, &terms, limit)
+            .await;
+        candidates.sort();
+
+        let total_count = candidates.len();
```
jsummerfield, line 261: [P2] `truncated` is still computed from the size of a bounded overfetch window, not from the number of available search results. Since Tantivy only receives `limit * 5` hits below, anything past that window is invisible here; after dedupe/capping this can return `truncated=false` even when more matches exist, which contradicts the proto contract that says this means more results were available than returned.

crates/coral-app/src/search/service.rs
```diff
@@ -0,0 +1,1230 @@
...
+            provider_statuses: vec![
+                SearchProviderStatus {
+                    provider: SearchProvider::CatalogMetadata as i32,
+                    state: catalog_status.state as i32,
+                    note: catalog_status.note,
+                },
+                SearchProviderStatus {
+                    provider: SearchProvider::ObservedValues as i32,
+                    state: SearchProviderState::NotEnabled as i32,
+                    note: "Observed-value search is not enabled in this PR".to_string(),
```
jsummerfield: [P3] This note is returned in the runtime API response, so it should not refer to "this PR". Please make this product/runtime wording, e.g. that observed-value search is not enabled for this provider or build.

crates/coral-app/src/search/index.rs
```diff
@@ -0,0 +1,1129 @@
...
+            self.fields.surface_name,
+            self.fields.name,
+            self.fields.data_type,
+        ];
+        terms
+            .iter()
+            .filter(|term| term.chars().count() < 3)
+            .flat_map(|term| {
+                fields.into_iter().map(move |field| {
+                    Box::new(TermQuery::new(
```
jsummerfield: [P3] Short identifier matching is case-sensitive here even though queries are normalized to lowercase before reaching the index. For one- or two-character terms, the trigram path is skipped and these exact `STRING` fields store original casing, so query `id` will miss a valid column or argument named `ID`. Please normalize the stored exact fields or query both original and lowercase variants.

crates/coral-app/src/search/service.rs
mystic123: Fixed in b9db66d7 by always quoting schema, function, and named argument identifiers in `sql_call_example`. I also added a unit regression for the case-preserving `Search.Search_Issues(Q)` shape and updated the gRPC native-search expectation to the quoted SQL form.

crates/coral-app/src/search/service.rs
mystic123: Fixed in b9db66d7 by making `search_catalog_page` fetch one extra Tantivy hit and carry a backend `has_more` bit through catalog provider status. `SearchResponse.truncation.truncated` now stays true when the backend retrieval window saturated, even if dedupe/capping leaves the visible candidates under the request limit. Added an index regression for the paged `has_more` signal.

crates/coral-app/src/query/manager.rs
```diff
@@ -234,6 +248,35 @@ impl QueryManager {
...
+    fn load_stored_query_sources(
+        &self,
+        workspace_name: &WorkspaceName,
+    ) -> Result<Vec<QuerySource>, AppError> {
+        let catalog = self.config_store.load_catalog()?;
+        let mut query_sources = Vec::new();
+        for source in catalog.workspace_sources(workspace_name) {
+            let query_source = self
+                .load_stored_query_source(workspace_name, &source)
+                .map_err(|error| {
```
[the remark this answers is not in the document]
mystic123: This fail-closed behavior is intentional for the catalog refresh path. The earlier stale/partial-catalog thread asked us to either propagate hard failures or carry skipped-source diagnostics through as PARTIAL/ERROR; this PR takes the hard-failure option so SearchService cannot refresh Tantivy and report complete coverage from an incomplete installed-source catalog. `search_fails_when_a_stored_source_cannot_load` covers this. A softer partial-source response would need explicit skipped-source diagnostics in the search contract rather than silently reintroducing partial indexing.

crates/coral-app/src/query/manager.rs
mystic123: Addressed in b9db66d7 by removing the catch-all skip in stored catalog loading. A stored-source load error now propagates before SearchService can refresh the Tantivy catalog index or report normal provider coverage from a partial catalog. Added `search_fails_when_a_stored_source_cannot_load` for the missing stored material case.

crates/coral-app/src/search/service.rs
mystic123: Addressed in b9db66d7. The runtime provider note now says `Observed-value search is not enabled in this build`, so the API response no longer references PR context.

crates/coral-app/src/search/index.rs
mystic123: Addressed in b9db66d7. Short exact-term queries now include lower-case, upper-case, and title-case variants for one- and two-character identifiers, and `catalog_tantivy_supports_case_preserving_short_identifier_matches` covers query `id` matching a stored `ID` result column.

crates/coral-app/src/search/service.rs
```diff
@@ -0,0 +1,1294 @@
...
+            .list_stored_catalog(workspace_name, None)
+            .await?;
+        let (mut candidates, catalog_status) = self
+            .catalog_metadata_candidates(workspace_name, &catalog, &terms, limit)
+            .await;
+        candidates.sort();
+
+        let total_count = candidates.len();
+        let max_results = usize::try_from(limit).unwrap_or(usize::MAX);
+        let truncated = total_count > max_results || catalog_status.has_more;
```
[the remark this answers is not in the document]
mystic123: Addressed in e3cdf953. Raw Tantivy page overflow now marks the catalog provider as `PARTIAL` with a bounded-window note instead of setting response-level `truncated=true`. `truncated` is only true when typed candidates exceed the response limit. Added `search_marks_catalog_partial_without_response_truncation_for_raw_index_overflow` to cover the dedupe/cap case.

crates/coral-app/src/search/index.rs
```diff
@@ -0,0 +1,1241 @@
...
+            table.guide.as_str(),
+            required_filters.as_str(),
+            columns.as_str(),
+        ]),
+    });
+
+    for column in &table.columns {
+        table_column_record(table, column, records);
+    }
+    for filter in &table.required_filters {
```
jsummerfield: Could we carry the full table filter metadata into the search index here? The Search API says `TABLE_FILTER` covers required or optional filters, but this loop only emits standalone filter hints from `required_filters`, and `CatalogInfo::TableInfo` has already dropped optional filters. Optional filters that are not also columns will never be suggested as filter hints.

crates/coral-app/src/search/index.rs
```diff
@@ -0,0 +1,1241 @@
...
+    }
+    Ok(Index::create_in_dir(path, search_schema())?)
+}
+
+fn index_is_usable(path: &Path) -> Result<bool, SearchIndexError> {
+    let meta_file = path.join("meta.json");
+    if !meta_file.exists() {
+        return Ok(false);
+    }
+    let index = Index::open_in_dir(path)?;
```
jsummerfield: Could we treat `Index::open_in_dir` failures as an unusable derived cache instead of returning the error from freshness checks? A malformed `meta.json` or missing segment file still prevents the rebuild path from running, even though the Tantivy index can be regenerated from catalog truth. The current stale-schema test covers an openable index, not a corrupt/unreadable one.

crates/coral-app/src/search/service.rs
```diff
@@ -0,0 +1,1270 @@
...
+            Err(error) => return (Vec::new(), catalog_index_error_status(&error)),
+        };
+        let capabilities = index.capabilities();
+        tracing::debug!(
+            workspace = %workspace_name,
+            tantivy_version = %capabilities.tantivy_version,
+            tokenizer = %capabilities.tokenizer,
+            "using Tantivy catalog search index"
+        );
+        let search_limit = usize::try_from(limit)
```
jsummerfield, line 315: This bounds the raw Tantivy candidate window before the app-level boosts run. If an exact source/surface/field match lands outside this raw page, the later scorer can never promote it even though `SearchResponse.results` is documented as ordered hints. Could we either move the important boosts into candidate generation or use a retrieval window/rerank strategy that preserves the final ordering contract?

jsummerfield (commented):
Leaving the actionable findings from the refreshed search backend review.

jsummerfield (approved):

crates/coral-app/src/search/index.rs
mystic123: Addressed in f13b6b44. `TableInfo` now carries `filters: Vec<TableFilterInfo>` from registered table metadata, and the catalog index writes every table filter, required or optional, as a `TABLE_FILTER` hint with `required`, `data_type`, and `description` preserved. It falls back to legacy `required_filters` only when richer filter metadata is absent, and it skips duplicate ordinary column records for declared filters. Added `catalog_tantivy_indexes_optional_non_column_filters` for an optional non-column `status` filter returned as `TABLE_FILTER` with `required=false`.

crates/coral-app/src/search/service.rs
mystic123: Addressed in f13b6b44. The important exact-field boosts now happen inside Tantivy candidate generation via boosted exact identifier term queries for schema/surface/name/qualified-name/data-type fields, so exact matches can enter `TopDocs` before app-level scoring runs. The optional-filter regression intentionally uses `limit = 1` to prove an exact `status` table-filter hint survives the raw retrieval window.

crates/coral-app/src/search/index.rs
mystic123: Addressed in f13b6b44. `Index::open_in_dir` failures are now treated as an unusable derived cache for freshness checks, so `catalog_index_is_fresh` falls into the rebuild path instead of returning the open error. `open_or_create_index` also logs, removes, and recreates a corrupt Tantivy directory. Added `open_workspace_recreates_corrupt_search_index` for the malformed cache case.

PR #1019, opened 2026-05-29
feat(search): index observed values from source scans

mystic123:
## Summary

This PR adds observed-value memory to Universal Search, but the implementation has been substantially refactored from the original final-SQL-result observer design.

The current shape indexes values from typed HTTP/MCP source-scan observations emitted by the shared `JsonExec` path, before SQL projection removes unselected fields. App-level observed-value indexing no longer uses final `QueryResultObserver` payloads, no longer reparses SQL for provenance, and no longer depends on `sqlparser`.

## What Changed

- Adds a shared `SourceObservationPublisher` / `SourceScanObservation` seam in `coral-engine`.
- Wires HTTP and MCP table/table-function scans through that shared source-observation path.
- Adds an app-owned `ObservedSourceScanIndexer` worker that converts source-scan batches into observed-value records.
- Adds bounded foreground drainage in the shared app query/search path, so CLI and MCP both get the same lifecycle behavior without MCP-specific logic.
- Keeps observed values enabled by default, with `search.observed_values_enabled = false` as the explicit disable switch.
- Adds workspace/source clear via `SearchService.ClearObservedValues`.
- Adds source/surface/column exclusion config for observed-value indexing.
- Adds last-observed retention: values stay eligible when re-observed and age out only after their `last_observed_at` becomes stale.
- Adds `source_scope_id` filtering based on the live source shape/configuration plus app-mediated credential generation.
- Strengthens HTTP source scopes to include request-shape inputs: base URL, auth/header shape, table filters, request routes, response mapping, pagination, function args, and function request/response/pagination.
- Makes workspace clear-all depend only on installed source names from config, not source manifest resolution, so broken/missing manifests do not block clearing local observed memory.
- Rechecks source generation while holding the search-index mutation lock immediately before enqueueing collected observations.
- Removes public observed-value `observed_count` and `last_observed_at` from the gRPC response; count/timestamps remain internal for upsert, ranking, and retention.
- Removes read-side stale cleanup/reconcile behavior from Universal Search. Search filters by live scope and staleness, while purge remains write-side maintenance.

## Why Source-Scan Observation

The original PR indexed final SQL output. That was too late for value memory: if a query selected only `name`, the observer could only learn `name`, even when Coral had already fetched and mapped other useful source fields such as `email`, `state`, or `url`.

Source-scan observation indexes values Coral actually received and mapped from HTTP/MCP source rows before SQL projection. A gRPC integration test now covers this explicitly: SQL returns only `name`, then Universal Search can find an unprojected `email` value from the same observed HTTP source scan.

This is still not raw-body indexing. The observed-value path consumes typed, declared, table-shaped values after source mapping and sensitivity filtering.

## Reviewer-Relevant Boundaries

- MCP stays thin. There is no MCP-adapter-specific indexing or drainage logic.
- Delivery is best-effort plus bounded foreground drainage, not a durable cross-process event bus.
- CLI and MCP both use the same app-layer drainage path.
- File-backed sources do not participate in observed-value memory in this version. The old final-result fallback was removed deliberately; files need their own typed observation point before being included.
- HTTP local exact-filter scans currently observe after local exact filtering but before SQL projection/residual filtering. The observation scope records this distinction.
- Account scoping is credential-install isolation, not true provider-account identity. Records are scoped to source shape/configuration plus credential generation; no raw secrets/tokens are hashed.
- Clear-all no longer loads source manifests. It reads installed source names from config only, marks those generations, and clears the local observed sidecar/projection.
- Source generation is checked before collection and again under the mutation lock before enqueue. If clear/reimport wins the race, stale collected observations are skipped or removed by the clear path.
- HTTP live-scope filtering now invalidates old observations when a table/function keeps the same name and columns but changes endpoint, request routing, filter binding, response mapping, or pagination.

## Governance and Safety

- `search.observed_values_enabled` defaults to `true`.
- `observed_value_excluded_sources`, `observed_value_excluded_surfaces`, and `observed_value_excluded_columns` allow app-owned exclusion policy.
- `SearchService.ClearObservedValues` clears observed memory for one source or the whole workspace.
- `observed_value_stale_after_days` is based on `last_observed_at`, so durable truths can stay alive when re-observed.
- Sensitivity filtering skips obvious credential, PII, payment, token-like, nested JSON, and key-value-bag values before they enter the observed-value queue.
- Observed search only returns records whose source scope matches the current installed source snapshot.

## Stack Position

Part 3 of the Universal Search stack:

4. #1020 MCP `search` tool
5. #1021 CLI `coral search`
6. #1180 provider fan-out orchestration

## Validation

Latest validation for commit `1447c0bc`:

- `cargo fmt --all`
- `cargo test -p coral-app search::observed -- --nocapture`
- `cargo test -p coral-app --test grpc search_tests:: -- --nocapture`
- `cargo clippy --locked -p coral-app -p coral-engine --all-targets -- -D warnings`
- `git diff --check`


mystic123:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/withcoral/coral/1019?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1019** <a href="https://app.graphite.com/github/pr/withcoral/coral/1019?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1019?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-app/src/search/observed.rs
```diff
@@ -0,0 +1,1102 @@
...
+}
+
+fn observed_records_from_batches(
+    schema: &Schema,
+    batches: &[RecordBatch],
+    provenance: &[Option<FieldProvenance>],
+) -> Result<Vec<ObservedValueRecord>, ObservedValueIndexError> {
+    let mut records = BTreeMap::<ObservedValueRecordKey, ObservedValueRecord>::new();
+    for batch in batches {
+        for row in record_batch_rows(batch)? {
```
jsummerfield, line 1185: This needs an explicit v1 budget before it runs on every successful `execute_sql`. As written, a broad/high-cardinality query can walk every returned row/field, expand nested JSON/arrays, accumulate one record per unique candidate, and synchronously commit redb + Tantivy writes before the SQL response returns. Could we cap rows/cells/candidates/value bytes/container depth and skip or sample once the budget is exhausted?

crates/coral-app/src/search/observed.rs
```diff
@@ -0,0 +1,1102 @@
...
+    surfaces: &[ObservedSurface],
+) -> Vec<Option<FieldProvenance>> {
+    let empty = vec![None; schema.fields().len()];
+    let Ok(statements) = Parser::parse_sql(&GenericDialect {}, sql) else {
+        return empty;
+    };
+    let [Statement::Query(query)] = statements.as_slice() else {
+        return empty;
+    };
+    let Some(select) = query.body.as_select() else {
```
jsummerfield: Can we skip CTE queries here, at least for v1? `WITH messages AS (...) SELECT service FROM messages` can currently resolve the outer `FROM messages` against an installed source table named `messages`, even though the values came from the CTE. That would persist derived values under the wrong source provenance.

crates/coral-app/src/search/observed.rs
```diff
@@ -0,0 +1,1102 @@
...
+            );
+            return Ok(());
+        }
+
+        let records = observed_records_from_batches(schema, batches, &provenance)?;
+        if records.is_empty() {
+            return Ok(());
+        }
+
+        let store = SearchIndexStore::open_workspace(&self.layout, &self.workspace_name)?;
```
jsummerfield, line 213: This observer-side open can replace an old-schema catalog index with an empty new-schema index, but it does not invalidate the catalog fingerprint. If the first post-upgrade action is `execute_sql`, the next search can treat that empty catalog index as fresh and return no catalog metadata. Could schema replacement either dirty/clear the catalog fingerprint or stay out of the observer path?

crates/coral-app/src/search/service.rs
```diff
@@ -336,13 +379,76 @@ impl UniversalSearch {
...
+        terms: &QueryTerms,
+        limit: u32,
+    ) -> (Vec<Candidate>, ObservedProviderStatus) {
+        let index = match SearchIndexStore::open_workspace(&self.indexes.layout, workspace_name) {
+            Ok(index) => index,
+            Err(error) => return (Vec::new(), observed_index_error_status(&error)),
+        };
+        let retention_cutoff = observed_value_retention_cutoff();
+        let purge_error =
+            match index.purge_observed_values_before(workspace_name, &retention_cutoff) {
```
jsummerfield: Search is doing retention cleanup inline before reading results. That makes the read path take the observed-value mutation lock and commit redb/Tantivy deletes, while the later `last_observed_at >= retention_cutoff` filter already protects freshness. Can cleanup move to write-time/startup/background/explicit maintenance so search stays retrieval + filtering?

jsummerfield (commented):
Left a few v1 safety-rail comments inline. The overall direction looks right to me; the blockers are about bounding, provenance conservatism, and lifecycle/privacy controls rather than needing a larger redesign.

jsummerfield:
This looks directionally right to me for a v1: observed-value search belongs in `coral-app`, and the `QueryResultObserver` seam is a reasonable place to start. I would not push this toward a larger engine-provenance redesign before landing an initial version.

The v1 bar I think we still need is mostly safety rails:

- Bound the indexing work. Passive indexing should have hard caps for rows/cells/candidates/value bytes/container depth before it runs on every successful SQL query.
- Keep provenance conservative. Skipping CTEs/ambiguous SQL is fine for v1; persisting values under the wrong source is not.
- Fix the index lifecycle edge case where observer-side opening can replace an old catalog index without invalidating the catalog fingerprint.
- Keep search read-only. Filtering stale observed values on read is enough for freshness; cleanup can happen elsewhere.
- Scope observed values to the access boundary that produced them. `source_name` alone is too coarse: values learned under one credential/account/tenant should not be suggested under another. I would key by workspace/source instance plus credential set or equivalent authorization context, or purge/ignore values when that context changes.
- Make the privacy model explicit. The secret-name heuristic is a useful guardrail, but it is not a privacy model: emails, issue titles, URLs, customer IDs, and long text can still be sensitive. For v1, I would want observed-value search to be feature-gated or otherwise clearable/disableable, with local retention documented.

So my read is: good product direction, but v1 should ship bounded, conservative, access-scoped, and explicit about local persistence.

crates/coral-app/src/search/observed.rs
mystic123: Addressed in a505a835. Observed writes still create/repair the search index so `execute_sql` followed by `search` keeps working, but they now first clear the catalog fingerprint whenever the workspace index is currently unusable. That prevents an observer-created empty/new-schema Tantivy index from being treated as fresh catalog metadata; the next search refresh rebuilds catalog docs and preserves observed values from the redb sidecar. Added `clears_catalog_fingerprint_when_building_search_index` to cover the SQL-first path.

crates/coral-app/src/search/service.rs
mystic123: Addressed in a505a835. Observed-value search no longer calls `purge_observed_values_before` on the read path. It opens the existing search index without creating/repairing it, retrieves hits, and filters by `last_observed_at >= retention_cutoff` plus live source membership before returning results. The purge helper is now test-only until there is an explicit maintenance/write-time caller.

crates/coral-app/src/search/observed.rs
mystic123: Addressed in 0a0056f5 for v1. `resolve_projection_provenance` now treats any query with a `WITH` clause as unknown provenance, so observed-value indexing skips it instead of resolving an outer CTE name against an installed source table with the same name. Added `skips_cte_queries_to_avoid_derived_value_misattribution`. Follow-up for CTE-aware provenance is tracked in BENCH-562.

crates/coral-app/src/search/observed.rs
mystic123: Partially addressed in the current stack, but not fully enough to resolve this thread.

This PR no longer does the expensive redb + Tantivy projection write synchronously in the SQL response path: successful SQL now enqueues observed records durably, and SQL/search only spend the configured foreground drain budget (`search.observed_queue_foreground_drain_ms`, default 1000ms) trying to project queued jobs. That removes the biggest synchronous write-path risk from the original version.

The remaining budget question is still real: collection can still walk the returned rows/fields/JSON before enqueueing, and we have not added row/cell/candidate/container-depth caps here because the current product direction is to avoid deterministic row/column caps and instead handle this through a broader observed-search storage/work budget. Leaving this unresolved until that budget is implemented or the v1 scope is explicitly narrowed.

mystic123: Addressed in 3a1a3676.

The observer now has explicit v1 budgets:

- `[search].observed_collection_max_candidates` (default 10000), `[search].observed_collection_max_candidate_bytes` (default 8 MiB), and `[search].observed_collection_max_json_depth` (default 8) bound the SQL-result collection pass before enqueueing. When exhausted, collection stops and only the accepted prefix is enqueued. Deep JSON containers are also not indexed as raw strings when their subtree exceeds the depth budget, so skipped deep values do not leak through the parent blob.
- `[search].observed_max_storage_mb` (default 256 MiB) bounds the durable observed DB payload. Enforcement accounts the observed records, source index, last-observed index, and queue payload; it drops queued jobs first, then prunes oldest observed records by `last_observed_at` and removes their Tantivy projection entries.
- SQL and search paths enforce the storage budget around the foreground queue drain; the queue drain remains bounded by `[search].observed_queue_foreground_drain_ms` (default 1000ms).

I intentionally did not add row/cell caps here. The v1 bound is candidate count, candidate bytes, JSON depth, foreground drain time, and durable storage size, matching the current direction to avoid deterministic row/column truncation.

Coverage added: `observed_collection_candidate_budget_limits_sql_result_ingestion`, `observed_collection_depth_budget_does_not_index_deep_json_via_container`, `observed_storage_budget_drops_queue_before_records`, and `observed_storage_budget_prunes_oldest_records`.

mystic123: Follow-up after the latest #1019 update (`5c42f48a`): the earlier reply on this thread is now outdated.

Current behavior:

- Obvious high-risk PII/payment fields are excluded before observed-value indexing (`ssn`, `social_security_number`, `credit_card`, `card_number`, CVV/CVC, passport/license/tax IDs, etc.). Added `skips_obvious_pii_and_payment_columns`.
- Candidate/byte/depth budget exhaustion no longer drops the rest of the SQL result. Collection now flushes bounded record chunks into the durable queue and continues scanning, so overflow becomes recoverable backlog instead of a silent truncated warm index. Added `observed_collection_candidate_budget_flushes_overflow_to_queue`.
- Storage budget pressure is visible to search clients: if observed storage is over budget, the observed provider note says indexing is paused because the storage budget is exhausted. Added `search_reports_observed_storage_budget_exhaustion`.
- `observed_storage_bytes()` now includes the Tantivy/search tree bytes as well as observed redb logical payload, so the configured budget accounts for the projection side too. Added `observed_storage_budget_counts_tantivy_search_tree_bytes`.

Validation run locally included observed/index unit slices, serial gRPC search tests, CLI/MCP search tests, `cargo clippy -p coral-app --all-features --all-targets -- -D warnings`, `cargo fmt --check`, and `git diff --check`.

mystic123:
Thanks, this framing makes sense to me. I agree with the v1 bar: bounded, conservative, read-only on search, access-scoped, and explicit about local persistence.

Current state after the follow-up patches:

- Provenance is intentionally conservative for v1. Ambiguous projections are skipped, and CTE queries now resolve to unknown provenance instead of risking source misattribution.
- The observer-side index lifecycle issue is fixed by clearing the catalog fingerprint before observer-side repair can create a new-schema/empty catalog projection.
- Search no longer performs retention cleanup on the read path. It opens the existing index, filters by retention/live source, and leaves mutation/cleanup to write/maintenance paths.
- Source add/import/delete now marks an observed-source generation and discards observed values for that source; the observer rechecks that generation before enqueueing so values from a query racing a source mutation are skipped.
- Indexing now has explicit configurable budgets for candidate count, candidate bytes, JSON depth, foreground queue drain time, and observed-search storage size. Overflow is queued in bounded chunks rather than synchronously projected.

Two caveats I do not want to hand-wave:

- We still are not doing deterministic row/cell caps. The current bound is candidates/bytes/depth/storage/foreground time, not rows and cells. If row/cell caps are a hard v1 requirement, that is still a gap.
- Access scoping is improved for source mutation/reuse, but it is not yet keyed by credential/account/tenant identity. It is effectively workspace + source generation/source name, with purge on source mutation. I agree `source_name` alone is not the final access boundary.
- The privacy model is still a local-memory + retention + storage-budget + denylist model, not a full privacy model. We added filters for obvious credential/PII/payment fields and credential-looking raw strings, but I agree this should be documented and paired with explicit clear/disable controls before calling the surface mature.

So I think the current direction is right, and several of the concrete v1 safety rails are now in place, but I would keep this as partially addressed until we decide whether row/cell caps, credential/account scoping, and explicit clear/disable controls are required before landing this PR.

PR #1020, opened 2026-05-29
feat(mcp): expose universal search tool

mystic123:
## Summary
- Exposes Universal Search through the MCP `search` tool.
- Removes the separate MCP catalog-search surface so agents use one search entry point.
- Deletes the legacy gRPC `SearchCatalog` RPC/messages from the catalog service contract; out-of-tree callers should use `SearchService.Search` or catalog/list/describe APIs instead.
- Updates the MCP tool description to explain supported keyword-style queries and the typed result shapes agents should expect.
- Renders provider statuses, per-result provider provenance, truncation metadata, field-role provenance, observed-value surface-kind/field-path/count/freshness provenance, catalog matched fields, and compact table-column previews in MCP results.
- Updates the MCP structured output schema so agents can rely on typed catalog metadata, column hints, native search paths, observed values, and provider statuses.
- Keeps missing truncation metadata schema-valid if a malformed response omits it.
- Updates MCP tests, reference docs, and guidance to match the single-tool Universal Search surface.

## Stack Position
Part 4 of the current Universal Search stack:
4. #1020 MCP `search` tool
5. #1021 CLI `coral search`
6. #1180 provider fan-out orchestration

## Validation
Validated for this layer:
- `cargo fmt --check`
- `cargo check -p coral-mcp -p coral-cli`
- `cargo test -p coral-mcp search`
- `cargo test -p coral-mcp list_catalog_surfaces_table_functions`


mystic123:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/withcoral/coral/1020?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1020** <a href="https://app.graphite.com/github/pr/withcoral/coral/1020?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1020?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

docs/guides/use-coral-over-mcp.mdx
```diff
@@ -15,8 +15,8 @@ Before setting up a client, make sure you have [connected at least one source](/
 | Type     | Name             | Description                                             |
 | -------- | ---------------- | ------------------------------------------------------- |
 | Tool     | `sql`            | Execute read-only SQL against the Coral database        |
+| Tool     | `search`         | Route clues to likely catalog items, columns, filters, observed values, and native search paths |
```
AlbertQM: Should we conform to the search PR? #1021 

```suggestion
| Tool     | `search`         | Find likely catalog items, columns, filters, observed values, and native search paths |
```

docs/project/architecture.mdx
```diff
@@ -68,6 +68,7 @@ The local server and core orchestrator.
 | Feedback persistence | Workspace-scoped blocked-agent feedback reports (`feedback/`) |
 | Query orchestration | Loads sources, builds the engine, executes SQL (`query/`) |
 | Catalog discovery | Lists/searches/describes query-visible tables and columns (`catalog/`) |
+| Universal Search | Routes clues to typed catalog, column, and native-search-path hints (`search/`) |
```
AlbertQM: Coral 🤝  Sentence case

```suggestion
| Universal search | Routes clues to typed catalog, column, and native-search-path hints (`search/`) |
```

docs/guides/use-coral-over-mcp.mdx
mystic123: Updated #1020 to use this wording in `docs/guides/use-coral-over-mcp.mdx`; I also kept the matching MCP table in `docs/reference/cli-reference.mdx` in sync.

docs/guides/use-coral-over-mcp.mdx
```diff
@@ -37,9 +37,12 @@ Agents should treat the discovery tools as catalog helpers, not as replacement p
 - For table functions, use `sql_call_example` and fill in the required arguments.
 - Do not quote the whole `schema.item` string: write `github.pulls` or `"github"."pulls"`, not `"github.pulls"`.
 
-**`search_catalog`**: deterministic regex matching over database catalog metadata.
+**`search`**: one-query clue routing over Coral discovery metadata and previously observed values.
 
-- Arguments: required `pattern` plus optional `schema`, `kind`, `ignore_case`, `limit`, `offset`.
+- Arguments: required `query` only.
+- Returns typed hints such as `catalog_item`, `column_hint`, `observed_value`, and `native_search_path`, plus provider-status and truncation metadata.
+- Use it when the agent has a clue but does not know which source, table, table function, column, or filter to try first.
```
AlbertQM: Have I been reading too much docs, and now everything starts to sound wrong? This sentence says "use it" (I assume myself, a human) "when the agent has a clue" (???).

Is the intended suggestion here something like this? Or am I missing something? 😄 

```suggestion
- It's intended to be used by an agent when it has a clue, but does not know which source, table, table function, column, or filter to try first.
```

docs/guides/use-coral-over-mcp.mdx
```diff
@@ -37,9 +37,12 @@ Agents should treat the discovery tools as catalog helpers, not as replacement p
...
+- Results are not evidence. Agents must verify useful hints with ordinary `sql` queries or explicit table-function calls before answering.
```
AlbertQM: Same comment as #1021 (and similar as the line above): this sentence looks like it was written for an agent (ie should live in the AGENTS.md not here -- but please don't add it there either 😂 )

WDYT? Does a human need to care that "results are not evidence", whatever that means?

docs/guides/use-coral-over-mcp.mdx
mystic123: Updated this to human-facing guidance. It now says to use `search` to turn a plain-text query, identifier, or phrase into candidate sources, SQL surfaces, columns, filters, or table-function calls.

AlbertQM (approved):
Thank you! This looks very powerful 😄

sauldhernandez (approved):

Bradley-Butcher (approved):
I'm approving this. I pulled this down, installed the binary, and tried the MCP `search` tool locally. I'd like you to consider a couple of contract questions though.

There are two separate questions I think we should answer.

First: what is the `provider_statuses` contract?

- Is `provider_statuses` reporting provider candidates before the final merge/sort/truncation?
- If yes, is it expected that `observed_values: results_found` can appear even when there is no `observed_value` in the returned `results[]`?
- In that case, what is the agent meant to do with that signal?
- Since the MCP tool only accepts `query` today, with no pagination and no exposed larger limit, is the intended recovery just "try a narrower query"?

Second: how much do we trust the final ranking?

- If we fully trust the global ranking, then the verbose result shape seems fine. Give the agent rich details for the top hits.
- If we do not fully trust it, then I'm less sure about spending this much payload on detailed individual hints, because useful candidates can exist but still fall below the returned top 50.

One concrete thing I saw: after seeding an observed value from this PR, I hit a broad query where observed values reported `results_found`, but the returned top 50 had no `observed_value`. That may be totally intentional, but I want to confirm that is the contract rather than an accidental artifact of the current merge/truncation behavior.

mystic123:
Thanks, this is a real ambiguity in the current contract. I updated the MCP description/docs to make the intended behavior explicit.

The intended contract is:

- `provider_statuses` are informational provider coverage/error metadata, computed before final global ranking/truncation.
- `results_found` means that provider produced candidates in its own retrieval window; it does not guarantee that one of those candidates appears in the returned `results[]` window.
- `results[]` is the final ranked/truncated result window. In #1020 the service builds catalog and observed-value candidate sets/statuses first, then merges candidates, sorts them, and takes `max_results`.
- For MCP specifically, `search` currently only accepts `query`; the server sends `limit: 0`, so there is intentionally no MCP pagination/offset contract here.

The recovery behavior for v1 is query refinement, not paging. If `truncation.truncated` is true, or if a provider reports matches but none of that provider's rows survive into `results[]`, the agent should retry with a more targeted keyword/identifier query. Real pagination would need a stable materialized result set and cursors; rerunning the search and skipping rows would not be stable because catalog/index/provider state and ranking can change between calls.

I updated:

- `crates/coral-mcp/src/surface/tools.rs` search tool description
- `crates/coral-mcp/src/surface/search.rs` output schema descriptions for `provider_statuses` and `truncation`
- `crates/coral-mcp/src/guide_template.md` agent guide text
- `docs/guides/use-coral-over-mcp.mdx` public MCP guide

PR #1021, opened 2026-05-29
feat(cli): add coral search

mystic123:
## Summary
- Adds `coral search` as the CLI entry point for Universal Search.
- Supports compact text output and structured JSON output for typed search results.
- Renders catalog metadata, column hints, native search paths, observed values, provider statuses, truncation metadata, and provider-native rows when present later in the stack.
- Moves stable search enum labels onto the shared `coral-api` enum types so CLI/MCP output do not duplicate provider/state/role stringification.
- Adds shared always-quoted SQL identifier/reference formatting in `coral-client` and uses it from CLI/MCP surfaces.
- Adds `coral search --limit` with the same 1..100 result window as the Search API.
- Updates CLI tests and reference documentation for the new command.

## Stack Position
Part 5 of the current Universal Search stack:
4. #1020 MCP `search` tool
5. #1021 CLI `coral search`
6. #1180 provider-native fanout

## Validation
Validated while restacking this layer:
- `cargo fmt --check`
- `cargo test -p coral-api search_ --lib`
- `cargo test -p coral-client sql_formatting_always_quotes_identifier_parts --lib`
- `cargo test -p coral-mcp search --lib`
- `cargo test -p coral-mcp list_catalog_surfaces_table_functions --lib`
- `cargo test -p coral-cli search_output --lib`
- `cargo test -p coral-cli --features cli-test-server search_command`
- `cargo test -p coral-cli --all-features mcp_stdio_lists_tools_and_resources --no-run`


mystic123:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/withcoral/coral/1021?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1021** <a href="https://app.graphite.com/github/pr/withcoral/coral/1021?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1021?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

docs/reference/cli-reference.mdx
```diff
@@ -34,6 +34,25 @@ coral sql "<SQL>"
...
+## `coral search`
+
+Find likely Coral catalog items, columns, filters, and provider-native search paths.
+
+```shellscript
+coral search "payments-api timeout"
+coral search --json "payments-api timeout"
+```
+
+The text output is compact and shows each result's type, name, and most useful payload detail. Treat results as routing hints; verify them with `coral sql` before using retrieved data as evidence.
```
AlbertQM: >Treat results as routing hints; verify them with `coral sql` before using retrieved data as evidence.

This sounds like a hint to an agent more than a human, WDYT?

I read this, but I'm not sure what I'd need to verify with `coral sql`. What would be useful here? An example of data you might get back? An explanation of why that data isn't "good enough" to be used as evidence? Or something else 🤔

Otherwise :chef-kiss:

docs/guides/use-coral-over-mcp.mdx
```diff
@@ -15,7 +15,7 @@ Before setting up a client, make sure you have [connected at least one source](/
 | Type     | Name             | Description                                             |
 | -------- | ---------------- | ------------------------------------------------------- |
 | Tool     | `sql`            | Execute read-only SQL against the Coral database        |
-| Tool     | `search`         | Route clues to likely catalog items, columns, filters, observed values, and native search paths |
+| Tool     | `search`         | Find likely catalog items, columns, filters, observed values, and native search paths |
```
AlbertQM, line 18: Everywhere else in #1020 we use "Route clues", should we conform this one too?
Or keep the "Find" (I prefer this one FWIW), but not a mix of both 😄

docs/reference/cli-reference.mdx
mystic123: Updated this paragraph to be more concrete for humans: it now explains what table/column, native search path, observed value, and provider search row results mean, and says to run `coral sql` when current source rows are needed.

AlbertQM (approved):
Thanks!

crates/coral-cli/src/search_output.rs
```diff
@@ -0,0 +1,709 @@
...
+        ]),
+        Payload::NativeSearchPath(path) => native_search_path_row(path),
+    }
+}
+
+fn catalog_metadata_row(metadata: &CatalogMetadata) -> Option<[String; 3]> {
+    match metadata.item.as_ref()?.item.as_ref()? {
+        catalog_item::Item::Table(table) => Some([
+            "catalog_item".to_string(),
+            qualified_name(&table.schema_name, &table.name),
```
sauldhernandez, line 97: P1: Text mode renders the surface name with raw `schema.table` concatenation, but table names are only required to be non-empty in `coral-spec`. Valid names like `Pull.Requests`, `message-events`, or case-preserving identifiers will print as non-runnable or ambiguous SQL references, even though the docs tell users to run `coral sql` with the returned reference. The text output needs to expose the quoted `sql_reference` behavior too, either as the displayed name for queryable surfaces or as an explicit Reference column.

crates/coral-cli/src/lib.rs
```diff
@@ -619,6 +651,24 @@ fn run_features(
...
+fn print_search_response(
+    response: &coral_api::v1::SearchResponse,
+    json: bool,
+) -> Result<(), anyhow::Error> {
+    if json {
+        println!("{}", search_output::search_json(response)?);
+        return Ok(());
+    }
+
+    let rows = search_output::search_rows(response);
```
sauldhernandez, line 667: P2: Text output derives the whole command result from rows and drops `provider_statuses` plus `truncation`. The Search service reports provider/index failures, partial coverage, and truncation as successful response metadata, so this can print `No search results.` when a provider is actually in `Error`, or show degraded/partial results with no warning. Text mode should surface non-healthy provider states and truncation notes.

crates/coral-cli/src/search_output.rs
```diff
@@ -0,0 +1,709 @@
...
+    json!({
+        "provider": provider,
+        "type": "catalog_item",
+        "kind": "table_function",
+        "schema_name": function.schema_name,
+        "name": qualified_name(&function.schema_name, &function.name),
+        "sql_reference": format_sql_reference(&function.schema_name, &function.name),
+        "description": function.description,
+        "matched_fields": metadata.matched_fields,
+        "table_function": {
```
sauldhernandez: P2: This hand-rolled CLI JSON contract diverges from the existing MCP search output schema. Catalog table-function hits omit the required `sql_call_example`, and this nested `table_function` object adds a `kind` field the MCP schema does not allow. That contradicts the CLI docs' claim that `--json` contains the same shared typed-result fields. Prefer sharing the existing renderer/schema, or explicitly document and test a CLI-specific contract.

sauldhernandez (changes requested):

crates/coral-cli/src/search_output.rs
mystic123: Addressed in the updated #1021 stack layer. Text output now has an explicit `Reference` column populated with the same quoted SQL-reference logic for queryable surfaces, while `Name` remains the human-readable catalog/display name. Column hints and observed values point the reference at their parent surface.

crates/coral-cli/src/lib.rs
mystic123: Addressed. Text mode now prints warnings for non-healthy provider states (`skipped`, `partial`, `error`, `unknown`) and for truncation metadata, so an empty or degraded response no longer silently renders as just ordinary rows or `No search results.`

crates/coral-cli/src/search_output.rs
mystic123: Addressed. `coral search --json` now delegates to the shared MCP search response renderer instead of maintaining a hand-rolled CLI JSON shape. That keeps `sql_call_example` on catalog table-function hits, removes the stray nested `table_function.kind`, and keeps source-native/provider-result optional fields aligned with the MCP schema.

simonwhitaker (approved):

sauldhernandez (approved):

Bradley-Butcher (approved):
Approving this. I pulled the branch into a clean worktree, checked the current code against the earlier review threads, and ran the new CLI surface locally.

What I verified:

- text output now has an explicit `Reference` column, so users get the quoted/queryable surface separately from the display name
- `coral search --json` delegates to the shared MCP search renderer instead of maintaining a separate hand-rolled JSON contract
- text output surfaces non-healthy provider status and truncation warnings
- real local `coral search "github search_issues q"` produced the expected table with a provider partial warning
- real local `coral search --json "github search_issues q"` produced the shared `{ provider_statuses, truncation, results }` shape

Validation I ran:

- `cargo test -p coral-cli search_output --lib`
- `cargo test -p coral-mcp search_value_renders`
- `cargo test -p coral-cli --features cli-test-server search_command`
- `cargo fmt --check`

Brad said I approved too early he was still looking, but it looks fine. I am a big dumb robot.  - Codex

crates/coral-cli/src/search_output.rs
```diff
@@ -0,0 +1,653 @@
...
+    if !(first.is_ascii_lowercase() || first == '_') {
+        return true;
+    }
+    if !chars.all(|char| char.is_ascii_lowercase() || char.is_ascii_digit() || char == '_') {
+        return true;
+    }
+    RESERVED_SQL_IDENTIFIERS.contains(&identifier)
+}
+
+const RESERVED_SQL_IDENTIFIERS: &[&str] = &[
```
jsummerfield: Please don't do this. And don't do it again in `values.rs`

jsummerfield (changes requested):
Taking a closer look at this one before it merges — the functionality is fine but the
implementation duplicates a lot of logic that already exists one crate over, and I don't
think it should land as-is.

**1\. The enum→label mappers are verbatim copies of** #1020**.**
`provider_name`, `provider_state`, `surface_kind`, and `field_role` in
`crates/coral-cli/src/search_output.rs` are line-for-line reimplementations of the same
four functions already in `crates/coral-mcp/src/surface/search.rs` — same names, same
match arms, same string literals (`"table_column"`, `"catalog_metadata"`, …). These
strings are the canonical wire/JSON names of the enums (they're pinned in the MCP JSON
schema), so they need exactly one home. The right place is `coral-api`, next to the enum
definitions — add a `display_name()`/label helper there and have both the MCP JSON path
and this text path consume it.

**2\. The SQL-identifier quoting is duplicated** **_and_** **incomplete.**
`identifier_needs_quotes` + `format_sql_identifier` + the 48-entry
`RESERVED_SQL_IDENTIFIERS` const are copy-pasted into both `search_output.rs` and
`crates/coral-mcp/src/surface/values.rs`. On top of the duplication, the hand-rolled
keyword list is a partial subset of the reserved set we already vendor via DataFusion 53 /
sqlparser 0.59 — it's missing common collidable identifiers (`table`, `values`, `using`,
`default`, `primary`, `references`, `window`, `partition`, …), so any source whose
table/column hits one of those will render an unquoted reference that breaks when pasted
into `coral sql`. Drop the hand list entirely and either always-quote or look up
`sqlparser`'s `ALL_KEYWORDS`, in one shared helper.

**3\. The duplication is inconsistent within this same PR.**
The JSON path correctly reuses `coral_mcp::search_response_json` across the crate
boundary — so the sharing mechanism clearly exists and is understood. The text path then
reimplements that same boundary by hand. Please make the text path delegate the same way;
`search_output.rs` should be left holding only table layout + printing.

**4\.** **`--limit`** **is referenced but doesn't exist.**
The request hardcodes `limit: 0` and `SearchArgs` exposes no `--limit` flag, yet the
truncation warning can tell users to "increase --limit to inspect more candidates."
Either add the flag or drop the note

crates/coral-cli/src/search_output.rs
mystic123: Addressed in the restacked #1021 layer. CLI/MCP no longer own duplicate search enum stringification or SQL quoting: stable search labels now live on the shared `coral-api` enum types, and SQL identifiers/references use the shared always-quoting `coral-client` helpers. I also removed the duplicated keyword/quoting logic from both `search_output.rs` and `values.rs`, and added `coral search --limit` so the truncation guidance maps to a real CLI flag.

mystic123:
Replying to the top-level review as well: agreed, these were the right issues to force clean before landing this layer.

Current restacked #1021 addresses the four points this way:

1. Search enum labels now have one shared home on the generated API enum types (`SearchProvider::label`, `SearchProviderState::label`, `SearchSurfaceKind::label`, `SearchFieldRole::label`) and both MCP/CLI consume those instead of carrying duplicate string mappers.
2. SQL identifier/reference rendering now uses shared `coral-client` helpers (`format_sql_identifier`, `format_sql_reference`) that always quote each identifier part. The duplicated keyword-list quoting logic was removed from both CLI search output and MCP values.
3. `coral search --json` delegates to `coral_mcp::search_response_json`; the CLI-specific code is limited to text rows/warnings/table rendering.
4. `coral search --limit` now exists with a `1..=100` parser. When omitted, the CLI sends `0` intentionally as the service-default sentinel.

Validation after the restack: #1021 is green for Rust checks, Windows smoke, performance, and validate. The stack head (#1180) is green on those checks as well.

PR #1026, opened 2026-05-29
feat: add baseten community source spec

Rishikesh63:
## Summary

Adds a Coral source spec for [Baseten](https://www.baseten.co) :
 A  ML model deployment platform. Enables SQL queries over your deployed models,
deployments, and secrets via the Baseten Management API.

## Tables

| Table | Description |
|---|---|
| `baseten.models` | All models in your Baseten account |
| `baseten.deployments` | Deployments for a model (requires `model_id`) |
| `baseten.secrets` | Registered secret names and metadata (values never exposed) |

## Authentication

API key via `BASETEN_API_KEY` (from app.baseten.co → Settings → API Keys).
Uses `Api-Key` header format as required by the Baseten Management API.

## Notable Features

- `deployments` table flattens nested `autoscaling_settings` fields
  (`min_replicas`, `max_replicas`, `target_concurrency`) using
  `expr: kind: path` with depth-2 paths
- Cross-table JOIN: models + deployments to find active production replicas
- `secrets` table exposes only metadata — secret values are never returned
  by the Baseten API

## Example Query

```sql
SELECT m.name, d.status, d.active_replica_count, d.instance_type_name
FROM baseten.models m
JOIN baseten.deployments d ON d.model_id = m.id
WHERE d.is_production = true AND d.status = 'ACTIVE';


jsummerfield (changes requested):
Findings
- High: `sources/community/baseten/manifest.yaml:183`, `sources/community/baseten/manifest.yaml:190`, and `sources/community/baseten/manifest.yaml:197` read autoscaling paths that do not match Baseten’s schema. The OpenAPI schema uses `min_replica`, `max_replica`, and `concurrency_target`, not `min_replicas`, `max_replicas`, and `target_concurrency`, so these columns will come back null even when autoscaling settings are present. Please point the columns at the documented fields, and update the README wording at `sources/community/baseten/README.md:30` and `sources/community/baseten/README.md:61` if you keep the friendlier column names. Docs: https://api.baseten.co/v1/spec
- Medium: `sources/community/baseten/manifest.yaml:53`, `sources/community/baseten/manifest.yaml:134`, and `sources/community/baseten/manifest.yaml:232` model Baseten `date-time` fields as `Utf8`. Please expose these as `Timestamp` with `format_timestamp`/`iso8601` where Coral can parse them, matching the pattern used by other sources for provider timestamps. Docs: https://api.baseten.co/v1/spec
- Medium: `sources/community/baseten/manifest.yaml:23` includes a `test_queries` entry with `model_id = 'YOUR_MODEL_ID'`. That query cannot run unchanged and is not live-test evidence. Please replace it with a smoke query that can run as-is, or remove it and document how to manually test deployments after selecting a real model ID. Also add captured live Coral output from `coral source add` / `coral source test` or a representative query against a real Baseten account.
- Low: `sources/community/baseten/README.md:40` labels the example as “Cross-source”, but it only joins two Baseten tables. Please rename it to a Baseten models/deployments join or add an actual cross-source example.

Thanks for the useful ML deployment source. The main fixes are the documented autoscaling field names and proof that the source runs with real Baseten data.

Rishikesh63:
image
Screenshot 2026-05-29 162946
image

jsummerfield (changes requested):
Findings
- High: `sources/community/baseten/manifest.yaml:23` declares a source test query that uses `WHERE model_id = (SELECT id FROM baseten.models LIMIT 1)` against `baseten.deployments`. That table has a required `model_id` filter, and Coral required filters need a concrete pushed-down value; the subquery does not satisfy the required filter at planning time. Please remove this from `test_queries` or replace it with a test that does not require a dynamic value.
- High: `sources/community/baseten/README.md:40` teaches the same unsupported join pattern for deployments. A user should instead query `baseten.models` first, copy a model id, then query `baseten.deployments WHERE model_id = '...'`. Otherwise the first product example points users into a query that cannot plan cleanly.
- Medium: `sources/community/baseten/README.md:57` and `sources/community/baseten/manifest.yaml:11` do not explain the required Baseten API key type or permission level. Please state which API key users need and where to create it so the install path is self-contained. Docs: https://docs.baseten.co/organization/api-keys

Thanks for adding this. Once the examples and test queries respect Coral's required-filter behavior, this should be much easier for users to get working.

jsummerfield (approved):
Thanks for the updates. I re-reviewed current head `f435c3472cb2ee920b9a8653bccd88786b6654c8`: the prior required-filter/query-shape issues are resolved, the README now documents the API key path and first-query flow, live Coral evidence is present in the PR comments, and local/static checks plus the now-approved GitHub workflows are green.

One non-blocking polish note for a future pass: a short README note on Baseten Management API rate limits/429 behavior would make the docs more complete.

jsummerfield (approved):
No blocking findings on this current head. The source shape, README guidance, and live-test evidence reviewed here are approval-ready, and `Validate` / `Lint PR Title` are green.

Rishikesh63:
thank you @jsummerfield

PR #1027, opened 2026-05-29
Add incident.io community source

pooja-bhavani:
Summary

Adds a read-only incident.io community source for SRE and incident response workflows.

What changed

Adds sources/community/incident_io with:

- manifest.yaml defining incident.io tables over read-only GET endpoints
- README.md with setup steps and example SQL queries

Tables included:

- incident_io.incidents
- incident_io.actions
- incident_io.users
- incident_io.schedules
- incident_io.incident_roles
- incident_io.severities

Why

incident.io is used by SRE, DevOps, and platform teams to coordinate incidents, follow-up actions, users, schedules, roles, and severity configuration. Making it queryable through Coral SQL helps teams inspect incidents, audit action items, review on-call context, and join incident data with other operational sources.

Validation

- coral source lint sources/community/incident_io/manifest.yaml
- Result: Manifest is valid

Notes

- Only read-only GET endpoints are used.
- v2 endpoints use cursor pagination with after.
- incident_io.severities uses the v1 severities endpoint because incident.io documents severity configuration there.

### DIFF
--- a/sources/community/incident_io/README.md
+++ b/sources/community/incident_io/README.md
@@ -1,0 +1,89 @@
+# incident.io source
+
+Query incident.io incident response data through Coral SQL.
+
+This community source exposes read-only incident.io API data for SRE,
+DevOps, platform engineering, incident response, and operational review
+workflows.
+
+## Configuration
+
+Create an incident.io API key from:
+
+`Settings -> API keys`
+
+Then configure the source:
+
+```bash
+export INCIDENT_IO_API_TOKEN="incident_io_api_key"
+
+coral source add --file sources/community/incident_io/manifest.yaml
+coral source test incident_io
+```
+
+## Example Queries
+
+Recent incidents:
+
+```sql
+SELECT
+  reference,
+  name,
+  status__category,
+  severity__name,
+  permalink,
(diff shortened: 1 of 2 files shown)


jsummerfield (changes requested):
Findings
- High: `sources/community/incident_io/manifest.yaml:1` adds a parallel incident.io community source, but `sources/core/incident_io` already exists and is the canonical Coral source for this provider. The core source already exposes incidents, alerts, escalations, schedules, status pages, workflows, teams, users, and catalog resources, so this should not land as a second community source. Please remove `sources/community/incident_io/**` and, if this PR contains useful improvements, scope them as a focused change to `sources/core/incident_io` instead.

Thanks for contributing this. Keeping incident.io in the existing core source gives users one clear schema to learn and avoids splitting the same provider across competing source names.

pooja-bhavani:
Thanks for the review. I missed that incident.io already exists as the canonical source under sources/core/incident_io.

I am withdrawing this PR to avoid introducing a duplicate community source for the same provider. If I contribute incident.io changes later, I will scope them as focused improvements to the existing core source instead.

PR #1034, opened 2026-05-29
feat(sources/community/nvd): add NVD community source

Prem4777:
## Summary

Adds a new community source for the NIST National Vulnerability Database (NVD) under `sources/community/nvd/`.

## Why

The NVD is the U.S. government's authoritative repository for CVE records and CVSS severity scores. It's a natural complement to the existing `cisa_kev` source — KEV tells you what's being exploited, NVD tells you the full severity and technical details.

## Tables

| Table | Description |
|---|---|
| `vulnerabilities` | Core CVE records — id, status, published date, description |
| `cvss_v3_1_metrics` | CVSS v3.1 metric array as JSON |
| `cvss_v3_0_metrics` | CVSS v3.0 metric array as JSON |
| `cvss_v2_metrics` | CVSS v2 metric array as JSON |
| `references` | Advisory and patch links per CVE |

## Features

- No authentication required (unauthenticated only, 5 req/30s)
- CVSS tables return full metric arrays as JSON — filter by `source = 'nvd@nist.gov'` and `type = 'Primary'` for the authoritative score
- Date filters pushed server-side, NVD's 120-day window cap documented
- `cvss_v3_severity` pushed server-side across all tables
- CVSS v4.0 explicitly out of scope

## Example Queries

```sql
-- Lookup a specific CVE
SELECT cve_id, description, vuln_status
FROM nvd.vulnerabilities
WHERE cve_id = 'CVE-2021-44228'
LIMIT 1;
```

```sql
-- Critical CVEs via server-side severity filter
SELECT cve_id, vuln_status
FROM nvd.vulnerabilities
WHERE cvss_v3_severity = 'CRITICAL'
LIMIT 10;
```

```sql
-- CVEs published in a date window (max 120 days)
SELECT cve_id, published
FROM nvd.vulnerabilities
WHERE pub_start_date = '2024-01-01T00:00:00.000'
AND pub_end_date = '2024-01-31T23:59:59.999'
LIMIT 10;
```

## Validation

```
coral source lint sources/community/nvd/manifest.yaml  ✓
coral source add --file sources/community/nvd/manifest.yaml
coral source test nvd
```

```
✓ nvd connected successfully
nvd (5 tables)
├─ cvss_v2_metrics
├─ cvss_v3_0_metrics
├─ cvss_v3_1_metrics
├─ references
└─ vulnerabilities

Query tests
5 declared · 5 passed · 0 failed
✓ SELECT cve_id, vuln_status FROM nvd.vulnerabilities WHERE cve_id = 'CVE-2021-44228' LIMIT 1   1 row
✓ SELECT cve_id, metrics FROM nvd.cvss_v3_1_metrics WHERE cve_id = 'CVE-2021-44228' LIMIT 1   1 row
✓ SELECT cve_id, metrics FROM nvd.cvss_v2_metrics WHERE cve_id = 'CVE-2021-44228' LIMIT 1   1 row
✓ SELECT cve_id FROM nvd.vulnerabilities WHERE pub_start_date = '2024-01-01T00:00:00.000' AND pub_end_date = '2024-01-07T23:59:59.999' LIMIT 1   1 row
✓ SELECT cve_id, vuln_status FROM nvd.vulnerabilities WHERE cvss_v3_severity = 'CRITICAL' LIMIT 1   1 row
```


sources/community/nvd/README.md
```diff
@@ -0,0 +1,138 @@
...
+WHERE c.base_severity = 'CRITICAL'
+LIMIT 10;
+```
+
+Recent CVEs published in 2024:
+
+```sql
+SELECT cve_id, published, vuln_status
+FROM nvd.vulnerabilities
+WHERE published >= '2024-01-01T00:00:00.000'
```
jsummerfield: High: this example filters on the materialized `published` column, so Coral has to fetch pages from NVD and filter locally instead of pushing `pubStartDate` / `pubEndDate` to the API. NVD requires those date filters in pairs and caps any date range at 120 consecutive days. Please rewrite the date examples and notes to use `pub_start_date` + `pub_end_date` (and similarly `last_mod_start_date` + `last_mod_end_date`) with a <=120-day window. Docs: https://nvd.nist.gov/developers/vulnerabilities

sources/community/nvd/manifest.yaml
```diff
@@ -0,0 +1,575 @@
...
+      - name: base_score
+        type: Float64
+        description: CVSS v3 base score (0.0 – 10.0).
+        nullable: true
+        expr:
+          kind: first_array_item_path
+          path:
+            - cve
+            - metrics
+            - cvssMetricV31
```
jsummerfield: High: this table is documented as CVSS v3.x, but every metric column reads only `cvssMetricV31`. The NVD 2.0 schema exposes both `cvssMetricV31` and `cvssMetricV30`, so CVEs with only CVSS 3.0 metrics will return null score/severity fields here. Please either coalesce v3.1 then v3.0, or split the schema into explicit v3.1/v3.0 surfaces. Schema: https://csrc.nist.gov/schema/nvd/api/2.0/cve_api_json_2.0.schema

sources/community/nvd/manifest.yaml
```diff
@@ -0,0 +1,575 @@
...
+        type: Utf8
+        nullable: true
+        virtual: true
+        description: Filter by keyword search across CVE descriptions and metadata.
+        expr: { kind: from_filter, key: keyword_search }
+
+  - name: cvss_v3
+    description: >-
+      CVSS v3.x severity scores for CVE records. Each row contains the primary
+      NVD scoring for one CVE. Rows are absent when NVD has not yet assigned a
```
jsummerfield: High: with `rows_path: vulnerabilities`, this table returns one row per fetched CVE; if NVD has not assigned CVSS v3 metrics, the metric columns are null. The description says rows are absent when no score exists, which will mislead users doing counts or joins. Please either model only scored rows, or document this as nullable per-CVE metric data and update examples accordingly. Schema: https://csrc.nist.gov/schema/nvd/api/2.0/cve_api_json_2.0.schema

sources/community/nvd/README.md
```diff
@@ -0,0 +1,138 @@
...
+Query CVE records, CVSS severity scores, and advisory references from the [NIST National Vulnerability Database](https://nvd.nist.gov/) — the U.S. government repository of standards-based vulnerability management data. No authentication required.
+
+```bash
+coral source add --file sources/community/nvd/manifest.yaml
+```
+
+## Rate Limits
+
+Without an API key: 5 requests per 30 seconds.
+With an API key: 50 requests per 30 seconds.
```
jsummerfield: Medium: the README tells users an API key increases the rate limit, but the manifest has no `NVD_API_KEY` input or `apiKey` request header support. NVD 2.0 requires API keys in the `apiKey` request header. Please either add supported secret input/header wiring or remove the API-key setup guidance so users do not expect auth to work. Docs: https://nvd.nist.gov/general/news/api-20-announcements

sources/community/nvd/manifest.yaml
```diff
@@ -0,0 +1,575 @@
...
+    response:
+      rows_path:
+        - vulnerabilities
+      row_strategy: direct
+    pagination:
+      mode: offset
+      offset_param: startIndex
+      page_size:
+        default: 2000
+        max: 2000
```
jsummerfield: Medium: 2,000-row pages with no `fetch_limit_default` or required source-side filters make ordinary exploratory queries easy to turn into very broad NVD scans, especially when users filter locally on columns like `published`. Given the README's 5-requests/30s unauthenticated limit and the 350k+ corpus, please add conservative fetch limits or require scoped filters for high-cardinality tables. NVD API paging/rate-limit guidance: https://nvd.nist.gov/developers/vulnerabilities

sources/community/nvd/manifest.yaml
```diff
@@ -0,0 +1,575 @@
...
+        required: false
+      - name: cvss_v3_severity
+        required: false
+      - name: keyword_search
+        required: false
+    request:
+      method: GET
+      path: /rest/json/cves/2.0
+      query:
+        - name: cveId
```
jsummerfield: Medium: NVD now marks `cveId` as deprecated in favor of `cveIds`, which also supports up to 100 comma-separated CVE IDs. Please map this filter to `cveIds` and document the supported single/multiple-ID shape, or explain why this source intentionally keeps the deprecated parameter. Docs: https://nvd.nist.gov/developers/vulnerabilities

jsummerfield (changes requested):
Findings:

- High: the NVD date-filter guidance and examples need to use source-pushed date filters and NVD's paired, <=120-day date-window rules.
- High: the CVSS v3 table currently reads only CVSS v3.1 metrics even though it is documented as v3.x.
- High: the CVSS tables are described as score-only rows, but the current `rows_path` returns one nullable row per CVE.
- Medium: README API-key guidance does not match manifest auth support, and broad unfiltered scans need a safer default/filter story.

Thanks for the quick iteration on the reference extraction issue. Once these API semantics are tightened up, this will be much closer to a reliable Coral source.

jsummerfield (changes requested):
Findings

- High: `sources/community/nvd/manifest.yaml:237` and the analogous CVSS columns flatten the first item from the NVD metric arrays as the score/severity. NVD metric entries carry `source` and `type` semantics, and the first array entry is not guaranteed to be the NVD primary metric. That can return the wrong CVSS score/severity for real CVEs. Please expose the metric arrays as JSON or model primary/secondary metrics explicitly rather than treating the first metric as canonical. Docs: https://nvd.nist.gov/developers/vulnerabilities.

- High: `sources/community/nvd/manifest.yaml:188` declares `pub_start_date`, `pub_end_date`, and `keyword_search` filters for `cvss_v3_1`, but the table only exposes virtual columns for `cve_id_filter` and `cvss_v3_severity` later in that table. The same pattern appears in `cvss_v3_0`, `cvss_v2`, and `references`. In Coral, users need same-named SQL columns to push those request filters from `WHERE`; otherwise the filter wiring is not usable. Please add virtual `from_filter` columns for every declared filter, or remove the dead filter/query wiring. NVD also documents paired date filters and the 120-day cap here: https://nvd.nist.gov/developers/vulnerabilities.

- Medium: `sources/community/nvd/README.md:126` shows a critical-CVE query using `WHERE c.base_severity = 'CRITICAL'`, which is local filtering over returned rows rather than the provider-pushed `cvss_v3_severity` filter. With `LIMIT 10`, that query can stop after a small unfiltered slice. Please change examples to use `c.cvss_v3_severity = 'CRITICAL'` plus `c.base_score IS NOT NULL`, or clearly label `base_severity` filtering as local/post-fetch.

- Medium: `sources/community/nvd/manifest.yaml:5` and `sources/community/nvd/README.md:20` present the source as covering CVSS severity scores, but NVD now exposes CVSS v4 filters/metrics and this source has no `cvss_v4_0` table or documented omission. Please either add CVSS v4 support or explicitly document it as out of scope. Docs: https://nvd.nist.gov/developers/vulnerabilities.

Thanks for the fast iteration. The prior `cveIds`, CVSS v3.0/v3.1 split, nullable score wording, and API-key docs are materially improved; the remaining blockers are about returning the correct NVD metric and making the pushed filters actually usable.

jsummerfield (changes requested):
Findings:

- Medium: `sources/community/nvd/README.md:18` still describes an API-key setup path that Coral cannot actually perform. `coral source add --interactive` does not let users add an arbitrary `apiKey` request header when the manifest has no `inputs`, `auth`, or `request_headers`, while NVD API 2.0 expects the optional key in the `apiKey` request header. Please either wire an `NVD_API_KEY` secret into the source or remove the workaround and state that this version is unauthenticated only. Docs: https://nvd.nist.gov/general/news/api-20-announcements

- Medium: `sources/community/nvd/README.md:160` and `sources/community/nvd/manifest.yaml:187`-`:190` still steer users toward `json_get(metrics, 0)` for CVSS extraction. NVD metric arrays carry `source` and `type` fields, and primary NVD scores should be selected by those fields, not by first-array-item position. Please replace this with a concrete safe example that verifies `source = 'nvd@nist.gov'` and `type = 'Primary'`, or avoid presenting index `0` as the normal path. Docs: https://nvd.nist.gov/developers/vulnerabilities

Most of the earlier NVD issues are fixed: the paired date filters, `cveIds`, CVSS v3.0/v3.1 split, nullable metric semantics, fetch limit, and CVSS v4.0 scope note all look much better. These two remaining docs/semantics issues are the blockers before approval.

Prem4777:
Both blockers are resolved. README no longer describes any API key setup path — auth section correctly states unauthenticated only at 5 req/30s. No json_get(metrics, 0) in either file; the manifest exposes raw JSON arrays and the README explicitly warns against position-based selection, directing to source = 'nvd@nist.gov' and type = 'Primary'. Good to approve.

jsummerfield (changes requested):
Findings

- Medium: `sources/community/nvd/manifest.yaml:24`-`:27` gives a date-window example from `2024-01-01T00:00:00.000` to `2024-04-30T23:59:59.999`, but NVD rejects that because the range exceeds its 120-day cap. Please make the guide example use a valid <=120-day window so users can copy it successfully. Docs: https://nvd.nist.gov/developers/vulnerabilities

- Low: `sources/community/nvd/manifest.yaml` is checked in with CRLF line endings/trailing whitespace, and `git diff --check` reports whitespace errors throughout the file. Please normalize the manifest to the repo's usual line endings before merge.

This source is close, and the current live checks look much better than earlier heads, but the shipped guide should not contain an NVD query shape the provider rejects. Thanks for the quick iteration on the API-key and CVSS semantics.

Prem4777:
The two flagged issues are fixed. Also did a full pass against the NVD API docs while I was in here and cleaned up several other things I found along the way.

**Fixes for the reported issues:**

- Date-window example in the guide (lines 24–27) updated from `2024-01-01` → `2024-04-30` (120+ days, rejected by NVD) to `2024-01-01` → `2024-03-31` (90 days, safely within the cap).
- Normalized manifest to LF line endings and stripped all trailing whitespace — `git diff --check` now passes.

**Additional fixes from the API docs review:**

- `cve_id` set to `nullable: true` across all 5 tables to match NVD reality.
- `references` table was missing the `cvss_v3_severity` filter, request query mapping, and virtual column — added to match the other four tables.
- `cvss_v3_1_metrics` table description was missing `baseSeverity` from the `cvssData` field list.
- `cvss_v3_0_metrics` and `cvss_v2_metrics` guides now advise filtering to `source = 'nvd@nist.gov'` and `type = 'Primary'` for the authoritative score, consistent with the `cvss_v3_1_metrics` guide.
- `keyword_search` virtual column description now carries the whole-word exact-match warning on all 5 tables, not just `vulnerabilities`.
- README 30-day window example corrected from `2024-01-30` (29 days) to `2024-01-31` (30 days).
- Test queries updated to cover a date-window and a severity filter query alongside the existing `cve_id` lookups, capped at 5 total due to rate limits.

jsummerfield (commented):
Findings

- Low: `sources/community/nvd/README.md:158` and the `keyword_search` guide text in `sources/community/nvd/manifest.yaml` still describe keyword search as exact whole-word matching, including the example that `buf` does not match `buffer`. NVD's `keywordSearch` behavior is broader than that; it supports suffix-style matching such as matching pluralized/extended terms, so this wording is likely to surprise users. Please reword this as provider keyword search with implicit expansion rather than exact whole-word matching. Docs: https://nvd.nist.gov/developers/vulnerabilities

The previous date-window, line-ending, API-key, and CVSS primary-metric findings look addressed in the current static review. Thanks for the careful iteration here.

jsummerfield (changes requested):
Findings

- High: the current head now defines 5 test queries and the README describes a 5-table source, but I do not see current live Coral evidence for this exact head. The existing pasted validation still appears to reflect the earlier 4-table / 4-test shape, and the latest author comment describes the changes without including fresh `coral source add --file ...` / `coral source test nvd` output. Please rerun validation on this head and paste the current output so we know the final manifest shape works end to end.

- Medium: `sources/community/nvd/README.md` links the NVD homepage, but it does not link the exact API/rate-limit docs users need for the date-window, pagination, keyword, and unauthenticated-limit behavior this source depends on. Please add the NVD CVE API 2.0 docs and developer/rate-limit docs to the README. Useful targets are https://nvd.nist.gov/developers/vulnerabilities and https://nvd.nist.gov/developers/start-here.

The keyword-search wording and prior YAML line-length issue look fixed in the current static pass. This is close; the remaining blocker is proving the current head and tightening the provider docs links for users.

Prem4777:
Reran validation on the current head (5 tables, 5 test queries):

```
✓ nvd connected successfully
nvd (5 tables)
├─ cvss_v2_metrics
├─ cvss_v3_0_metrics
├─ cvss_v3_1_metrics
├─ references
└─ vulnerabilities

Query tests
5 declared · 5 passed · 0 failed
✓ SELECT cve_id, vuln_status FROM nvd.vulnerabilities WHERE cve_id = 'CVE-2021-44228' LIMIT 1   1 row
✓ SELECT cve_id, metrics FROM nvd.cvss_v3_1_metrics WHERE cve_id = 'CVE-2021-44228' LIMIT 1   1 row
✓ SELECT cve_id, metrics FROM nvd.cvss_v2_metrics WHERE cve_id = 'CVE-2021-44228' LIMIT 1   1 row
✓ SELECT cve_id FROM nvd.vulnerabilities WHERE pub_start_date = '2024-01-01T00:00:00.000' AND pub_end_date = '2024-01-07T23:59:59.999' LIMIT 1   1 row
✓ SELECT cve_id, vuln_status FROM nvd.vulnerabilities WHERE cvss_v3_severity = 'CRITICAL' LIMIT 1   1 row
```

Also added links to the [NVD CVE API 2.0 docs](https://nvd.nist.gov/developers/vulnerabilities) and [rate limit / best practices](https://nvd.nist.gov/developers/start-here) in the README.

jsummerfield (approved):
Static review is clear on the current head, the previous API/auth findings are resolved, live Coral validation evidence is present, and the paused workflows are now approved and green. Thanks for the careful follow-up here.

PR #1046, opened 2026-05-29
feat: add Docker Engine API source spec

ayanasarkar:
## Docker Engine HTTP API Source

This source spec enables querying Docker container metadata as SQL using Coral.

### What it does
Connects to the Docker Engine HTTP API (default: `http://localhost:2375`) and exposes two tables:
- `docker.containers` — all containers including stopped/exited ones
- `docker.running` — only currently running containers

### Example queries
```sql
-- List all containers
SELECT id, image, status, state FROM docker.containers LIMIT 20;

-- Find exited containers
SELECT id, image, status FROM docker.containers WHERE state = 'exited' LIMIT 10;

-- Cross-source JOIN with GitHub to correlate deploys with crashes
SELECT
  dc.id,
  dc.image,
  dc.status,
  gp.title        AS last_pr,
  gp.user__login  AS deployed_by
FROM docker.containers dc
LEFT JOIN github.pulls gp
  ON gp.owner = 'your-owner'
  AND gp.repo  = 'your-repo'
  AND gp.state = 'closed'
WHERE dc.state != 'running'
LIMIT 10;
```

### Setup
See README.md for three setup options: socat proxy (recommended for local use), Docker Desktop, and remote TLS-secured daemon.

### Tables

| Table | Description |
|---|---|
| `containers` | All containers (`/containers/json?all=true`) |
| `running` | Running containers only (`/containers/json`) |

### Columns

| Column | Type | Description |
|---|---|---|
| `id` | Utf8 | Full 64-character container ID |
| `image` | Utf8 | Image name and tag |
| `status` | Utf8 | Human-readable status (e.g. `Exited (255) 2 days ago`) |
| `state` | Utf8 | Machine-readable state: `running`, `exited`, `dead`, `paused`, `restarting` |
| `created_epoch` | Int64 | Unix timestamp of creation |
| `command` | Utf8 | Entrypoint command |

### DIFF
--- a/sources/community/docker/README.md
+++ b/sources/community/docker/README.md
@@ -1,0 +1,149 @@
+# Docker
+
+**Version:** 0.2.0
+**Backend:** HTTP
+**Tables:** 2
+**Base URL:** `http://localhost:2375` (configurable)
+
+Query Docker container metadata as SQL via the [Docker Engine HTTP API](https://docs.docker.com/reference/api/engine/). No authentication required for local use.
+
+```bash
+coral source add --file sources/community/docker/manifest.yaml
+```
+
+## Tables
+
+| Table | Description |
+|---|---|
+| `containers` | All containers including stopped, exited, paused, and dead ones |
+| `running` | Only currently running containers |
+
+---
+
+### `containers`
+
+All containers on the Docker host. Equivalent to `docker ps --all`.
+
+#### Columns
+
+| Column | Type | Description |
+|---|---|---|
+| `id` | Utf8 | Full 64-character container ID |
+| `image` | Utf8 | Image name and tag (e.g. `nginx:latest`) |
+| `status` | Utf8 | Human-readable status string (e.g. `Exited (255) 2 days ago`) |
+| `state` | Utf8 | Machine-readable state: `running`, `exited`, `dead`, `paused`, `restarting` |
+| `created_epoch` | Int64 | Container creation time as Unix epoch seconds |
+| `command` | Utf8 | Entrypoint command string |
+
+---
+
+### `running`
+
+Only currently running containers. Equivalent to `docker ps`.
+
+#### Columns
+
+| Column | Type | Description |
+|---|---|---|
+| `id` | Utf8 | Full 64-character container ID |
+| `image` | Utf8 | Image name and tag |
+| `status` | Utf8 | Human-readable status string |
+| `state` | Utf8 | Machine-readable state |
+| `created_epoch` | Int64 | Container creation time as Unix epoch seconds |
+
+---
+
+## Setup
+
+### Option 1 — Local via socat (recommended)
+
+Proxy the Unix socket over TCP without exposing a network port:
+
+```bash
+socat TCP-LISTEN:2375,reuseaddr,fork UNIX-CLIENT:/var/run/docker.sock
+```
+
+Set `DOCKER_API_BASE=http://localhost:2375`.
+
+### Option 2 — Docker Desktop (Windows / macOS)
+
+Docker Desktop → Settings → General → **"Expose daemon on tcp://localhost:2375 without TLS"**
+
+> ⚠️ Only enable on trusted local machines. Never expose port 2375 on a network interface.
+
+### Option 3 — Remote daemon with TLS
+
+Follow Docker's [remote access guide](https://docs.docker.com/engine/daemon/remote-access/) to configure TLS certificates, then set:
+
+```
+DOCKER_API_BASE=https://your-docker-host:2376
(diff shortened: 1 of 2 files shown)


jsummerfield (changes requested):
Findings

- High: `sources/community/docker/manifest.yaml:7` defaults users to `http://localhost:2375`. That makes the unsafe Docker Engine TCP daemon path the happy path; Docker's remote-access docs warn that remote daemon access must be secured and that unauthenticated/TLS-less access is not recommended: https://docs.docker.com/engine/daemon/remote-access/. Please make the safe path explicit instead of asking users to expose the daemon without TLS.

- High: `sources/community/docker/manifest.yaml:20` and `sources/community/docker/manifest.yaml:41` call `/containers/json` but do not expose Docker's native `filters` or `limit` query parameters. Docker documents server-side filters such as `status`, `name`, `id`, `label`, `ancestor`, and `network`: https://docs.docker.com/reference/api/engine/version/v1.53/#tag/Container/operation/ContainerList. As written, normal operational queries like filtering by status/name/label fetch the container list first and only filter locally in Coral.

- Medium: `sources/community/docker/manifest.yaml:26` exposes raw CamelCase provider fields (`Id`, `Image`, `Status`, etc.). Please map these to Coral-style snake_case columns (`id`, `image`, `status`, `state`, `created_epoch` or `created_at`, `command`) so users do not need quoted identifiers in every query.

- Medium: `sources/community/docker/manifest.yaml:1` has no top-level description, no checked-in README/guide, and no captured live Coral evidence. Please add source docs covering setup/security, the Docker API version/limitations, schema orientation, example queries, and a pasted successful `coral source add`, `coral source test`, or representative `coral sql` output.

Thanks for the contribution. The provider is useful, but the source needs a safer setup story and Docker-native query semantics before it should become Coral product surface.

ayanasarkar:
Thank you for the detailed review! I've addressed all four points:

Mapped CamelCase Docker API fields to snake_case using path: mappings (id, image, status, state, created_epoch, command) so users never need quoted identifiers
Added top-level description field
Added README with three setup options: socat proxy (recommended), Docker Desktop local-only with security warning, and remote TLS setup linking to Docker's official docs
Added live coral source add and coral sql output as evidence in the README

jsummerfield (changes requested):
Findings

- High: `sources/community/docker/manifest.yaml:35` and the other column entries still use `path:` directly on columns, which is not valid source-spec schema. `coral source lint sources/community/docker/manifest.yaml` rejects every column `path` field as an unexpected property, so this source cannot be installed on the current head. Please switch these to the supported column expression shape, for example `expr: {kind: path, path: [Id]}`. Because the manifest does not parse, the README live output cannot be current-head evidence yet.

- High: `sources/community/docker/manifest.yaml:13` still makes unauthenticated `http://localhost:2375` the default, and `sources/community/docker/README.md:60` says the `socat TCP-LISTEN:2375,reuseaddr,fork UNIX-CLIENT:/var/run/docker.sock` recipe avoids exposing a network port. That command listens on TCP port 2375, and Docker's guidance is to use SSH or TLS for daemon access rather than making unauthenticated TCP the happy path. Please make the safe path the default, bind any local proxy explicitly to loopback if you keep it, and document TLS/SSH as the remote path. Docker docs: https://docs.docker.com/engine/security/protect-access/ and https://docs.docker.com/engine/daemon/remote-access/.

- Medium: `sources/community/docker/manifest.yaml:31` and `sources/community/docker/manifest.yaml:66` call `/containers/json` without exposing Docker's native `filters`/`limit` query params. For a Docker host with many containers, pushing status/name/id/label filters into the Docker Engine API is much better than always fetching the whole list and filtering locally. Docker documents those params on ContainerList: https://docs.docker.com/reference/api/engine/version/v1.43/#tag/Container/operation/ContainerList.

- Medium: `sources/community/docker/README.md:68` says the Docker Desktop unauthenticated `2375` setting applies to Windows/macOS. Docker Desktop documents that setting as Windows Hyper-V only and warns that exposing the daemon without TLS can lead to remote code execution: https://docs.docker.com/desktop/settings-and-maintenance/settings/.

After the manifest installs again, please refresh the captured `coral source add` / query output from that fixed head. Thanks for tightening this up.

jsummerfield:

ayanasarkar:
Thank you so much for your time and for reviewing my work thus far. I completely understand closing the pull request due to the recent inactivity.
​Please rest assured that I will be resuming work on these requested changes very soon. I plan to reopen this pull request shortly once the updates are complete. I truly appreciate your patience and your continued willingness to review the updated source code when it is ready.

PR #1051, opened 2026-05-29
feat(sources/community/ansible): add Ansible facts source

### LINKED ISSUE #1048: design: Add Ansible facts as a file-backed community source
## Summary

I would like to add a new community source named `ansible` that exposes curated Ansible-derived infrastructure data as SQL-queryable Coral tables.

The source is intentionally file-backed and read-only. Users gather infrastructure facts with Ansible, normalize them into JSONL datasets, and Coral exposes those datasets through the `ansible.*` schema.

The goal is to make infrastructure inventory, service state, package inventory, filesystem metadata, network information, and selected security posture data queryable through Coral SQL without exposing raw Ansible payloads, credentials, secrets, or direct host access.

---

## Motivation

Ansible is already widely used as an inventory, configuration management, and infrastructure automation layer across Linux environments.

Operational teams frequently need to answer questions such as:

* Which hosts are running a specific distribution?
* Which hosts use `systemd`, `openrc`, or another service manager?
* Which services are failed, stopped, or missing?
* Which package versions are installed across the fleet?
* Which hosts are under disk pressure?
* Which expected services are missing from a role assignment?
* Which hosts are enforcing SELinux or using AppArmor?

Today, this information often exists inside gathered facts, inventory metadata, or exported operational data, but it is usually spread across multiple files and not easily queryable through a common interface.

By exposing normalized Ansible-derived data through Coral SQL, users and agents can inspect infrastructure state using a single query layer without requiring direct SSH access, playbook execution, or exposure of raw `ansible_facts`.

Without this source, many of these workflows require manual inspection of inventory files, fact exports, or custom scripts.

---

## Proposed design

The proposed source is a file-backed community source named `ansible`.

### Initial tables

```text
ansible.hosts
ansible.services
ansible.packages
ansible.mounts
ansible.interfaces
ansible.security
ansible.roles
```

### High-level flow

```text
Ansible facts
    ↓
Normalizer
    ↓
JSONL datasets
    ↓
Coral source
    ↓
ansible.* SQL tables
```

The source reads normalized JSONL files and exposes them through Coral as SQL tables.

The implementation intentionally avoids exposing raw `ansible_facts` and instead exposes curated operational fields that are useful for:

* inventory inspection
* service-state analysis
* package inspection
* filesystem and disk checks
* network visibility
* security posture inspection
* role-drift detection

The initial implementation is intentionally simple:

* file-backed
* read-only
* local-first
* no SSH execution
* no playbook execution
* no AWX / Automation Platform integration

The implementation also includes:

* source manifest
* synthetic fixtures
* example SQL queries
* fixture validation script
* README
* design notes
* security notes
* example gather/normalize workflow

---

## Drawbacks and trade-offs

This approach is snapshot-based rather than live.

Users must gather and normalize facts before Coral can query them.

The source intentionally excludes large portions of raw Ansible data to reduce security risk and avoid exposing sensitive information.

Because the source uses normalized datasets instead of raw fact payloads, some advanced Ansible-specific metadata may not be available through the initial table design.

The first version also does not include AWX / Automation Platform integration.

---

### Exposing raw `ansible_facts`

I considered exposing raw fact payloads directly.

I rejected this approach because raw facts may contain environment-specific information, sensitive metadata, or values that are difficult to review and query safely.

A curated table design produces a smaller and more reviewable surface while keeping common operational workflows straightforward.

### AWX / Automation Platform API integration

I considered building the source directly on top of AWX / Automation Platform APIs.

I chose a file-backed implementation first because it is simpler to validate locally, works well for homelab and self-hosted environments, and avoids introducing API authentication complexity in the initial version.

### SSH-based fact collection

I considered allowing Coral to gather facts directly from managed hosts.

I rejected this because it significantly expands scope and moves the source beyond inventory querying into infrastructure execution and remote access management.

---

## Open questions

* Should future versions expose AWX / Automation Platform integration as a separate source?
* Should playbook execution history be represented as additional tables in a later version?
* Should container-related metadata such as Podman inventory remain a separate source or eventually integrate with Ansible-derived datasets?
* Are there additional curated tables maintainers would prefer beyond the initial seven-table design?

---

Yash-2003P:
Adds a new community source named `ansible`.

This source exposes sanitized and normalized Ansible fact exports as SQL-queryable Coral tables. It is file-backed and read-only: users gather facts with Ansible, normalize them into JSONL datasets, and Coral exposes those datasets through the `ansible.*` schema.

## What changed

This PR adds a new source under `sources/community/ansible/`. It includes:

* A Coral source manifest
* Seven SQL tables under the `ansible` schema
* Synthetic JSONL fixtures
* A fixture validation script
* An example Ansible inventory
* An example fact-gathering workflow
* A normalizer script
* README documentation
* Design notes
* Security notes
* Example SQL queries
* Contribution workflow notes

## Tables exposed

```text
ansible.hosts
ansible.services
ansible.packages
ansible.mounts
ansible.interfaces
ansible.security
ansible.roles

```

## Design overview

```text
Ansible facts ──> Normalizer ──> JSONL datasets ──> Coral source ──> ansible.* SQL tables

```

The source exposes curated operational data instead of raw `ansible_facts`. The first version is intentionally:

* File-backed and read-only
* Local-first
* Focused on infrastructure inventory and operational visibility
* Designed to avoid direct infrastructure execution from Coral

## Safety model

This source intentionally avoids:

* Raw `ansible_facts`
* Secrets and credentials
* SSH execution
* Playbook execution from Coral
* AWX / Automation Platform API integration
* Real infrastructure data in fixtures (all included fixtures are synthetic)

## Validation performed

**Repository source checks:**

```bash
ryl sources
make lint-sources

```

**Manifest validation:**

```bash
cargo run --locked -p coral-cli -- source lint sources/community/ansible/manifest.yaml

```

**Fixture validation:**

```bash
python sources/community/ansible/tests/validate-fixtures.py sources/community/ansible/fixtures

```

**Local source test with fixture data:**

```bash
cargo run --locked -p coral-cli -- source test ansible

```

## Checklist

* [x] Added source manifest
* [x] Added table definitions
* [x] Added validation queries
* [x] Added synthetic fixtures
* [x] Added fixture validation script
* [x] Added example Ansible inventory
* [x] Added gather-facts example
* [x] Added normalizer script
* [x] Added README
* [x] Added design notes
* [x] Added security notes
* [x] Added example SQL queries
* [x] Added contribution workflow notes
* [x] Verified ryl sources
* [x] Verified make lint-sources
* [x] Verified manifest linting
* [x] Verified fixture validation
* [x] Verified local source test
* [x] Confirmed no real infrastructure facts are committed
* [x] Confirmed no credentials or secrets are committed

## Review focus

Please review:

* Source manifest structure
* Table names and column design
* Fixture quality
* Example SQL usefulness
* README clarity
* Security boundaries
* Validation workflow
* Cross-platform file-path notes

## Future work

Not included in this PR:

* AWX / Automation Platform integration
* Playbook execution history tables
* Playbook run analytics
* Container inventory integrations
* Additional API-backed Ansible sources

Closes #1048


jsummerfield (changes requested):
Findings

- Medium: `sources/community/ansible/manifest.yaml:22` makes every table read from `file:///tmp/coral-ansible-facts/`. That works for a local smoke test, but it is a volatile Unix-specific location for the checked-in product default, and the source's own security notes say durable snapshots should not live in `/tmp`. Please move the default to a durable home-relative location, or make the data directory configurable in the source setup and keep `/tmp` only as an example test path.

- Medium: `sources/community/ansible/scripts/normalize-ansible-facts.py:171` uses `selinux.get("policyvers")` before `selinux.get("type")` for `selinux_policy`. Ansible's SELinux facts expose the policy name separately from the policy version, so this can put a numeric version into a column documented as the policy name. Please map `selinux_policy` from the policy type/name field and add a separate policy-version column only if that value is useful. Reference: https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_vars_facts.html

- Medium: `sources/community/ansible/examples/gather-facts.yml:36` interpolates `inventory_hostname` directly into the output filename. Inventory names can contain path separators or other characters that are poor local filenames, so the example can write outside the intended directory or fail unexpectedly. Please derive a sanitized filename, while keeping the original inventory hostname inside the JSON payload.

- Low: `sources/community/ansible/queries/examples.sql:1` starts with `## --`, which is not SQL comment syntax. Anyone running the example file as SQL will fail before the first query. Please make it a normal SQL comment such as `-- ansible Coral source example SQL`.

- Low: `sources/community/ansible/manifest.yaml:10` only smoke-tests five of the seven advertised tables. Since `interfaces` and `roles` are part of the documented workflow, please add declared test queries for them too.

I like the direction here: file-backed, allowlisted Ansible facts are a useful Coral surface, and I verified the current fixture path with `coral source add --file ...` plus `coral source test ansible`. Once the packaging/default path and the normalizer/example issues are tightened up, this should be much easier for users to trust.

Yash-2003P:
Thanks for the detailed review .

I’ve updated the PR to address the requested changes:
- moved the default data location away from `/tmp`
- fixed SELinux policy mapping
- sanitized the Ansible output filename
- fixed the SQL comment syntax
- added declared source test queries for `interfaces` and `roles`

I also reran validation, and `coral source test ansible` now reports `7 declared · 7 passed · 0 failed`.

Please review

jsummerfield (changes requested):
Findings

- Medium: `sources/community/ansible/scripts/normalize-ansible-facts.py:147` JSON-encodes `ipv6_addresses`, but `sources/community/ansible/manifest.yaml:193` declares the column as `Json` and the fixtures now use real arrays. Users following the normalizer will generate a JSON string instead of the JSON array the schema and examples imply. Please emit the list directly, or change the schema to `Utf8` and document the string shape.

- Medium: `sources/community/ansible/queries/examples.sql:261` checks service names without the systemd suffix, for example `postgresql`, `rabbitmq-server`, and `datadog-agent`, while Ansible `service_facts` commonly reports unit names such as `postgresql.service`, and this PR's own fixtures use `.service` names. That makes the role/service drift example report services as missing even when the data contains them. Please either normalize a service basename column or make the example match the actual `name` values. Docs: https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/service_facts_module.html

- Medium: `sources/community/ansible/README.md:257`-`:277` has a Windows workflow that copies fixture data to `C:\tmp\coral-ansible-facts` while the checked-in manifest still reads `file://~/.coral/ansible-facts/`; the replacement command shown there is a no-op. Please either copy fixtures to the manifest path or have the Windows commands update the manifest to the temp path they actually use.

- Low: `sources/community/ansible/README.md:216`-`:222` and the later validation section still say `5 declared query tests passed`, but the manifest now declares seven tests. Please refresh the captured output so the live evidence matches the current source.

The earlier `/tmp` default, SELinux mapping, filename sanitization, SQL comment, and test-query coverage findings look addressed. This source is useful, but the generated data shape and copy-paste setup path need to match the manifest before it is ready.

jsummerfield (changes requested):
Findings

- Medium: `sources/community/ansible/examples/gather-facts.yml:4` still uses `gather_facts: true`, and `sources/community/ansible/examples/gather-facts.yml:32` exports the entire `ansible_facts` object before the allowlist normalizer runs. The normalized Coral tables are much narrower now, but the user-facing example still writes a broad raw fact dump that can include buckets such as environment, local facts, user facts, and SSH host key facts. Please gather only the subsets needed by this source, or otherwise avoid writing the full raw fact object in the recommended workflow. Ansible setup docs: https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/setup_module.html

- Low: the docs still use `python` in setup/validation commands at `sources/community/ansible/README.md:293`, `sources/community/ansible/README.md:456`, `sources/community/ansible/docs/DESIGN.md:188`, and `sources/community/ansible/SECURITY_NOTES.md:356`, while the scripts use `python3` and local validation failed when `python` was not installed. Please update these commands to `python3` so the first-success path matches the scripts.

The previous blockers look fixed: the source no longer defaults to `/tmp`, and the normalizer now emits real JSON arrays for the JSON columns. The remaining issue is about keeping the example collection workflow as narrow as the product surface. Thanks for tightening this up.

sources/community/ansible/examples/gather-facts.yml
```diff
@@ -0,0 +1,62 @@
+---
+- name: Gather selected facts for Coral normalization
+  hosts: all
+  gather_facts: false
+  become: true
```
jsummerfield: This makes privilege escalation the default for the whole copy-paste gathering path. Please make `become` opt-in, move it only to tasks that truly require it, or document why each privileged task is needed so the example matches the least-privilege guidance.

sources/community/ansible/tests/validate-fixtures.py
```diff
@@ -0,0 +1,79 @@
...
+                    raise ValueError(f"{path}:{lineno} missing required key: {key}")
+            if path.name == "interfaces.jsonl" and not isinstance(row.get("ipv6_addresses", []), list):
+                raise ValueError(f"{path}:{lineno} ipv6_addresses must be a JSON array")
+            count += 1
+    if count == 0:
+        raise ValueError(f"{path} has no rows")
+    return count
+
+
+def main() -> None:
```
jsummerfield, line 128: This validator checks the already-normalized JSONL files, but the user workflow depends on `scripts/normalize-ansible-facts.py`. A small selected-payload fixture that asserts generated JSONL would cover the real transformation path.

jsummerfield (changes requested):
Thanks for the update. The broad raw-facts concern is resolved: the play now disables automatic full fact gathering and exports an allowlisted payload before normalization. The committed docs also now use `python3` in the setup and validation paths.

Findings
- Medium: `sources/community/ansible/examples/gather-facts.yml:5` sets `become: true` for the entire example play, so the copy-paste path gathers setup, service, and package facts with privilege escalation by default. That cuts against the source's own least-privilege guidance in `SECURITY_NOTES.md` and raises the bar for accepting this as a safe community setup path. Please make privilege escalation opt-in, move it only to tasks that truly require it, or document exactly why each privileged task is required. Ansible's privilege-escalation behavior is documented here: https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_privilege_escalation.html
- Low: `sources/community/ansible/tests/validate-fixtures.py:68` validates normalized JSONL fixtures, but it does not exercise `scripts/normalize-ansible-facts.py`. A small selected-payload fixture that asserts the generated JSONL would make future regressions in the real workflow much easier to catch.

I locally verified `coral source lint`, the fixture validator, and `coral source test ansible` against the submitted fixtures; those all pass. Once the default privilege-escalation behavior is tightened up, this should be straightforward to re-review.

sources/community/ansible/tests/validate-fixtures.py
Yash-2003P: - @jsummerfield It would be best if you may help me over the review process , must I look up for the ansible docs more thoroughly . If you can list more which I must look such that it is sorted clearly .
-  Furthermore , any other checking which I must do based on what you delivers me for correction each time , how do I look it myself first before asking you for review . 
- Also , sorry for any inconvience I am causing if any. Anything specified would be more helpful.

jsummerfield (approved):
Thanks for iterating on this. I rechecked the current head and the remaining source-review blockers look resolved: the example play now keeps privilege escalation opt-in and scoped to the remote collection tasks, the fixture validator now exercises the normalizer workflow, and the Ansible docs use the `python3` path that matches the scripts.

Verified locally with `coral source lint`, `make lint-sources`, the Ansible fixture validator, and `coral source test ansible` against a temporary source install. Approved from the community source review side; James will still merge manually.

PR #1054, opened 2026-05-29
Feat source temporal

prasannakumar414:
## Adds a new community source - Temporal
This pull request introduces a new community source for Temporal, enabling users to query Temporal workflow runtime data (namespaces, workflow executions, and schedules) via the Temporal HTTP API. It includes comprehensive documentation and a detailed `manifest.yaml` for source configuration, authentication, and table schemas. The source supports both self-hosted Temporal Server (v1.24+) with support for authentication and various query options.

Key additions and changes:

**Source Introduction and Documentation**
* Added a new `README.md` describing how to set up, configure, and use the Temporal source, including requirements, authentication, example queries, and limitations.

**Source Manifest and Table Schemas**
* Introduced `manifest.yaml` defining the Temporal source, including:
  - Source metadata, input variables, and authentication (support for API key or open clusters).
  - Table definitions for `namespaces`, `workflows`, and `schedules`, each mapping to the Temporal HTTP API endpoints with detailed column schemas, pagination, and filtering options.
  - Example test queries for validating connectivity and configuration.

**Authentication and Connectivity**
* Implemented bearer token authentication using the `TEMPORAL_API_KEY` for secure clusters or Temporal Cloud, with support for unauthenticated self-hosted clusters. [[1]](diffhunk://#diff-85cfac7c4ac3b150184dc7da7969d5e5c08ed7eb6df89b9b78cf1ac562d989a4R1-R340) [[2]](diffhunk://#diff-3a85e710b361a0ab73019bb84ebdb36c8a5437e4291fb887c1d96bef01be11afR1-R192)

**Query and Filtering Support**
* Enabled advanced querying of workflow executions using Temporal Visibility query language via the `query` filter, and enforced required filters for `namespace` where applicable. [[1]](diffhunk://#diff-85cfac7c4ac3b150184dc7da7969d5e5c08ed7eb6df89b9b78cf1ac562d989a4R1-R340) [[2]](diffhunk://#diff-3a85e710b361a0ab73019bb84ebdb36c8a5437e4291fb887c1d96bef01be11afR1-R192)

**Usage Guidance and Examples**
* Provided extensive usage instructions, including setup for both self-hosted and cloud environments, and a variety of example SQL queries for common Temporal data inspection tasks.


jsummerfield (changes requested):
Findings:

- High: `sources/community/temporal/manifest.yaml:293`-`:309` maps schedule `state`, `action`, and `spec` from a non-existent `schedule` object, and `sources/community/temporal/manifest.yaml:320`-`:335` maps `created_at` / `updated_at` from fields that are not returned by `ListSchedules`. Temporal's `ListSchedulesResponse` returns `ScheduleListEntry` rows with `schedule_id`, `memo`, `search_attributes`, and `info`; the list `info` contains `spec`, `workflow_type`, `notes`, `paused`, `recent_actions`, and `future_action_times`, but not the full `schedule` object or create/update timestamps. As written, the most important schedule columns will be null. Please map list-safe fields from `info` or switch to a different pattern that describes each schedule before exposing fields only available from describe responses. Docs: https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/request_response.proto#L1484-L1498 and https://github.com/temporalio/api/blob/master/temporal/api/schedule/v1/message.proto#L365-L393

- High: `sources/community/temporal/README.md:3`, `sources/community/temporal/README.md:32`-`:53`, and `sources/community/temporal/manifest.yaml:7`-`:28` claim Temporal Cloud namespace support through this HTTP source, but Temporal Cloud namespace access is documented around namespace/regional endpoints on `:7233` and the workflow-service HTTP API docs warn that APIs in this file may not be compatible with Temporal Cloud. Please remove Temporal Cloud support from this source unless there is a tested, documented Temporal Cloud HTTP gateway URL for these `/api/v1/...` workflow-service endpoints. Docs: https://docs.temporal.io/cloud/namespaces#what-is-a-cloud-grpc-endpoint, https://docs.temporal.io/cloud/namespaces#how-to-access-a-namespace-in-temporal-cloud, and https://api-docs.temporal.io/

- Medium: `sources/community/temporal/README.md:23`-`:28` shows the wrong self-hosted config shape for enabling the HTTP API. Temporal's documented config nests the HTTP port under `services.frontend.rpc.httpPort`, not a top-level `frontend.httpPort`. Please correct the setup snippet and mention when the user's Docker/Helm distribution already enables the HTTP API. Docs: https://github.com/temporalio/temporal/blob/main/docs/architecture/nexus.md#L271-L277

- Medium: `sources/community/temporal/manifest.yaml:184`-`:190` describes workflow `status` as friendly values like `Running` and `Completed`, but the API returns `WorkflowExecutionStatus` enum strings such as `WORKFLOW_EXECUTION_STATUS_RUNNING`, and the current enum also includes `WORKFLOW_EXECUTION_STATUS_PAUSED`. Please make the description match the returned enum strings or add a derived friendly status column. Docs: https://github.com/temporalio/api/blob/master/temporal/api/workflow/v1/message.proto#L30-L45 and https://github.com/temporalio/api/blob/master/temporal/api/enums/v1/workflow.proto#L72-L82

- Medium: live Coral evidence is missing. I do not see redacted output from `coral source add --file sources/community/temporal/manifest.yaml`, `coral source test temporal`, or representative queries against `namespaces`, `workflows`, and `schedules`. Please add captured Coral output from a real Temporal instance; if Cloud remains documented, include evidence from Cloud too, because that is the riskiest setup claim here.

The manifest lint passes, and Temporal is a useful source for Coral users. The main fix is to make the schedule table match the actual list response, tighten the setup/auth claims, and add live Coral evidence before we treat this as product-ready.

prasannakumar414:
Sure thing! Thanks for the review! @jsummerfield 
I will fix the issues you have raised and will also add few more tables by following temporal's open api documentation.

prasannakumar414:
- Updated the README and manifest based on latest temporal open api documentation.
- I have also added and verified all the queries against local temporal instance by running sample workflows from temporal.
- fixed the issues mentioned in the previous comment.
- Added verified examples in README that I have run against my local temporal instance.

jsummerfield (changes requested):
Findings:

- High: `sources/community/temporal/README.md:7`, `sources/community/temporal/README.md:258`, and `sources/community/temporal/manifest.yaml:6` claim Temporal Server `v1.24+`, but the manifest now exposes newer APIs such as `activities`, `nexus_operations`, `workflow_rules`, and `worker_deployments`. A v1.24 cluster can satisfy the older workflow-service HTTP surface but will not satisfy all current test queries, so users can follow the README and still get 404s or failed tables. Please either raise the documented minimum Temporal Server version to one that supports every exposed table, or narrow this first version back to the tables actually supported by the stated minimum. Docs: https://temporal.io/changelog/v1-24-0 and https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/service.proto

- High: `sources/community/temporal/manifest.yaml:297` maps `schedules.workflow_type` from `[info, workflowType]`, but Temporal's `ScheduleListInfo.workflow_type` is a `WorkflowType` message; the useful string is the nested `name`. This will expose an object/null rather than the workflow type string, and the README's schedule verification has no schedule rows, so it does not prove this mapping. Please change this expression to `[info, workflowType, name]` and add live evidence with at least one schedule row. Docs: https://github.com/temporalio/api/blob/master/temporal/api/schedule/v1/message.proto and https://github.com/temporalio/api/blob/master/temporal/api/common/v1/message.proto

- Medium: `sources/community/temporal/manifest.yaml:190` and `sources/community/temporal/manifest.yaml:523` enumerate workflow statuses but omit `WORKFLOW_EXECUTION_STATUS_PAUSED`, which exists in the current Temporal API. Please include it or avoid closed inline enum wording if the source is targeting a moving Temporal API surface. Docs: https://github.com/temporalio/api/blob/master/temporal/api/enums/v1/workflow.proto

- Medium: `sources/community/temporal/README.md:54` documents only `namespaces`, `workflows`, and `schedules`, while the manifest exposes 10 tables. That leaves most of the product surface undocumented: activities, archived workflows, batch operations, Nexus operations/endpoints, workflow rules, and worker deployments. Please either document the actual table set with setup/version caveats, or narrow the manifest to the three documented tables.

- Low: `sources/community/temporal/README.md:40` and `sources/community/temporal/manifest.yaml:18` use `TEMPORAL_API_KEY` but do not explicitly say Temporal Cloud namespace endpoints are not supported by this source. The source now says self-hosted, which is good, but Cloud users are likely to try their namespace endpoint plus API key. Please add a short note that Temporal Cloud is not supported unless there is tested support for the `/api/v1/...` workflow-service HTTP surface. Docs: https://temporal.io/changelog/namespace-endpoints-ga and https://temporal.io/change-log/cloudops-http-api-public-preview

The source is useful for self-hosted Temporal operators, and the previous Cloud claim has mostly been removed. The remaining blockers are version/API truth, schedule mapping, and documenting the full table surface before this becomes product-ready. Thanks for continuing to iterate on it.

prasannakumar414:
> Findings:
> 
> * High: `sources/community/temporal/README.md:7`, `sources/community/temporal/README.md:258`, and `sources/community/temporal/manifest.yaml:6` claim Temporal Server `v1.24+`, but the manifest now exposes newer APIs such as `activities`, `nexus_operations`, `workflow_rules`, and `worker_deployments`. A v1.24 cluster can satisfy the older workflow-service HTTP surface but will not satisfy all current test queries, so users can follow the README and still get 404s or failed tables. Please either raise the documented minimum Temporal Server version to one that supports every exposed table, or narrow this first version back to the tables actually supported by the stated minimum. Docs: https://temporal.io/changelog/v1-24-0 and https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/service.proto
> * High: `sources/community/temporal/manifest.yaml:297` maps `schedules.workflow_type` from `[info, workflowType]`, but Temporal's `ScheduleListInfo.workflow_type` is a `WorkflowType` message; the useful string is the nested `name`. This will expose an object/null rather than the workflow type string, and the README's schedule verification has no schedule rows, so it does not prove this mapping. Please change this expression to `[info, workflowType, name]` and add live evidence with at least one schedule row. Docs: https://github.com/temporalio/api/blob/master/temporal/api/schedule/v1/message.proto and https://github.com/temporalio/api/blob/master/temporal/api/common/v1/message.proto
> * Medium: `sources/community/temporal/manifest.yaml:190` and `sources/community/temporal/manifest.yaml:523` enumerate workflow statuses but omit `WORKFLOW_EXECUTION_STATUS_PAUSED`, which exists in the current Temporal API. Please include it or avoid closed inline enum wording if the source is targeting a moving Temporal API surface. Docs: https://github.com/temporalio/api/blob/master/temporal/api/enums/v1/workflow.proto
> * Medium: `sources/community/temporal/README.md:54` documents only `namespaces`, `workflows`, and `schedules`, while the manifest exposes 10 tables. That leaves most of the product surface undocumented: activities, archived workflows, batch operations, Nexus operations/endpoints, workflow rules, and worker deployments. Please either document the actual table set with setup/version caveats, or narrow the manifest to the three documented tables.
> * Low: `sources/community/temporal/README.md:40` and `sources/community/temporal/manifest.yaml:18` use `TEMPORAL_API_KEY` but do not explicitly say Temporal Cloud namespace endpoints are not supported by this source. The source now says self-hosted, which is good, but Cloud users are likely to try their namespace endpoint plus API key. Please add a short note that Temporal Cloud is not supported unless there is tested support for the `/api/v1/...` workflow-service HTTP surface. Docs: https://temporal.io/changelog/namespace-endpoints-ga and https://temporal.io/change-log/cloudops-http-api-public-preview
> 
> The source is useful for self-hosted Temporal operators, and the previous Cloud claim has mostly been removed. The remaining blockers are version/API truth, schedule mapping, and documenting the full table surface before this becomes product-ready. Thanks for continuing to iterate on it.

Just to be clear in these issues mentioned, the grpc documentation of temporal cloud has been used instead of http api documentation provided by temporal. Please use this https://github.com/temporalio/api/tree/master/openapi

jsummerfield (changes requested):
Findings

- High: `sources/community/temporal/manifest.yaml:45`-`:54` still makes `coral source test temporal` exercise every version-gated table, while `sources/community/temporal/README.md:7` says the core source works on Temporal Server v1.24+. A v1.24 cluster can cover the older namespace/workflow/schedule/batch APIs, but not the newer activities, workflow-rules, worker-deployment, and Nexus surfaces. Please either raise the source's minimum version/feature requirements to match every default test query, or keep default tests to a v1.24-safe core and document newer tables as optional/manual checks. Docs: https://github.com/temporalio/temporal/blob/v1.24.0/go.mod#L47, https://github.com/temporalio/api/blob/master/openapi/openapiv3.yaml, and https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/service.proto.

- High: `sources/community/temporal/README.md:103`-`:107` documents `nexus_endpoints` as Temporal Server v1.24+, and `manifest.yaml:53` includes it in the default tests. The Nexus endpoint list API is on `OperatorService` and I see it in the API dependency used by Temporal Server v1.27, not the v1.24 API dependency. Please correct the version claim, likely to v1.27+ unless you have a more exact server-version source, and remove it from the v1.24 core/default-test path. Docs: https://github.com/temporalio/temporal/blob/v1.27.0/go.mod#L60 and https://github.com/temporalio/api/blob/v1.44.1/temporal/api/operatorservice/v1/service.proto#L134-L143.

- High: `sources/community/temporal/manifest.yaml:1003`-`:1031` maps `worker_deployments.current_version` and `ramping_version` from `routingConfig.currentDeploymentVersion.buildId` / `rampingDeploymentVersion.buildId`, but `README.md:143`-`:147` says the table works on Temporal Server v1.27+. The v1.27 API shape exposes `routingConfig.currentVersion` and `routingConfig.rampingVersion` strings, not those nested deployment-version objects, so the advertised v1.27 path would return null for key rollout columns. Please either map the v1.27-compatible string fields or raise the table requirement to the server/API version that actually returns the nested objects. Docs: https://github.com/temporalio/api/blob/v1.44.1/temporal/api/deployment/v1/message.proto#L222-L251 and https://github.com/temporalio/api/blob/master/temporal/api/deployment/v1/message.proto#L282-L312.

- Low: `sources/community/temporal/README.md:93`-`:101` says `batch_operations` includes a `reason` column, but `manifest.yaml:615`-`:647` exposes only `job_id`, `state`, `start_time`, `close_time`, and `raw`. Please remove that column from the README list or add a real decoded column if it comes from a different response. Docs: https://github.com/temporalio/api/blob/master/temporal/api/batch/v1/message.proto#L18-L30.

- Low: `sources/community/temporal/README.md:103`-`:111` says `nexus_endpoints` includes `description`, but `manifest.yaml:888`-`:949` does not expose it. Temporal's endpoint spec has `description`; please either add the decoded column or remove it from the README table list. Docs: https://github.com/temporalio/api/blob/master/temporal/api/nexus/v1/message.proto#L154-L169.

The Cloud-support claim, HTTP config shape, schedule mapping, paused workflow status, and broad README coverage from the earlier reviews look improved. The remaining blockers are version/API truth and copyable table docs. Thanks for tightening this up; once those are fixed I can take another pass.

prasannakumar414:
Here is the detailed description of latest updates made
High: default tests vs v1.24 core claim
Addressed. Default test queries now cover only the v1.24-safe core path (namespaces, workflows, schedules, batch_operations). Version-gated tables were removed from default test execution and are now documented as optional/manual checks.

High: nexus_endpoints version claim and core path
Addressed. Updated nexus_endpoints requirement to v1.27+ and removed it from the default v1.24 core test path. It is now included under optional/manual checks with explicit query guidance and expected unavailable behavior.

High: worker_deployments mapping vs advertised v1.27 support
Addressed. Updated current_version/ramping_version mappings to v1.27-compatible routingConfig.currentVersion and routingConfig.rampingVersion, instead of nested deployment-version buildId objects.

Low: batch_operations README mentions reason column
Addressed. Removed reason from the README column list to match exposed manifest columns.

Low: nexus_endpoints README mentions description column
Addressed. Removed description from the README column list and replaced it with the actual exposed columns.

jsummerfield (changes requested):
Findings

- Medium: `sources/community/temporal/README.md:123`, `sources/community/temporal/README.md:185`, and `sources/community/temporal/manifest.yaml:778` describe `workflow_rules` as requiring only Temporal Server v1.28+, but Temporal gates `ListWorkflowRules` behind `frontend.workflowRulesAPIsEnabled`, which defaults to false in v1.28. A user on a v1.28+ server can still get `Unimplemented` unless that dynamic config is enabled. Please document the feature-flag requirement anywhere `workflow_rules` is listed or queried, and consider keeping it in the optional/manual-check path. Docs: https://github.com/temporalio/api/blob/6c8bd137c96b5e1a80b0291fa5e7e70b79726421/temporal/api/workflowservice/v1/service.proto#L1513-L1519, https://github.com/temporalio/temporal/blob/280a0349228c432c892824069c9ddbb33b6086da/service/frontend/workflow_handler.go#L7041-L7049, and https://github.com/temporalio/temporal/blob/v1.28.0/common/dynamicconfig/constants.go#L2707-L2710

- Low: `sources/community/temporal/README.md:400` says earlier Temporal versions do not expose the `/api/v1` HTTP gateway. That is broader than the evidence supports: v1.24 has HTTP bindings for core routes, and older server configs already had an HTTP port. If this source only supports v1.24+, please say earlier versions are unsupported or unverified instead of saying the gateway does not exist. Docs: https://github.com/temporalio/api/blob/v1.24.0/temporal/api/workflowservice/v1/service.proto#L516-L518 and https://github.com/temporalio/api/blob/v1.24.0/temporal/api/workflowservice/v1/service.proto#L602-L605

The previous schedule mapping, Cloud wording, default-test version path, worker deployment mapping, and table-coverage docs look substantially improved. Local `coral source lint` and the repo YAML style check pass at this head. This should be close after the remaining Temporal version/feature-flag wording is corrected. Thanks for continuing to tighten a tricky provider surface.

jsummerfield (changes requested):
Findings

- Medium: `sources/community/temporal/manifest.yaml:805` adds `pageSize` pagination to `workflow_rules`, but Temporal's `ListWorkflowRulesRequest` only has `namespace` and `next_page_token`, and the generated OpenAPI parameters for `GET /api/v1/namespaces/{namespace}/workflow-rules` list only `namespace` and `nextPageToken`. Please remove the `page_size` block for this table and keep cursor pagination to `nextPageToken` only. Docs: https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/request_response.proto#L2879-L2886, https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/service.proto#L1513-L1519, and https://github.com/temporalio/api/blob/master/openapi/openapiv3.yaml#L3346-L3362

The prior workflow-rules feature-flag wording is now documented, and the README no longer overclaims that earlier Temporal versions lack `/api/v1`; it now says they are unsupported/unverified by this source. Local `coral source lint` and `ryl` pass at this head. This should be down to making the `workflow_rules` request match Temporal's HTTP API parameters. Thanks for the quick cleanup.

jsummerfield (changes requested):
Findings

- High: `sources/community/temporal/manifest.yaml:805` still sends `pageSize` for `temporal.workflow_rules`. Temporal's `ListWorkflowRulesRequest` only includes `namespace` and `next_page_token`, and the HTTP binding is `/api/v1/namespaces/{namespace}/workflow-rules`; it does not document a `page_size` / `pageSize` parameter for this endpoint. Please remove the `page_size` block for `workflow_rules` and keep cursor pagination to `nextPageToken` only, or point to an API contract that documents `pageSize` for this endpoint. Docs: https://github.com/temporalio/api/blob/6c8bd137c96b5e1a80b0291fa5e7e70b79726421/temporal/api/workflowservice/v1/request_response.proto#L2879-L2882 and https://github.com/temporalio/api/blob/6c8bd137c96b5e1a80b0291fa5e7e70b79726421/temporal/api/workflowservice/v1/service.proto#L1513-L1520

- Medium: `sources/community/temporal/manifest.yaml:50` hard-codes `namespace = 'default'` in the default `test_queries`. The README now scopes this source to self-hosted Temporal clusters, but valid self-hosted clusters do not have to define a `default` namespace, so `coral source test temporal` can fail even when the source is configured correctly. Please keep the default validation path to namespace discovery only, or add an explicit namespace input and use that in the test queries.

`coral source lint` and repo YAML lint pass on this head, and the self-hosted/Temporal Cloud guidance is much clearer. These are the remaining provider-contract and first-success blockers. Thanks for continuing to refine it.

sources/community/temporal/manifest.yaml
```diff
@@ -0,0 +1,1055 @@
...
+          label: Paste API key
+
+base_url: "{{input.TEMPORAL_ADDRESS}}"
+
+auth:
+  type: HeaderAuth
+  headers:
+    - name: Authorization
+      from: template
+      template: "Bearer {{input.TEMPORAL_API_KEY}}"
```
jsummerfield: This template still makes `TEMPORAL_API_KEY` effectively required at registration time. Coral omits empty optional secrets, then validates top-level auth templates before running requests, so the README’s unauthenticated self-hosted setup path cannot work with this `HeaderAuth` block.

jsummerfield (changes requested):
Findings

- High: `sources/community/temporal/manifest.yaml:38`-`43` still makes the open self-hosted setup path unusable. The README says users can leave `TEMPORAL_API_KEY` blank for unauthenticated clusters, but the source has top-level `HeaderAuth` that always renders `Bearer {{input.TEMPORAL_API_KEY}}`. Coral omits empty optional secrets during `source add`, and HTTP source registration validates auth templates before requests run, so an open cluster install will fail with a missing `TEMPORAL_API_KEY` instead of querying `/api/v1/namespaces`. Please choose one supported path: either make the token required and remove the open-cluster docs, or remove the top-level auth path and document auth-enabled clusters as unsupported until the source can express conditional auth.

The prior `workflow_rules` `pageSize` issue is resolved, the default source test no longer hard-codes `default`, and the current manifest passes local source lint. This remaining issue is about making the documented first-success path actually install for the common unauthenticated self-hosted case.

sources/community/temporal/manifest.yaml
prasannakumar414: understood now working on it!

prasannakumar414: I am making changes to support open clusters for now. Since self hosted temporal most of the time will be behind a private network.

prasannakumar414:
@cubic-dev-ai review

jsummerfield (approved):

The previous auth/setup blocker is resolved: the manifest no longer declares `TEMPORAL_API_KEY` or sends a top-level `Authorization` header, and the README now scopes this source to open unauthenticated self-hosted Temporal clusters. That makes the documented first-success path installable for the common local/self-hosted setup.

I also rechecked the current head: changed files are limited to `sources/community/temporal/**`, there is no existing core/community Temporal source conflict, live Coral evidence is present in the README, `coral source lint` passes locally, YAML style lint passes, and the current checks are green. Thanks for the steady iteration on a tricky provider surface.

Issue #1045, opened 2026-05-29
Direction for AWS temporary credentials in `aws_sigv4` sources


ShamithaReddy:
### Summary

The bundled AWS sources (`sources/core/cloudwatch_logs`, `sources/core/cloudwatch_metrics`, and the in-progress `sources/community/aws_cost_explorer`) only support **long-term IAM user keys** — the two-piece `(AKID, secret)` credential pattern.

They cannot use **temporary credentials** — the three-piece `(AKID, secret, session_token)` pattern that AWS hands out for `sts assume-role`, IAM Identity Center, EKS IRSA, EC2 instance roles, and GitHub Actions OIDC. Those are the credential models AWS itself recommends.

The signer side already supports session tokens. The manifest path to opt into them does not, and the obvious workaround (declare an optional `AWS_SESSION_TOKEN` input) hits a validation error.

I'd like a direction signal from maintainers before opening a PR, because the choice has follow-on implications for `CustomAuth` more broadly.

---

### Why session tokens matter

AWS issues two flavors of credentials:

| Flavor | Shape | When you get it |
|---|---|---|
| **Long-term** | `AKID + secret` (AKID starts with `AKIA`) | IAM user with permanent access keys |
| **Temporary** | `AKID + secret + session_token` (AKID starts with `ASIA`) | `sts assume-role`, IAM Identity Center / SSO, EKS IRSA, EC2 instance role, GitHub Actions OIDC |

The session token is sent to AWS as the `X-Amz-Security-Token` HTTP header. Without it, AWS rejects requests signed with temporary credentials.

---

### What works today and what doesn't

**The signer side is already wired up.** `crates/auth/aws/src/lib.rs` defines:

```rust
struct AwsSigV4Config {
    service: String,
    region: ParsedTemplate,
    access_key_id: ParsedTemplate,
    secret_access_key: ParsedTemplate,
    #[serde(default)]
    session_token: Option<ParsedTemplate>,   // already optional
}
```

…and `authenticate(...)` already forwards the rendered token into `Credentials::new`, where `aws-sigv4` emits `X-Amz-Security-Token` whenever it's `Some`.

**The manifest path is not.** Neither bundled manifest declares `AWS_SESSION_TOKEN` as an input or threads it into `auth.session_token`. So the renderer never has a chance to produce a session token, and the long-term-keys-only behavior is what users get.

[...]

PR #1062, opened 2026-05-30
feat(sources/community/aws_cost_explorer): add AWS Cost Explorer source

ShamithaReddy:
# feat(sources/community/aws_cost_explorer): add Cost Explorer source

Adds a new read-only community source for AWS Cost Explorer under `sources/community/aws_cost_explorer/`, exposing cost and usage metrics, rightsizing recommendations, Savings Plans coverage, anomaly detection, and cost forecasting via the Cost Explorer API.

## What's included

`manifest.yaml` — six tables, all using the `aws_sigv4` authenticator with `service: ce`, the same auth pattern as `cloudwatch_logs` and `cloudwatch_metrics`:

| Table | Purpose | API call |
|---|---|---|
| `cost_by_service` | Per-billing-period cost grouped by AWS service | `GetCostAndUsage` (group by `SERVICE`) |
| `cost_by_tag` | Per-billing-period cost grouped by a resource tag value | `GetCostAndUsage` (group by `TAG`) |
| `rightsizing_recommendations` | EC2 downsize / terminate recommendations | `GetRightsizingRecommendation` |
| `savings_plans_coverage` | Savings Plans coverage by service | `GetSavingsPlansCoverage` |
| `anomalies` | ML-detected cost spikes with root-cause breakdown | `GetAnomalies` |
| `cost_forecast` | Projected spend with prediction-interval bounds | `GetCostForecast` |

`README.md` walks the user through:

1. **Authentication setup** — IAM policy with the five required `ce:*` actions (read-only, `Resource: "*"` because Cost Explorer doesn't support resource-level scoping).
2. **Enabling Cost Explorer in the console** — required because the API returns `OptInRequired` until activated, with up to 24 hours of activation lag.
3. **Generating IAM access keys** for an IAM user with the policy attached.
4. **Adding the source** — `coral source add --file sources/community/aws_cost_explorer/manifest.yaml` with environment variables, including partition-aware `AWS_REGION` and `AWS_ENDPOINT_SUFFIX` overrides for GovCloud and China.

The README also documents:
- A summary table of every table's required filters.
- The per-metric NULL contract (`metric = 'AMORTIZED_COST'` populates `amortized_cost` and leaves `unblended_cost` NULL on the same row, etc.) with a concrete example.
- One worked SQL example per table, plus `json_get_json` patterns for the `groups` and `root_causes` JSON arrays.
- A "Notes" section covering data lag, the exclusive-end-date convention, retention windows per table, the Utf8 cast contract, tag activation, and the Cost Explorer enable requirement.

## Implementation notes

- **Region pinning.** Cost Explorer is global but routed through `us-east-1` for standard partitions. The manifest defaults `AWS_REGION` to `us-east-1` and `AWS_ENDPOINT_SUFFIX` to `amazonaws.com`, with the README documenting `us-gov-west-1` / `cn-northwest-1` overrides.
- **Per-metric typed columns on `cost_by_service` / `cost_by_tag`.** AWS keys the response under the requested metric (`Total.UnblendedCost.Amount`, `Total.AmortizedCost.Amount`, etc.). The source exposes seven typed columns (`unblended_cost`, `amortized_cost`, `blended_cost`, `net_unblended_cost`, `net_amortized_cost`, `usage_quantity`, `normalized_usage_amount`) plus a coalesced `period_total_cost`. Whichever metric the caller filters on populates its matching column; the other six are NULL on the same row. `period_total_cost` and `unit` always reflect the requested metric.
- **Filter-bound metric / granularity.** `metric` and `granularity` are filter-bound with virtual columns on `cost_by_service` and `cost_by_tag`, so `WHERE metric = 'AMORTIZED_COST'` routes the value into `Metrics[0]` of the request body. Defaults are `UNBLENDED_COST` and `MONTHLY` to match the AWS CLI canon. AWS rejects lowercase enums, so values are documented in upper case.
- **`rightsizing_recommendations` projections.** `cpu_utilization_pct` and `memory_utilization_pct` project from `CurrentInstance.ResourceUtilization.EC2ResourceUtilization.{Max…UtilizationPercentage}`. `estimated_monthly_savings` coalesces `ModifyRecommendationDetail.TargetInstances[0].EstimatedMonthlySavings` with `TerminateRecommendationDetail.EstimatedMonthlySavings` so TERMINATE recommendations expose savings instead of NULL. `recommendation_target` and `benefits_considered` are filter-bound for queries like `WHERE recommendation_target = 'CROSS_INSTANCE_FAMILY'`.
- **`savings_plans_coverage`.** Exposes `spend_covered_by_savings_plans`, `total_cost`, `period_start`, `period_end` alongside `coverage_pct` and `on_demand_cost`. Pagination is `cursor_body` over `NextToken` with `MaxResults: 100` to avoid the server-side default of 20.
- **Required-filter enforcement.** All seven required filters across the six tables surface `MISSING_REQUIRED_FILTER` from the engine before any HTTP request reaches AWS.
- **Unit-less aggregates.** `usage_quantity` and `normalized_usage_amount` aggregates are unit-less by AWS design (the API sums quantities across services with different natural units), so `unit` returns `'N/A'` for those metrics. Documented on the column descriptions and in the README Notes section.
- **Retention.** 14 months daily / 38 months monthly retention is documented in the README. Queries with `time_period_start` older than the retention window surface AWS's `ValidationException` cleanly.

## Verification

`coral source add`:

```text
Added source aws_cost_explorer (secrets: keychain)
✓ aws_cost_explorer connected successfully
Secrets: keychain
aws_cost_explorer (6 tables)
├─ anomalies
├─ cost_by_service
├─ cost_by_tag
├─ cost_forecast
├─ rightsizing_recommendations
└─ savings_plans_coverage
Query tests
✓ SELECT instance_id, instance_type, recommendation_type, monthly_cost FROM aws_cost_explorer.rightsizing_recommendations LIMIT 5
0 rows
```

`coral source test`:

```text
✓ aws_cost_explorer connected successfully
Secrets: keychain
aws_cost_explorer (6 tables)
├─ anomalies
├─ cost_by_service
├─ cost_by_tag
├─ cost_forecast
├─ rightsizing_recommendations
└─ savings_plans_coverage
Query tests
```

**Live query — per-metric NULL contract on `cost_by_service`:**

```sql
SELECT period_total_cost, unit, amortized_cost, unblended_cost
FROM aws_cost_explorer.cost_by_service
WHERE time_period_start = '2025-09-01'
  AND time_period_end   = '2025-10-01'
  AND metric            = 'AMORTIZED_COST'
```

| period_total_cost | unit | amortized_cost | unblended_cost |
|---|---|---|---|
| 0 | USD | 0 | NULL |

Switching to the default metric (no `metric` filter) populates `unblended_cost` and leaves `amortized_cost` NULL — same row shape, opposite NULL pattern.

**Live query — DAILY granularity row multiplication:**

```sql
SELECT period_start, period_end, unblended_cost
FROM aws_cost_explorer.cost_by_service
WHERE time_period_start = '2025-09-01'
  AND time_period_end   = '2025-09-04'
  AND granularity       = 'DAILY'
```

| period_start | period_end | unblended_cost |
|---|---|---|
| 2025-09-01 | 2025-09-02 | 0 |
| 2025-09-02 | 2025-09-03 | 0 |
| 2025-09-03 | 2025-09-04 | 0 |

**Live query — `recommendation_target` filter binding:**

```sql
SELECT instance_id, recommendation_target
FROM aws_cost_explorer.rightsizing_recommendations
WHERE recommendation_target = 'CROSS_INSTANCE_FAMILY'
LIMIT 5
```

Empty (test account has no EC2). The same query with an invalid value surfaces AWS's enum validation:

```text
ValidationException: 1 validation error detected: Value 'NOT_A_REAL_VALUE' at 'configuration.recommendationTarget'
failed to satisfy constraint: Member must satisfy enum value set: [CROSS_INSTANCE_FAMILY, SAME_INSTANCE_FAMILY]
```

Confirms the new filter binding routes through the request body.

**Required-filter enforcement (Coral-side, before HTTP):**

```text
Error: aws_cost_explorer.cost_forecast requires `WHERE granularity = <constant>`
Detail: aws_cost_explorer.cost_forecast requires a constant equality filter on granularity
```

**12-month retention boundary (AWS-side validation surfaced cleanly):**

```sql
SELECT period_total_cost
FROM aws_cost_explorer.cost_by_service
WHERE time_period_start = '2020-01-01'
  AND time_period_end   = '2020-02-01'
```

```text
Source rejected the request (400)
Detail: ValidationException: You haven't enabled historical data beyond 14 months.
```

**`anomalies` cartesian — 8 combinations of optional filters** (`feedback ∈ {YES, NO, PLANNED_ACTIVITY}` × `total_impact_min` × `monitor_arn`), plus invalid-feedback rejection by AWS:

```text
ValidationException: 1 validation error detected: Value 'BOGUS' at 'feedback' failed to satisfy constraint:
Member must satisfy enum value set: [NO, YES, PLANNED_ACTIVITY]
```

**`cost_forecast` cartesian — 5 valid metrics × {DAILY, MONTHLY} = 10 round-trips.** All reach AWS. All return `DataUnavailableException` on this test account because it lacks historical data, but the 10 identical errors prove uniform filter binding across the metric × granularity matrix. `NORMALIZED_USAGE_AMOUNT` is rejected with the documented "does not support requests for NORMALIZED_USAGE_AMOUNT" message; `HOURLY` granularity is rejected with "Forecast requests not currently supported for HOURLY".

## Acknowledged limitation: temporary AWS credentials

This source supports only **long-term IAM user access keys** (`AWS_ACCESS_KEY_ID` + `AWS_SECRET_ACCESS_KEY`). It cannot authenticate with temporary credentials from `aws sts assume-role`, IAM Identity Center / SSO, EKS IRSA, EC2 instance roles, or GitHub Actions OIDC.

The signer side already supports session tokens — `AwsSigV4Authenticator` accepts an optional `session_token: ParsedTemplate` and forwards it to `aws-sigv4`'s `Credentials::new`. The blocker is in the renderer: declaring `AWS_SESSION_TOKEN` as `kind: secret, required: false` and threading it into `auth.session_token` fails at `coral source add` with:

```text
custom authenticator 'aws_sigv4' failed: missing source input 'AWS_SESSION_TOKEN' for template token
```

`required: false` on a secret input is correctly accepted by `coral_spec::resolve_inputs`, but `render_input_template` in the SigV4 authenticator treats every `{{input.X}}` reference as mandatory regardless of how the input was declared. So "configured but absent" and "not configured" are indistinguishable to the renderer.

The same gap affects `cloudwatch_logs` and `cloudwatch_metrics`. Reproduction and design discussion in **#1045**. Once the renderer fix lands, all three AWS sources can add `AWS_SESSION_TOKEN` as `required: false` in a single follow-up PR — no schema changes here would need to be revisited.

This PR ships the long-term-keys path because it's the existing supported pattern across the bundled AWS sources; deferring the source until #1045 resolves would block usable Cost Explorer querying on a discussion that may take a while to settle. The README documents the IAM-user setup explicitly so users aren't surprised when assume-role / SSO doesn't work today.

## Why ship this now vs. wait for #1045

- The renderer gap is repo-wide, not Cost-Explorer-specific. Two existing bundled sources (`cloudwatch_logs`, `cloudwatch_metrics`) ship with the same limitation.
- The chosen direction in #1045 (relax renderer vs. typed value sources vs. `aws-config` chain) is a maintainer call with follow-on implications for `CustomAuth` more broadly. Whichever direction wins, the manifest in this PR remains valid — only the input declaration and `auth.session_token` line would be added.
- Users running on long-term keys (still common for Cost Explorer specifically, since Cost Explorer is often consumed from an admin-only billing IAM user) are unblocked today.

### DIFF
--- a/sources/community/aws_cost_explorer/README.md
+++ b/sources/community/aws_cost_explorer/README.md
@@ -33,8 +33,17 @@ Create a policy with the following permissions (all read-only):
 }
 ```
 
-Cost Explorer IAM actions only support `Resource: "*"` — resource-level
-scoping is not available.
+The example policy uses `Resource: "*"` for simplicity, which works for
+every action this source calls. Some Cost Explorer actions also support
+resource-level scoping and tag-based conditions if you want to tighten the
+policy: `ce:GetCostAndUsage` and `ce:GetCostForecast` support the
+`billingview` resource type, `ce:GetAnomalies` is scoped to the
+`anomalymonitor` resource type, and these actions accept the
+`aws:ResourceTag/${TagKey}` condition key. `ce:GetRightsizingRecommendation`
+and `ce:GetSavingsPlansCoverage` have no resource type and require
+`Resource: "*"`. See the
+[Cost Explorer service authorization reference](https://docs.aws.amazon.com/service-authorization/latest/reference/list_awscostexplorerservice.html)
+for the full action-to-resource mapping.
 
 ### Step 2 — Enable Cost Explorer
 
@@ -74,6 +83,16 @@ and AWS GovCloud (US) does not expose a Cost Explorer endpoint. Coral
 does not read your AWS CLI profile, AWS SSO cache, or `AWS_PROFILE` —
 credentials must be supplied via the four environment variables above
 (or `coral source add --interactive`).
+
+> **Billable API.** Cost Explorer API requests are billed by AWS at
+> $0.01 per paginated request ([pricing](https://aws.amazon.com/aws-cost-management/aws-cost-explorer/pricing/)),
+> independent of the free Cost Explorer console. Coral follows
+> `NextPageToken`/`NextToken` until the result set is exhausted, so a
+> single broad query can issue many billable requests: `cost_by_service`
+> and `cost_by_tag` page up to 100 requests, `anomalies` and
+> `savings_plans_coverage` up to 50, and `rightsizing_recommendations` up
+> to 20. Narrow the time window, filter, or add `LIMIT` to keep wide scans
+> from fanning out into dozens of charges.
 
 ## Tables
 
@@ -121,7 +140,7 @@ WHERE time_period_start = '2026-05-01'
   AND time_period_end   = '2026-06-01'
 ```
 
-To work with individual services, use `json_get_json` on the `groups` column:
+To inspect a single element, use `json_get_json` on the `groups` column:
 
 ```sql
 SELECT
@@ -133,6 +152,32 @@ FROM aws_cost_explorer.cost_by_service
 WHERE time_period_start = '2026-05-01'
   AND time_period_end   = '2026-06-01'
 ```
+
+To analyze **all** services rather than just the first element, expand the
+`groups` array into one row per service with
+`unnest(json_get_array(groups))`, then project the fields you need. This is
+how you answer questions like "top services by spend":
+
+```sql
+SELECT
+    json_get_str(g, 'Keys', 0)                              AS service,
+    CAST(json_get_str(g, 'Metrics', 'UnblendedCost', 'Amount') AS DOUBLE) AS cost_usd
+FROM (
+    SELECT unnest(json_get_array(groups)) AS g
+    FROM aws_cost_explorer.cost_by_service
+    WHERE time_period_start = '2026-05-01'
+      AND time_period_end   = '2026-06-01'
+)
+ORDER BY cost_usd DESC
+LIMIT 10
+```
+
+`json_get_array(groups)` returns the array as a list, `unnest` explodes it to
+one row per element, and `json_get_str(g, 'Keys', 0)` / `json_get_str(g,
+'Metrics', '<metric>', 'Amount')` read each element. Match the metric key to
+the `metric` filter you bound (e.g. `'AmortizedCost'` when
+`metric = 'AmortizedCost'`); it defaults to `UnblendedCost`. Wrap the
+extracted amount in `CAST(... AS DOUBLE)` before sorting or aggregating.
 
 ### `aws_cost_explorer.cost_by_tag`
 
@@ -152,6 +197,26 @@ FROM aws_cost_explorer.cost_by_tag
 WHERE time_period_start = '2026-05-01'
   AND time_period_end   = '2026-06-01'
   AND tag_key           = 'aws:cloudformation:stack-name'
+```
+
+To rank all tag values by spend, expand the `groups` array the same way as
+`cost_by_service` — `unnest(json_get_array(groups))` gives one row per tag
+value. Resources with no value for the tag appear with an empty string as
+the `Keys` value:
+
+```sql
+SELECT
+    json_get_str(g, 'Keys', 0)                              AS tag_value,
+    CAST(json_get_str(g, 'Metrics', 'UnblendedCost', 'Amount') AS DOUBLE) AS cost_usd
+FROM (
+    SELECT unnest(json_get_array(groups)) AS g
+    FROM aws_cost_explorer.cost_by_tag
+    WHERE time_period_start = '2026-05-01'
+      AND time_period_end   = '2026-06-01'
+      AND tag_key           = 'aws:cloudformation:stack-name'
+)
+ORDER BY cost_usd DESC
+LIMIT 10
 ```
 
 ### `aws_cost_explorer.anomalies`
@@ -210,8 +275,11 @@ a prediction beyond. Supported metrics are narrower than `cost_by_service`
 `NET_AMORTIZED_COST`, and `NET_UNBLENDED_COST`.
 
 ```sql
--- Forecast from yesterday through end of next month. Use yesterday's
--- date as the start because earlier starts are rejected by AWS.
+-- Forecast from yesterday through the end of next month. Replace the
+-- dates below with your own run: time_period_start must be "yesterday"
+-- (AWS rejects earlier starts), and time_period_end is exclusive. The
+-- literals here are illustrative for a run on 2026-05-31; substitute the
+-- current yesterday/next-month-start when you query.
 SELECT
     period_start,
     CAST(mean_value AS DOUBLE)                       AS expected_usd,
@@ -223,6 +291,15 @@ WHERE time_period_start         = '2026-05-30'
   AND metric                    = 'UNBLENDED_COST'
   AND prediction_interval_level = 80
 ```
+
+The filters also accept computed expressions, so you can avoid hand-editing
+dates — for example
+`time_period_start = CAST(current_date - INTERVAL '1 day' AS VARCHAR)`.
+Note that "yesterday" is evaluated against AWS's service clock: if the host
+clock lags UTC, a computed `current_date - 1 day` can land one day before
+AWS's earliest supported start and return
+`ValidationException ("Earliest supported Start is …")`. If that happens,
+use the date from the error message (AWS's "yesterday") as the start.
 
 ### `aws_cost_explorer.rightsizing_recommendations`
 
@@ -276,6 +353,11 @@ ORDER BY CAST(on_demand_cost AS DOUBLE) DESC
 
 ## Notes
 
+- **Billable requests:** Cost Explorer API calls cost $0.01 per paginated
+  request. Broad queries paginate (up to 100 requests for `cost_by_service`
+  and `cost_by_tag`), so narrow the window or add `LIMIT` to avoid
+  unexpected charges. See the billable-API note under
+  [Authentication](#authentication).
 - **Data lag:** Cost Explorer data has a ~24-hour lag. The current open month
   is marked `estimated = true`.
 - **End date is exclusive:** `time_period_end = '2026-06-01'` includes all of
--- a/sources/community/aws_cost_explorer/README.md
+++ b/sources/community/aws_cost_explorer/README.md
@@ -46,6 +46,16 @@ After activation, the API becomes available within 24 hours.
 
 Attach the policy to an IAM user and generate an access key pair.
 
+> **Long-term IAM keys only.** This source signs requests with the
+> `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` pair as-is. Temporary
+> credentials that require a session token — STS `AssumeRole`, IAM
+> Identity Center / SSO, EC2 instance roles, EKS / IRSA — are **not
+> supported** because `AWS_SESSION_TOKEN` is not threaded through the
+> SigV4 authenticator yet. Access keys whose ID starts with `ASIA` are
+> session-scoped and will fail with `InvalidClientTokenId` or
+> `SignatureDoesNotMatch`. Use a long-term IAM user key (the ID starts
+> with `AKIA`).
+
 ### Step 4 — Add the source
 
 ```sh
--- a/sources/community/aws_cost_explorer/README.md
+++ b/sources/community/aws_cost_explorer/README.md
@@ -1,0 +1,277 @@
+# AWS Cost Explorer
+
+Query AWS billing data including cost by service, cost by resource tag,
+detected cost anomalies, cost forecasts, EC2 rightsizing recommendations, and
+Savings Plans coverage from
+[AWS Cost Explorer](https://aws.amazon.com/aws-cost-management/aws-cost-explorer/).
+
+## Authentication
+
+AWS Cost Explorer uses AWS Signature Version 4. You need an IAM user with
+read-only Cost Explorer permissions.
+
+### Step 1 — Create an IAM policy
+
+Create a policy with the following permissions (all read-only):
+
+```json
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ce:GetCostAndUsage",
+        "ce:GetAnomalies",
+        "ce:GetCostForecast",
+        "ce:GetRightsizingRecommendation",
+        "ce:GetSavingsPlansCoverage"
+      ],
+      "Resource": "*"
+    }
+  ]
+}
+```
+
+Cost Explorer IAM actions only support `Resource: "*"` — resource-level
+scoping is not available.
+
+### Step 2 — Enable Cost Explorer
+
+Cost Explorer must be activated before the API returns data. In the AWS
+console, go to **Billing and Cost Management → Cost Explorer → Enable**.
+After activation, the API becomes available within 24 hours.
+
+### Step 3 — Create access keys
+
+Attach the policy to an IAM user and generate an access key pair.
+
+### Step 4 — Add the source
+
+```sh
+export AWS_ACCESS_KEY_ID="AKIA..."
+export AWS_SECRET_ACCESS_KEY="..."
+# Optional — defaults shown:
+export AWS_REGION="us-east-1"             # us-gov-west-1 for GovCloud, cn-northwest-1 for China
+export AWS_ENDPOINT_SUFFIX="amazonaws.com"  # amazonaws.com.cn for China
+coral source add --file sources/community/aws_cost_explorer/manifest.yaml
+```
+
+`AWS_REGION` and `AWS_ENDPOINT_SUFFIX` default to `us-east-1` and
+`amazonaws.com`. Override them only for AWS GovCloud (US) or the China
+partition. Coral does not read your AWS CLI profile, AWS SSO cache, or
+`AWS_PROFILE` — credentials must be supplied via the four environment
+variables above (or `coral source add --interactive`).
+
+## Tables
+
+| Table | Description | Required filters |
+|---|---|---|
+| `aws_cost_explorer.cost_by_service` | Spend grouped by AWS service | `time_period_start`, `time_period_end` |
+| `aws_cost_explorer.cost_by_tag` | Spend grouped by a resource tag key | `time_period_start`, `time_period_end`, `tag_key` |
+| `aws_cost_explorer.anomalies` | Cost anomalies detected by AWS's ML models | `date_interval_start`, `date_interval_end` |
+| `aws_cost_explorer.cost_forecast` | Projected cost for a future or in-progress period | `time_period_start`, `time_period_end`, `granularity`, `metric` |
+| `aws_cost_explorer.rightsizing_recommendations` | EC2 downsize/terminate recommendations | — |
+| `aws_cost_explorer.savings_plans_coverage` | Savings Plans coverage by service | `time_period_start`, `time_period_end` |
+
+`cost_by_service` and `cost_by_tag` accept an optional `metric` filter
+(`UNBLENDED_COST`, `BLENDED_COST`, `AMORTIZED_COST`, `NET_AMORTIZED_COST`,
+`NET_UNBLENDED_COST`, `USAGE_QUANTITY`, `NORMALIZED_USAGE_AMOUNT`) and an
+optional `granularity` filter (`MONTHLY`, `DAILY`, `HOURLY`). Both default to
+`UNBLENDED_COST` and `MONTHLY` when omitted. `HOURLY` is opt-in on the AWS
+account and limited to the last 14 days. Per-metric typed columns
+(`unblended_cost`, `blended_cost`, `amortized_cost`, `net_amortized_cost`,
+`net_unblended_cost`, `usage_quantity`, `normalized_usage_amount`) are
+populated only for the requested `metric`; the other six are NULL on the
+same row. `period_total_cost` and `unit` always reflect the requested
+metric.
+
+For example, `metric = 'AMORTIZED_COST'` produces:
+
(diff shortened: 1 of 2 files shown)


jsummerfield (changes requested):
Findings

- High: `sources/community/aws_cost_explorer/manifest.yaml:90`-`:122` and `:329`-`:361` use `UNBLENDED_COST` / `AMORTIZED_COST` enum values for `GetCostAndUsage.Metrics`. AWS documents `GetCostAndUsage` metric names as `UnblendedCost`, `AmortizedCost`, and similar CamelCase values, so the default `cost_by_service` / `cost_by_tag` queries are likely to fail validation before returning data. The README examples at `sources/community/aws_cost_explorer/README.md:87`-`:107` teach the same wrong values. Docs: https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_GetCostAndUsage.html

- High: `sources/community/aws_cost_explorer/manifest.yaml:1135`-`:1142` reads Savings Plans coverage from `Coverage.CoverageHours.CoverageHoursPercentage`, but `GetSavingsPlansCoverage` returns `Coverage.CoveragePercentage` directly. This makes `coverage_pct` null and breaks the README's primary coverage example. Docs: https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_GetSavingsPlansCoverage.html

- Medium: `sources/community/aws_cost_explorer/README.md:49`-`:57` says temporary AWS credentials are unsupported because `AWS_SESSION_TOKEN` is not threaded through SigV4. Coral's AWS SigV4 auth supports a session token, so this source should add an optional `AWS_SESSION_TOKEN` secret input and wire it into `auth.aws_sigv4.session_token`. Without that, SSO/AssumeRole/CI users are pushed toward long-lived IAM user keys.

- Medium: `sources/community/aws_cost_explorer/manifest.yaml:757`-`:762` and `sources/community/aws_cost_explorer/README.md:192`-`:196` say `GetCostForecast` start dates must be today or future, but AWS documents the start date as equal to or no later than the current date. Please align the guide with the provider docs or add live evidence if AWS's wording behaves differently in practice. Docs: https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_GetCostForecast.html

- Medium: `sources/community/aws_cost_explorer/manifest.yaml:300`-`:304` and `sources/community/aws_cost_explorer/README.md:141`-`:144` say `aws:cloudformation:stack-name` is automatically activated for `cost_by_tag`. AWS-generated tags still need to be activated as cost allocation tags before they appear in Cost Explorer, so please document that activation step. Docs: https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/cost-alloc-tags.html

- Low: local source lint also reports `sources/community/aws_cost_explorer/manifest.yaml:54` as too long, so please run the repo source lint before the next review pass.

AWS Cost Explorer is useful product surface, but these provider-shape mismatches would make important default workflows fail or return nulls. Thanks for the broad first pass; once the API names/paths and credential guidance are corrected, this should be much easier to re-review.

ShamithaReddy:
@jsummerfield 
Thanks for the thorough review. Pushed 37ef83f addressing the findings end-to-end against live AWS Cost Explorer with a least-privilege IAM key. Breakdown below.

## Resolved per the review

**High — UNBLENDED_COST / AMORTIZED_COST enum case in `cost_by_service` / `cost_by_tag`.** Switched to CamelCase (`UnblendedCost`, `AmortizedCost`, etc.) at the call sites and in the surrounding docs to match the documented `GetCostAndUsage.Metrics` shape. Verified live: a 1-year DAILY query against the test account returned 365 distinct rows with `Metrics.UnblendedCost.Amount` populated. UPPER_SNAKE remains in the `cost_forecast` section because that operation's `Metric` field is genuinely a UPPER_SNAKE enum (`Metric` shape in `service-2.json`), distinct from `MetricName` used by `GetCostAndUsage`. Also dug into the AWS error surface — `GetCostAndUsage` actually accepts both forms in practice (its `ValidationException` lists UPPER_SNAKE values), but CamelCase is what the public reference documents and what the AWS SDKs send, so that's what the manifest uses.

**High — Savings Plans coverage path mismatch.** Repointed `coverage_pct`, `on_demand_cost`, `spend_covered_by_savings_plans`, and `total_cost` from the deeper `Coverage.CoverageHours.*` / `Coverage.CoverageCost.*` paths to the top-level `Coverage.<member>` paths that `SavingsPlansCoverageData` actually returns. Confirmed live: AWS now reaches the data layer (returns `DataUnavailableException` because the test account has no SP, not a field-not-found error).

**Medium — `GetCostForecast` start date wording.** This one needed live evidence to settle. Findings:
- AWS rejects `time_period_start` earlier than yesterday with "Earliest supported Start is …".
- AWS accepts future starts in practice — sending `time_period_start = '2026-06-15'` returned `DataUnavailableException` (no history to forecast from), not the `ValidationException` the docs imply.
- AWS rejects `time_period_end` more than ~3 months ahead at DAILY (or ~18 months at MONTHLY) with "Latest supported End is …".

The manifest guide and README now describe live runtime behavior with the verbatim AWS error strings, instead of the docs-only "today or earlier" wording. Updated the README example to start from yesterday so it works against real accounts.

**Medium — `aws:cloudformation:stack-name` auto-activation claim.** Removed. Live testing across multiple tag keys (`aws:cloudformation:stack-name`, `Environment`, `Project`, `aws:createdBy`) all returned `Keys: ["<tag>$"]` — the `$` suffix is AWS's untagged-resource bucket. Live evidence that even AWS-generated tags need explicit activation under Billing → Cost Allocation Tags. Both the manifest guide and the README now document the activation step, the non-retroactive nature, and the management-account-only restriction in AWS Organizations.

**Low — `make lint-sources` line-length.** `make lint-sources` (and direct `ryl sources/community/aws_cost_explorer`) now exits clean. The originally-flagged content on line 54 was reformatted away as part of the `test_queries` cleanup.

## What else was added or removed during the contract review

**GovCloud guidance dropped.** Cost Explorer has no `ce` endpoint in the `aws-us-gov` partition (DNS NXDOMAIN; absent from the boto3 endpoints model). Description, `AWS_REGION` hint, and README region note all simplified to "global service with one regional endpoint per partition; standard partition `us-east-1`, China partition `cn-northwest-1`; AWS GovCloud (US) does not expose a Cost Explorer endpoint."

**Fixed the `Total: {}` hazard on grouped tables.** Both `cost_by_service` and `cost_by_tag` hardcode a `GroupBy`, and AWS returns the response `Total` block as an empty object whenever a `GroupBy` is set (verified with raw `aws ce` CLI: `Total: {}` with 9 groups under `GroupBy=DIMENSION,Key=SERVICE`; `Total: {"UnblendedCost": ...}` without `GroupBy`). Removed `period_total_cost`, `unit`, and the seven per-metric typed columns (`unblended_cost`, `amortized_cost`, `blended_cost`, `net_unblended_cost`, `net_amortized_cost`, `usage_quantity`, `normalized_usage_amount`) from both tables, since they would have been NULL on every row. Updated the table descriptions, the `groups` column docstrings, and the README example to teach summing across `groups` JSON for period totals.

**Documented the HOURLY date-format gotcha.** Live testing surfaced an undocumented contract: `granularity = 'HOURLY'` with `YYYY-MM-DD` dates returns `ValidationException: "Time period is invalid. Valid format is yyyy-MM-ddThh:mm:ssZ."` In addition to the existing payer-account opt-in requirement. Both filter descriptions, both virtual-column descriptions, and the README now call this out with the verbatim AWS error.

**Added `anomalies.date_interval_end` upper bound.** Live testing surfaced "Latest supported detectionDate for GetRecentAnomalies is [yesterday]". Both the guide and the README now document end-date ≤ yesterday.

**Added `savings_plans_coverage.time_period_end` upper bound.** Live testing surfaced "end date should NOT be after [today]". Both the guide and the README document end-date ≤ today.

**Tightened rightsizing wording.** Dropped the unverified "highest-savings target instance" claim from `estimated_monthly_savings` and `target_instance_type`. AWS marks the recommended option with `DefaultTargetInstance: true`, but the manifest projects the array's first element regardless of that flag — made that explicit.

**Replaced the unverified `OptInRequired` README claim** with the operationally observable `AccessDeniedException ce:GetCostAndUsage`, plus the activation path through Billing → Cost Explorer.

**Added an explanatory comment** above tables pointing at `botocore service-2.json` as the source of the `AWSInsightsIndexService` `targetPrefix`. The prefix is internal-looking and absent from the user-facing API reference, so a future reviewer might try to "fix" it without the comment.

**Per-call page-size literals now have inline rationale comments** (`MaxResults`/`PageSize` 100 across `GetAnomalies`, `GetRightsizingRecommendation`, `GetSavingsPlansCoverage`).

**Tightened "String columns to cast" in the README** to the columns that actually exist after the `cost_by_*` pruning, and added a note that grouped tables route per-period totals through the `groups` JSON (extract with `json_get_str` then cast).

Live test sweep covered every documented enum value, every required and optional filter, boundary date values, and bound-violating values across all 6 tables. Bundled `test_queries` passes 1/1 against real AWS.

## `AWS_SESSION_TOKEN` (high) — intentionally deferred

I dug into this and the situation is more involved than just adding the input. The `aws_sigv4` signer does support `session_token` (`lib.rs` 25-58), but the obvious manifest pattern — `AWS_SESSION_TOKEN: { kind: secret, required: false }` plus `auth.session_token: "{{input.AWS_SESSION_TOKEN}}"` — fails at first request with `missing source input 'AWS_SESSION_TOKEN' for template token`. `coral_spec::resolve_inputs` (`inputs.rs` 316-336) and the CLI's env collector (`source_ops.rs` 599-637) both correctly drop missing optional secrets, so registration succeeds. But `render_input_template` in the SigV4 authenticator (`lib.rs` 162-194) treats every `{{input.X}}` reference as mandatory regardless of how the input was declared. So "configured but absent" and "not configured" are indistinguishable to the renderer, and any `auth.session_token` template fires every request even when the caller never provided a token.

Reproduced against `manifest.yaml` on `main` to confirm it isn't `aws_cost_explorer`-specific. The same gap affects `cloudwatch_logs`, `cloudwatch_metrics`, and the in-flight `aws_cost_explorer` source. Filed #1045 with the reproduction and three possible fix shapes for maintainer direction.

**Proposal for this PR** given the May 31 hackathon deadline: ship the long-term IAM user keys path (which works today and matches the existing `cloudwatch_logs` / `cloudwatch_metrics` pattern) and follow up with one PR adding `AWS_SESSION_TOKEN` across all four AWS sources once the renderer side of #1045 lands. The README is already narrowed to long-term keys (Step 3 of the auth section), the manifest interface stays the same, and no schema decisions in this PR get revisited.

Happy to mark this draft and wait if you'd rather hold the merge — but with the hackathon ending Saturday, shipping now and following up after #1045 settles seems like the smaller round-trip.

ShamithaReddy:
@cubic-dev-ai review

jsummerfield (changes requested):
Findings

- Medium: `sources/community/aws_cost_explorer/README.md:36`-`:37` overstates the IAM policy requirement by saying Cost Explorer actions only support `Resource: "*"`. AWS now documents resource scoping for some Cost Explorer APIs, including `BillingViewArn` on `GetCostAndUsage` / `GetCostForecast` and Cost Explorer resource types in the service authorization reference. Keeping `Resource: "*"` as the simple first-success policy is fine, but please do not present it as the only possible scoping. If practical, also consider optional `billing_view_arn` filters for the APIs that support it. Docs: https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_GetCostAndUsage.html, https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_GetCostForecast.html, and https://docs.aws.amazon.com/service-authorization/latest/reference/list_awscostexplorerservice.html

- Medium: `sources/community/aws_cost_explorer/manifest.yaml:179`-`:192` and `:342`-`:355` make the two main cost tables hard to use for actual SQL analysis: `cost_by_service` and `cost_by_tag` return one row per period with the useful service/tag breakdown buried in a JSON array, and the README examples at `sources/community/aws_cost_explorer/README.md:124`-`:135` only extract array element 0. That does not let a user answer normal questions like "top services by spend" or "all tag values by spend" without manual JSON array handling. Please either expose a more directly queryable table/function shape, or narrow/rename the current tables as grouped-period JSON tables and add a first-success path for totals plus practical guidance for working with all groups. AWS does return grouped values under `ResultsByTime[].Groups[]`, so this is a product-surface issue rather than an AWS contract mismatch. Docs: https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_GetCostAndUsage.html

- Medium: The README omits a billable-request warning. Cost Explorer API calls can be charged, and `sources/community/aws_cost_explorer/manifest.yaml:144`-`:148` / `:303`-`:307` allow up to 100 paginated requests for broad `cost_by_service` and `cost_by_tag` queries. Please add a cost/rate note near setup and the pagination-heavy tables so users understand that broad scans can incur multiple billable requests. Docs: https://aws.amazon.com/aws-cost-management/aws-cost-explorer/pricing/

- Low: `sources/community/aws_cost_explorer/README.md:213`-`:224` says to use yesterday's date for `cost_forecast`, but the example uses `2026-05-30`; for this review date, yesterday is `2026-05-29`. Please avoid a fixed "yesterday" example or make the dates internally consistent. Docs: https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_GetCostForecast.html

The SigV4 shape, metric casing, Savings Plans coverage paths, and long-term-key limitation now look directionally sound. Live evidence in the PR body is useful. I still would not approve until the IAM wording, request-cost warning, and core cost-table ergonomics are tightened.

ShamithaReddy:
@jsummerfield 
Thanks for the detailed review. Pushed `742f95c` addressing the findings. Everything in this commit is documentation/guidance only , the request/response contract and table shapes are unchanged.

## Changes addressing review comments

### IAM Resource `"*"` overstatement (Medium)

Reworded so `"*"` stays the simple first-success policy but is no longer presented as the only option. Now documents that:

- `ce:GetCostAndUsage` / `ce:GetCostForecast` support the `billingview` resource type
- `ce:GetAnomalies` is scoped to `anomalymonitor`
- These accept `aws:ResourceTag/${TagKey}`, with the service-authorization reference cited.

### Cost-table ergonomics (Medium)

Kept the tables as grouped-period JSON tables (matches `ResultsByTime[].Groups[]`), but fixed the docs gap: the README only showed element 0. Added "top services by spend" and "all tag values by spend" recipes using `unnest(json_get_array(groups))`, which answers those questions today without manual array handling or an engine change. Verified live against AWS for both tables.

> **Note on the alternatives suggested:** a flattened one-row-per-group `RowStrategy` would be a `coral-engine` change (out of scope for a community source), and a scalar totals column isn't possible because AWS returns `Total: {}` empty whenever a `GroupBy` is set.

### Billable-request warning (Medium)

Added a $0.01-per-paginated-request note near setup and in Notes (pricing), including the per-table page caps (100 for `cost_by_service`/`cost_by_tag`, 50/20 elsewhere) so broad scans don't surprise anyone.

### `cost_forecast` "yesterday" date (Low)

Removed the hardcoded-date drift trap. The literal is now labeled illustrative for the review date, with a computed `current_date - INTERVAL '1 day'` alternative and a note that "yesterday" is evaluated against AWS's service clock (a lagging host clock can land one day early and return `ValidationException`).

### Source lint (Low)

Ran `ryl sources`: passes clean. Longest line in the manifest is 131 chars against the repo's 150-char limit (`.yamllint.yaml`). The 80-char violation looked like a pre-edit/stricter-local-config artifact; current tree is clean.

---

## Extra changes (not in the review list)

- **`metric` / `granularity` virtual columns** — Clarified in both the manifest column descriptions that these are NULL unless bound explicitly in `WHERE`; they echo the bound filter, not the applied request default (which still defaults to `UnblendedCost` / `MONTHLY`). Caught while validating the cost tables: querying `metric` unbound returned null, which read as "no metric applied."
- **`unnest` guidance mirrored into the manifest** — Beyond the README, the `cost_by_service` / `cost_by_tag` guide blocks and `groups` column descriptions now point at the `unnest` pattern, so agents reading the catalog/MCP surface get the same recipe.

---

## Validation

- `coral source lint` → valid
- `ryl sources` → clean
- Documented `unnest` recipes run verbatim against live AWS for both tables (e.g. `cost_by_service` returned per-service rows ordered by spend)

jsummerfield (approved):
Thanks for the detailed follow-up. I re-reviewed the current head (`742f95c43e89a4ea824ac548c8fa5fb56ea339ed`) and the previous blockers look resolved:

- The IAM policy text now keeps `Resource: "*"` as the simple first-success path without claiming it is the only possible policy.
- The grouped Cost Explorer response shape is now documented as grouped-period JSON, with examples that expand every `groups` entry instead of only showing element 0.
- The README now warns clearly that Cost Explorer API calls are billable and broad paginated scans can fan out into many charged requests.
- The forecast example now uses date guidance that matches the review date and explains the AWS boundary behavior.

I did not independently run live AWS queries, but the PR discussion includes useful live Coral evidence and the current static source/docs review has no remaining findings from me.

PR #1077, opened 2026-05-31
feat(sources/community/chesscom): add Chess.com source

Photon079:
Adds a new read-only community source for the Chess.com Public API under `sources/community/chesscom/`, exposing player rating statistics, profiles, game history, and club memberships.

### What's included
`manifest.yaml` — four tables, requiring no authentication.

| Table | Purpose | API call |
|---|---|---|
| `stats` | Player rating statistics across rapid, blitz, bullet, and daily formats | `/player/{username}/stats` |
| `profile` | Player public profile information | `/player/{username}` |
| `games` | Game history for a given year and month. Includes PGN, results, and accuracy | `/player/{username}/games/{year}/{month}` |
| `clubs` | Chess clubs the player is a member of | `/player/{username}/clubs` |

`README.md` walks the user through setup and includes working SQL examples for querying player statistics and recent games.

### Implementation notes
- **Authentication:** The Chess.com API is entirely public and requires no API keys or OAuth setup.
- **Dynamic Inputs:** The source accepts `CHESSCOM_USERNAME`, `YEAR`, and `MONTH` as variables to dynamically route endpoints without hardcoded values.

### Verification
`coral source lint` passes with `Manifest is valid`.

**`coral source test chesscom`:**


jsummerfield (changes requested):
Thanks for adding this source. I found a few blockers on the current head that need another pass before this can be approved.

Findings

- High: `sources/community/chesscom/manifest.yaml:7`-`16` models `CHESSCOM_USERNAME`, `YEAR`, and `MONTH` as install-time inputs, then uses them in paths at `manifest.yaml:22`, `:156`, `:183`, and `:262`. This is a public no-auth API where those path values should be query filters, not source-install inputs; otherwise one installed source can only query one configured player/month. Please move these to required table filters such as `username`, `year`, and `month`, and use `{{filter.*}}` in the paths. Chess.com documents these as URL path variables for player/profile/stats/games/clubs: https://www.chess.com/news/view/published-data-api

- High: required local checks fail. `make lint-sources` fails across `sources/community/chesscom/manifest.yaml` starting at line 19 because list indentation does not match repo style, and `git diff --check` fails on trailing whitespace at `sources/community/chesscom/README.md:44`.

- High: mandatory live-test evidence is still missing. The PR body has an empty `coral source test chesscom:` section, and `README.md:26`-`31` only tells users what to run. Please paste successful current-head `coral source add`, `coral source test chesscom`, or representative `coral sql` output.

- Medium: `sources/community/chesscom/manifest.yaml:268` exposes `clubs.id`, but the Chess.com player-clubs response uses `@id`, not `id`, so this column will be null. Please map `@id` explicitly or rename the column to something like `club_api_url`. Docs: https://www.chess.com/news/view/published-data-api

- Medium: `sources/community/chesscom/manifest.yaml:113`-`128` and `:137`-`:152` expose bullet/daily win/loss counts but omit the corresponding draw counts and last-date fields that the stats response provides. Since the README describes player rating statistics across formats, the partial shape is misleading; please either expose the rest of the important fields or document the intentional subset.

The endpoint set is useful, but the install-time input shape will be painful for users until it becomes query-filter driven.

Photon079:
Hey, thanks for the review on this and the other pr too , I will soon update these changes requested.

Photon079:
Hey i am pasting the output of `coral source test chesscom `:   
```
  ✓ chesscom connected successfully

    chesscom (4 tables)
    ├─ clubs
    ├─ games
    ├─ profile
    └─ stats
    Query tests

    ✓ SELECT chess_blitz__last__rating, chess_rapid__last__rating, chess_blitz__record__win, chess_blitz__record__loss FROM chesscom.stats WHERE username = 'hikaru'
      1 row

    ✓ SELECT url, time_class, white__username, white__result, black__username, black__result FROM chesscom.games WHERE username = 'hikaru' AND year = '2026' AND month = '01' LIMIT 5
      5 rows
```

jsummerfield (changes requested):
Thanks for the update and for adding the live `coral source test` output. The previous blockers around install-time inputs, lint/whitespace, missing live evidence, the club identifier field, and the stats columns look resolved on this head.

I found one remaining correctness issue:

Medium: `sources/community/chesscom/manifest.yaml:199` maps the `profile.username` column from the SQL filter instead of from Chess.com's profile response. That means a query like `WHERE username = 'HIKARU'` returns `HIKARU` from Coral even though the Chess.com API canonicalizes the profile URL and returns `username: "hikaru"`. The provider also exposes `player_id` as a stable account identifier for username changes, so the profile table should preserve the canonical response identity instead of only echoing the requested filter.

Please map `profile.username` from the response field, and consider adding `player_id` as well. If you still want to keep the requested value, add it as a separate `requested_username`/filter echo column rather than using it as the profile identity.

Docs checked: https://www.chess.com/news/view/published-data-api

Once that is corrected, this should be close.

Photon079:
Hey @jsummerfield, thanks for catching that! I have pushed a fix to address the canonical username mapping issue.

Summary of Updates
- **`profile.username`**: Removed the `from_filter` expression so Coral now maps `username` directly from Chess.com's profile response, preserving canonical casing (e.g., `hikaru`).
- **`player_id`**: Added the stable `player_id` (`Int64`) field as recommended in the API docs.
- **`requested_username`**: Preserved the exact SQL filter echo in a separate `requested_username` column using `from_filter`.

Let me know if there's anything else needed to get this merged!

jsummerfield (approved):
I re-reviewed the current head (`ef73499`). The previous profile identity blocker is resolved: `profile.username` now comes from the provider response, and the requested input is preserved separately. Local validation and the live Chess.com query evidence look good.

Approved.

PR #1083, opened 2026-05-31
feat(sources/community/render): add Render source

ritwikareddykancharla:
## Summary

Adds a community source for **Render** (managed cloud platform). Query services, deploys, events, logs, PostgreSQL databases, Key Value instances, projects, and environments through Coral via the Render API.

Built for two workflows:
- **Operational debugging** — correlate deploys, events, and build/runtime logs around an incident.
- **Cross-source root cause** — `render.deploys.commit_id` is exposed so you can JOIN against a source-control source (e.g. GitHub) to trace a failed deploy back to the commit/PR that caused it.

## Tables (8)

`services`, `deploys`, `events`, `logs`, `postgres`, `key_value`, `projects`, `environments`.

Required filters are documented per table (`deploys`/`events` need `service_id`, `logs` needs `owner_id`+`resource`, `environments` needs `project_id`).

## Auth

Render API key as `Authorization: Bearer <RENDER_API_KEY>` (read-only key sufficient).

## Validation

Validated live against a real Render account on 2026-05-31 (Coral 0.3.0). `coral source test render` connects with all 8 tables and passing query tests; the README includes representative `coral sql` output (services, deploys, events, postgres, key_value) from disposable `coral-validate-throwaway` fixtures — no real customer data.

```
  ✓ render connected successfully
    render (8 tables)
    Query tests: 2 declared · 2 passed · 0 failed
```

## Notes

- `coral source lint` passes.
- Scoped to `sources/community/render/` only; no other files touched.
- Provider quirks documented: `suspended`/`auto_deploy` are returned as strings (not booleans) by Render and typed `Utf8` accordingly; `region` lives in `service_details` JSON; the per-row pagination cursor workaround is explained in the README.

### DIFF
--- a/sources/community/render/README.md
+++ b/sources/community/render/README.md
@@ -1,0 +1,271 @@
+# Render source
+
+Query Render services, deploys, events, logs, PostgreSQL databases, Key Value
+instances, projects, and environments through Coral using the Render API.
+
+This source is designed for operational debugging as well as infrastructure
+inventory. It can answer questions like which services exist, what deployed
+recently, what events happened around an incident, which log lines mention an
+error, and which datastore resources are attached to the account. Because it
+exposes `deploys.commit_id`, it is built to **join against a source-control
+source** (GitHub/GitLab) so you can trace a failed deploy back to the commit or
+pull request that caused it.
+
+## Authentication
+
+This source uses a Render API key sent as:
+
+```text
+Authorization: Bearer <RENDER_API_KEY>
+```
+
+Create an API key under **Account Settings → API Keys**
+(<https://dashboard.render.com/u/settings#api-keys>). A read-only key is
+sufficient for every table here. The key authorizes access to every workspace
+your Render account can see.
+
+## Setup
+
+```bash
+RENDER_API_KEY=<token> coral source add --file sources/community/render/manifest.yaml
+```
+
+Or add it interactively:
+
+```bash
+coral source add --interactive --file sources/community/render/manifest.yaml
+```
+
+## Validation
+
+Validated live against a real Render account on 2026-05-31 (Coral 0.3.0). All
+rows below come from disposable `coral-validate-throwaway` fixtures.
+
+`coral source test render`:
+
+```text
+  ✓ render connected successfully
+
+    render (8 tables)
+    ├─ deploys
+    ├─ environments
+    ├─ events
+    ├─ key_value
+    ├─ logs
+    ├─ postgres
+    ├─ projects
+    └─ services
+    Query tests
+    2 declared · 2 passed · 0 failed
+```
+
+Representative `coral sql` results:
+
+```text
+coral sql "SELECT id, name, type FROM render.services LIMIT 5"
++--------------------------+--------------------------+-------------+
+| id                       | name                     | type        |
++--------------------------+--------------------------+-------------+
+| srv-d8cqi8l7vvec73bbnq3g | coral-validate-throwaway | static_site |
++--------------------------+--------------------------+-------------+
+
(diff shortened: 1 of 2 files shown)


jsummerfield (changes requested):
Findings

- Medium: `sources/community/render/manifest.yaml:478` exposes `render.logs` as a single-page fetch. Render's logs API paginates with `hasMore`, `nextStartTime`, and `nextEndTime`; this manifest only sends `limit=100`, so wider time windows or `LIMIT 500` will silently drop logs after the first page. Please either implement the Render log pagination shape Coral can support or document the table as first-page-only with a narrower intended workflow. Docs: https://api-docs.render.com/reference/list-logs

- Medium: `sources/community/render/manifest.yaml:509` and `:515` expose `logs.level` and `logs.type` as virtual echoes of the filter inputs rather than row data. Render log entries carry these values in per-row `labels`, so unfiltered queries return null and filtered queries return the predicate value rather than the provider value. Please extract the provider label values or avoid exposing these as row columns. Docs: https://api-docs.render.com/reference/list-logs

- Medium: `sources/community/render/README.md:230` uses `github.pull_requests`, but Coral's GitHub source exposes `github.pulls` with required `owner` and `repo` filters. Please update the cross-source example so it runs as written.

- Low: `sources/community/render/manifest.yaml:352` omits useful event filters that Render supports (`type`, `startTime`, `endTime`). Adding those would make the incident-window workflow in the README much more precise. Docs: https://api-docs.render.com/reference/list-events

The source is useful and the broad scope fits community sources, but the logs table and copy-paste examples need to be correct before approval. Thanks for putting this together.

ritwikareddykancharla:
Thanks for the detailed review @jsummerfield (and cubic) — pushed fixes for all four findings:

- **logs.level / logs.type (Medium):** removed both as output columns. Render returns per-row level/type inside the `labels` array with no fixed position, so the path walker can't extract them deterministically. The `level`/`type` **filters** are retained, and the docs now point users to read these values out of `labels`.
- **logs pagination (Medium):** documented `render.logs` as first-page-only — Render uses response-level `hasMore`/`nextStartTime`/`nextEndTime` cursors (not the per-row cursor the other tables use), which this manifest does not follow yet. README now recommends narrowing the time window instead of a large `LIMIT`.
- **GitHub example (Medium):** updated to `github.pulls` with required `owner`/`repo` filters so the cross-source example runs as written.
- **events filters (Low):** added optional `type`, `start_time`, and `end_time` filters to the `events` table for incident-window queries, documented in the README and table guide.

`coral source lint sources/community/render/manifest.yaml` → `Manifest is valid`.

ritwikareddykancharla:
Follow-up: pushed a second commit (4a2d55b) — the `start_time`/`end_time` filters on `events` needed virtual echo columns to be usable in SQL WHERE. Without them, Coral registered the filters but rejected the names in queries. Fixed and live-tested against the Render API:

- `type='deploy_ended'` → 1 row (correct)
- `start_time`/`end_time` window covering known events → 4 rows (correct)
- `start_time`/`end_time` window with no activity → 0 rows (correct, proves server-side filtering)

`coral source lint` still passes.

jsummerfield (changes requested):
Findings

- Medium: the current head is still failing the required source manifest lint job. I approved the paused workflow and `Validate / Lint source manifests` now reports many YAML style errors in `sources/community/render/manifest.yaml`, for example `668:70 [braces] too many spaces inside braces`, followed by the same `[braces]` issue across the inline objects later in the file. Please run `make lint-sources` and normalize those inline braces so CI can pass.

The previous Render-specific findings look addressed: the log table is now documented as first-page-only, the virtual `logs.level` / `logs.type` output columns were removed, the GitHub example uses `github.pulls`, and the event filters were added. `coral source lint sources/community/render/manifest.yaml` passes locally, but the repo-level source lint gate still has to be green before this can be approved.

jsummerfield:

PR #769, opened 2026-05-25
feat(sources/community/npm): add npm registry source

git-sakshii:
## Summary
Adds a new community source for the [npm Public Registry API](https://github.com/npm/registry/blob/master/docs/REGISTRY-API.md) (https://github.com/npm/registry/blob/master/docs/REGISTRY-API.md). This source allows querying npm packages by keyword, retrieving package metadata, and accessing popularity/quality scores directly via Coral SQL.

## Changes
- `npm.search` table: Exposes keyword-based search with pagination support. Supports filtering by text (required), as well as weights for popularity, quality, and maintenance.
- Includes parsing of nested JSON fields like keywords and maintainers.
- Added complete README.md with configuration details, example queries, and a Validation section demonstrating live test queries and their expected tabular outputs.

## Validation
- [x] Linted successfully (`coral source lint`)
- [x] Tested against the live registry (`coral source test npm`)
- [x] Query outputs verified and documented in the README.

*Note: The npm registry does not require authentication or API keys.*

Fixes issue #766

PR #1084, opened 2026-05-31
feat(sources/community/npm): add npm registry source spec

mrayhankhan:
## Summary

Adds a community source spec for the **npm registry** (`registry.npmjs.org`) — package metadata over the public, no-auth API.

- `sources/community/npm/manifest.yaml` — `packages` table (GET `/{package_name}`)
- `sources/community/npm/README.md`

One row per package, filtered by `package_name` (scoped packages like `@types/node` supported). Surfaces canonical name, `latest_version` (`dist-tags.latest`), `repository__url`, first/last publish timestamps, maintainers, keywords, dist-tags, license, bugs URL and README.

Nested fields use explicit `expr: { kind: path }` (the `__` convention isn't auto-applied).

> Download counts live on a different host (`api.npmjs.org`) and are submitted as a separate spec, `npm_downloads`.

## Test

```sql
SELECT name, latest_version FROM npm.packages WHERE package_name = 'react' LIMIT 1;
```

Built during the Pirates of the Coral-bean hackathon (powers a dependency-risk auditor that JOINs OSV + npm + npm downloads in one query).

### DIFF
--- a/sources/community/npm/README.md
+++ b/sources/community/npm/README.md
@@ -1,12 +1,14 @@
 # npm (npm)
 
-**Version:** 0.1.0
+**Version:** 0.2.0
 **Backend:** HTTP
-**Tables:** 1
+**Tables:** 2
 **Base URL:** `https://registry.npmjs.org`
 
 Query the [public npm registry REST API](https://github.com/npm/registry/blob/master/docs/REGISTRY-API.md)
-to search for packages and retrieve search-result summary data (such as latest version, publisher, downloads, and quality scores).
+to search for packages (`npm.search`) and retrieve full metadata for a single
+package by name (`npm.packages`) — latest version, repository, publish
+timestamps, maintainers, license, README and more.
 
 ```bash
 coral source add --file sources/community/npm/manifest.yaml
--- a/sources/community/npm/README.md
+++ b/sources/community/npm/README.md
@@ -18,9 +18,17 @@ This source does not require any authentication or input configuration. The npm 
 
 ## Tables
 
-| Table        | Description                                            | Key filters               |
-| ------------ | ------------------------------------------------------ | ------------------------- |
-| `npm.search` | Search npm packages by keyword, ranked by popularity   | `text` (**required**)     |
+| Table          | Description                                            | Key filters                   |
+| -------------- | ------------------------------------------------------ | ----------------------------- |
+| `npm.search`   | Search npm packages by keyword, ranked by popularity   | `text` (**required**)         |
+| `npm.packages` | Full metadata for one package, fetched by name         | `package_name` (**required**) |
+
+The `packages` table returns one row per package via `GET /{package_name}`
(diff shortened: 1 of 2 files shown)


jsummerfield (changes requested):
Thanks for the contribution. I noticed one scope issue before doing a full source review.

`sources/community/npm` already exists on `main` from #769, so this PR should rebase and extend the existing npm source rather than add/review a parallel replacement for the same provider. As written, this branch defines `npm.packages` only, while the merged source already exposes `npm.search`; after rebase, please preserve the existing `npm.search` table, add `packages` as an additional table if that is the intended improvement, and bump the source version for the user-visible schema change.

Once that is rebased into an additive update, I’ll be happy to review the package-metadata table itself.

mrayhankhan:
Thanks @jsummerfield — rebased onto `main` and reworked this as an additive change:

- **Preserved** the existing `npm.search` table from #769 untouched.
- **Added** `npm.packages` as a second table — fetches full metadata for a single package by name via `GET /{package_name}` (scoped packages like `@types/node` supported): `latest_version`, `repository__url`, publish timestamps (as real `Timestamp` columns), maintainers, keywords, dist-tags, license, bugs URL, README, and a `raw` column.
- **Bumped** the source version `0.1.0 → 0.2.0` for the schema change.
- Updated `test_queries` and the README to cover the new table.

Validated locally: `coral source lint` passes, and `coral source test npm` reports 3/3 query tests passing (search ×2 + packages). Ready for the full review when you are.

jsummerfield (changes requested):
Thanks for rebasing this into an additive update. The source ownership issue is resolved: `npm.search` is preserved, `npm.packages` is now additive, and the manifest version bump is in place.

One small user-facing doc fix before this is ready:

- `sources/community/npm/README.md` still says `Version: 0.1.0` and `Tables: 1`, but the manifest is now `version: 0.2.0` and exposes both `npm.search` and `npm.packages`. Please update those header values so the README matches the source users will install.

I did not find substantive API-shape issues with the new package metadata table. After that README header is corrected, this should be straightforward to re-check.

mrayhankhan:
Good catch — updated the README header to **Version: 0.2.0** and **Tables: 2**, and reworded the intro to cover both `npm.search` and `npm.packages`. README now matches the manifest.

jsummerfield (approved):
Reviewed the current npm head. The previous README drift is resolved: the README now says version `0.2.0` and two tables, matching the manifest and the added `npm.packages` table.

I also re-checked the current source shape against the npm registry API, and local validation passed: `coral source lint sources/community/npm/manifest.yaml` and `ryl --config-file .yamllint.yaml sources/community/npm/manifest.yaml sources/community/npm/README.md` both pass. A focused re-review also ran live source add/test and a scoped package query successfully.

Thanks for rebasing this into an additive update.

mrayhankhan:
Thank you for the thorough review, @jsummerfield — appreciate you walking it through the additive rebase. 🙏

PR #1085, opened 2026-05-31
feat(sources/community/npm_downloads): add npm downloads source spec

mrayhankhan:
## Summary

Adds a community source spec for the **npm downloads API** (`api.npmjs.org`) — package download counts over the public, no-auth API.

- `sources/community/npm_downloads/manifest.yaml` — two tables:
  - `downloads_last_month` — total downloads, trailing 30 days
  - `downloads_last_week` — total downloads, trailing 7 days
- `sources/community/npm_downloads/README.md`

One row per package, filtered by `package_name`. Useful for popularity and short-term trend detection (compare a week ×4 against the month).

> This is intentionally separate from the `npm` registry spec: download counts live on `api.npmjs.org` while metadata lives on `registry.npmjs.org`, and a Coral source has a single `base_url`.

## Test

```sql
SELECT downloads FROM npm_downloads.downloads_last_month WHERE package_name = 'lodash' LIMIT 1;
SELECT downloads FROM npm_downloads.downloads_last_week  WHERE package_name = 'lodash' LIMIT 1;
```

Built during the Pirates of the Coral-bean hackathon (powers a dependency-risk auditor that JOINs OSV + npm + npm downloads in one query).

### DIFF
--- a/sources/community/npm_downloads/README.md
+++ b/sources/community/npm_downloads/README.md
@@ -4,8 +4,15 @@ Query the public **npm downloads API** (`api.npmjs.org`) for package download
 counts over the trailing week and month. No authentication required.
 
 > Package metadata (versions, repository, maintainers…) lives on a different
-> host (`registry.npmjs.org`) and is provided by the separate
-> [`npm`](../npm) source spec.
+> host (`registry.npmjs.org`) and is provided by the separate `npm` source spec.
+
+## Setup
+
+```bash
+coral source add --file sources/community/npm_downloads/manifest.yaml
+```
+
+No authentication required.
 
 ## Tables
 
@@ -21,13 +28,13 @@ Both tables expose the same columns:
 
 | Column | Type | Notes |
 |---|---|---|
-| `package_name` | Utf8 | Echoes the `WHERE` filter |
+| `package_name` | Utf8 | Virtual — echoes the `WHERE` filter |
 | `package` | Utf8 | Canonical package name from the API |
 | `downloads` | Int64 | Total downloads in the period |
 | `start` | Utf8 | First day of the window (YYYY-MM-DD) |
 | `end` | Utf8 | Last day of the window (YYYY-MM-DD) |
 
-## Examples
+## Example queries
 
 ```sql
 -- Monthly popularity
@@ -47,4 +54,6 @@ WHERE m.package_name = 'express' AND w.package_name = 'express';
 |---|---|---|
 | `NPM_DOWNLOADS_BASE` | `https://api.npmjs.org` | Base URL for the npm download-counts API |
 
-No authentication required.
+## References
+
+- npm download counts API: https://github.com/npm/registry/blob/main/docs/download-counts.md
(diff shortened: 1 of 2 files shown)


mrayhankhan:
Thanks for the review — both points fixed:

1. **Broken cross-link** — removed the relative `[npm](../npm)` link (the sibling `npm` spec lives in a separate PR, so the path doesn't resolve here). It's now referenced by name in plain text.
2. **Missing install section** — added a `## Setup` section with `coral source add --file sources/community/npm_downloads/manifest.yaml`, matching the convention in existing community sources.

I also proactively applied the same two fixes to the companion `npm` PR (#1084), which had the mirror-image cross-link.

jsummerfield (changes requested):
Thanks for the updates here. I checked this as a separate source from `npm`, since download counts are served from `api.npmjs.org` and the package metadata source uses `registry.npmjs.org`.

A couple of things need tightening before this is ready:

- `sources/community/npm_downloads/manifest.yaml:43` passes `package_name` straight into npm's point downloads endpoint, while `sources/community/npm_downloads/manifest.yaml:45` maps only the single-package response shape. npm's download-count docs also support comma-separated bulk point queries, where the response is keyed by package name instead of returning `downloads`, `package`, `start`, and `end` at the root: https://github.com/npm/registry/blob/main/docs/download-counts.md#point-values and https://github.com/npm/registry/blob/main/docs/download-counts.md#bulk-queries. With `package_name = 'npm,express'`, this source returns one row with null mapped fields instead of either modeling the bulk response or clearly rejecting/documenting that only a single package name is supported. Please either model bulk correctly or make the single-package restriction explicit in the table guide/README so users do not get silent nulls.
- I don't see captured live Coral output in the PR body or README yet. The PR includes example SQL, but community sources need evidence like `coral source add`, `coral source test`, or representative `coral query` output from the source. Please add that output to the PR description or README.

The basic single-package endpoint shape looks good, including scoped packages, and `coral source lint` passes locally. Happy to take another look once those two pieces are addressed.

mrayhankhan:
Thanks for the detailed review — both points addressed:

**1. Bulk vs single-package nulls.** Confirmed the behavior you described (`package_name = 'npm,express'` → one row with null fields). Rather than model npm's name-keyed bulk response shape, I've made the single-package restriction **explicit** so users don't hit silent nulls:
- Both table `guide`s now state that exactly one package is supported, that the comma-separated bulk point endpoint returns a differently-shaped object keyed by package name (not modeled here), and that a comma-separated value yields null fields.
- The README has a matching **"One package per query"** callout.

**2. Captured live Coral output.** Added a **Validation** section to the README with real output:
- `coral source add` → `npm_downloads (2 tables)`, `2 declared · 2 passed · 0 failed`
- a representative query: `express` → `425540092` downloads for `2026-05-02 … 2026-05-31`

`coral source lint` passes locally after the edits. Ready for another look.

jsummerfield (approved):
Reviewed the current npm downloads head. The previous blockers are resolved: the source now clearly models the single-package point endpoints, documents the comma-separated bulk-response caveat, and includes captured live Coral output for source add/test plus a representative query.

Local validation passed: `coral source lint sources/community/npm_downloads/manifest.yaml` and `ryl --config-file .yamllint.yaml sources/community/npm_downloads/manifest.yaml sources/community/npm_downloads/README.md` both pass. The separate `npm` source reference is acceptable here because `sources/community/npm` already exists on `main`.

Thanks for tightening the scope and adding the live evidence.

mrayhankhan:
Thanks for the careful review, @jsummerfield — much appreciated. 🙏

PR #1091, opened 2026-05-31
feat(sources/community): add obsidian and onepiece source specs

Kshitij-K-Singh:
## What changed
Added two new community source specs:
- **obsidian**: Query notes, commands, tags, and search from the Obsidian Local REST API plugin
- **onepiece**: Query Devil Fruits from the One Piece API (no auth required)
## Why it changed
Community contribution to expand the source catalog with two new integrations.
## User-visible impact
- New `coral source add` options for users who want to query their Obsidian vault or One Piece data
- `onepiece` source is ready to use with `coral sql "SELECT ... FROM onepiece.fruits"`
- `obsidian` source requires the Local REST API plugin installed in Obsidian
## Notes
- Added comprehensive READMEs with setup instructions and example queries

### DIFF
--- a/sources/community/obsidian/README.md
+++ b/sources/community/obsidian/README.md
@@ -1,8 +1,8 @@
 # Obsidian Local REST API (obsidian)
 
-**Version:** 0.1.0
+**Version:** 0.2.0
 **Backend:** HTTP
-**Tables:** 4
+**Tables:** 3
 **Functions:** 2
 **Base URL:** `https://127.0.0.1:27124`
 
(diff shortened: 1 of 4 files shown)


jsummerfield (changes requested):
Findings

- High: this PR adds two independent community sources, `sources/community/obsidian/**` and `sources/community/onepiece/**`. Community source PRs need to stay scoped to one provider/product so review, docs, product fit, live evidence, and future ownership are clear. Please split this down to a single generally useful provider source before review continues.

- High: the One Piece source is primarily an entertainment/novelty data feed rather than a durable product, work, research, developer, operational, or user-owned data surface. Under the current community source acceptance bar, that source should be removed rather than iterated on in this repo. If you want to continue with this PR, the useful path is to focus only on Obsidian and provide current live Coral validation for that source.

Thanks for contributing this. Once the PR is narrowed to one generally useful source, I can take a normal source-review pass on the remaining manifest and README.

PR #1092, opened 2026-05-31
feat: add OpenSearch source spec connector to community sources

OpswhizMir:


jsummerfield (changes requested):
Thanks for adding an OpenSearch source. I reviewed the current head (`53ef0d2`) and the manifest does not load yet.

Blocking issues:

- `sources/community/opensearch/manifest.yaml` fails `coral source lint` because `test_queries` uses object entries, but the current source-spec schema expects each test query to be a string. Please change those to supported string queries.
- `sources/community/opensearch/manifest.yaml` also puts `pagination` inside `tables[0].request`, which the schema rejects as an unexpected request property. Please move pagination to the supported table-level pagination shape or remove it if the table is intentionally unpaginated.
- `sources/community/opensearch/README.md` has trailing whitespace in the current diff. `git diff --check origin/main HEAD -- sources/community/opensearch` reports it on the introduction and SQL example lines.

Once the manifest lints, please include sanitized live `coral source test opensearch` evidence from this exact revision.

PR #1095, opened 2026-05-31
feat(sources/community): add obsidian source spec

Kshitij-K-Singh:
## What changed
Added a new community source spec for obsidian — queries notes, commands, tags, and search from the Obsidian Local REST API plugin.
## Why it changed
Community contribution to expand the source catalog with Obsidian integration.
## User-visible impact
- New coral source add option for users who want to query their Obsidian vault
- Requires the Local REST API (https://github.com/coddingtonbear/obsidian-local-rest-api) plugin installed in Obsidian
## Notes
- Added comprehensive READMEs with setup instructions and example queries


jsummerfield (changes requested):
Findings

- High: `sources/community/obsidian/manifest.yaml:63`-`:78` maps the tag name from `tag`, but the Obsidian Local REST API returns tag objects as `{ name, count }`. That means `obsidian.tags.tag` will come back null while `count` works. Please map the column from `name`, or rename the Coral column to `name`. Docs: https://github.com/coddingtonbear/obsidian-local-rest-api/blob/master/docs/openapi.yaml#L3046-L3078.

- High: I do not see live Coral evidence in the scoped files or PR body. The README has example commands at `sources/community/obsidian/README.md:80`-`:94`, but no captured successful `coral source add`, `coral source test`, or representative `coral sql` output. Please add a short live-test section with real output from a vault using the plugin.

- Medium: `sources/community/obsidian/README.md:98`-`:99` and `sources/community/obsidian/manifest.yaml:16`-`:20` understate the HTTPS/self-signed certificate setup. The provider defaults to HTTPS with a self-signed cert, and its docs say users must trust that cert or explicitly enable the HTTP endpoint. "Use HTTP if HTTPS fails" is not enough for first success because HTTP may not be enabled. Please document the exact path: trust/download the plugin cert, or enable the insecure HTTP server in Obsidian settings and set `OBSIDIAN_BASE_URL=http://127.0.0.1:27123`. Docs: https://github.com/coddingtonbear/obsidian-local-rest-api#quick-start and https://github.com/coddingtonbear/obsidian-local-rest-api/blob/master/docs/openapi.yaml#L60-L79.

The product fit is good here: this exposes user-owned knowledge-base data, not novelty content. Auth header, command listing, vault-root listing, search endpoint shape, and note JSON `Accept` usage broadly match the Local REST API docs. Thanks for adding this; once the tag mapping and first-success evidence are fixed I can take another look.

Kshitij-K-Singh:
The HTTPS server support was removed due to `Rustl` rejecting the CA certificate as `CaUsedAsEndEntity`, will try to make a PR about this issue on the API repo to have both CA and leaf/end-entity certificate. And will then update the source once accepted
Until then the recent commit has fixed all the known issues
and defaults to HTTP as stated in the README.md and manifest.yaml

jsummerfield (changes requested):
Findings

- High: `sources/community/obsidian/manifest.yaml:147` and `sources/community/obsidian/manifest.yaml:196` interpolate raw SQL arguments directly into URL path segments for `read_note` and `list_directory`. That breaks real Obsidian vault paths containing reserved URL characters. A focused mock check showed `read_note(path => 'Note #1.md')` sending `GET /vault/Note%20`, dropping the `#1.md` fragment. Since search/list results can naturally return filenames with spaces, `#`, and other reserved characters, users cannot reliably round-trip note paths into the table functions. Please URL-encode path arguments before putting them into the request path, or otherwise model the endpoint so vault-relative paths are transmitted safely. Docs/OpenAPI: https://coddingtonbear.github.io/obsidian-local-rest-api/openapi.yaml

- Medium: `sources/community/obsidian/README.md:107` says `4 declared · 4 passed · 0 failed`, but the manifest declares only three test queries at `sources/community/obsidian/manifest.yaml:207`-`:210`. A current-head local import also reports three declared tests. Please update the README evidence or add the missing declared test so the live-test section accurately reflects what was validated.

The prior tag mapping issue looks resolved, and the HTTP/default setup is much clearer now. `coral source lint` and `ryl` pass on the current head, but the path-encoding issue is a source-usability blocker because it affects ordinary note names. Thanks for continuing to tighten this source.

jsummerfield (changes requested):
Findings

- High: `sources/community/obsidian/manifest.yaml:95` still does not produce valid path values for `read_note` and `list_directory`. The new encoded helper replaces reserved characters first and `%` last, so a path like `Note #1.md` becomes `Note%2520%25231.md` instead of `Note%20%231.md`. That double-encoded value will not address the upstream `/vault/{filename}` endpoint correctly. The same replacement order appears in the other encoded path helpers, including `sources/community/obsidian/manifest.yaml:208`, `:321`, `:445`, and `:523`. Please encode `%` before introducing `%XX` escape sequences, or avoid hand-rolled path encoding if the source-spec/template layer can provide a safer path-parameter escape. Docs/OpenAPI: https://github.com/coddingtonbear/obsidian-local-rest-api/blob/master/docs/openapi.yaml#L3253-L3259 and https://github.com/coddingtonbear/obsidian-local-rest-api/blob/master/docs/openapi.yaml#L3326-L3328

- Low: `sources/community/obsidian/manifest.yaml:297` and `sources/community/obsidian/manifest.yaml:426` exceed the repo source-style line length limit, so `ryl --config-file .yamllint.yaml sources/community/obsidian/manifest.yaml` fails even though `coral source lint` passes. Please wrap those descriptions so the repository source-style gate is green.

The README live evidence now matches the three declared test queries, and the earlier tag mapping plus HTTP setup issues look resolved. The remaining blocker is still ordinary note-path round-tripping for filenames with reserved URL characters. Thanks for iterating on this.

Kshitij-K-Singh:
The encoding chain is already correct — % is encoded first (innermost), not last.

The replace expressions nest inside-out (inner expr evaluates first, outer replace runs last). In all five locations (:95, :208, :321, :445, :523), the YAML nesting from outermost to innermost is:
replace(: → %3A)          outermost — runs LAST
  replace(= → %3D)
    ...
      replace(# → %23)
        replace(  → %20)
          replace(% → %25)   innermost — runs FIRST
            current_row / path
            
**Live Execution:**
image

jsummerfield (changes requested):
Findings

- Medium: `sources/community/obsidian/README.md:74` documents the reserved-character path workflow, but the live evidence at `sources/community/obsidian/README.md:138` only demonstrates plain paths such as `Welcome.md`, `Folder_Test/`, and `File1.md`. Because path encoding was the disputed behavior on the last review, please add durable live evidence for a real reserved-character filename, for example a `Note #1.md` row exposing `encoded_filename = Note%20%231.md` and a successful `read_note(path => 'Note%20%231.md')` result.

- Low: `sources/community/obsidian/manifest.yaml:297` is still one character over the repo source-style line-length limit. `coral source lint` passes, but `ryl --config-file .yamllint.yaml sources/community/obsidian/manifest.yaml sources/community/obsidian/README.md` fails with `line too long (151 > 150 characters)`. Please wrap that description so the repository source lint path is green.

- Low: `sources/community/obsidian/README.md:169` links the `list_directory` live-evidence block to `./images/source_test1.png`, which is already used for the `read_note` example above it. The directory screenshot appears to be `source_test2.png`; please point the evidence block at the right image or remove the duplicate image reference.

The previous path-encoding implementation concern looks resolved on this head: the nested replacement expression applies `%` first before introducing `%20` / `%23`, and the source exposes encoded path helper columns for path-based functions. Once the evidence/style cleanup is in place, I can take another pass. Thanks for clarifying the replacement order.

Kshitij-K-Singh:
**Fixes:**

1. The description is now well under 150 characters.
2. Fixed the image reference path.
3. Added more encoded live examples.

jsummerfield (changes requested):
Findings

- Medium: `sources/community/obsidian/manifest.yaml:409` and `sources/community/obsidian/manifest.yaml:417` expose `tags` and `frontmatter` as `Utf8`, but the Obsidian Local REST API `NoteJson` schema defines `tags` as an array and `frontmatter` as an object. Please make these `Json` columns so Coral preserves the provider shape instead of presenting structured fields as strings. Docs: https://github.com/coddingtonbear/obsidian-local-rest-api/blob/main/docs/openapi.yaml#L15-L58

The previous reserved-character live evidence, YAML style, and duplicate-screenshot issues look resolved on this head. Two small cleanups while you are updating: `README.md:214` uses `coral/sources/community/obsidian/manifest.yaml`, which is not the repo-root path used elsewhere, and `manifest.yaml:297` says only `encoded_path` even though the source exposes encoded values under names like `encoded_name` / `encoded_filename` depending on the table.

Thanks for sticking with the review; the remaining blocker is the JSON typing for the note metadata.

jsummerfield (approved):
Thanks for continuing to iterate on this. I do not see remaining blocking issues on this head.

The earlier findings are resolved: tag rows now map from the provider `name` field, live Coral evidence is present, HTTP setup is clear, reserved-character paths now round-trip through encoded helper columns, the declared test count matches the README evidence, YAML style passes, and `tags` / `frontmatter` are modeled as JSON.

I also verified locally with `coral source lint`, the repo YAML style check, `coral source test obsidian` against a mock Local REST API server, and explicit mock queries for `NOTE #1.md` path encoding. One non-blocking cleanup remains: the two setup screenshot files without `.png` extensions would be clearer if renamed with normal image extensions, but that is polish and should not block this source.

PR #967, opened 2026-05-28
perf(catalog): reduce startup catalog/runtime work

jsummerfield:
## Summary

Reduce startup catalog/runtime work on the short-lived MCP path while keeping catalog aggregation and runtime lookup owned by the app/engine layers.

This removes avoidable repeated work by:

- adding app-computed per-kind `CatalogCounts` to `ListCatalogResponse`, letting callers request one bounded catalog item (`limit=1`) when they only need counts
- using that count contract in MCP `tools/list` instead of transferring/counting broader catalog pages
- moving exact `describe_table` lookup into the engine runtime snapshot, so hits clone only the matched table while misses still use one runtime build for suggestions
- adding startup tracing spans around catalog config loading, query-source loading, runtime builds, and source registration

## Validation

- `cargo test -p coral-cli --features cli-test-server mcp_stdio_lists_tools_and_resources`
- `cargo test -p coral-app --test grpc list_catalog`
- `make rust-checks` - 889 tests passed after merging current `origin/main`

## Benchmark

Measured process spawn through MCP `initialize` and `tools/list` response using debug binaries, 10 runs after 1 warmup:

- `origin/main` (`16087f2`): 4.176 s +/- 0.228 s
- this branch: 3.769 s +/- 0.078 s
- result: branch ran 1.11x +/- 0.06x faster

PR #1105, opened 2026-06-01
fix(mcp): reduce catalog discovery memory

sauldhernandez:
**Intent**

Reduce Coral MCP stdio memory growth during automation startup and repeated catalog/tool calls, while keeping the MCP behavior unchanged.

This PR is stacked directly on #967 (`james/reduce-mcp-startup-catalog-work`).

**What it enables**

- Reuses catalog summary metadata instead of forcing provider-backed table descriptions during broad catalog discovery.
- Keeps MCP `tools/list` on count-only catalog work instead of loading sources for the SQL tool description.
- Converts Arrow batches to JSON rows one batch at a time to avoid holding an extra full-result JSON string.

**Usage examples**

- Run `coral mcp-stdio` from an automation that calls `tools/list`, `list_catalog`, or `search_catalog`.
- Call `list_catalog` repeatedly and expect stable results with lower transient RSS.
- Use `describe_table` or `list_columns` when full column metadata is needed.

**Benchmark**

Local RSS benchmark on macOS 26.4.1 arm64, using detached dev-profile `cargo build -p coral-cli` binaries:

- `origin/main` at `c41060274`
- #967 parent at `cdb664439`
- this PR at `2317a406a`

Fixture: one local file-backed source with 120 tables x 50 columns (6,000 declared columns). Workload per run: MCP initialize, `tools/list`, 9 `list_catalog` calls, and 8 `search_catalog` calls. RSS was sampled with `ps` every 10ms. Each comparison used 5 runs.

| Comparison | Baseline median peak RSS | This PR median peak RSS | Median savings |
| --- | ---: | ---: | ---: |
| `origin/main` -> #1105 | 179.7 MB | 82.6 MB | 97.1 MB / 54.0% |
| #967 parent -> #1105 | 106.9 MB | 82.8 MB | 24.0 MB / 22.5% |

The main-to-#1105 run had peak RSS ranges of 177.5-181.3 MB on `main` and 82.5-83.6 MB on this PR. The parent-to-#1105 run had peak RSS ranges of 101.6-107.8 MB on #967 and 82.5-83.8 MB on this PR.

**Real config benchmark**

I also ran the same MCP workload against an existing local Coral config instead of the synthetic fixture. Source names and credential details are intentionally omitted; only aggregate RSS numbers are reported.

Workload per run: MCP initialize, `tools/list`, 9 `list_catalog` calls, and 8 `search_catalog` calls. RSS was sampled with `ps` every 10ms. Each comparison used 5 runs.

| Comparison | Baseline median peak RSS | This PR median peak RSS | Median savings |
| --- | ---: | ---: | ---: |
| `origin/main` -> #1105 | 1163.8 MB | 142.9 MB | 1020.9 MB / 87.7% |
| #967 parent -> #1105 | 4436.1 MB | 142.7 MB | 4293.4 MB / 96.8% |

The real-config main-to-#1105 run had peak RSS ranges of 853.2-1355.4 MB on `main` and 142.3-143.0 MB on this PR. The real-config parent-to-#1105 run had peak RSS ranges of 4382.9-4471.2 MB on #967 and 142.6-143.6 MB on this PR.

**Validation**

- `/Users/saul/.cargo/bin/cargo build -p coral-cli`
- `/Users/saul/.cargo/bin/cargo test -p coral-client --lib formatting_helpers_render_batches`
- `/Users/saul/.cargo/bin/cargo test -p coral-mcp --lib mcp_surface_refreshes_and_renders_dynamic_guide`
- `/Users/saul/.cargo/bin/cargo test -p coral-engine list_catalog_summaries --test engine`
- `/Users/saul/.cargo/bin/cargo test -p coral-app --test grpc catalog_discovery`
- `/Users/saul/.cargo/bin/cargo fmt --check`


sauldhernandez:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/withcoral/coral/1105?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1105** <a href="https://app.graphite.com/github/pr/withcoral/coral/1105?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1105?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#967** <a href="https://app.graphite.com/github/pr/withcoral/coral/967?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

jsummerfield:
I think this needs one design correction before it lands.

The MCP adapter changes look fine; `coral-mcp` stays thin. The issue is that `runtime/catalog_summary.rs` now has a second, runtime-side implementation of what HTTP/MCP sources expose in the catalog. That bypasses the backend registration path that decides whether a table/function is actually queryable.

There is already a concrete drift case: HTTP source specs can have empty/omitted `columns`. The full runtime rejects those during provider/function construction via `schema_from_columns`, so SQL cannot query them and normal registration skips the source. The new summary path still synthesizes table/function metadata directly, so `list_catalog`, `search_catalog`, and MCP tool counts can advertise items that `sql`, `describe_table`, and full `CoralQuery::list_catalog` disagree with.

I would move the lightweight summary construction behind a backend-owned seam, or otherwise reuse the same backend validation/construction rules that full registration uses. Please also add a regression that compares `list_catalog` vs `list_catalog_summaries` for an HTTP table/function with empty or omitted `columns`, so summary discovery cannot publish unqueryable surfaces.

Bradley-Butcher:
I dug into the memory behavior here because I was confused by the cache making RSS lower. I think the current PR body may now be stale relative to the branch.

The PR description says the benchmarked commit is `2317a406a`, but the current head is `f9823eea`. That matters: the current head appears to have moved from the earlier lightweight summary path to a `build_runtime_with_options(..., RuntimeBuildOptions::catalog_summary())` path.

I tested three release binaries:

- #967 parent: `cdb664439`
- current #1105: `f9823eea`
- current #1105 with only the `CatalogDiscovery::catalog_items` app cache disabled

On my real local config, median of 3 runs for: MCP initialize, `tools/list`, 9 `list_catalog`, 8 `search_catalog`:

| build | peak RSS | final RSS | `tools/list` | repeated `list_catalog` | repeated `search_catalog` |
|---|---:|---:|---:|---:|---:|
| #1105 current | 3153.0 MiB | 3150.9 MiB | 2461.9 ms | 1.3 ms | 1.3 ms |
| #1105 current, no cache | 4415.1 MiB | 4222.6 MiB | 2386.9 ms | 2209.4 ms | 2262.1 ms |
| #967 parent | 4654.3 MiB | 4472.2 MiB | 2450.9 ms | 2190.9 ms | 2171.6 ms |

So current #1105 definitely makes repeated catalog calls fast, but that mostly comes from caching the derived catalog items after the first broad discovery. The underlying cold path is still very expensive on the real config: the first `tools/list` warms the cache and leaves ~3.1 GiB RSS behind.

I also profiled with `sample`, `vmmap`, `heap`, and `leaks`.

The sample stack for current #1105 during `tools/list` still goes through:

```text
CatalogDiscovery::catalog_items
  -> build_runtime_with_options
  -> build_runtime_inner
```

and is dominated by cloning source/runtime registration state: `QuerySource`, `HttpSourceClient`, `HttpSourceManifest`, `TableCommon`, `ExprSpec`, and nested `Vec` / `String` data.

`heap` only reported ~6-7 MiB of live malloc nodes and `leaks` reported 0 leaked bytes, so this does not look like a classic live-object leak. It looks like large temporary allocation churn where macOS malloc keeps dirty/empty pages resident. `vmmap` showed multi-GB dirty/empty malloc regions after the workload.

My read: the cache is useful and probably worth keeping, but on the current head it is mostly avoiding repeated exercise of an expensive runtime-registration-shaped path. It does not fix low startup/broad-discovery RSS for real configs.

I think the cleaner root fix is to separate canonical catalog metadata construction from execution-provider construction. I do not mean maintaining two independent interpretations of the source manifest. I mean extracting the catalog metadata builder that registration already effectively performs, then making both registration and summary discovery consume that same builder.

Concretely, the shape I would expect is something like:

```rust
struct SourceCatalogModel {
    schema_name: String,
    tables: Vec<TableInfo>,
    table_functions: Vec<TableFunctionInfo>,
    inputs: Vec<SourceInputInfo>,
}

fn http_catalog_model(
    manifest: &HttpSourceManifest,
    source: &QuerySource,
) -> Result<SourceCatalogModel, CoreError> {
    // Single backend-owned source of truth for:
    // - table names/descriptions
    // - required filters
    // - function names/args/result columns
    // - input metadata
    // - cheap validation that determines whether the surface is query-visible
}
```

Then normal registration would use that same model while still building executable providers:

```rust
async fn register_http_source(...) -> Result<BackendRegistration> {
    let catalog = http_catalog_model(manifest, source)?;

    let providers = build_http_table_providers(...)?;
    let functions = build_http_table_functions(...)?;

    Ok(BackendRegistration {
        tables: providers,
        table_functions: functions,
        source: catalog.into_registered_source(),
    })
}
```

And broad summary discovery would use only the catalog-model half:

```rust
fn summarize_http_source(...) -> Result<RegisteredSource, CoreError> {
    Ok(http_catalog_model(manifest, source)?.into_registered_source())
}
```

That would keep one backend-owned definition of what the source exposes, but avoid creating `SessionContext`, `TableProvider`s, `HttpSourceClient`s, or per-table/per-function provider clones for `list_catalog`, `search_catalog`, and `tools/list`.

This also lines up with James's drift concern: summary discovery should not synthesize a catalog that full registration would reject. The shared `*_catalog_model` builder should be where those backend-owned validation/construction rules live, so `list_catalog_summaries` cannot publish unqueryable surfaces.

At minimum, I think the PR description should be refreshed with benchmarks from current head `f9823eea`, because the existing numbers appear to describe an older implementation.

PR #1116, opened 2026-06-01
feat(sources): per-method credential hints

AlbertQM:
- Add an optional per-method `hint` so each credential method's guidance renders next to the field it fills, both in the CLI and the generated source docs (and in the future, in the UI). 
- Split GitHub, Slack, and Google Calendar's long single hints into focused per-method hints.

## Test plan

image
image
image

Now you get per-method hints:

image

image

Input with no methods still show hints:
image

### DIFF
--- a/crates/coral-client/src/sources.rs
+++ b/crates/coral-client/src/sources.rs
@@ -291,7 +291,7 @@ mod tests {
                         SourceCredentialMethod {
                             label: "Connect".to_string(),
                             description: String::new(),
-                            hint: String::new(),
+                            hint: "Authorize in your browser.".to_string(),
                             method: Some(ProtoCredentialMethod::Oauth(Box::new(
                                 OAuthCredentialMethod {
                                     flow: OauthCredentialFlowType::AuthorizationCode as i32,
@@ -340,6 +340,15 @@ mod tests {
             ManifestCredentialMethodKind::OAuth
         );
         assert_eq!(credential.methods[0].label.as_deref(), Some("Connect"));
+        assert_eq!(
+            credential.methods[0].hint.as_deref(),
+            Some("Authorize in your browser."),
+            "non-empty method hint should survive proto decoding"
+        );
+        assert_eq!(
+            credential.methods[1].hint, None,
+            "empty proto hint should decode to None"
+        );
         assert_eq!(
             credential.methods[0]
                 .oauth
--- a/crates/coral-client/src/sources.rs
+++ b/crates/coral-client/src/sources.rs
@@ -126,6 +126,7 @@ fn credential_method_from_proto(
         kind,
         label: (!method.label.is_empty()).then(|| method.label.clone()),
         description: (!method.description.is_empty()).then(|| method.description.clone()),
+        hint: (!method.hint.is_empty()).then(|| method.hint.clone()),
         oauth,
     })
 }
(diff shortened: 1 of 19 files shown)


sources/core/github/manifest.yaml
```diff
@@ -20,36 +20,17 @@ inputs:
...
-      `gh auth token`, or create a new personal access token (PAT).
-      Grant read access to the repositories and organizations you plan
-      to query; some admin tables require elevated scopes.
-      See [GitHub PAT docs](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens).
     credential:
       methods:
         - type: oauth
           label: Connect with GitHub device code
           description: Sign in to GitHub with a device code.
+          hint: |
```
jsummerfield, line 32: Could we keep a concise input-level summary here, or expose the method metadata through the inspection surfaces before removing this hint entirely? Moving the detailed setup text into `credential.methods[].hint` improves interactive add and generated docs, but `coral source info --verbose` and `coral.inputs` still only expose `input.hint`, so agent/catalog inspection loses the useful `GITHUB_TOKEN` guidance.

plugins/coral/skills/coral-create-source-spec/SKILL.md
```diff
@@ -89,6 +89,7 @@ Only switch to Coral repo layout when the user is explicitly editing the Coral r
 - OAuth endpoint URLs may template declared `kind: variable` inputs with `{{input.KEY}}` for non-secret endpoint components such as tenant IDs or domains. Do not reference secret inputs, filters, function arguments, state, or inline defaults from OAuth endpoint URLs.
 - If a provider also supports manually pasted tokens, include a `type: source_config` fallback after the OAuth method. When the provider's token endpoint requires client authentication with a client secret, prompt for both OAuth client values: declare `client.id.input`, `client.secret.input`, and `client.secret.transport` (`basic_auth` or `request_body`).
 - Do not add top-level source inputs solely for OAuth client credentials; `client.id.input` and `client.secret.input` are collected during OAuth setup.
+- Each credential method accepts optional `label`, `description`, and `hint` fields, surfaced during interactive install and in the generated source docs. When an input offers more than one method, put the how-to-get-it guidance in each method's `hint` (rendered next to that method's fields) instead of in one long input-level `hint`, and scope each hint to the inputs that method collects.
```
jsummerfield, line 92: This new guidance points authors toward method-specific hints, but the OAuth-backed HTTP source example below still teaches the old pattern: broad setup guidance in the input-level `hint`, with no `hint` on either credential method. Since this skill is the maintained source-authoring path, it should be updated to show method-specific hints too, otherwise agents will keep generating specs that contradict the new guidance.

crates/coral-client/src/sources.rs
```diff
@@ -126,6 +126,7 @@ fn credential_method_from_proto(
         kind,
         label: (!method.label.is_empty()).then(|| method.label.clone()),
         description: (!method.description.is_empty()).then(|| method.description.clone()),
+        hint: (!method.hint.is_empty()).then(|| method.hint.clone()),
```
jsummerfield, line 129: The production mapping now carries `credential.methods[].hint`, but the app/client tests only populate empty hints and do not assert preservation. Please add a non-empty method hint to the proto/app mapping tests and assert it survives the round trip.

jsummerfield (commented):
Leaving the actionable review comments from James's requested pass.

sources/core/github/manifest.yaml
AlbertQM: Makes sense, thanks for flagging!

I've tweaked the logic a bit to _not_ print this hint along with the generic "paste token" message, as it was confusing IMO. So now we only get this hint if that's the only input hint, and we get it in the `info --verbose`, too 😄

jsummerfield (approved):

PR #1126, opened 2026-06-02
perf(engine): stream HTTP/MCP scans into bounded RecordBatches

sauldhernandez:
## Intent — why

HTTP- and MCP-backed table scans go through `JsonExec`, which used `stream::once`: it fetched the **entire** result set as one `Vec<serde_json::Value>` and converted it into a **single giant `RecordBatch`**. At peak the whole dataset is held twice — the heavy serde_json buffer (each row re-stores its column-name keys as owned `String`s, ~3–5× the wire JSON) plus one contiguous Arrow batch — and the scan ignored DataFusion's `batch_size`. On large or unfiltered scans this drives multi-GB peak memory and was a contributor to the ~10 GB RSS we observed on real queries.

## What it enables — what

`JsonExec::execute` now emits the fetched rows as a **lazy stream of `batch_size`-row `RecordBatch`es** (default 8192) via `futures::stream::try_unfold`, dropping each chunk's `Vec<Value>` as soon as it is converted. Effects:

- **Caps the Arrow batch** at one `batch_size` chunk instead of one batch spanning the whole result.
- **Releases the heavy serde_json rows incrementally** as the consumer pulls, instead of holding the full `Vec<Value>` alongside the full Arrow batch.
- **Lets downstream operators pipeline / short-circuit** — e.g. a `LIMIT` stops pulling, so the remaining rows are never converted.
- **Respects `datafusion.execution.batch_size`**, making HTTP/MCP scans behave like the file-format scans the pipeline already supports.

Behavior-preserving details: per-chunk conversion uses the same fixed schema (`convert_items`), so every batch is schema-consistent and projection is applied identically; an **empty result still emits exactly one schema-carrying batch**.

**Scope is deliberately limited to the `JsonExec` conversion/emission boundary.** The upstream `fetch_rows` still fetches all pages up front — page-level fetch streaming and a bounded DataFusion `MemoryPool` are separate follow-ups (see #886 and the predicate-pushdown stack #399–#607). This is an independent branch and does **not** touch the catalog-memory PR (#1105).

## Usage examples

Transparent — no API or config change. Any HTTP/MCP source query now converts and emits in `batch_size` chunks rather than one mega-batch, e.g.:

```sql
-- previously: whole result -> one Vec<Value> + one giant RecordBatch
-- now:        emitted as 8192-row RecordBatches, freed incrementally
SELECT number, title FROM github.pulls WHERE owner = 'withcoral' AND repo = 'coral';
```

Batch size follows the session config (`datafusion.execution.batch_size`).

## Verification

- New unit tests in `json_exec.rs`: bounded chunking with order preserved (`[2,2,1]` for 5 rows at `batch_size=2`), and empty result → single empty batch.
- `cargo test -p coral-engine` — 227 lib + 124 integration pass (incl. HTTP `LIMIT`/`ORDER BY`/pagination); `coral-client` (28) and `coral-mcp` (21) pass; clippy clean.
- Consumer audit: every production consumer of the result already iterates the full `Vec<RecordBatch>` (no `batches[0]` assumption), and multiple batches were already possible (file scans, sorts, aggregates), so nothing downstream regresses.

## Benchmark — peak heap, old vs new

Measured with [`dhat`](https://docs.rs/dhat) (peak live heap, `max_bytes`), 200,000 synthetic rows, isolating the `JsonExec` scan stage. `new` is the real `JsonExec::execute`; `old` is a faithful inline reproduction of the prior `execute()` (whole `Vec<Value>` converted into one batch, both alive together). The harness was temporary and is **not** part of this PR.

| Row shape | fetched `Vec<Value>` floor | **old** peak | **new** peak | reduction |
|---|--:|--:|--:|--:|
| narrow — 16 fields × ~16 B | 473.8 MiB | 550.0 MiB | 476.9 MiB | **−13.3%** |
| medium — 8 fields × ~48 B | 277.7 MiB | 411.8 MiB | 282.5 MiB | **−31.4%** |
| wide — 3 fields × ~256 B | 302.5 MiB | 496.8 MiB | 309.2 MiB | **−37.8%** |

**Takeaways**

- The win = eliminating the second full copy of the result (the single giant Arrow `RecordBatch` that used to coexist with the whole fetched `Vec<serde_json::Value>`). New peak ≈ `Vec<Value>` + one `batch_size` chunk.
- Magnitude scales with row shape: wide rows make the Arrow batch a larger fraction of the total (~38% saved); narrow rows let serde_json key/enum overhead dominate the `Vec<Value>`, so the eliminated batch is a smaller fraction (~13% saved). Numeric-heavy columns (denser Arrow) would push the saving higher.
- The remaining floor **is** the un-streamed `Vec<serde_json::Value>` from `fetch_rows` (the "floor" column; new peak is within ~2% of it). Bringing that down is the separate fetch-streaming + bounded-memory-pool follow-up — out of scope here.
- `collect` and `drain` consumers measured identically because `fetch()` builds the full `Vec<Value>` before streaming begins, so the peak is hit up front regardless of consumer (a `LIMIT` short-circuit wouldn't move it either).

Run (harness not included): `cargo test -p coral-engine --lib mem_bench -- --ignored --nocapture --test-threads=1`.

### DIFF
--- a/crates/coral-engine/src/backends/shared/json_exec.rs
+++ b/crates/coral-engine/src/backends/shared/json_exec.rs
@@ -251,7 +251,7 @@ mod tests {
         Arc::new(move |items: &[Value]| {
             let values: Int32Array = items
                 .iter()
-                .map(|value| value.as_i64().map(|n| n as i32))
+                .map(|value| value.as_i64().and_then(|n| i32::try_from(n).ok()))
                 .collect();
             RecordBatch::try_new(schema.clone(), vec![Arc::new(values) as Arc<dyn Array>]).map_err(
                 |error| datafusion::error::DataFusionError::ArrowError(Box::new(error), None),
--- a/crates/coral-engine/src/backends/shared/json_exec.rs
+++ b/crates/coral-engine/src/backends/shared/json_exec.rs
@@ -7,7 +7,7 @@ use std::sync::Arc;
 use async_trait::async_trait;
 use datafusion::arrow::array::RecordBatch;
 use datafusion::arrow::datatypes::SchemaRef;
-use datafusion::error::Result;
+use datafusion::error::{DataFusionError, Result};
 use datafusion::execution::TaskContext;
 use datafusion::physical_expr::EquivalenceProperties;
 use datafusion::physical_plan::execution_plan::{Boundedness, EmissionType};
@@ -16,7 +16,7 @@ use datafusion::physical_plan::{
     DisplayAs, DisplayFormatType, ExecutionPlan, Partitioning, PlanProperties,
     SendableRecordBatchStream,
 };
-use futures::stream;
+use futures::{TryStreamExt, stream};
 use serde_json::Value;
 
 /// Fetches raw JSON rows for one logical table scan.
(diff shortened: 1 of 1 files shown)


antonmry (approved):
Nice, good catch!

PR #1128, opened 2026-06-02
feat(codex): expose JSONL event provenance

Bradley-Butcher:
## Intent

Codex JSONL logs need stable file/session provenance before higher-level transcript surfaces can group events reliably. The raw `codex.events` table is append-only event data, but not every event payload carries a session id. The JSONL file is the real session boundary.

This PR adds opt-in file metadata for file-backed sources, keeps row-scoped `line_number` JSONL-only, and uses those columns in the bundled Codex source.

## What Changed

- Added `source.metadata` for `backend: file` tables.
- Supported file-scoped metadata kinds across file formats:
  - `relative_path`
  - `file_name`
  - `file_stem`
- Supported row-scoped metadata kind for JSONL only:
  - `line_number`
- Added a shared file-group scan path so Parquet, JSONL, JSON, and CSV can attach file-scoped metadata as per-file constants.
- Kept `line_number` injection in the custom JSONL row reader.
- Appended metadata fields to the table schema so they are queryable and discoverable through `coral.columns`.
- Updated `sources/core/codex/manifest.yaml` so `codex.events` exposes:
  - `session_path`
  - `session_file`
  - `event_index`

## Ownership / Boundaries

`coral-spec` owns the manifest contract and validation. Both the Rust parser and JSON schema allow file-scoped metadata across file formats and reject `line_number` outside JSONL file tables.

`coral-engine` owns runtime derivation because these values come from scan context: the matched object path and, for JSONL only, the line reader.

The Codex source only declares the metadata columns; it does not gain derived session/message/tool-call tables in this PR.

Existing file-backed tables are unchanged unless they opt into `source.metadata`.

For Codex, `codex.events` gains three columns that agents can use to group and order rows by JSONL session file:

- `session_path`
- `session_file`
- `event_index`

## Not In This PR

- No `codex.sessions`, `codex.messages`, or tool-call surfaces.
- No file-backed SQL views.
- No HTTP or MCP source metadata support.
- No row-scoped metadata beyond JSONL `line_number`.

Those are stack follow-ups.

## Validation

- `cargo test -p coral-engine --test engine file_format_tests`
- `cargo test -p coral-engine --test engine jsonl_tests`
- `cargo test -p coral-engine --test engine parquet_tests`
- `cargo test -p coral-engine --lib backends::file`
- `cargo test -p coral-spec metadata_for_non_jsonl`
- `cargo test -p coral-spec line_number_metadata`
- `cargo test -p coral-spec`
- `cargo clippy -p coral-engine --all-targets --all-features --locked -- -D warnings`
- `cargo clippy -p coral-spec --all-targets --all-features --locked -- -D warnings`
- `make rust-checks`
- `make docs-check`

### DIFF
--- a/crates/coral-engine/src/backends/file/metadata.rs
+++ b/crates/coral-engine/src/backends/file/metadata.rs
@@ -1,0 +1,249 @@
+//! File metadata columns for file-backed tables.
+
+use std::collections::HashSet;
+use std::sync::Arc;
+
+use datafusion::arrow::datatypes::{DataType, Field, FieldRef, SchemaBuilder, SchemaRef};
+use datafusion::common::ScalarValue;
+use datafusion::datasource::listing::ListingTableUrl;
+use datafusion::error::{DataFusionError, Result};
+use datafusion_datasource::table_schema::TableSchema;
+use object_store::path::Path as ObjectPath;
+use serde_json::Value;
+
+use coral_spec::backends::file::{FileMetadataColumnSpec, FileMetadataKind};
+
+#[derive(Debug, Clone)]
+pub(super) struct FileMetadataColumns {
+    file_columns: Vec<FileMetadataColumn>,
+    row_columns: Vec<FileMetadataColumn>,
+}
+
+impl FileMetadataColumns {
+    pub(super) fn try_new(specs: &[FileMetadataColumnSpec]) -> Result<Self> {
+        let mut seen = HashSet::with_capacity(specs.len());
+        let mut file_columns = Vec::new();
+        let mut row_columns = Vec::new();
+
+        for spec in specs {
+            if !seen.insert(spec.name.as_str()) {
+                return Err(DataFusionError::Plan(format!(
+                    "duplicate metadata column '{}'",
+                    spec.name
+                )));
+            }
+
+            let column = FileMetadataColumn {
+                name: spec.name.clone(),
+                kind: spec.kind,
+            };
+            if is_file_scoped(column.kind) {
+                file_columns.push(column);
+            } else {
+                row_columns.push(column);
+            }
+        }
+
+        Ok(Self {
+            file_columns,
+            row_columns,
+        })
+    }
+
+    pub(super) fn has_file_columns(&self) -> bool {
+        !self.file_columns.is_empty()
+    }
+
+    pub(super) fn has_row_columns(&self) -> bool {
+        !self.row_columns.is_empty()
+    }
+
+    pub(super) fn reject_schema_collisions(
+        &self,
+        file_schema: &SchemaRef,
+        source_schema: &str,
+        table_name: &str,
+    ) -> Result<()> {
+        for field in file_schema.fields() {
+            if self.column_named(field.name()).is_some() {
+                return Err(DataFusionError::Plan(format!(
+                    "{source_schema}.{table_name} metadata column '{}' duplicates a file column",
+                    field.name()
+                )));
+            }
+        }
+        Ok(())
+    }
+
+    pub(super) fn row_contains(&self, name: &str) -> bool {
+        self.row_columns
+            .iter()
+            .any(|metadata| metadata.name == name)
+    }
+
+    /// Extends a table schema with configured metadata columns, when present.
+    ///
+    /// Row-scoped metadata is read by the JSON decoder as part of the file
+    /// schema. File-scoped metadata is modeled as table partition fields so
+    /// `DataFusion` can attach one constant value per file.
+    pub(super) fn extend_table_schema_if_present(&self, table_schema: TableSchema) -> TableSchema {
+        if !self.has_row_columns() && !self.has_file_columns() {
+            return table_schema;
+        }
+
+        let file_schema = self.append_row_fields(Arc::clone(table_schema.file_schema()));
+        let partition_fields = table_schema
+            .table_partition_cols()
+            .iter()
+            .cloned()
+            .chain(self.file_fields())
+            .collect();
+        TableSchema::new(file_schema, partition_fields)
+    }
+
+    pub(super) fn file_values(
+        &self,
+        table_path: &ListingTableUrl,
+        location: &ObjectPath,
+    ) -> Result<Vec<ScalarValue>> {
+        if self.file_columns.is_empty() {
+            return Ok(vec![]);
+        }
+
+        let relative_path = relative_file_path(table_path, location);
+        self.file_columns
+            .iter()
+            .map(|column| column.file_value(&relative_path))
+            .collect()
+    }
+
+    pub(super) fn insert_row_values(
+        &self,
+        location: &ObjectPath,
+        line_number: usize,
+        object: &mut serde_json::Map<String, Value>,
+    ) -> Result<()> {
+        for column in &self.row_columns {
+            object.insert(
+                column.name.clone(),
+                column.row_value(location, line_number)?,
+            );
+        }
+        Ok(())
+    }
+
+    fn file_fields(&self) -> Vec<FieldRef> {
+        self.file_columns
+            .iter()
+            .map(FileMetadataColumn::arrow_field)
+            .collect()
+    }
+
+    fn column_named(&self, name: &str) -> Option<&FileMetadataColumn> {
+        self.file_columns
+            .iter()
+            .chain(self.row_columns.iter())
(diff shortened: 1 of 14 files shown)


Bradley-Butcher:
* **#1168** <a href="https://app.graphite.com/github/pr/withcoral/coral/1168?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1164** <a href="https://app.graphite.com/github/pr/withcoral/coral/1164?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1128** <a href="https://app.graphite.com/github/pr/withcoral/coral/1128?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1128?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

AlbertQM (approved):

crates/coral-engine/src/backends/file/metadata.rs
```diff
@@ -0,0 +1,234 @@
...
+fn metadata_arrow_type(kind: FileMetadataKind) -> DataType {
+    match kind {
+        FileMetadataKind::RelativePath
+        | FileMetadataKind::FileName
+        | FileMetadataKind::FileStem => DataType::Utf8,
+        FileMetadataKind::LineNumber => DataType::Int64,
+    }
+}
+
+fn relative_file_path(table_path: &ListingTableUrl, location: &ObjectPath) -> Result<String> {
```
jsummerfield: This returns incorrect metadata for supported single-file JSONL tables. If `source.location` is `file:///.../events.jsonl`, DataFusion treats it as a single object and `meta.location` equals the table prefix. `strip_prefix` then yields no remaining segments, so `relative_path`, `file_name`, and `file_stem` become `""`. Could we handle the exact-prefix case by deriving from the scanned location itself, and add a single-file metadata test?

crates/coral-engine/src/backends/file/metadata.rs
```diff
@@ -0,0 +1,234 @@
...
+    let Some(segments) = table_path.strip_prefix(location) else {
+        return Ok(location.to_string());
+    };
+
+    segments
```
jsummerfield: This decoding rewrites valid object-store keys. `object_store::Path` preserves literal percent sequences, so a file named `L%3ABC.jsonl` is a distinct path from `L:ABC.jsonl`; decoding here reports the wrong provenance, and invalid percent text can fail metadata projection. The file metadata should preserve the object-store path text after prefix stripping.

crates/coral-engine/src/backends/file/metadata.rs
Bradley-Butcher: Fixed in 8d6896e4. `relative_path` now handles the exact-prefix single-file case by falling back to the scanned file name, so `relative_path`, `file_name`, and `file_stem` are populated for `source.location = file:///.../events.jsonl`.

Added `jsonl_metadata_columns_work_for_single_file_locations` to lock this path.

crates/coral-engine/src/backends/file/metadata.rs
Bradley-Butcher: Fixed in 8d6896e4. File metadata now preserves the object-store path text after prefix stripping; it no longer URL-decodes literal percent sequences.

Added `jsonl_metadata_preserves_object_store_path_text`, covering both `L%3ABC.jsonl` and invalid percent text like `bad%ZZ.jsonl`.

crates/coral-spec/src/backends/file.rs
```diff
@@ -229,6 +231,20 @@ impl FileSourceSpec {
                 )));
             }
         }
+        if !self.metadata.is_empty() && format != FileFormat::Jsonl {
```
antonmry: I’m concerned this makes both the spec and implementation narrower than the feature really needs to be. `relative_path`, `file_name`, and `file_stem` are file-scoped values derived from the matched object path, so they should be valid for CSV/Parquet/JSON as well as JSONL. Only `line_number` seems truly JSONL-specific.

Could we split this by metadata kind instead of gating all `source.metadata` on `format=jsonl`? Ideally the file-scoped metadata path would live in the shared file listing/provider layer, while JSONL keeps the row-scoped `line_number` injection. That would avoid baking in a JSONL-only public contract and a JSON-provider-specific implementation for metadata that is mostly format-agnostic.

Bradley-Butcher: makes sense - will do

antonmry (approved):

crates/coral-spec/src/backends/file.rs
Bradley-Butcher: Fixed in f10739e0.

Split this by metadata kind now:
- `relative_path`, `file_name`, and `file_stem` are valid for file-backed tables across formats.
- `line_number` remains JSONL-only and still uses the row rewrite path.
- file-scoped metadata now runs through a shared file-group scan path, so CSV/Parquet/JSON/JSONL can all attach per-file constants instead of relying on JSON-provider-specific logic.

Added schema/parser coverage for non-JSONL file metadata vs `line_number`, plus a runtime test that queries file-scoped metadata across Parquet, JSONL, JSON, and CSV.

Validation: `make rust-checks` and `make docs-check` both pass.

crates/coral-engine/src/backends/file/provider.rs
```diff
@@ -115,6 +122,153 @@ impl FileTableProvider {
...
+            listing_options,
+            partition_columns,
+        } = prepare_listing_table(&ctx, source_schema, table, home_dir, resolved_inputs).await?;
+
+        let file_schema =
+            file_schema_for_table(&ctx, table, source_schema, &listing_options, &table_path)
+                .await?;
+        let file_schema = strip_partition_columns(file_schema, &partition_columns);
+        let table_schema = TableSchema::new(file_schema, partition_columns.arrow_fields());
+        let table_schema = metadata_columns.extend_table_schema_if_present(table_schema);
```
jsummerfield, line 180: For inferred Parquet tables, this can append metadata fields without checking the actual inferred file schema for name collisions. `validate_derived_column_overlap` only sees manifest-declared columns, so a Parquet dataset that already has a physical `file_path` column can still declare metadata `file_path` and end up with a duplicate/ambiguous table schema here. Could we check the stripped inferred schema before extending it with metadata fields, or reject metadata names that already exist in the file schema?

docs/reference/source-spec-reference.mdx
```diff
@@ -396,6 +396,39 @@ Hive-style partitioning. Positional segment partitioning is currently limited to
 JSON and JSONL tables because those formats already use Coral's custom JSON row
 mapper.
 
+### JSONL file metadata
```
jsummerfield: This section is now stale relative to the implementation. The pushed code and tests support file-scoped metadata (`relative_path`, `file_name`, `file_stem`) across Parquet, JSON, JSONL, and CSV; only `line_number` is JSONL-only. Could we reword this from “JSONL file metadata” to file metadata generally, and call out the JSONL-only restriction just for `line_number`?

jsummerfield (approved):

crates/coral-engine/src/backends/file/provider.rs
Bradley-Butcher: Addressed in the current head. The metadata-backed provider now checks the stripped inferred file schema before appending metadata fields:

`metadata_columns.reject_schema_collisions(&file_schema, source_schema, table.name())?;`

That catches inferred Parquet physical columns like `file_path` before the metadata fields are added. The regression test is `backends::file::tests::parquet_inferred_schema_rejects_metadata_name_collision`, which builds an inferred Parquet table with a physical `file_path` column and verifies registration fails with the metadata/file-column collision error.

docs/reference/source-spec-reference.mdx
Bradley-Butcher: Addressed in the current docs. The section is now titled `File metadata`, says file-backed tables can expose metadata columns derived from each scanned file, and calls out the JSONL-only restriction only for `line_number`:

- `relative_path`, `file_name`, and `file_stem` are file-scoped metadata.
- `line_number` is row-scoped and only supported for `format: jsonl`.

It also documents that metadata names must not duplicate declared payload columns, partition columns, or physical file columns.

PR #1138, opened 2026-06-02
fix(engine): explain table-function describe errors

Bradley-Butcher:
## Intent

Coral has two queryable object kinds inside source schemas:

- tables, which can be referenced directly in `FROM` and described with `DESCRIBE <schema>.<table>`
- table functions, which must be called with parentheses and any required arguments before they produce rows

On `main`, using a table function where SQL expects a table falls through to DataFusion's generic missing-table error. For example, `DESCRIBE datadog.metrics` says `datadog.metrics` is not a table, but it does not tell the user that `metrics` is actually registered as a table function or where to find its arguments.

This PR makes that failure explicit without changing DataFusion's normal `DESCRIBE` output for real tables.

## What Changed

When DataFusion reports a missing relation during planning, Coral checks the registered table-function catalog before returning the normal `TABLE_NOT_FOUND` diagnostic.

If the schema-qualified relation matches a table function, Coral returns a structured query error with:

- reason: `TABLE_FUNCTION_NOT_TABLE`
- summary: `` `<schema>.<function>` is a table function, not a table ``
- detail: query it as `FROM <schema>.<function>(...)`; inspect arguments/result columns in `coral.table_functions`; or run `DESCRIBE SELECT * FROM <schema>.<function>(...)` once required arguments are known
- hint: a concrete `coral.table_functions` query for that schema/function

The same reason is used for `DESCRIBE datadog.metrics` and `SELECT * FROM datadog.metrics`, because both hit the same object-kind mistake. The remediation names the valid `DESCRIBE SELECT * FROM ...` form so the `DESCRIBE` case does not imply unsupported `DESCRIBE schema.function(args...)` syntax.

## User Impact

After this change:

```text
Error: `datadog.metrics` is a table function, not a table
Detail: `datadog.metrics` is registered as a table function. Query it as `FROM datadog.metrics(...)`; inspect arguments and result columns in `coral.table_functions`, or run `DESCRIBE SELECT * FROM datadog.metrics(...)` after filling any required arguments.
Hint: Inspect table functions with `SELECT schema_name, function_name, arguments_json, result_columns_json FROM coral.table_functions WHERE schema_name = 'datadog' AND function_name = 'metrics'`.
```

Users can inspect the callable shape before knowing the arguments:

```sql
SELECT schema_name, function_name, arguments_json, result_columns_json
FROM coral.table_functions
WHERE schema_name = 'datadog'
  AND function_name = 'metrics';
```

Existing successful paths are unchanged:

- `DESCRIBE datadog.hosts` still returns DataFusion's three-column describe output: `column_name`, `data_type`, `is_nullable`.
- `DESCRIBE SELECT * FROM datadog.metrics(...)` still describes the result schema of the called table function.
- real missing tables still use the existing `TABLE_NOT_FOUND` path and table suggestions.

## Boundaries

This stays in `coral-engine`'s DataFusion-to-Core error normalization layer. DataFusion still owns SQL planning and `DESCRIBE` behavior; Coral only enriches a failed relation lookup with catalog knowledge it already has.

This PR does not add a statement-planner hook, does not parse SQL a second time to distinguish `DESCRIBE` from `SELECT`, and does not add API/proto surface.

The special diagnostic only fires for schema-qualified references such as `datadog.metrics`. Bare names such as `metrics` remain ordinary missing-table errors because they are ambiguous without a schema.

## Edge Cases Covered

- a full DataFusion catalog-qualified ref such as `datafusion.datadog.metrics`
- a source schema literally named `datafusion`, such as `datafusion.metrics`
- quoted source schema pieces such as `"foo.bar".metrics`
- case-insensitive matching for unquoted references

## Not in This PR

- No custom `DESCRIBE` output shape for table functions.
- No `DESCRIBE FUNCTION` SQL extension.
- No new public API/proto field for table-function describe metadata.
- No change to the `coral.table_functions` catalog schema.

## Validation

Local validation:

- `cargo fmt --check`
- `cargo test -p coral-engine --lib table_function_ref --all-features --locked`
- `cargo test -p coral-engine --lib table_function_not_table_points_at_table_functions_catalog --all-features --locked`
- `cargo test -p coral-engine --test engine table_function_used_as_table_returns_guided_error --all-features --locked`
- earlier branch validation: `cargo check -p coral-engine --lib --locked`, `cargo clippy -p coral-engine --lib --all-features --locked -- -D warnings`, focused describe/table-function integration tests, and manual CLI smoke for `DESCRIBE datadog.metrics` and `SELECT * FROM datadog.metrics`

GitHub Validate was green before the review-feedback amendment; it should rerun on the new commit.

### DIFF
--- a/crates/coral-engine/src/contracts/query_error.rs
+++ b/crates/coral-engine/src/contracts/query_error.rs
@@ -4,7 +4,7 @@ use std::collections::HashMap;
 
 use datafusion::common::utils::quote_identifier;
 
-use super::catalog::TableInfo;
+use super::catalog::{TableFunctionInfo, TableInfo};
 use super::error::StatusCode;
 
 /// Wire-stable `reason` code for unknown-column errors.
@@ -12,6 +12,9 @@ pub(crate) const UNKNOWN_COLUMN_REASON: &str = "UNKNOWN_COLUMN";
 
 /// Wire-stable `reason` code for table-not-found errors.
 pub(crate) const TABLE_NOT_FOUND_REASON: &str = "TABLE_NOT_FOUND";
+
+/// Wire-stable `reason` code for table functions used where a table is needed.
+pub(crate) const TABLE_FUNCTION_NOT_TABLE_REASON: &str = "TABLE_FUNCTION_NOT_TABLE";
 
 /// Minimum `strsim::normalized_levenshtein` score for a "did you mean?"
 /// suggestion to surface.
@@ -205,6 +208,32 @@ impl StructuredQueryError {
             hint,
             false,
             StatusCode::NotFound,
+            metadata,
+        )
+    }
+
+    /// Builds a structured error for a table function used where SQL expects a table.
+    pub(crate) fn table_function_not_table(function: &TableFunctionInfo) -> Self {
+        let display_ref = format_schema_function(function);
+        let mut metadata = HashMap::new();
+        metadata.insert("object_kind".to_string(), "table_function".to_string());
+        metadata.insert("schema".to_string(), function.schema_name.clone());
+        metadata.insert("function".to_string(), function.function_name.clone());
+
+        let schema_sql = sql_string_literal(&function.schema_name);
+        let function_sql = sql_string_literal(&function.function_name);
+
+        Self::new(
+            TABLE_FUNCTION_NOT_TABLE_REASON,
+            format!("`{display_ref}` is a table function, not a table"),
+            format!(
+                "`{display_ref}` is registered as a table function. Query it as `FROM {display_ref}(...)`; inspect arguments and result columns in `coral.table_functions`, or run `DESCRIBE SELECT * FROM {display_ref}(...)` after filling any required arguments."
+            ),
+            Some(format!(
+                "Inspect table functions with `SELECT schema_name, function_name, arguments_json, result_columns_json FROM coral.table_functions WHERE schema_name = {schema_sql} AND function_name = {function_sql}`."
+            )),
+            false,
+            StatusCode::InvalidArgument,
             metadata,
         )
     }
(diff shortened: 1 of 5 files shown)


antonmry (approved):

AlbertQM (approved):

crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -209,6 +212,32 @@ impl StructuredQueryError {
...
+        metadata.insert("function".to_string(), function.function_name.clone());
+
+        let schema_sql = sql_string_literal(&function.schema_name);
+        let function_sql = sql_string_literal(&function.function_name);
+
+        Self::new(
+            TABLE_FUNCTION_NOT_TABLE_REASON,
+            format!("`{display_ref}` is a table function, not a table"),
+            format!(
+                "`{display_ref}` is registered as a table function. Call it with parentheses and any required arguments before querying it."
```
jsummerfield: Could we make this remediation cover the `DESCRIBE` case explicitly? This error also fires for `DESCRIBE searchy.search_issues`, and “Call it with parentheses…” can point users toward `DESCRIBE searchy.search_issues(q => ...)`, while the valid schema-inspection form is `DESCRIBE SELECT * FROM searchy.search_issues(...)` or querying `coral.table_functions`. Suggested shape: use it in a query as `FROM schema.function(args...)`; to inspect its result schema, run `DESCRIBE SELECT * FROM schema.function(args...)` or query `coral.table_functions`.

jsummerfield (commented):
I left one inline comment on the remediation wording.

jsummerfield (approved):

crates/coral-engine/src/contracts/query_error.rs
Bradley-Butcher: Yep, fixed in 009ea6a. The detail now says to query it as `FROM schema.function(...)`, inspect `coral.table_functions`, or run `DESCRIBE SELECT * FROM schema.function(...)` after filling required arguments. The integration test now asserts the `FROM ...` and `DESCRIBE SELECT * FROM ...` guidance for both `DESCRIBE` and `SELECT * FROM` uses.

PR #1140, opened 2026-06-02
feat(sources/community/discord): add thread channel coverage

kyracheng:
## Summary
- add Discord thread-channel coverage for active threads, archived public/private threads, and joined private archived threads
- document how forum/media posts map to thread IDs before querying discord.messages
- expose additional forum/media channel fields on discord.channels

## Validation
- coral source lint sources/community/discord/manifest.yaml
- git diff --check
- coral source test discord passed against the currently installed source, but that installed source still reports 6 tables, so it does not validate the new manifest tables

## Validation gap
- live validation of the updated local manifest was not run because coral source add --file sources/community/discord/manifest.yaml requires DISCORD_BOT_TOKEN, which is not set in this shell
- cargo tests were not run because cargo is not installed in this environment

### DIFF
--- a/sources/community/discord/README.md
+++ b/sources/community/discord/README.md
@@ -1,6 +1,6 @@
 # Discord
 
-Query Discord bot identity, guilds, channels, messages, members, and roles from the [Discord REST API v10](https://discord.com/developers/docs/intro).
+Query Discord bot identity, guilds, channels, threads, messages, members, and roles from the [Discord REST API v10](https://discord.com/developers/docs/intro).
 
 ## Prerequisites
 
@@ -17,6 +17,9 @@ Configure the following by table:
 | `current_user` | Token only | none | none |
 | `guilds` | Token only | none | none |
 | `channels` | `guild_id` filter | none | `VIEW_CHANNEL` (`0x400`) |
+| `active_threads` | `guild_id` filter | none | `VIEW_CHANNEL` (`0x400`), `READ_MESSAGE_HISTORY` (`0x10000`) |
+| `archived_threads` | `channel_id`, `visibility` filters | none | `VIEW_CHANNEL` (`0x400`), `READ_MESSAGE_HISTORY` (`0x10000`); private archives may require `MANAGE_THREADS` |
+| `joined_private_archived_threads` | `channel_id` filter | none | `VIEW_CHANNEL` (`0x400`), `READ_MESSAGE_HISTORY` (`0x10000`) |
 | `messages` | `channel_id` filter | `MESSAGE_CONTENT` | `VIEW_CHANNEL` (`0x400`), `READ_MESSAGE_HISTORY` (`0x10000`) |
 | `members` | `guild_id` filter | `GUILD_MEMBERS` | none |
 | `roles` | `guild_id` filter | none | `VIEW_CHANNEL` (`0x400`) |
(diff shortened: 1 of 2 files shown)


sources/community/discord/manifest.yaml
```diff
@@ -447,20 +474,312 @@ tables:
...
+    response:
+      rows_path:
+        - threads
+    pagination:
+      mode: none
+      page_size:
+        default: 50
+        max: 100
+        query_param: limit
+    columns: *thread_columns
```
jsummerfield: This reuses `*thread_columns`, but the table's required filters (`channel_id`, `visibility`, and optional `before`) are not columns. Coral builds the table schema from `columns`, then extracts filter values from DataFusion predicates over schema columns, so the documented `WHERE channel_id = ... AND visibility = ...` cannot bind and the required filters never reach the request. Please add table-specific virtual `from_filter` columns here; `joined_private_archived_threads` needs the same treatment for `channel_id`/`before`.

sources/community/discord/README.md
```diff
@@ -17,6 +17,9 @@ Configure the following by table:
 | `current_user` | Token only | none | none |
 | `guilds` | Token only | none | none |
 | `channels` | `guild_id` filter | none | `VIEW_CHANNEL` (`0x400`) |
+| `active_threads` | `guild_id` filter | none | `VIEW_CHANNEL` (`0x400`), `READ_MESSAGE_HISTORY` (`0x10000`) |
+| `archived_threads` | `channel_id`, `visibility` filters | none | `VIEW_CHANNEL` (`0x400`), `READ_MESSAGE_HISTORY` (`0x10000`); private archives may require `MANAGE_THREADS` |
```
jsummerfield: This should be exact rather than "may require": Discord's `/threads/archived/private` endpoint requires both `READ_MESSAGE_HISTORY` and `MANAGE_THREADS`; only the joined-private archived endpoint avoids `MANAGE_THREADS`. As written, a bot with the listed minimal permissions can still hit 403s for `visibility = 'private'`. See Discord's channel resource docs: https://docs.discord.com/developers/resources/channel

sources/community/discord/README.md
```diff
@@ -56,6 +64,9 @@ Use the returned IDs as required filters for the `channels`, `messages`, `member
 | `current_user` | The Discord bot user associated with the configured token | none |
 | `guilds` | Discord guilds that the configured bot can see | none |
 | `channels` | Channels in a Discord guild | `guild_id` |
+| `active_threads` | Active thread channels in a Discord guild | `guild_id` |
+| `archived_threads` | Public or private archived thread channels in a parent channel | `channel_id`, `visibility` |
+| `joined_private_archived_threads` | Private archived thread channels the bot has joined | `channel_id` |
```
jsummerfield, line 69: Since these thread tables are now documented, the validation section lower in the README also needs updating. It still says "all six tables" and shows `discord (6 tables)`, so this README now claims validation for the old manifest shape rather than the nine-table source.

sources/community/discord/manifest.yaml
```diff
@@ -3,8 +3,8 @@ version: 0.1.0
 dsl_version: 3
 backend: http
 description: >-
-  Query Discord bot identity, guilds, channels, messages, members, and roles
-  from the Discord REST API.
+  Query Discord bot identity, guilds, channels, threads, messages, members, and
```
jsummerfield, line 6: This source now exposes new user-visible thread tables and channel columns, so please bump the manifest `version` from `0.1.0` as part of this change.

jsummerfield (commented):
Leaving inline comments on the actionable review findings.

PR #1143, opened 2026-06-03
fix(spec): resolve DSL v4 response refs

### LINKED ISSUE #1121: Resolve OpenAPI $ref response objects during DSL v4 import
## Problem

DSL v4 OpenAPI import currently misses valid response objects that are expressed as `$ref`s. `import_response` calls `select_json_response`, and `select_json_response` looks for `content` directly on each response object without first resolving `$ref`.

Code references:
- `crates/coral-spec/src/v4/mod.rs:972`
- `crates/coral-spec/src/v4/mod.rs:1290`

A valid OpenAPI shape like this is treated as if the operation has no JSON response:

```yaml
responses:
  '200':
    $ref: '#/components/responses/Foo'
components:
  responses:
    Foo:
      content:
        application/json:
          schema:
            type: array
            items:
              type: object
```

## Impact

The importer falls through to `OutputCardinality::None`, which causes otherwise valid read operations to be hidden/dropped from the generated projection catalog. This can make major portions of a real OpenAPI descriptor disappear from DSL v4 sources.

## Acceptance criteria

- Response `$ref`s under operation responses are resolved before selecting JSON responses.
- Missing or external response refs produce diagnostics consistent with existing schema/parameter ref diagnostics.
- A regression test covers a `200` response that references `#/components/responses/...` and verifies the projection is published.

simonwhitaker:
## Summary

- Resolve OpenAPI response-object `$ref`s before selecting JSON response content in the DSL v4 importer.
- Keep response schema `$ref` handling unchanged after response selection.
- Add regression coverage for referenced `components.responses` objects and missing/external response refs.

## Why

A valid OpenAPI response like `200: { $ref: '#/components/responses/Foo' }` was skipped before schema resolution, causing otherwise valid read operations to fall through to `OutputCardinality::None` and disappear from generated projections.

Fixes #1121.

## Validation

- `cargo test -p coral-spec --lib importer_`

### DIFF
--- a/crates/coral-spec/src/v4/surfaces/openapi/responses.rs
+++ b/crates/coral-spec/src/v4/surfaces/openapi/responses.rs
@@ -20,9 +20,11 @@ impl OpenApiImporter<'_> {
         RestResponseAttachment,
         Option<IrEntityCandidate>,
     ) {
-        let Some((status_code, media_type, schema)) =
-            select_json_response(operation.get("responses").and_then(Value::as_object))
-        else {
+        let Some((status_code, media_type, schema)) = self.select_json_response(
+            operation.get("responses").and_then(Value::as_object),
+            operation_id,
+            diagnostics,
+        ) else {
             let response = ResponseSpec::default();
             return (
                 IrOperationOutput {
@@ -39,7 +41,7 @@ impl OpenApiImporter<'_> {
             );
         };
 
-        let Some(resolved) = self.resolve_ref(schema, operation_id, diagnostics) else {
+        let Some(resolved) = self.resolve_ref(&schema, operation_id, diagnostics) else {
             diagnostics.push(Diagnostic::warning(
                 "OPENAPI_RESPONSE_SCHEMA_UNRESOLVED",
                 format!("operation '{operation_id}' response schema could not be resolved"),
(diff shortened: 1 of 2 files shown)


crates/coral-spec/src/v4/surfaces/openapi/responses.rs
```diff
@@ -95,32 +97,39 @@ impl OpenApiImporter<'_> {
...
+        let responses = responses?;
+        let mut candidates = Vec::new();
+        for (status, response) in responses {
+            let Ok(status_code) = status.parse::<u16>() else {
+                continue;
+            };
+            if !(200..300).contains(&status_code) {
+                continue;
+            }
+            let Some(response) = self.resolve_ref(response, operation_id, diagnostics) else {
```
jsummerfield, line 115: This now skips an unresolved response-object `$ref` during response selection, so a declared 200 response can fall through to the no-response default (`OutputCardinality::None` plus a synthetic 204). That seems different from unresolved schema refs, which preserve the declared response context as `Unknown`. Could we choose the raw response first, resolve it separately, and return `Unknown` if the selected response ref cannot be resolved?

simonwhitaker: I'm not sure I understand. We're in this world:

```
200: { $ref: '#/components/responses/Foo' }
```

So the raw response would be:

```
{ $ref: '#/components/responses/Foo' }
```

Furthermore, we're in the world where this ref does not resolve. So in effect, the response object we're dealing with is:

```
{ }
```

You said "Could we [...] return `Unknown` if the selected response ref does not resolve". This function returns a tuple of `(u16, String, Value)` (status code, content type, response schema), so are you suggesting that in this case we treat this response candidate as `(200, "application/json", Value::Null)` instead of ignoring it? If so, is that better than the fall-through behaviour? They're both somewhat synthetic, no? (Who knows whether this route's content type is indeed JSON?) And it doesn't feel to me like either is the obvious right way to treat a broken spec.

What am I missing?

jsummerfield: You’re (directionally) right.

The concern I was trying to point at is that the current tuple can’t distinguish “there was no usable JSON response” from “the selected response object was a broken ref.” But I agree that fixing that properly would need a different selection state. I don’t think this should block the PR on its own.

jsummerfield (approved):

PR #1146, opened 2026-06-03
fix(spec): resolve OpenAPI response refs

### LINKED ISSUE #1121: Resolve OpenAPI $ref response objects during DSL v4 import

jsummerfield:
## Intent

OpenAPI response objects can legally be `$ref`s, and OpenAPI success responses can be declared with a wildcard key such as `2XX`. The DSL v4 importer resolved schema refs after choosing a JSON response, but it chose that response by reading `content.application/json` directly from unresolved response objects and only accepted numeric status-code keys. That meant valid operations with referenced or wildcard success responses were treated as having no JSON output and could disappear from generated projections.

## Changes

- Resolve OpenAPI response-object refs before selecting JSON content.
- Accept valid wildcard `2XX` success response keys.
- Prefer concrete success responses over wildcard fallbacks when both are present.
- Add focused regression fixtures for referenced responses and wildcard success responses.

## Verification

- `cargo test -p coral-spec importer_ -- --nocapture`
- `cargo fmt --package coral-spec`
- `git diff --check`

Fixes https://github.com/withcoral/coral/issues/1121

### DIFF
--- a/crates/coral-spec/src/v4/mod.rs
+++ b/crates/coral-spec/src/v4/mod.rs
@@ -969,9 +969,11 @@ impl<'a> OpenApiImporter<'a> {
         RestResponseAttachment,
         Option<IrEntityCandidate>,
     ) {
-        let Some((status_code, media_type, schema)) =
-            select_json_response(operation.get("responses").and_then(Value::as_object))
-        else {
+        let Some((status_code, media_type, schema)) = self.select_json_response(
+            operation.get("responses").and_then(Value::as_object),
+            operation_id,
+            diagnostics,
+        ) else {
             let response = ResponseSpec::default();
             return (
                 IrOperationOutput {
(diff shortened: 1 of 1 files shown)


simonwhitaker:
Ha, [snap](https://github.com/withcoral/coral/pull/1143). Please mark this one as resolving issue #1121

simonwhitaker (commented):
I'd like to get #1149 out first, then I've got a long list of todo items, this one included.

PR #1164, opened 2026-06-03
feat(engine): add source-scoped SQL view runtime

Bradley-Butcher:
## Intent

Add private engine runtime support for source-scoped SQL views: table-shaped relations backed by SQL over the provider tables for the same source.

This gives the runtime a clean place to host derived source relations without asking agents to repeat raw JSON parsing or source-specific query shapes every time.

## Relation to prepared statements / #889

This is adjacent to source-declared prepared statements, but it is solving a narrower table-shaping problem.

Prepared statements are an invocation surface: "run this saved query", usually through `EXECUTE`. They are useful for reusable workflows, parameterized queries, and source-authored actions that happen to return rows.

Source SQL views are a relation surface: "this source has another table-shaped relation". They live in the normal table universe, so callers can use ordinary `SELECT`, filters, joins, limits, catalog discovery, and whatever agents already know how to do with tables.

The motivating case here is raw audit/event data:

- raw table: `codex.events(timestamp, type, payload, ...)`
- derived relations: `codex.sessions`, `codex.messages`, `codex.tool_calls`

For that shape, I do not want every agent or recipe redoing JSON archaeology over `payload`. I want the source to expose useful semantic tables.

This PR is intentionally not a general memory surface, not agent-authored saved views, and not public manifest syntax yet. It only adds the private engine runtime needed to register source-owned SQL-backed relations. The source-spec surface and Codex transcript tables are stacked separately so we can get mileage before widening the contract.

## What changed

- `BackendRegistration` now carries registered source table implementations, not only provider tables.
- A registered source relation can be provider-backed or SQL-backed.
- Source registration builds provider tables first, plans SQL views against those provider tables, then publishes one immutable source schema.
- SQL view planning uses DataFusion logical planning and `DataFrame::into_view()` rather than string-built `CREATE VIEW` statements.
- Temporary planning schemas restore any pre-existing DataFusion schema after planning, including failed planning.
- View validation rejects non-read-only SQL, multiple statements, constant selects with no source table scan, unqualified references, cross-source references, three-part references, duplicate view names, provider/view name conflicts, and declared schema drift.

## Boundaries

- No public source manifest, docs, CLI, MCP, or API surface is added in this PR.
- File manifest `views:` support is stacked in the next PR.
- Codex transcript tables are stacked after that.
- View-on-view support is intentionally not included: source views are planned against provider-backed source tables only.
- Views over table functions are intentionally not included: source views are planned before UDTFs are registered, and PR2 only supports SQL views over provider-backed source tables.
- Constant/static views are intentionally not included: a source view must reference at least one source-local provider table.
- Existing file, HTTP, and MCP backends continue to register provider tables in this PR.

## Validation

- `make rust-checks`: 993 tests passed.
- `make docs-check`.
- `git diff --check`.
- Focused engine coverage for source view planning, schema validation, table reference restrictions, constant-view rejection, duplicate/conflict handling, planning-schema restoration, and published static schemas.
- Real-config binary smoke on `codex-source-sql-views-runtime` (`coral 0.4.1+b336e004`) using Brad's installed Coral config, without adding sources: `coral.tables`/`coral.table_functions` registered `datadog`, `github`, `google_docs`, `google_meet`, `linear`, `sentry`, `slack`, and `wandb`; live queries succeeded for GitHub `search_issues`, Linear `teams` and `team_issues`, Slack `channels` and `messages`, Datadog `monitors`, and Sentry `projects`. Slack also surfaced a real provider error (`not_in_channel`) for an inaccessible channel before succeeding against `general`.

### DIFF
--- a/crates/coral-engine/src/runtime/registry.rs
+++ b/crates/coral-engine/src/runtime/registry.rs
@@ -1,18 +1,21 @@
 //! Registers compiled backend sources into a shared `DataFusion` session.
 
+use std::collections::HashSet;
 use std::sync::Arc;
 
+use datafusion::catalog::{CatalogProvider, MemorySchemaProvider, SchemaProvider};
 use datafusion::error::{DataFusionError, Result as DataFusionResult};
 use datafusion::prelude::SessionContext;
 use tracing::{Instrument as _, info_span};
 
 use crate::backends::{
     BackendRegistration, BackendRegistrationContext, CompiledBackendSource, RegisteredSource,
-    SourceTableFunctions,
+    RegisteredSourceTable, RegisteredTableImplementation, SourceTableFunctions,
 };
 use crate::runtime::error::{datafusion_to_core, source_decorator_error_to_core};
 use crate::runtime::schema_provider::StaticSchemaProvider;
-use crate::{CoreError, QuerySource, SourceDecorator, SourceFailurePolicy};
+use crate::runtime::source_views::{SourceSqlView, build_source_views};
+use crate::{CoreError, QuerySource, SourceDecorator, SourceFailurePolicy, SourceTables};
 
 const RESERVED_SCHEMA_NAMES: &[&str] = &["coral", "coral_admin"];
 
@@ -24,6 +27,11 @@ const RESERVED_SCHEMA_NAMES: &[&str] = &["coral", "coral_admin"];
 pub(crate) struct CompiledQuerySource {
     pub(crate) source: QuerySource,
     pub(crate) compiled: Box<dyn CompiledBackendSource>,
+}
+
+struct SourceTableImplementations {
+    provider_tables: SourceTables,
+    sql_views: Vec<SourceSqlView>,
 }
 
 /// One selected source's readiness for runtime registration.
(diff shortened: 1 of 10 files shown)


Bradley-Butcher:
* **#1168** <a href="https://app.graphite.com/github/pr/withcoral/coral/1168?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1164** <a href="https://app.graphite.com/github/pr/withcoral/coral/1164?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1164?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1128** <a href="https://app.graphite.com/github/pr/withcoral/coral/1128?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

crates/coral-engine/src/runtime/registry.rs
```diff
@@ -215,6 +251,99 @@ async fn register_source(
...
+fn register_source_planning_schema(
+    catalog: &Arc<dyn CatalogProvider>,
+    schema_name: &str,
+    provider_tables: &SourceTables,
+) -> DataFusionResult<()> {
+    let planning_schema = MemorySchemaProvider::new();
+    for (table_name, provider) in provider_tables {
+        planning_schema.register_table(table_name.clone(), Arc::clone(provider))?;
+    }
+    catalog.register_schema(schema_name, Arc::new(planning_schema))?;
```
antonmry: Could we guard against replacing an existing schema here? My read is that `register_schema` can return a previous schema for the same name, but this path always removes `schema_name` after view planning. If a source name collides with an existing DataFusion/runtime schema and planning fails, cleanup could remove the pre-existing schema. Would it be safer to treat `Some(_)` as an error here, or restore the previous schema during cleanup?

crates/coral-engine/src/runtime/source_views.rs
```diff
@@ -0,0 +1,589 @@
...
+    validate_source_view_plan(schema_name, view, &plan)?;
+    Ok(plan)
+}
+
+fn validate_source_view_plan(
+    schema_name: &str,
+    view: &SourceSqlView,
+    plan: &LogicalPlan,
+) -> Result<()> {
+    plan.apply_with_subqueries(|node| match node {
```
antonmry, line 87: Could we document whether constant views are intentionally supported? As written, a view like `SELECT 1 AS id` can pass validation if its declared schema matches, which may be useful for source authors. It would be helpful for the spec/docs to make that explicit so users know whether SQL views are allowed to define constant rows, not only projections over provider tables.

antonmry (approved):

crates/coral-engine/src/runtime/source_views.rs
Bradley-Butcher: We ended up making constants invalid for this runtime slice. Source SQL views must derive from at least one source-local table, so `SELECT 1 AS id` now fails validation. The public DSL3 docs for the stacked spec PR describe this constraint.

crates/coral-engine/src/runtime/registry.rs
Bradley-Butcher: Fixed in the current head by making the temporary planning schema restore the previous schema when DataFusion had one. Added `invalid_source_view_restores_existing_schema` to lock the failure path.

Bradley-Butcher:
^ github dragged in other codeowners due to graphite restack. Have removed. Feel free to comment if you have opinions tho

Bradley-Butcher:
Closing this stack for now. We decided we don't need this source-view path.

PR #1171, opened 2026-06-03
feat(engine): substitute input variables in test_queries

Kvnpsiddhartha:
## Intent

Manifest `test_queries` were executed literally, so tokens like
`billingAccounts/{{input.GCP_BILLING_ACCOUNT_ID}}` reached the API
unexpanded during `coral source add` validation.

## What it enables

- `test_queries` may use `{{input.VARIABLE_KEY}}` with the same syntax as
  HTTP path and header templates.
- Values come from installed `kind: variable` bindings (including manifest
  defaults applied at install). Secrets are not substituted.
- Render failures are reported per query test; the overall validate call
  still returns a report (same as execution failures today).

## Usage example

```yaml
inputs:
  BILLING_ACCOUNT_ID:
    kind: variable
    required: true
test_queries:
  - >-
    SELECT name FROM my_source.budgets(
      billing_account => 'billingAccounts/{{input.BILLING_ACCOUNT_ID}}'
    ) LIMIT 1


Kvnpsiddhartha:
@jsummerfield, I wanted to follow up on this PR as it's been open for a few days. Could you please take a look and share any feedback?

Kvnpsiddhartha:
Fixes #1253

PR #1179, opened 2026-06-04
fix(spec): require omitted OpenAPI path params

### LINKED ISSUE #1155: Treat OpenAPI path parameters as required in v4 import
This tracks unresolved feedback identified by cubic on PR #1149:
https://github.com/withcoral/coral/pull/1149#discussion_r3347190640

Context:
This appears to be pre-existing DSL v4 behavior surfaced by the refactor review, not a regression from the module split.

Problem:
OpenAPI path parameters are imported as optional when `required` is omitted, even though path parameters are required by OpenAPI. This can misclassify required path inputs as optional SQL filters.

Acceptance criteria:
Import path parameters as required regardless of an omitted `required` field, preserve explicit behavior for non-path parameters, and add focused importer/projection coverage.

simonwhitaker:
## Summary

- Treat OpenAPI path parameters as required when their `required` flag is omitted during DSL v4 import.
- Preserve omitted non-path parameters as optional.
- Add focused importer and projection coverage so singleton path operations become table functions with required args.

Fixes #1155.

## Tests

- `cargo test -p coral-spec importer_treats_path_parameters_as_required_when_required_is_omitted`
- `cargo test -p coral-spec projection_generation_uses_omitted_path_required_for_table_function_args`
- `make rust-checks`

## Risks

Low. This changes DSL v4 OpenAPI import behavior only for path parameters with omitted `required`; OpenAPI treats path parameters as required, and explicit non-path behavior is preserved.

### DIFF
--- a/crates/coral-spec/src/v4/surfaces/openapi/operations.rs
+++ b/crates/coral-spec/src/v4/surfaces/openapi/operations.rs
@@ -155,10 +155,7 @@ impl OpenApiImporter<'_> {
                 Some(IrOperationInput {
                     name,
                     location,
-                    required: parameter_obj
-                        .get("required")
-                        .and_then(Value::as_bool)
-                        .unwrap_or(false),
+                    required: parameter_is_required(parameter_obj, location),
                     data_type: scalar,
                     default_value: schema.get("default").map(openapi_default_to_string),
                     description: parameter_obj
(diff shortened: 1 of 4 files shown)


jsummerfield (commented):
One inline comment on path parameter requiredness.

crates/coral-spec/src/v4/surfaces/openapi/operations.rs
```diff
@@ -268,6 +265,16 @@ fn parse_parameter_location(location: &str) -> Option<OpenApiParameterLocation>
     }
 }
 
+fn parameter_is_required(
+    parameter_obj: &Map<String, Value>,
+    location: OpenApiParameterLocation,
+) -> bool {
+    parameter_obj
```
jsummerfield, line 275: Could we make path requiredness unconditional for `OpenApiParameterLocation::Path` here? This fixes omitted `required`, but an explicit `required: false` still leaks through, and projection derivation can also clear requiredness when a schema default exists. Since path rendering still needs a concrete value, Coral can publish an optional-looking table/filter or function arg that fails later during request rendering.

jsummerfield (approved):

crates/coral-spec/src/v4/surfaces/openapi/operations.rs
simonwhitaker: Addressed in dd90a216. Path parameters are now imported as required even when OpenAPI says `required: false`, and projection requiredness no longer clears path params that have schema defaults. Added importer/projection regressions for both cases; `make rust-checks` passes locally.

PR #1184, opened 2026-06-04
fix(app): reject SQL-controlled v4 base URLs

### LINKED ISSUE #1150: DSL v4 base_url can be SQL-controlled and leak auth secrets
_New Codex Security finding for withcoral:coral at commit <https://github.com/withcoral/coral/commit/d4b4bd0a0fd9637af34f8b18725d9ef3692016ff|d4b4bd0>_ :large_orange_circle: _High:_ DSL v4 base_url can be SQL-controlled and leak auth secrets &gt; DSL v4 surfaces are validated with validate_input_references, which only checks that input tokens are declared and does not reject other template namespaces such as filter._ or arg._. The app then converts the v4 surface directly into an HttpSourceManifest, preserving the authored base_url and auth configuration. At query time, the HTTP backend renders base_url with the full per-request RenderContext, including SQL-controlled filters and table-function arguments, and then sends the request with the source auth headers. As a result, a v4 manifest can set base_url to a value such as https://{{filter.host}} and define an OpenAPI operation exposing host as a required filter; an untrusted agent or local caller can run a query with host='attacker.example' and cause Coral to send Authorization or other secret-backed headers to that attacker-controlled origin. _Relevant files_
• <https://github.com/withcoral/coral/blob/d4b4bd0a0fd9637af34f8b18725d9ef3692016ff/crates/coral-spec/src/v4/mod.rs#L178-L196|crates/coral-spec/src/v4/mod.rs:L178-L196>
• <https://github.com/withcoral/coral/blob/d4b4bd0a0fd9637af34f8b18725d9ef3692016ff/crates/coral-spec/src/inputs.rs#L1236-L1254|crates/coral-spec/src/inputs.rs:L1236-L1254>
• <https://github.com/withcoral/coral/blob/d4b4bd0a0fd9637af34f8b18725d9ef3692016ff/crates/coral-app/src/sources/runtime_package.rs#L134-L158|crates/coral-app/src/sources/runtime_package.rs:L134-L158>
• <https://github.com/withcoral/coral/blob/d4b4bd0a0fd9637af34f8b18725d9ef3692016ff/crates/coral-engine/src/backends/http/fetch.rs#L88-L110|crates/coral-engine/src/backends/http/fetch.rs:L88-L110>
• <https://github.com/withcoral/coral/blob/d4b4bd0a0fd9637af34f8b18725d9ef3692016ff/crates/coral-engine/src/backends/shared/template.rs#L318-L337|crates/coral-engine/src/backends/shared/template.rs:L318-L337> <https://chatgpt.com/codex/security/findings/36d268974f70819196c837a8af8bd848|View finding →> Ask @Codex: "how can this be exploited?" or "suggest a fix"  •  <https://chatgpt.com/codex/security/scans/github-1200798950|Update notification rule>

[Slack Message](https://withcoral.slack.com/archives/C0B4LFC035Y/p1780467927685039?thread_ts=1780467927.685039&cid=C0B4LFC035Y)

simonwhitaker:
## Summary
- reject DSL v4 OpenAPI `base_url` templates that use SQL-controlled runtime namespaces such as `filter`, `arg`, `state`, `expr`, or unknown namespaces
- apply the same guard defensively when deriving `base_url` from OpenAPI `servers[0].url` and server variable defaults
- document the v4 `base_url` source-input-only constraint in the security guidance

## Tests
- `cargo test -p coral-spec rejects_v4_openapi_base_url_runtime_controlled_tokens`
- `cargo test -p coral-app derived_openapi_server_url_rejects_runtime_controlled_tokens`
- `make rust-checks`
- `make docs-check`

## Risks
- Low-to-moderate. This intentionally rejects v4 OpenAPI sources that put query-time tokens in `base_url`; declared `input` tokens and literal base URLs continue to work.

Fixes #1150

### DIFF
--- a/crates/coral-app/src/sources/runtime_package.rs
+++ b/crates/coral-app/src/sources/runtime_package.rs
@@ -7,7 +7,7 @@ use coral_spec::backends::http::{HttpSourceManifest, HttpTableSpec};
 use coral_spec::v4::{
     ProjectionKind, ProjectionVisibility, V4MaterializedSource, V4SourceManifest,
     openapi_document_metadata, projection_arg_specs, projection_column_specs,
-    projection_filter_specs, request_spec_for_projection,
+    projection_filter_specs, request_spec_for_projection, validate_openapi_base_url_template,
 };
 use coral_spec::{ParsedTemplate, SourceManifestCommon, SourceTableFunctionSpec, TableCommon};
 
@@ -139,7 +139,7 @@ fn http_manifest_for_surface(
             description: manifest.common.description.clone(),
             test_queries: Vec::new(),
         },
-        base_url: surface_base_url(surface, materialized_surface)?,
+        base_url: surface_base_url(manifest, surface, materialized_surface)?,
         auth: surface.openapi_runtime.auth.clone(),
         request_headers: surface.openapi_runtime.request_headers.clone(),
         rate_limit: surface.openapi_runtime.rate_limit.clone(),
@@ -150,11 +150,14 @@ fn http_manifest_for_surface(
 }
 
 fn surface_base_url(
+    manifest: &V4SourceManifest,
     surface: &coral_spec::v4::V4Surface,
     materialized_surface: &coral_spec::v4::MaterializedSurface,
 ) -> Result<ParsedTemplate, AppError> {
     if !surface.openapi_runtime.base_url.raw().trim().is_empty() {
-        return Ok(surface.openapi_runtime.base_url.clone());
+        let base_url = surface.openapi_runtime.base_url.clone();
+        validate_surface_base_url_template(manifest, surface, &base_url, "authored")?;
+        return Ok(base_url);
     }
     let bytes = std::fs::read(&materialized_surface.raw_source_document_path).map_err(|error| {
         AppError::FailedPrecondition(format!(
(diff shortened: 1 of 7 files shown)


AlbertQM (approved):

crates/coral-spec/src/v4/manifest.rs
```diff
@@ -161,6 +162,9 @@ impl V4SourceManifest {
             }
             let inputs = collect_declared_inputs(surface_value)?;
             validate_input_references(surface_value, &inputs)?;
+            if let Some(base_url) = raw_surface.base_url.as_ref() {
```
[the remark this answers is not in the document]
simonwhitaker: I think this is invalid. Existing v4 input reference validation is already surface-local: `parse_manifest_value` calls `collect_declared_inputs(surface_value)` and then `validate_input_references(surface_value, &inputs)` for each surface. The later `declared_inputs` union is for app-side binding/prompting after compatible per-surface input definitions are merged; it is not a cross-surface reference scope. This PR's namespace-only `base_url` guard preserves that existing surface-local input validation rather than narrowing it.

iaptsiauri (approved):

crates/coral-app/src/sources/runtime_package.rs
```diff
@@ -174,10 +178,121 @@ fn surface_base_url(
...
+}
+
+fn validate_surface_base_url_template(
+    surface: &coral_spec::v4::V4Surface,
+    base_url: &ParsedTemplate,
+    provenance: &str,
+) -> Result<(), AppError> {
+    for token in base_url.tokens() {
+        match token.namespace() {
+            TemplateNamespace::Input => {
```
jsummerfield: Derived OpenAPI base URLs should reject input-token defaults too. This path accepts `{{input.API_BASE|https://fallback.example}}` as long as `API_BASE` is declared, and the engine treats that default as satisfying a missing input. Authored `base_url` templates already reject input defaults; please apply the same rule here so derived server URLs cannot bypass required input enforcement.

crates/coral-app/src/sources/runtime_package.rs
```diff
@@ -174,10 +178,121 @@ fn surface_base_url(
...
     })?;
-    ParsedTemplate::parse(server_url).map_err(|error| {
+    let base_url = ParsedTemplate::parse(server_url).map_err(|error| {
         AppError::FailedPrecondition(format!(
             "failed to parse derived base_url for DSL v4 surface '{}': {error}",
             surface.id
         ))
-    })
+    })?;
+    validate_surface_base_url_template(surface, &base_url, "derived OpenAPI server")?;
```
jsummerfield: This catches derived OpenAPI server URLs only during runtime package assembly, not during source add/materialization. If this rule defines v4 source validity, materialization should validate the derived server URL before artifacts are persisted so `coral source add` fails instead of installing a source that later breaks query-source loading.

jsummerfield (commented):
I left two focused inline comments from the re-review.

jsummerfield (approved):

crates/coral-app/src/sources/runtime_package.rs
simonwhitaker: Fixed in 90dcf3de: the shared v4 base_url validator now rejects `{{input.*|...}}` token defaults, and runtime assembly calls that same validator for derived OpenAPI server URLs.

crates/coral-app/src/sources/runtime_package.rs
simonwhitaker: Fixed in 90dcf3de: materialization now derives and validates the OpenAPI server URL immediately after reading the descriptor, before artifacts are written or installed. Added a SourceManager import regression that proves `coral source add` rejects the invalid derived base_url and leaves no materialization artifacts behind.

PR #1185, opened 2026-06-04
fix(app): wrap v4 artifact read errors

### LINKED ISSUE #1124: Fail loudly when DSL v4 materialized artifacts cannot be read
## Problem

Some materialized artifact read failures escape as plain `AppError::Io` instead of the DSL v4 incompatible-materialization error. In particular, reading the raw materialized source document uses `std::fs::read(...)?` during loaded-artifact validation.

Code references:
- `crates/coral-app/src/sources/materialization.rs:373`
- `crates/coral-app/src/query/manager.rs:185`

## Impact

`load_query_sources` treats most app errors as skippable source load failures. An unreadable, missing, or raced-away DSL v4 artifact can silently remove the source from query/catalog results instead of failing loudly with the repo-required re-add guidance.

## Acceptance criteria

- All missing, unreadable, corrupted, or incompatible DSL v4 materialized artifact reads return the consistent incompatible-materialization error.
- Query source loading fails loudly for these v4 artifact failures instead of silently skipping the source.
- Regression coverage includes an unreadable or missing raw materialized source document path.

simonwhitaker:
## Summary
- wrap raw DSL v4 materialized source document read failures in the existing incompatible-materialization error
- add a regression test for unreadable raw source-document artifacts preserving re-add guidance

## Tests
- `cargo test -p coral-app load_v4_materialization_rejects_unreadable_raw_source_document_with_readd_guidance`
- `make rust-checks`

## Risks
- Low. This only changes the error type/detail for a failed raw artifact read path so query loading fails closed consistently.

Fixes #1124

### DIFF
--- a/crates/coral-app/src/sources/materialization.rs
+++ b/crates/coral-app/src/sources/materialization.rs
@@ -340,6 +340,23 @@ fn require_file(source_name: &SourceName, path: &Path) -> Result<(), AppError> {
     }
 }
 
+fn read_raw_source_document_artifact(
+    source_name: &SourceName,
+    surface: &coral_spec::v4::V4Surface,
+    path: &Path,
+) -> Result<Vec<u8>, AppError> {
+    std::fs::read(path).map_err(|error| {
+        incompatible_materialization_error(
+            source_name,
+            format!(
+                "failed to read raw source document artifact for surface '{}' '{}': {error}",
+                surface.id,
+                path.display()
+            ),
+        )
+    })
+}
+
 fn validate_loaded_materialization(
     source_name: &SourceName,
     manifest: &V4SourceManifest,
@@ -370,7 +387,11 @@ fn validate_loaded_materialization(
                 format!("fingerprint is missing surface '{}'", surface.id),
             ));
         };
-        let raw_bytes = std::fs::read(&materialized_surface.raw_source_document_path)?;
+        let raw_bytes = read_raw_source_document_artifact(
+            source_name,
+            surface,
+            &materialized_surface.raw_source_document_path,
+        )?;
         let observed_raw_hash = sha256_hex(&raw_bytes);
         if observed_raw_hash != fingerprint_surface.descriptor_sha256 {
             return Err(incompatible_materialization_error(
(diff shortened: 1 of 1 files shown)


crates/coral-app/src/sources/materialization.rs
```diff
@@ -1017,6 +1038,49 @@ surfaces:
...
+            &workspace_name(),
+            &source_name(),
+            &manifest_yaml,
+            &manifest,
+        );
+
+        std::fs::set_permissions(&raw_path, original_permissions)
+            .expect("restore raw descriptor permissions");
+        let error = result.expect_err("unreadable raw descriptor should fail");
+        let message = error.to_string();
```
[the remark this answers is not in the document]
simonwhitaker: Applied the suggested guard in `8650feb1`: if the platform can still read the chmodded file, the test now returns instead of failing spuriously. Local validation after the change: `cargo test -p coral-app load_v4_materialization_rejects_unreadable_raw_source_document_with_readd_guidance`, `cargo fmt --all -- --check`, and `make rust-checks` all passed.

crates/coral-app/src/sources/materialization.rs
```diff
@@ -1017,6 +1038,51 @@ surfaces:
...
+            &source_name(),
+            &manifest_yaml,
+            &manifest,
+        );
+
+        std::fs::set_permissions(&raw_path, original_permissions)
+            .expect("restore raw descriptor permissions");
+        let message = match result {
+            Err(error) => error.to_string(),
+            Ok(_) => return,
```
jsummerfield: This regression test silently passes if `load_v4_materialization` accepts the unreadable raw artifact. If chmod-based unreadability is unreliable in some environments, probe `std::fs::read(&raw_path)` before invoking the loader and skip only when the file is still readable; once the loader is called, `Ok(_)` should fail.

jsummerfield (approved):

crates/coral-app/src/sources/materialization.rs
simonwhitaker: Fixed in 150aa1d5: the test now probes `std::fs::read(&raw_path)` immediately after chmod and skips only if the file is still readable. If the probe confirms unreadability, the loader is called and `Ok(_)` fails via `expect_err`. Local validation: `cargo test -p coral-app load_v4_materialization_rejects_unreadable_raw_source_document_with_readd_guidance` and `make rust-checks`.

PR #1193, opened 2026-06-04
fix(spec): preserve path parameter defaults

### LINKED ISSUE #1153: Preserve path parameter defaults when generating v4 request paths
This tracks unresolved feedback identified by cubic on PR #1149:
https://github.com/withcoral/coral/pull/1149#discussion_r3347190629

Context:
This appears to be pre-existing DSL v4 behavior surfaced by the refactor review, not a regression from the module split.

Problem:
Path parameters with defaults are lowered into request path templates without fallback handling, so generated optional inputs can still fail URL rendering at runtime.

Acceptance criteria:
When a path parameter has a default, generated request paths should preserve an equivalent fallback for filters and function args, with focused projection/runtime tests.

simonwhitaker:
Summary:
- Preserve authored path parameter defaults when rendering generated v4 request path templates.
- Cover both table filter tokens and table-function arg tokens.

Tests:
- cargo test -p coral-spec request_paths_preserve_path_parameter_defaults
- make rust-checks
- git diff --check

Risks:
- Low. Path parameters without defaults render unchanged; defaulted path parameters now use the existing template fallback form.

Closes #1153

### DIFF
--- a/crates/coral-spec/src/v4/projections/runtime.rs
+++ b/crates/coral-spec/src/v4/projections/runtime.rs
@@ -130,8 +130,12 @@ pub fn request_spec_for_projection(
     for input in &projection.inputs {
         if input.source_location == IrInputLocation::Path {
             let replacement = match input.sql_exposure {
-                SqlInputExposure::Filter => format!("{{{{filter.{}}}}}", input.name),
-                SqlInputExposure::FunctionArg => format!("{{{{arg.{}}}}}", input.name),
+                SqlInputExposure::Filter => {
+                    path_template_token("filter", &input.name, input.default_value.as_deref())
+                }
+                SqlInputExposure::FunctionArg => {
+                    path_template_token("arg", &input.name, input.default_value.as_deref())
+                }
                 SqlInputExposure::Internal => continue,
             };
             path = path.replace(&format!("{{{}}}", input.wire_name), &replacement);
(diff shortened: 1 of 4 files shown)


iaptsiauri (approved):
if default value has escape chars `default: "foo}}bar"` this will change the default value. But I am not sure this is relevant to escape as well.

crates/coral-spec/src/v4/projections/runtime.rs
```diff
@@ -152,3 +156,10 @@ pub fn request_spec_for_projection(
         headers: Vec::new(),
     })
 }
+
+fn path_template_token(namespace: &str, key: &str, default: Option<&str>) -> String {
+    default.map_or_else(
+        || format!("{{{{{namespace}.{key}}}}}"),
+        |default| format!("{{{{{namespace}.{key}|{default}}}}}"),
```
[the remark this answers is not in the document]
simonwhitaker: Addressed in 78db8e5f. Generated path defaults are now URL-encoded before being embedded in `{{filter.*|...}}` / `{{arg.*|...}}` tokens, so template delimiters like a leading `|` or `}}` cannot corrupt the parsed path template. The regression now uses `|public}}` for both filter and arg paths; `make rust-checks` passes locally.

crates/coral-spec/src/v4/projections/runtime.rs
```diff
@@ -108,8 +108,12 @@ pub fn request_spec_for_projection(
     for input in &projection.inputs {
         if input.source_location == IrInputLocation::Path {
             let replacement = match input.sql_exposure {
-                SqlInputExposure::Filter => format!("{{{{filter.{}}}}}", input.name),
-                SqlInputExposure::FunctionArg => format!("{{{{arg.{}}}}}", input.name),
+                SqlInputExposure::Filter => {
+                    path_template_token("filter", &input.name, input.default_value.as_deref())
```
jsummerfield, line 134: I think this still misses the runtime behavior the issue asks for. The generated template now has a fallback, but the same path input remains required via `projection_input_required`: path inputs are required even when `default_value` exists. `projection_filter_specs` / `projection_arg_specs` then propagate that flag, and the HTTP table/function paths reject missing required filters/args before `render_template` can ever use this fallback. Could we make defaulted path inputs non-required in the generated SQL surface, or otherwise exempt them from the pre-render required check, and add an engine-level test that omitting `tenant` actually sends the defaulted path?

crates/coral-spec/src/v4/projections/runtime.rs
```diff
@@ -152,3 +156,13 @@ pub fn request_spec_for_projection(
         headers: Vec::new(),
     })
 }
+
+fn path_template_token(namespace: &str, key: &str, default: Option<&str>) -> String {
+    default.map_or_else(
+        || format!("{{{{{namespace}.{key}}}}}"),
+        |default| {
+            let encoded_default = urlencoding::encode(default);
```
jsummerfield: This encoding is not path-segment-safe. `urlencoding::encode` leaves `.` unescaped, so a default of `..` renders as a literal dot segment; when reqwest/url parses `https://api.example.com/tenants/../items`, it normalizes that to `/items`. That changes the endpoint instead of preserving the path parameter value. Could this use an encoder or path-building path that preserves dot-segment defaults as literal parameter values, and cover `.` / `..` in the regression test?

crates/coral-spec/src/v4/projections/runtime.rs
simonwhitaker: Addressed in `8de6fe7e`. Defaulted path inputs now derive `required: false` for generated SQL surfaces, so `projection_filter_specs` / `projection_arg_specs` no longer reject omitted filters/args before `render_template` can apply the fallback. I also added engine regressions for omitted table filter and table-function arg path defaults. `make rust-checks` passes locally.

crates/coral-spec/src/v4/projections/runtime.rs
simonwhitaker: Addressed in `8de6fe7e`. I replaced `urlencoding::encode` with a path-segment encoder and special-cased exact `.` / `..` defaults to emit `%252E` / `%252E%252E`; `reqwest` normalizes single-encoded dot segments before sending, so double-encoding is needed to preserve them as literal parameter values. The projection regression now covers `.` and `..`, and the engine tests assert the raw request target. `make rust-checks` passes locally.

sauldhernandez (approved):

PR #1196, opened 2026-06-04
fix(app): make source re-add atomic

### LINKED ISSUE #1123: Make DSL v4 manifest and artifact swaps atomic during source re-add
## Problem

When re-adding an installed DSL v4 source, source persistence writes the new manifest before swapping the materialized artifact directory. Query loading reads the manifest and artifacts independently.

Code references:
- `crates/coral-app/src/sources/manager.rs:553`
- `crates/coral-app/src/sources/manager.rs:618`
- `crates/coral-app/src/query/manager.rs:202`

## Impact

A concurrent query/list/test during source re-add can observe the new manifest with the old artifacts. That can produce a false missing/incompatible materialization error and misleading re-add guidance even though a re-add is already in progress.

## Acceptance criteria

- Source re-add exposes either the old manifest+artifacts or the new manifest+artifacts to concurrent readers, never a mixed pair.
- The fix fits the existing app-owned state locking model or introduces a clearly scoped source lifecycle lock.
- Regression coverage exercises a re-add/read interleaving or otherwise proves manifest/artifact replacement is atomic from query loading's perspective.

simonwhitaker:
Fixes #1123.

Summary:
- Hold the app state lock across query-source config, manifest, v4 artifact, and credential-material reads.
- Hold the exclusive app state lock across source persistence: rollback snapshot, manifest write, credential update, v4 materialization swap, and config upsert.
- Add state-lock-held credential helpers for file-backed material writes to avoid recursive lock acquisition.
- Add a regression covering a source re-add racing a shared app-state reader.

Tests:
- cargo test -p coral-app readd_source_waits_for_state_lock_before_replacing_manifest
- cargo test -p coral-app import_v4_source_writes_materialized_artifacts
- cargo test -p coral-app load_query_source_passes_present_optional_secrets_to_runtime
- cargo test -p coral-app load_query_sources_fails_closed
- make rust-checks

Risks:
- Source re-add now serializes more filesystem reads and writes under the app state lock; this is intentional but may increase wait time during large v4 materialization swaps.

### DIFF
--- a/crates/coral-app/src/credentials/mod.rs
+++ b/crates/coral-app/src/credentials/mod.rs
@@ -151,33 +151,39 @@ impl CredentialManager {
             .read_material(workspace_name, credential_set_id, storage)
     }
 
-    /// Read persisted credential material for the declared inputs, refreshing
-    /// provider-managed credentials before returning when needed.
-    pub(crate) async fn read_material_for_inputs(
+    /// Refresh provider-managed credentials already captured for a runtime.
+    ///
+    /// This updates the supplied material in memory only. Runtime input
+    /// resolution uses source snapshots, so it must not re-read or write live
+    /// source credential material after query-source loading has finished.
+    pub(crate) async fn refresh_material_for_inputs(
+        &self,
+        inputs: &[ManifestInputSpec],
+        material: &mut BTreeMap<String, String>,
+    ) -> Result<(), AppError> {
+        for input in inputs {
+            self.refresh_oauth_input_material(input, material).await?;
+        }
+        Ok(())
+    }
+
+    pub(crate) async fn refresh_and_persist_material_for_inputs_with_refresh_lock_held(
         &self,
         workspace_name: &WorkspaceName,
         credential_set_id: &CredentialSetId,
         storage: CredentialStorageKind,
         inputs: &[ManifestInputSpec],
-    ) -> Result<BTreeMap<String, String>, AppError> {
-        if !has_oauth_credential_inputs(inputs) {
-            return self.read_material(workspace_name, credential_set_id, storage);
-        }
-
-        let _refresh_file_lock = self
-            .credential_refresh_lock(workspace_name, credential_set_id)
-            .await?;
-        let mut material = self.read_material(workspace_name, credential_set_id, storage)?;
+        material: &mut BTreeMap<String, String>,
+    ) -> Result<(), AppError> {
         self.refresh_and_persist_oauth_material(
             workspace_name,
             credential_set_id,
             storage,
             inputs,
-            &mut material,
+            material,
         )
         .await
-        .map_err(credential_refresh_error)?;
-        Ok(material)
+        .map_err(credential_refresh_error)
     }
 
     pub(crate) fn default_write_storage(&self) -> Result<CredentialStorageKind, AppError> {
@@ -199,6 +205,32 @@ impl CredentialManager {
         })
     }
 
+    async fn refresh_oauth_input_material(
+        &self,
+        input: &ManifestInputSpec,
+        material: &mut BTreeMap<String, String>,
+    ) -> Result<bool, AppError> {
+        if input.kind != ManifestInputKind::Secret {
+            return Ok(false);
+        }
+        let Some(credential) = input.credential.as_ref() else {
+            return Ok(false);
+        };
+        let Some(oauth) = credential
+            .methods
+            .iter()
+            .find_map(|method| method.oauth.as_ref())
+        else {
+            return Ok(false);
+        };
+        self.oauth_credential_service
+            .refresh_if_needed(
+                RefreshOAuthCredentialRequest::for_source_input(&input.key, oauth),
+                material,
+            )
+            .await
+    }
+
     async fn refresh_and_persist_oauth_material(
         &self,
         workspace_name: &WorkspaceName,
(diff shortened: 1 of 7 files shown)


crates/coral-app/src/sources/manager.rs
```diff
@@ -1565,6 +1629,79 @@ tables:
...
+        let stored_before = std::fs::read_to_string(&manifest_path).expect("stored manifest");
+        let state_lock = config_store.state_lock_shared().expect("shared state lock");
+
+        let updated_manifest =
+            original_manifest.replace("https://example.com", "https://replacement.example.com");
+        let import_manager = manager.clone();
+        let (started_tx, started_rx) = std_mpsc::channel();
+        let (done_tx, done_rx) = std_mpsc::channel();
+        let handle = thread::spawn(move || {
+            started_tx.send(()).expect("send started");
```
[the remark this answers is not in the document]
simonwhitaker: Addressed in d4dc47a9. The regression now builds the exact PersistSourceRequest before signaling the test thread as started, then calls persist_source directly while the shared state lock is held. This removes the public import preflight gap the comment identified. Re-ran cargo test -p coral-app readd_source_waits_for_state_lock_before_replacing_manifest and make rust-checks.

crates/coral-app/src/sources/manager.rs
```diff
@@ -548,13 +548,14 @@ impl SourceManager {
         let credential_guard = self
```
sauldhernandez, line 578: <p class='graphite__hidden'><i>[Re: line +340]</i></p>
sources/manager.rs:341 and sources/manager.rs:348 validate required secrets and choose credential storage before `persist_source` takes the exclusive lock at sources/manager.rs:551. Concrete race: re-add A validates using old stored `API_TOKEN`; re-add B wins the lock first and removes that material; then A persists a manifest still requiring `API_TOKEN`, but sources/manager.rs:611 records `credential_storage: None`. Import succeeds, later queries fail. Move validation/storage selection into the locked transaction or revalidate under the lock.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/withcoral/coral/1196?utm_source=unchanged-line-comment">Graphite</a>.</p>

sauldhernandez, line 578: <p class='graphite__hidden'><i>[Re: line +467]</i></p>
This can delete the credential before the config folder is removed, and is outside a lock. This can be racy vs a query that takes the shared lock, sees the config entry, but then tries to load secrets that are no longer there.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/withcoral/coral/1196?utm_source=unchanged-line-comment">Graphite</a>.</p>

crates/coral-app/src/credentials/mod.rs
```diff
@@ -309,51 +309,49 @@ impl CredentialMaterialGuard<'_> {
         })
     }
 
-    pub(crate) fn update_material_or_empty_on_parse<F>(
+    pub(crate) fn update_material_or_empty_on_parse_with_state_lock_held<F>(
```
sauldhernandez: nit: naming could be simpler.

sauldhernandez (changes requested):
`extensions.rs` still re-reads the live source config/material, even after query manager has dropped the shared lock and built QuerySource.  
  
Also left a few inline comments

crates/coral-app/src/sources/manager.rs
simonwhitaker: Addressed in `43cab194`. `persist_source` now takes the exclusive app state lock before choosing credential storage, then revalidates required secret material through `source_persist_storage_with_state_lock_held` under that same lock. The earlier import/OAuth preflight can still prepare materialization inputs, but it is no longer the final authority for whether required stored material exists or which storage gets persisted.

Verified with `make rust-checks` after the rebase.

simonwhitaker: Addressed in `43cab194`. `delete_source` now acquires the exclusive app state lock before reading the source config and keeps it held across rollback snapshotting, source-dir rename, config removal, and credential removal. The credential delete uses the state-lock-held credential path, so a query holding or waiting on the shared lock cannot observe config that still points at credentials already removed by delete.

Verified with `make rust-checks` after the rebase.

crates/coral-app/src/credentials/mod.rs
simonwhitaker: Addressed in `43cab194`. I simplified the public guard-facing name to `update_material_with_state_lock`, and kept the more explicit `*_with_state_lock_held` suffix on lower-level store/backend calls where the caller’s lock ownership is the important contract.

Verified with `make rust-checks` after the rebase.

crates/coral-app/src/query/manager.rs
```diff
@@ -291,27 +315,54 @@ impl QueryManager {
...
             workspace_name.clone(),
             self.config_store.clone(),
             self.credential_manager.clone(),
+            selected_sources
+                .iter()
+                .map(|source| {
+                    (
+                        source.query_source.source_name().to_string(),
+                        SourceCredentialSnapshot {
+                            source: source.source.clone(),
```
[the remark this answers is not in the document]
simonwhitaker: Addressed in `43cab194`. Query loading now captures credential material alongside each loaded `QuerySource`, and runtime input resolution uses that snapshot. OAuth refresh only persists back to live material when the current installed-source config still matches the loaded snapshot; if a manifest-only re-add changed the source under us, refresh falls back to the in-memory snapshot instead of mutating live credentials.

Verified with `make rust-checks` after the rebase.

sauldhernandez (approved):

PR #1197, opened 2026-06-04
perf(mcp): compact success tool text

Bradley-Butcher:
## Intent

Make successful MCP tool-result text cheaper without changing the structured MCP contract. This changes only the fallback `content[0].text` rendering from pretty JSON to compact JSON. `structuredContent` remains the same value.

## Why this slice

Headroom's useful lesson here is the boring one: first remove non-information-bearing formatting, then measure on real tool calls. This is lossless text formatting compaction; it does not drop keys, rows, or values.

## Real live token measurement

Command used from the benchmark branch after this stack was restacked:

```shell
cargo run --locked -p xtask -- mcp-token-bench --coral-bin /tmp/coral-token-bins/coral-main --json
cargo run --locked -p xtask -- mcp-token-bench --coral-bin /tmp/coral-token-bins/coral-pr1 --json
```

- Config: default live Coral config
- Transport: real `coral mcp-stdio`
- Tokenizer: `gpt-5` / `o200k_base`
- Cases: 16 live MCP `tools/call` requests covering discovery, SQL, empty result, and error result

Totals:

| Metric | Before | After | Saved | Savings |
|---|---:|---:|---:|---:|
| Text tokens | 57524 | 38898 | 18626 | 32.4% |
| Protocol result tokens | 107432 | 80312 | 27120 | 25.2% |

Per-case text tokens:

| case | tool | baseline text | pr1 text | savings | protocol savings |
|---|---|---:|---:|---:|---:|
| `discovery.search.github_pull` | `search_catalog` | 3612 | 2372 | 1240 | 1768 |
| `discovery.search.slack_messages` | `search_catalog` | 8343 | 6172 | 2171 | 3228 |
| `discovery.search.table_functions` | `search_catalog` | 15356 | 10222 | 5134 | 7682 |
| `discovery.list.github_tables` | `list_catalog` | 2335 | 1614 | 721 | 1014 |
| `discovery.list.table_functions` | `list_catalog` | 12436 | 8195 | 4241 | 6271 |
| `discovery.columns.github_pulls` | `list_columns` | 3137 | 2030 | 1107 | 1534 |
| `discovery.columns.github_pulls_required` | `list_columns` | 188 | 123 | 65 | 90 |
| `discovery.columns.github_pulls_search` | `list_columns` | 1620 | 1030 | 590 | 843 |
| `discovery.describe.github_pulls` | `describe_table` | 207 | 176 | 31 | 44 |
[...]

PR #1202, opened 2026-06-04
perf(mcp): compact sql text with toon

Bradley-Butcher:
## Summary

- render successful non-empty MCP `sql` text fallback as TOON when it is shorter than compact JSON
- keep `structuredContent` as the unchanged JSON contract; TOON is only fallback text
- strict-roundtrip every candidate TOON string back to the original JSON value before using it, and fall back to compact JSON for empty rows, invalid TOON, or longer output

## Why

PR #1198's hand-rolled SQL row text did not move the aggregate benchmark enough and was closed. The follow-up TOON benchmark showed successful SQL responses were still a small but real hotspot after PR #1201. This slice applies TOON only where the live benchmark proves savings, without changing the machine-readable MCP contract.

## Actual live benchmark

Command:

`CARGO_TARGET_DIR=/tmp/coral-toon-target cargo run --locked -p xtask -- mcp-token-bench --coral-bin /tmp/coral-toon-target/debug/coral --json > /tmp/coral-token-pr5-toon-sql.json`

Config: default live Coral config.

Comparison: PR #1201 -> this PR.

| Case | Model before | Model after | Model saved | Protocol before | Protocol after | Protocol saved |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| Full 21-case benchmark | 117,362 | 114,978 | 2,384 / 2.0% | 163,386 | 160,862 | 2,524 / 1.5% |
| `sql.wide_coral_tables` | 1,413 | 1,135 | 278 | 2,901 | 2,657 | 244 |
| `sql.narrow_many_tables` | 595 | 347 | 248 | 1,310 | 963 | 347 |
| `sql.long_catalog_text` | 1,008 | 826 | 182 | 2,096 | 1,894 | 202 |
| `sql.github_pulls_columns` | 3,432 | 1,917 | 1,515 | 7,005 | 5,450 | 1,555 |
| `sql.empty_result` | 4 | 4 | 0 | 28 | 28 | 0 |
| `sql.headroom_contents` | 464 | 303 | 161 | 988 | 812 | 176 |
| `sql.invalid_statement` | 48 | 48 | 0 | 137 | 137 | 0 |

Non-SQL benchmark rows were unchanged.

## Actual output shape

`structuredContent` remains the same JSON object. Only `content[0].text` changes.

`sql.narrow_many_tables` before:

```json
[...]

PR #1200, opened 2026-06-04
perf(mcp): compact catalog tool text

Bradley-Butcher:
## Summary

- compact the MCP text fallback for catalog discovery tools:
  `search_catalog`, `list_catalog`, and `list_columns`
- keep the full machine-readable `structuredContent` unchanged
- leave `sql`, `describe_table`, and error text unchanged
- use the compact text renderer only when it is byte-shorter than compact JSON;
  token savings are measured separately below

Comparison base: this PR is measured against PR #1197, not `main`. The GitHub
diff includes #1197 because this is part of a Graphite stack.

This PR changes what the model sees in `content[0].text` for catalog discovery
results. It does not change the structured MCP result contract.

## What Changed

Before this PR, catalog discovery text was compact JSON. That preserved the data
but repeated field names for every row:

```json
{"items":[{"kind":"table_function","schema_name":"datadog","name":"datadog.apm_dependencies","sql_reference":"datadog.apm_dependencies","sql_call_example":"datadog.apm_dependencies(service => '<value>')","description":"Datadog APM service dependencies for a target service.","table_function":{"function_name":"apm_dependencies","arguments":[{"name":"service","required":true},...]}}]}
```

After this PR, the same text fallback is a compact tab-separated discovery page:

```text
items=20 total=983 limit=20 offset=0 has_more=true next_offset=20
kind	name	sql_reference	description	details	matched_fields
table_function	datadog.apm_dependencies	datadog.apm_dependencies	Datadog APM service dependencies for a target service.	function_name=apm_dependencies;call=datadog.apm_dependencies(service => '<value>');args=service!,env,start,end,primary_tag;result_columns=name:Utf8? Service name;called_by:Utf8? Comma-separated upstream services;calls:Utf8? Comma-separated downstream services	arguments
...
```

The first line is page metadata. The second line is the column header. Each
following line is one catalog item. `structuredContent` still contains the full
original JSON object.

## Why

The previous SQL-row compaction attempt did not save enough tokens. Catalog
discovery results were the measured hotspot because they repeat the same object
keys across many catalog rows.

This PR applies the Headroom lesson that helped here: route by output shape.
Catalog discovery output is tabular, so its text fallback can be tabular while
the structured result stays JSON.

## Live Token Measurement

Command used from the benchmark branch after this stack was restacked:

```shell
cargo run --locked -p xtask -- mcp-token-bench --coral-bin /tmp/coral-token-bins/coral-pr1 --json
[...]

PR #1213, opened 2026-06-05
perf(mcp): compact SQL tool text rendering

iaptsiauri:
## Summary
- stop emitting pretty-printed MCP success text and keep non-SQL tool text as compact JSON
- render MCP `sql` tool text as compact columnar JSON with `columns`, positional `rows`, and `row_count`
- preserve `structured_content` as the existing row-object shape for MCP client compatibility
- document the MCP SQL text format and cover the columnar contract in MCP tests

## Before / After Result Shape
For a SQL result with one `text` column and two rows, MCP text used to repeat row-object keys and pretty-print whitespace:

```json
{
  "rows": [
    {
      "text": "hello"
    },
    {
      "text": "world"
    }
  ]
}
```

After this PR, MCP `sql` text is compact columnar JSON:

```json
{"columns":["text"],"rows":[["hello"],["world"]],"row_count":2}
```

`structured_content` remains the existing object-row shape for MCP clients; only the text payload optimized for model context changes.

## Performance
Local MCP SQL text-size benchmark against the same SQL scenarios used during development:

| Version | Total chars | Estimated tokens | Reduction vs main | Reduction vs PR #1197 |
| --- | ---: | ---: | ---: | ---: |
| main, pretty object rows | 2,274 | 569 | baseline | n/a |
| PR #1197, compact object rows | 1,562 | 391 | 31.3% | baseline |
| this branch, compact columnar SQL text | 1,050 | 263 | 53.8% | 32.7% |

Wide `10x7` SQL result: PR #1197 compact object rows emitted 1,518 chars; columnar SQL text emitted 987 chars, a 35.0% reduction.

Tiny `2x1` SQL result: compact object rows remain smaller, 44 chars vs 63 chars, because column metadata overhead dominates. The intended win is wider or repeated-row SQL output where repeated object keys dominate context size.

## Validation
- `cargo test -p coral-mcp mcp_tool_error_does_not_end_session`
- `cargo test -p coral-mcp mcp_sql_returns_large_int64_as_string`

### DIFF
--- a/crates/coral-mcp/src/server.rs
+++ b/crates/coral-mcp/src/server.rs
@@ -7,7 +7,7 @@ use coral_api::v1::{
     TableSummary as ProtoTableSummary, catalog_item,
 };
 use coral_client::{
-    AppClient, CatalogClient, FeedbackClient, QueryClient, SourceClient,
+    AppClient, CatalogClient, CollectedQueryResult, FeedbackClient, QueryClient, SourceClient,
     batches_to_json_rows_json_safe_numbers, decode_execute_sql_response, default_workspace,
 };
 use rmcp::{
@@ -26,13 +26,13 @@ use tonic::Request;
 use crate::{
     McpOptions,
     surface::{
-        CatalogToolKind, build_tool_result, describe_table_arguments, describe_table_tool,
-        describe_table_value, feedback_tool, guide_resource, guide_resource_content,
-        initial_instructions, list_catalog_arguments, list_catalog_tool, list_catalog_value,
-        list_columns_arguments, list_columns_tool, list_columns_value, required_string_argument,
-        search_catalog_arguments, search_catalog_tool, search_catalog_value, sql_tool,
-        status_to_error_data, tables_resource, tables_resource_content, tool_error_from_status,
-        tool_error_result,
+        CatalogToolKind, build_tool_result, build_tool_result_with_text, describe_table_arguments,
+        describe_table_tool, describe_table_value, feedback_tool, guide_resource,
+        guide_resource_content, initial_instructions, list_catalog_arguments, list_catalog_tool,
+        list_catalog_value, list_columns_arguments, list_columns_tool, list_columns_value,
+        required_string_argument, search_catalog_arguments, search_catalog_tool,
+        search_catalog_value, sql_tool, status_to_error_data, tables_resource,
+        tables_resource_content, tool_error_from_status, tool_error_result,
     },
     telemetry,
 };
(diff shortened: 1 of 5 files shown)


iaptsiauri:
Another take on what additionally we can do with data representation to reduce token count with MCP. It does change the output format though

AlbertQM:
> Another take on what additionally we can do with data representation to reduce token count with MCP. It does change the output format though

Out of curiosity, why not use TOON?

iaptsiauri:
> Out of curiosity, why not use TOON?

For now this is an experimental PR, should try how it will look with TOON. Also need to figure out what are the drawbacks with this approach

Bradley-Butcher:
https://github.com/withcoral/coral/pull/1202

Bradley-Butcher:
Benchmarked this against the TOON SQL text PR (#1202) using the live MCP token benchmark from the stack.

Important caveat: full-benchmark totals are not apples-to-apples because this PR is based on `main` + #1197, while #1202 sits above #1200/#1199/#1201. The fair comparison for this PR is the SQL-only set.

| Scope | This PR columnar JSON | #1202 TOON | TOON delta |
| --- | ---: | ---: | ---: |
| Full 21-case model tokens | 152,478 | 114,978 | -37,500 |
| Full protocol tokens | 208,511 | 160,863 | -47,648 |
| SQL-only model tokens | 4,844 | 4,580 | -264 / -5.5% |
| SQL-only protocol tokens | 12,201 | 11,942 | -259 / -2.1% |

SQL case detail:

| Case | This PR model | #1202 TOON model | Winner |
| --- | ---: | ---: | --- |
| `sql.wide_coral_tables` | 1,128 | 1,135 | this PR by 7 |
| `sql.narrow_many_tables` | 358 | 347 | TOON |
| `sql.long_catalog_text` | 845 | 826 | TOON |
| `sql.github_pulls_columns` | 2,124 | 1,917 | TOON |
| `sql.empty_result` | 18 | 4 | TOON |
| `sql.headroom_contents` | 323 | 303 | TOON |
| `sql.invalid_statement` | 48 | 48 | tie |

Interpretation: columnar JSON is a good idea and has the advantage of staying JSON-shaped with no new dependency. But on the live SQL suite, TOON is slightly better overall and wins most cases. This PR also always emits columnar JSON, including tiny/empty results where the overhead hurts; #1202 only switches away from compact JSON when the candidate is shorter and strict-roundtrips back to the original structured value.

The one possible synthesis would be to choose the shortest of compact JSON / TOON / columnar JSON, but based on this run columnar only wins one SQL case by 7 model tokens, so I would not add that complexity unless we find a wider benchmark where it wins materially.

iaptsiauri:
Closing as PR on the benchmark didn't produce any significant improvement on token count

PR #1183, opened 2026-06-04
feat(mcp): add Slack and GitHub MCP sources

antonmry:
## Summary

Adds two community MCP sources backed by official remote MCP servers:

- `slack_mcp`: Slack's official remote MCP server over Streamable HTTP, with interactive OAuth setup through a Slack app, 1 table, and 11 read-oriented functions for user profile, public/private search, channel search, user search, emoji search, channel history, threads, channel members, files, and canvases.
- `github_mcp`: GitHub's official remote MCP server over Streamable HTTP, using the hosted all-toolsets read-only endpoint, with device-code OAuth, OAuth app, and pasted-token credential flows, 1 table, and 21 read-oriented functions for user context, issues, pull requests, repository trees, files, commits, releases, code search, repository search, and user search.

This also updates MCP result normalization so text content that is not valid JSON is preserved as a string instead of failing result decode. Some MCP servers return human-readable Markdown or rendered text, so this keeps those responses queryable through the `result` column.

## User Impact

Users can add Slack and GitHub MCP servers as Coral sources and query their read-oriented tool surfaces through SQL. These sources are complementary to the existing REST-backed sources: they expose provider MCP tools directly and keep the tool output available as `result` plus `raw`.

The GitHub MCP source intentionally uses GitHub's read-only MCP endpoint, and both new sources avoid modeling write-capable tools so SQL queries do not unexpectedly mutate provider state.

## Validation

- `cargo fmt --all`
- `cargo test -p coral-engine backends::mcp::response --all-features --locked`
- `cargo clippy -p coral-engine --all-targets --all-features --locked -- -D warnings`
- `make rust-checks`
- `make docs-check`
- `cargo run -p coral-cli -- source lint sources/community/slack_mcp/manifest.yaml`
- `cargo run -p coral-cli -- source lint sources/community/github_mcp/manifest.yaml`
- Isolated live GitHub MCP validation with temporary `CORAL_CONFIG_DIR` and `gh auth token`:
  - `source test github_mcp`
  - `github_mcp.search_repositories(...)`
  - `github_mcp.pull_request_read(...)`
  - `github_mcp.get_repository_tree(...)`
- Live Slack MCP validation with OAuth token stored by `coral source add --interactive`:
  - `source test slack_mcp`
  - `slack_mcp.search_channels(query => 'general', limit => 1)`
- `git diff --check`

PR #1215, opened 2026-06-05
feat: add MCP as a surface type in DSL v4

simonwhitaker:
## Summary

This PR makes MCP a first-class DSL v4 surface.

It adds support for declaring `type: mcp` surfaces in v4 manifests, discovering MCP tools during materialization, importing those tools into semantic IR, generating projections from MCP tool schemas, and assembling the resulting materialized artifacts into queryable runtime MCP sources.

## Why

DSL v4 materializes source surfaces into stable artifacts at source add/re-add time. Until now, v4 could model OpenAPI/REST surfaces, but MCP tools could not be discovered, frozen, projected, and registered through the same pipeline.

This change allows hybrid v4 sources, such as `github_v4`, to expose both REST and MCP surfaces through Coral while preserving deterministic materialization behavior.

## Key Changes

- Adds v4 MCP surface parsing, schema generation, validation, and runtime config.
- Adds MCP semantic IR support, including tool-argument inputs and MCP execution attachments.
- Adds MCP tool catalog discovery and normalized materialized artifacts.
- Generates SQL projections from MCP tool input/output schemas.
- Adds surface namespaces so multi-surface v4 sources can register distinct SQL schemas without accidental name collisions or breaking REST schema names.
- Extends app-owned runtime package assembly to produce MCP runtime components from materialized v4 artifacts.
- Updates engine registration/composite-source handling for multi-schema runtime components.
- Adds bounded MCP `tools/list` pagination with regression coverage for cursor progression, repeated cursors, and max-page failures.
- Updates the preview `github_v4` manifest to include an MCP surface.
- Adds regression tests for artifact schema compatibility, namespace behavior, MCP projection generation, materialization, pagination, and runtime registration.

## Validation

- `make rust-checks`

## How I implemented

I started by updating `sources/core-v4/github_vy/manifest.yaml`, adding an MCP surface:

```patch
diff --git a/sources/core-v4/github_v4/manifest.yaml b/sources/core-v4/github_v4/manifest.yaml
index 093b9d12..dff2530d 100644
--- a/sources/core-v4/github_v4/manifest.yaml
+++ b/sources/core-v4/github_v4/manifest.yaml
@@ -24,7 +24,28 @@ surfaces:
       - name: User-Agent
         from: literal
         value: coral-dsl-v4
+  - id: mcp
+    type: mcp
+    inputs:
+      GITHUB_TOKEN:
+        kind: secret
+        hint: |
+          GitHub personal access token used as a bearer token for the live
+          GitHub REST API. Set GITHUB_TOKEN when installing this preview
+          source.
+    server:
+      transport: streamable_http
+      url: https://api.githubcopilot.com/mcp/x/all/readonly
+      auth:
+        type: bearer
+        from: input
+        key: GITHUB_TOKEN
+
 test_queries:
-  - SELECT id, number, title, state FROM github_v4.repos_issues WHERE owner = 'octocat' AND repo = 'Hello-World' LIMIT 5
-  - SELECT id, number, title FROM github_v4.get_issue(owner => 'octocat', repo => 'Hello-World', issue_number => 1)
-  - SELECT id, number, title FROM github_v4.search_issues(q => 'repo:octocat/Hello-World is:issue') LIMIT 5
+  # MCP surface
+  - SELECT result FROM github_v4_mcp.get_me LIMIT 1
+  - SELECT result, result_json FROM github_v4_mcp.list_pull_requests(owner => 'withcoral', repo => 'coral', state => 'open', per_page => 10) LIMIT 5
+  # REST surface
+  - SELECT id, number, title, state FROM github_v4_rest.search_issues(q => 'repo:octocat/Hello-World is:issue') LIMIT 5
+  - SELECT id, number, title, state FROM github_v4_rest.repos_issues WHERE owner = 'octocat' AND repo = 'Hello-World' LIMIT 5
+  - SELECT id, number, title FROM github_v4_rest.get_issue(owner => 'octocat', repo => 'Hello-World', issue_number => 1)
```

Then I prompted Codex:

> I just committed a change to sources/core-v4/github_v4/manifest.yaml. You can attempt to install the manifest like so:
>
> ```
> GITHUB_TOKEN=$(gh auth token) cargo run --bin coral -- source add --file sources/core-v4/github_v4/manifest.yaml
> ```
> 
> This will fail, because the v4 DSL machinery currently only supports openapi surfaces.
> 
> /goal Add support for MCP surfaces. Success is achieved when the above command works and all test queries succeed. A few things to think about:
> 
> 1. The MCP machinery already exists. Re-use the existing code (e.g. in crates/coral-engine/src/backends/mcp and crates/coral-spec/src/backends/mcp.rs) wherever possible.
> 2. No massive [mod.rs](http://mod.rs/) files please! Look at the existing repo layout and adopt the conventions therein.
> 3. It's OK to change the test queries, e.g. if the natural shape of generated projections for the MCP surface varies from the shape I've used in those queries. Do not overfit your solution to those exact queries.


crates/coral-spec/src/v4/projections/derive.rs
```diff
@@ -159,15 +218,55 @@ fn projection_input_sql_exposure(
...
+            ProjectionColumn {
+                name: "result".to_string(),
+                data_type: ManifestDataType::Utf8,
+                source_path: Vec::new(),
+                nullable: true,
+                description: "Full decoded tool response row rendered as text.".to_string(),
+            },
+            ProjectionColumn {
+                name: "result_json".to_string(),
+                data_type: ManifestDataType::Json,
```
simonwhitaker, line 295: Adding both a Utf8 and a Json column was inspired by Antón's PR #1183. I'm not convinced we need both, though. As far as I understand, `Json` is just a `Utf8` that hints to the query planner that the Utf8 string is a serialised JSON doc, so it could try using `json_*` functions on it.

sources/core-v4/github_v4/manifest.yaml
```diff
@@ -24,7 +24,28 @@ surfaces:
       - name: User-Agent
         from: literal
         value: coral-dsl-v4
+  - id: mcp
```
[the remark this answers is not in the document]
simonwhitaker: This is fine, the v4 manifest is not yet in routine use.

crates/coral-engine/src/runtime/registry.rs
```diff
@@ -106,15 +106,14 @@ async fn register_sources_inner(
     prepare_source_decorators(source_decorators, &selected_sources)?;
 
     let mut result = SourceRegistrationResult::default();
-    let mut seen_schemas = std::collections::HashSet::new();
+    let mut seen_schemas = catalog.schema_names().into_iter().collect();
```
iaptsiauri, line 117: Pre-seeding `seen_schemas` from DataFusion catalog schemas makes `public` an implicit reserved source namespace, but `public` is not in `RESERVED_SCHEMA_NAMES` and manifest validation accepts it.

crates/coral-app/src/sources/manager.rs
```diff
@@ -1169,7 +1190,6 @@ fn source_needs_stored_material_for_validation(
     let supplied_secrets = collect_unique_secrets(&bindings.secrets)?;
     Ok(candidate.inputs.iter().any(|input| {
         input.kind == ManifestInputKind::Secret
-            && input.required
```
iaptsiauri, line 1172: Removing `input.required`  makes re-import/update hit credential storage whenever any optional secret is unsupplied, and a failing keychain read now aborts the install

crates/coral-spec/src/v4/surfaces/mcp/import.rs
```diff
@@ -0,0 +1,785 @@
...
+            id: id.clone(),
+            shape: IrTypeShape::Scalar(scalar),
+            nullable: true,
+            description: String::new(),
+        });
+        id
+    }
+}
+
+fn schema_scalar_type(schema: &Value) -> IrScalarType {
```
iaptsiauri: `schema_scalar_type` is the third hand-rolled `JSON-schema→IrScalarType` mapper in coral-spec, and the three already disagree on type arrays, format precedence, and a `datetime` alias nothing else accepts

crates/coral-engine/src/contracts/query.rs
```diff
@@ -97,18 +97,21 @@ impl QuerySource {
     ///
     /// # Errors
     ///
-    /// Returns [`CoreError`](crate::CoreError) when any component belongs to a
-    /// different logical source/schema.
+    /// Returns [`CoreError`](crate::CoreError) when the package is invalid.
     pub fn from_runtime_components(
```
iaptsiauri, line 101: `from_runtime_components` validates component schemas only within one runtime package, Maybe we can introduce cross-source namespace validation before install succeeds

crates/coral-app/src/sources/materialization.rs
```diff
@@ -492,44 +544,42 @@ fn write_materialization(
...
-        let bytes = read_descriptor(surface)?;
-        validate_materialized_surface_base_url(manifest, surface, &bytes)?;
-        let observed = sha256_hex(&bytes);
-        let semantic_ir = import_openapi_surface(manifest, surface, &bytes).map_err(|error| {
-            AppError::FailedPrecondition(format!(
-                "failed to import source '{}' surface '{}': {error}",
-                manifest.common.name, surface.id
-            ))
-        })?;
+        let materialized_surface = materialize_surface(manifest, surface, inputs)?;
```
iaptsiauri: Materialization is still all-or-nothing per source. For `github_v4`, any MCP tools/list failure against `api.githubcopilot.com` aborts the whole source, including the REST surface

crates/coral-spec/src/v4/surfaces/mcp/import.rs
```diff
@@ -0,0 +1,785 @@
...
+            }),
+            execution: IrExecutionAttachment::Mcp(McpExecutionAttachment {
+                tool_name: tool.name.clone(),
+                pagination,
+            }),
+            diagnostics: Vec::new(),
+        }
+    }
+
+    fn import_inputs(&mut self, tool: &McpToolDescriptor) -> Vec<IrOperationInput> {
```
iaptsiauri: `import_inputs` reads only top-level properties. Legal MCP input schemas using `$ref, allOf, or anyOf` import as zero inputs with no diagnostic. A read-only tool with required composed args becomes a no-arg SQL table/function and sends `{}` at query time. The importer’s diagnostics vec still has no writers, which is the silent-failure smell

crates/coral-engine/src/lib.rs
iaptsiauri, line 115: `list_catalog/search_catalog` still filter by exact SQL schema only, while the MCP tool description says `schema/source name`. For multi-surface v4 sources there may be no schema named after the installed source (github_v4), so filtering by source name returns an empty page.

crates/coral-spec/src/v4/manifest.rs
```diff
@@ -226,7 +365,31 @@ fn validate_surface_id(source_name: &str, id: &str) -> Result<()> {
     }
 }
 
-fn parse_descriptor(source_name: &str, surface: &RawV4Surface) -> Result<SurfaceDescriptor> {
+fn default_surface_namespace(source_name: &str, surface_id: &str, surface_count: usize) -> String {
```
jsummerfield: Could we avoid making this an implicit manifest-level default? `namespace` is being used as the SQL/runtime source name, but deriving it from `(source_name, surface_id, surface_count)` means adding a second surface silently renames the existing surface from `<source>` to `<source>_<surface>`, which can break existing SQL references even though the new surface is additive. I think this should either require explicit namespaces for multi-surface specs, or be derived through the SQL projection/runtime-package naming path instead of manifest parsing. That keeps provider/source identity separate from projection naming and avoids baking SQL identifier policy into the v4 source model.

jsummerfield (commented):
One inline comment on the namespace defaulting behavior.

crates/coral-spec/src/v4/manifest.rs
simonwhitaker: Yeah, fair point. Requiring explicit namespaces for multi-surface specs feels less magical, I'll do that.

crates/coral-spec/src/v4/projections/derive.rs
```diff
@@ -146,8 +146,72 @@ fn generate_projection(
...
+fn projection_kind(operation: &IrOperation, is_search: bool) -> ProjectionKind {
+    let function_kind = match &operation.execution {
+        IrExecutionAttachment::Rest(_) if is_search => Some(SourceTableFunctionKind::Search),
+        IrExecutionAttachment::Rest(_)
+            if operation.output.cardinality == OutputCardinality::Singleton
+                && operation.inputs.iter().any(|input| input.required) =>
+        {
+            Some(SourceTableFunctionKind::Table)
+        }
+        IrExecutionAttachment::Mcp(_) if operation.inputs.is_empty() && !is_search => None,
```
jsummerfield: This decides whether an MCP projection is a table before the pagination-owned cursor input is hidden below. A tool whose only input is the cursor will have `operation.inputs` non-empty here, so it becomes a zero-arg table function even though SQL callers cannot and should not pass the cursor. Could we base the MCP table/function decision on the public SQL inputs after excluding `mcp_pagination_owns_input`, and add a cursor-only pagination regression test?

crates/coral-app/src/sources/runtime_package.rs
```diff
@@ -134,28 +149,200 @@ fn http_manifest_for_surface(
...
+                    mcp,
+                    operation,
+                ));
+            }
+        }
+    }
+    Ok(McpSourceManifest {
+        common: SourceManifestCommon {
+            dsl_version: manifest.common.dsl_version,
+            name: surface.namespace.clone(),
```
jsummerfield: Now that one logical source can publish multiple schema names from surface namespaces, install-time collision checks need to validate those schema names too. Today `source_exists` only checks the candidate source name, so importing a v4 source whose surface namespace matches an already-installed schema can persist cleanly and then collide at query/catalog time. Can we reject workspace-wide surface namespace collisions before persistence?

crates/coral-engine/src/backends/mcp/transport.rs
```diff
@@ -362,6 +435,45 @@ fn value_to_env_string(value: Value) -> String {
...
+            .list_tools(Some(PaginatedRequestParams::default().with_cursor(cursor)))
+            .await
+            .map_err(|error| {
+                DataFusionError::External(Box::new(classify_error(source_name, &error)))
+            })?;
+        tools.extend(result.tools);
+        let Some(next_cursor) = result.next_cursor else {
+            return Ok(tools);
+        };
+        if !seen_cursors.insert(next_cursor.clone()) {
```
jsummerfield, line 458: This repeated-cursor guard is useful for discovery, but the query-time MCP fetch loop still lacks the same protection. `McpFetchPlan::fetch` follows `next_page_cursor` until `max_pages`, so a server returning the same cursor repeatedly burns all pages and reports only a max-pages failure. Could we add a `seen_cursors` check there as well so runtime pagination fails immediately with the repeated cursor?

crates/coral-spec/src/v4/manifest.rs
simonwhitaker: OK, this is done. I've gone for explicit namespace *suffixes* in the manifest. If a namespace suffix is not provided, you'll get the source name as the schema name. Two surfaces with the same schema name are not allowed. So you can do this:

```
name: github_v4
dsl_version: 4
surfaces:
  - id: rest
    type: openapi
  - id: mcp
    type: mcp
    namespace_suffix: mcp
```

And you'll get `github_v4.*` and `github_v4_mcp.*`. But this would be rejected:

```
surfaces:
  - id: rest
    type: openapi
  - id: mcp
    type: mcp # no namespace suffixes; both surfaces get the same schema
```

And so would this:

```
surfaces:
  - id: rest
    type: openapi
    namespace_suffix: api
  - id: mcp
    type: mcp
    namespace_suffix: api # two surfaces have the same namespace suffix
```

I went with suffixes partly to avoid having to repeat ourselves, and partly to avoid something like this:

```
name: github
dsl_version: 4
surfaces:
surfaces:
  - id: rest
    type: openapi
    namespace: github_rest
  - id: mcp
    type: mcp
    namespace_suffix: linear_mcp # OOPS!
```

That could happen via a bad copy-paste, but I suspect there's also a malicious use case there if we think through the possibilities.

crates/coral-engine/src/runtime/registry.rs
simonwhitaker: Addressed in https://github.com/withcoral/coral/commit/b277064da107c27626c671659a91eed0f7607993.

crates/coral-app/src/sources/manager.rs
simonwhitaker: Addressed in https://github.com/withcoral/coral/commit/6f6408e038da2de6b8710b31911938e3a60781e9.

crates/coral-spec/src/v4/surfaces/mcp/import.rs
simonwhitaker: Addressed in https://github.com/withcoral/coral/commit/c5c3e6b8addce86c72b8c5a3e95e62a405049ce1.

crates/coral-engine/src/contracts/query.rs
simonwhitaker: Addressed in https://github.com/withcoral/coral/commit/dbceda24b40bfc2eaa13787affe2e6ba5770c460.

crates/coral-app/src/sources/materialization.rs
simonwhitaker: Addressed in https://github.com/withcoral/coral/commit/af8b097b9564e3e12e8ea6d60c632c18589a7341.

crates/coral-spec/src/v4/surfaces/mcp/import.rs
simonwhitaker: Addressed in https://github.com/withcoral/coral/commit/6c990e598a0e6805b03bdb202f6158ce15105c79.

crates/coral-engine/src/lib.rs
simonwhitaker: Addressed in https://github.com/withcoral/coral/commit/6d9be02fb7ce66758413945fbdc5b206475da66a.

crates/coral-spec/src/v4/manifest.rs
simonwhitaker: Addressed in https://github.com/withcoral/coral/commit/753bc7d82d722b9311c7e3cc951b760bac7b15cb.

crates/coral-spec/src/v4/projections/derive.rs
simonwhitaker: Addressed in https://github.com/withcoral/coral/commit/9db524884f560e2ce7710d59886aff2bfd1ff5c8.

crates/coral-app/src/sources/runtime_package.rs
simonwhitaker: Addressed in https://github.com/withcoral/coral/commit/137639423a44bf6df2d0e76504b9f12449e569fe.

crates/coral-engine/src/backends/mcp/transport.rs
simonwhitaker: Addressed in https://github.com/withcoral/coral/commit/350cec10dddfd0c0aa234ea26841914cba073c21.

AlbertQM (commented):
Nice! I was expecting more than a 1-line tweak in the docs though? E.g. documenting the new type `mcp`

simonwhitaker:
> Nice! I was expecting more than a 1-line tweak in the docs though? E.g. documenting the new type `mcp`

We're not adding DSL v4 to the docs yet.

iaptsiauri (approved):

AlbertQM (approved):

Bradley-Butcher:
Nice direction overall. I like the `namespace_suffix` shape in particular: keeping surface schema names source-relative feels much safer than absolute namespace strings.

A few taste/design questions before this lands. These are less about individual bugs and more about the contracts this PR is making durable.

1. **MCP output typing feels half-committed**

   The importer reads structured MCP `outputSchema` into IR, including row shape, but the generated SQL projection deliberately exposes only `result` and `result_json`.

   I can see the argument for opaque MCP output, and I can see the argument for typed best-effort columns. What feels odd is doing both: preserving typed schema detail internally, then hiding it at the SQL boundary.

   The SQL shape I’d hope for, when the MCP output schema is projectable, is more like:

   ```sql
   select number, title, state
   from github_v4_mcp.list_issues(owner => 'withcoral', repo => 'coral');
   ```

   with `result_json` still available as the escape hatch. Is there a reason we can’t generate best-effort typed columns for object/list-object outputs, and fall back to `result` / `result_json` only when the shape is unknown or not safely projectable? If opaque output is the intended v1 contract, can we make that explicit in the PR/docs/comments?

2. **Partial materialization needs a clearer lifecycle contract**

   `write_materialization` now continues after a per-surface failure as long as at least one surface materializes. That means a source can install with a declared surface missing.

   That may be the right product choice, but it feels like a first-class lifecycle contract, not an implementation detail. If partial installs are intentional, I’d expect `coral source add` to print a very obvious warning naming the failed surface and reason. I’d also expect `source test`, catalog/source metadata, and docs to make the partial state visible.

   Otherwise, should `source add` fail unless all declared surfaces materialize, with an explicit optional-surface marker or `--allow-partial` escape hatch later?

3. **The MCP importer shape is inconsistent with the local v4 importer style**

   This is not “big file bad” by itself. It’s that the MCP importer is carrying several separate policy areas in one place: operation import, input schema `$ref` / `allOf`, schema conflict rules, output row typing, wrapped-list detection, and cursor pagination inference.

   The OpenAPI importer has a nicer local shape: thin orchestration plus separate operations/responses/schemas/shared JSON helpers. Could we split MCP along similar lines before landing? Something like `import.rs`, `input_schema.rs`, `output_schema.rs`, and `pagination.rs` would make future changes much easier to review.

4. **Pagination inference needs to be explicit about scope and control**

   I agree auto-generated MCP projections need inference. But inferred pagination becomes durable runtime behavior once materialized.

   Today this appears to be cursor-only and name-shape based: cursor-like input names plus response fields like `nextCursor` / `nextToken`. Is cursor-only pagination the intentional v1 scope? If so, can we document that non-cursor patterns like `page`/`per_page`, `offset`/`limit`, link-style next URLs, and keyset/window pagination are not inferred yet?

   I’d also like to understand whether source authors can inspect, override, or disable inferred pagination, and whether we surface a diagnostic when pagination was inferred.

Bradley-Butcher (approved):
if you just want the `mcp` crate reviewed tho - description changes LGTM

simonwhitaker:
> Nice direction overall. I like the `namespace_suffix` shape in particular: keeping surface schema names source-relative feels much safer than absolute namespace strings.

Gotta love Codex. This is like reviewing the plans for a nuclear power station and fixating on the colour they chose for the doors. 😆 

> A few taste/design questions before this lands. These are less about individual bugs and more about the contracts this PR is making durable.

<snip>

Good stuff, thanks. In reply:

1. I've gone back and forth on this. I think we should infer and store as much as we can from the MCP server's tool definitions. But unlike an OpenAPI spec, those tool definitions are not a durable, versioned contract. They can change at any time. So, for now, I'm sticking with only exposing raw JSON responses in the projection layer, since it's resistant to unannounced response shape changes. When agents query Coral they will be able to use that data easily enough. (A central tenet of DSL v4 is that users can write their own projections, so that's always an option for users who strongly prefer typed outputs.)
2. Agree this should be tightened up. I'm tempted to leave it for a follow-up PR though, this beast is already huge. Files #1299 
3. Ditto. I'll handle refactoring the code in a separate PR. Filed #1302 
4. Legit concern. Again, I'd rather merge this PR, get it into Coral employee's hands so we can start getting real, useful feedback, and address extending the pagination options in a follow-up PR. Files #1301

jsummerfield (approved):

PR #1216, opened 2026-06-05
feat(oauth): add dynamic client registration

sauldhernandez:
## Intent
Add OAuth Dynamic Client Registration and resource indicator support so Coral can connect to providers that issue per-install OAuth clients instead of requiring static client IDs.

## What it enables
OAuth credential methods can register native/public or confidential clients, carry resource indicators through authorization, device, token, and refresh requests, persist DCR metadata, and expose the configuration through proto/client/schema/docs/skills; ClickHouse Cloud now uses DCR with random loopback redirects.

## Usage examples
Run `coral source add --file sources/community/clickhouse_cloud/manifest.yaml --interactive` and choose `Connect with ClickHouse Cloud`.

## Validation
`python3 /Users/saul/.codex/skills/autoreview/scripts/autoreview --mode local --parallel-tests "make rust-checks && make schema-check"` passed with no accepted/actionable findings and `make rust-checks && make schema-check` exited 0.


crates/coral-spec/src/inputs.rs
```diff
@@ -526,15 +609,27 @@ fn validate_oauth_endpoint_templates_for_method(
...
             input_key,
-            "device_authorization_url",
+            "endpoints.device_authorization_url",
             template,
             declared,
             input_scope,
         )?;
     }
+    if let Some(template) = oauth.resource.as_deref() {
+        validate_oauth_endpoint_template(input_key, "resource", template, declared, input_scope)?;
```
jsummerfield: cubic flagged this too, and I agree: this only proves `resource` parses as a URL. RFC 8707 resource indicators must not contain fragments, so `https://api.example/#frag` can pass here and then get sent to the provider. Please reject fragments when validating/rendering `oauth.resource` so invalid manifests fail before the OAuth flow starts.

crates/coral-client/src/sources.rs
```diff
@@ -241,6 +252,39 @@ fn oauth_client_secret_from_proto(
...
+        Ok(OauthDynamicClientRegistrationAuthMethod::None) => {
+            Ok(ManifestOAuthDynamicClientRegistrationAuthMethod::None)
+        }
+        Ok(OauthDynamicClientRegistrationAuthMethod::ClientSecretBasic) => {
+            Ok(ManifestOAuthDynamicClientRegistrationAuthMethod::ClientSecretBasic)
+        }
+        Ok(OauthDynamicClientRegistrationAuthMethod::ClientSecretPost) => {
+            Ok(ManifestOAuthDynamicClientRegistrationAuthMethod::ClientSecretPost)
+        }
+        Ok(OauthDynamicClientRegistrationAuthMethod::Unspecified) | Err(_) => {
```
jsummerfield: cubic flagged this too, and I agree: `token_endpoint_auth_method` is optional in the manifest and defaults to `none`, but proto omission decodes as `UNSPECIFIED` and currently fails. The client should map `UNSPECIFIED` to `ManifestOAuthDynamicClientRegistrationAuthMethod::None` and reserve the error for unknown numeric enum values.

crates/coral-cli/src/source_ops.rs
```diff
@@ -1554,7 +1554,11 @@ fn prompt_oauth_credential_inputs(
 ) -> Result<Vec<OAuthCredentialInput>, anyhow::Error> {
     let mut values = Vec::new();
     if let Some(input_key) = oauth.client.id.input.as_deref()
-        && let Some(value) = prompt_oauth_client_id(input_key, oauth.client.id.default.as_deref())?
+        && let Some(value) = prompt_oauth_client_id(
+            input_key,
+            oauth.client.id.default.as_deref(),
+            oauth.client.dynamic_registration.is_some() && oauth.client.secret.is_none(),
```
jsummerfield: cubic flagged this too, and I agree: this makes DCR fallback depend on there being no static `client.secret`, but the app path falls back to dynamic registration whenever no client id is supplied. A manifest with `client.id.input` + `client.secret` + `dynamic_registration` is accepted by spec/app, yet the CLI rejects the empty client id before DCR can run. Either allow empty when `dynamic_registration` is present, or reject that combination in the spec.

crates/coral-app/src/credentials/oauth.rs
```diff
@@ -1337,6 +1715,11 @@ fn oauth_refresh_config(
         .filter(|value| !value.is_empty())
         .cloned()
         .unwrap_or_else(|| oauth.token_url.clone());
+    let resource = material
+        .get(&format!("{metadata_prefix}resource"))
+        .filter(|value| !value.is_empty())
+        .cloned()
+        .or_else(|| oauth.resource.clone());
```
jsummerfield: cubic flagged this too, and I agree: this fallback uses raw `oauth.resource`. If stored material predates the new `resource` metadata and the manifest resource uses `{{input.*}}`, refresh sends the template literally because `oauth_refresh_config` has no source inputs to render it. Please persist/use the rendered resource, or avoid falling back to an unrendered template.

jsummerfield (commented):
Inline comments on the DCR/resource issues I verified.

AlbertQM (approved):

docs/guides/write-a-custom-source.mdx
```diff
@@ -176,6 +176,42 @@ For field-level details and examples, see [Credential methods](/reference/source
 
 Keep each method's `hint` scoped to the fields that method collects. If OAuth endpoint URLs need non-secret install-time configuration, declare those values as source variables; see [Source inputs](/reference/source-spec-reference#source-inputs).
 
+When an API supports either a complete pasted API-key header or an OAuth access
```
AlbertQM: After #1235, these more in-depth explanations have moved to the Reference section. Could you please move these too? Feel free to make minor changes to this guide, and/or add links to the more in-depth reference if needed (though they should already be there)

crates/coral-app/src/credentials/oauth.rs
```diff
@@ -1253,14 +1587,48 @@ fn oauth_credential_material(
         format!("{prefix}token_url"),
         session.endpoints.token_url.clone(),
     );
-    if let Some(secret) = session.oauth.client.secret.as_ref() {
+    insert_optional_metadata(
+        &mut internal_metadata,
+        format!("{prefix}resource"),
```
jsummerfield: **Metadata key vocabulary is hand-encoded at every read and write site (worth fixing before merge).**

The set of metadata keys that make up a persisted OAuth credential (`resource`, `client_id`, `client_secret`, `token_url`, `refresh_token`, `method`, … plus the new DCR keys) is a single decision, but it's re-spelled as independent `format!("{prefix}<literal>")` string literals across three functions with no shared definition: the write site here in `oauth_credential_material`, the read sites in `oauth_refresh_config` / `oauth_refresh_resource` (~L1662–1783), and again in `apply_refreshed_token` (~L1786–1822).

Write and read have to agree character-for-character, so a typo or rename in one place compiles cleanly and silently breaks token refresh at runtime — there's no compiler help. This PR widens the surface by adding `resource` + the DCR keys to multiple sites.

**Suggestion:** give the key vocabulary one owner — either `const` suffix strings (`RESOURCE`, `CLIENT_ID`, …) referenced by all sites, or a typed `OAuthMetadata` struct with a `to_metadata` / `from_metadata` pair. Then adding a field is a one-line change and the writer/reader can't drift.

jsummerfield (commented):
A few design-level observations from a focused read of the OAuth changes. None are correctness bugs and the overall shape is good — unifying static-ID and DCR behind `ResolvedOAuthClient` (instead of `if is_dcr` branching) and folding the authorize args into `OAuthAuthorizationRequest` are both solid. Comments below are refinements, ordered roughly by value. The first is the one I'd act on before merge.

crates/coral-app/src/credentials/oauth.rs
```diff
@@ -143,12 +146,43 @@ struct TokenResponse {
...
 }
 
+struct ResolvedOAuthClient {
+    client_id: String,
+    client_secret: Option<String>,
+    client_secret_transport: Option<ManifestOAuthClientSecretTransport>,
+    dynamic_client_registration: Option<DynamicClientRegistrationMaterial>,
+}
+
+struct DynamicClientRegistrationMaterial {
```
jsummerfield: **These four fields are persisted but never read back — and one of them is a secret.**

`registration_client_uri`, `registration_access_token`, `client_id_issued_at`, and `client_secret_expires_at` are parsed from the DCR response, threaded through `ResolvedOAuthClient` → `OAuthSessionCommon`, and written into metadata in `oauth_credential_material` (~L1607–1632) — but a repo-wide grep shows nothing ever reads any of them back, including the refresh path (`oauth_refresh_config` reads only `client_id` / `client_secret` / `resource`). A reader follows the whole chain only to find a dead sink.

I assume the intent is forward-compat — these are exactly the RFC 7591/7592 fields a future re-registration/rotation feature would need. That's reasonable, but two things:
- **`registration_access_token` is a secret being written to the credential store with no consumer.** Worth a deliberate yes/no rather than an incidental write.
- If you keep them, a one-line doc comment here saying they're persisted for future client-management and intentionally not yet read would save the next reader the grep. Otherwise, dropping them until a consumer exists (keeping just the `dynamic_client_registration` marker) is cleaner.

crates/coral-app/src/credentials/oauth.rs
```diff
@@ -143,12 +146,43 @@ struct TokenResponse {
     expires_at: Option<DateTime<Utc>>,
 }
 
+struct ResolvedOAuthClient {
```
jsummerfield, line 146: **Minor:** `ResolvedOAuthClient` and `OAuthSessionCommon` (~L86–94) carry the same four-field client subset (`client_id`, `client_secret`, `client_secret_transport`, `dynamic_client_registration`), so the two `authorize_*` functions (~L287–295 and ~L416–424) re-list each field by hand when building the session. Adding a fifth client attribute then means editing both struct definitions plus both copy sites.

The two types do have a legitimate split (`ResolvedOAuthClient` is the short-lived resolution result, the session lives longer), so this is low-priority — but embedding `client: ResolvedOAuthClient` as a field of `OAuthSessionCommon` (or factoring a shared `OAuthClientMaterial`) would make "add a client attribute" a one-place change and delete the hand-copy at both call sites.

crates/coral-app/src/credentials/oauth.rs
```diff
@@ -1363,9 +1743,45 @@ fn oauth_refresh_config(
         client_secret,
         client_secret_transport,
         refresh_token,
+        resource,
     }))
 }
 
+fn oauth_refresh_resource(
```
jsummerfield, line 1744: **Minor:** this fallback parses the resource template twice — once here (`ParsedTemplate::parse`, ~L1767) and again inside `oauth.resource(&BTreeMap::new())` (~L1778) — and the two `map_err` closures (~L1769 and ~L1780) produce a byte-identical "manifest resource is invalid" message. Once the tokens-present check at ~L1772 passes, the L1778 path can no longer fail with a *parse* error, so that second message mis-describes the only failure it can actually surface (a URL/fragment render error), and that guard branch is effectively unreachable for the parse case.

Not a bug — worst case is a slightly misleading error string on an already-failing refresh — but de-duping the two error sites and correcting the second message to describe a render failure would make this read-through clearer.

jsummerfield (approved):

PR #1217, opened 2026-06-05
fix(catalog): expose system metadata tables to helpers

iaptsiauri:
## Summary

- add typed system catalog definitions for `coral.tables`, `coral.columns`, `coral.filters`, `coral.inputs`, and `coral.table_functions`
- include those system tables in `collect_tables`, `coral.tables`, and `coral.columns` so SQL and catalog helpers agree
- update MCP docs and regression coverage for the `coral` system schema

## Root Cause

`sql` could query the DataFusion-registered `coral` schema, but catalog helpers used the engine `TableInfo` snapshot from `collect_tables`, which only enumerated installed source tables. That made `coral.*` queryable by explicit SQL while still undiscoverable through `list_catalog`, `describe_table`, and `list_columns`.

## Validation

- `cargo test -p coral-engine --test engine system_catalog`
- `cargo test -p coral-mcp mcp_catalog_helpers_expose_coral_system_tables_from_sql_catalog`
- `cargo test -p coral-app --test grpc catalog_discovery_tests::describe_missing_table_returns_catalog_suggestions`
- `git diff --check`
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`
- `make docs-check`
- `make rust-checks`


sauldhernandez (approved):
Please update the mcp guide so it reflects that coral://tables also now includes the data for the coral catalog tables, not just sources.  
  
Also update the docs in catalog.proto  to reflect the changes. Current docs only state that it's source-related, but now it also includes catalog metadata.

AlbertQM (approved):

mystic123 (approved):

PR #1220, opened 2026-06-06
feat(sources): support OAuth client credentials

siiddhantt:
## Summary

- Add `flow.type: client_credentials` for OAuth source credential methods.
- Mint app-only access tokens with `grant_type=client_credentials`, store expiry metadata, and re-mint tokens from stored client credentials when the provider returns usable `expires_in` metadata.
- Thread the new flow through source spec parsing, schema validation, proto/client mappings, app materialization, and source info surfaces.
- Send the Coral default User-Agent on OAuth token requests so providers can identify token mint/refresh calls consistently.
- Update source-spec docs and maintained Coral source-authoring/review skills.

## Validation

- `make rust-checks`
- `make schema-check`
- `make docs-check`
- `git diff --check`
- Live Reddit source validation:
  - `source add --interactive --file /coral-reddit-live/manifest.yaml`: 3/3 query tests passed
  - `source test reddit`: 3/3 query tests passed


crates/coral-app/src/credentials/oauth.rs
```diff
@@ -1309,17 +1450,32 @@ fn oauth_refresh_config(
...
+        .get(&format!("{metadata_prefix}flow"))
+        .map(|value| {
+            oauth_flow_kind_from_label(value).ok_or_else(|| {
+                AppError::FailedPrecondition(format!(
+                    "stored OAuth flow for source secret '{access_token_material_key}' is invalid: {value}"
+                ))
+            })
+        })
+        .transpose()?
+        .unwrap_or(oauth.flow.kind);
```
sauldhernandez: Before this change refresh config would always go via refresh token. The default if no configuration is present should be to preserve that, not use the current flow type.

docs/reference/source-spec-reference.mdx
sauldhernandez: Needs to be update to also state that we support client_credentials.

crates/coral-app/src/credentials/oauth.rs
```diff
@@ -632,6 +724,10 @@ fn join_scope_values(delimiter: ManifestOAuthScopeDelimiter, values: &[String])
     values.join(separator)
 }
 
+fn oauth_scope_value(scopes: Option<&ManifestOAuthScopesSpec>) -> Option<String> {
```
sauldhernandez: Very simple function. Could be inlined on call sites.

crates/coral-app/src/credentials/oauth.rs
```diff
@@ -412,6 +503,7 @@ impl OAuthCredentialService {
 fn token_http_client(timeout: Duration) -> reqwest::Client {
     reqwest::Client::builder()
         .timeout(timeout)
+        .user_agent(DEFAULT_OAUTH_TOKEN_USER_AGENT)
```
sauldhernandez, line 670: Was this added because a specific provider required it?

crates/coral-app/src/credentials/oauth.rs
```diff
@@ -1357,12 +1513,20 @@ fn oauth_refresh_config(
             "OAuth access token for source secret '{access_token_material_key}' expired and cannot be refreshed because client secret metadata is missing"
         )));
     }
+    if matches!(grant, OAuthRefreshGrant::ClientCredentials)
+        && (client_secret_transport.is_none() || client_secret.is_none())
+    {
+        return Err(AppError::FailedPrecondition(format!(
+            "OAuth access token for source secret '{access_token_material_key}' expired and cannot be re-minted because client secret metadata is missing"
```
sauldhernandez: This error message should be clearer, with guidance to the user on how to take corrective action.

sauldhernandez (changes requested):

crates/coral-app/src/credentials/oauth.rs
siiddhantt: Thanks @sauldhernandez good catch, yeah that’s a backwards-compat miss. I’ll make missing flow metadata preserve the old refresh-token behavior instead of inferring from the current manifest flow.

crates/coral-app/src/credentials/oauth.rs
siiddhantt: Fair, I’ll inline it at the call sites.

crates/coral-app/src/credentials/oauth.rs
siiddhantt: Yeah, Reddit surfaced this during live testing. Their token endpoint returned a 403 network-policy block with no UA, while `coral/<version>` worked. I kept it as Coral’s generic token-client UA rather than adding any source-specific override.

crates/coral-app/src/credentials/oauth.rs
siiddhantt: Agreed. I’ll make this more actionable and tell users to reconnect/re-add the source so Coral can store the client credentials needed for re-minting.

docs/reference/source-spec-reference.mdx
siiddhantt: Fixed this wording to include `client_credentials` too.

sauldhernandez (approved):
LGTM, thank you!

siiddhantt:
Thanks @sauldhernandez! I pulled in latest `main` and resolved the conflict in the custom source guide.

sauldhernandez (approved):

docs/guides/write-a-custom-source.mdx
```diff
@@ -170,12 +170,49 @@ tables:
...
+When a source should offer OAuth setup, add `credential.methods` to the secret input while keeping runtime `auth` focused on how the stored secret is sent to the API. Use `type: oauth` for device-code, authorization-code, or client-credentials setup, and add a `type: source_config` fallback when users can paste an existing token.
 
 For field-level details and examples, see [Credential methods](/reference/source-spec-reference#credential-methods), [OAuth flow and client settings](/reference/source-spec-reference#oauth-flow-and-client-settings), [OAuth redirect URIs](/reference/source-spec-reference#oauth-redirect-uris), and [`HeaderAuth`](/reference/source-spec-reference#headerauth) for `from: bearer` and `from: one_of`.
 
 Keep each method's `hint` scoped to the fields that method collects. If OAuth endpoint URLs need non-secret install-time configuration, declare those values as source variables; see [Source inputs](/reference/source-spec-reference#source-inputs).
 
+When an API supports either a complete pasted API-key header or an OAuth access
```
AlbertQM: This guide should remain lean, and just link to the reference for users that are interested in looking under the hood. PR #1235 cleaned this up, so please update this guide accordingly (likely not making any changes to this file, but only to the souce-spec-reference if needed). Thanks!

AlbertQM (changes requested):

docs/guides/write-a-custom-source.mdx
siiddhantt: Good call, thanks! 
I restored the custom source guide back to the lean `main` version and kept the client-credentials details in the source-spec reference / skills where they fit better.

AlbertQM (approved):
Thank you! Docs LGMT

siiddhantt:
Hi, I've rebased onto the latest main and rebuilt the patch cleanly. 
I reran the relevant tests, schema and skill checks, and verified the flow against Reddit, including automatic token re-minting after expiry. Everything passes and it’s ready for re-review. Thanks!

sauldhernandez (approved):

crates/coral-app/src/credentials/oauth.rs
```diff
@@ -1766,11 +1932,49 @@ fn oauth_refresh_config(
         client_id,
         client_secret,
         client_secret_transport,
-        refresh_token,
+        scope: oauth
+            .scopes
+            .as_ref()
+            .map(|scopes| join_scope_values(scopes.scope.delimiter, &scopes.scope.values)),
```
sauldhernandez: Renewal can silently broaden the selected credential's permissions. Setup honors `method_index`, but `refresh_oauth_input_material` passes the first OAuth method here. With read/write scopes on the first method and read-only scopes on the selected client-credentials method, the next renewal requests a read/write token if the client is authorized for both. Persist the selected setup scope context with the credentials and reuse it, preserving an explicitly omitted scope instead of falling back to another method. Please cover renewal after selecting the second method.

sauldhernandez (changes requested):

crates/coral-app/src/credentials/oauth.rs
siiddhantt: Thanks, good catch! Fixed in 278b47bf. Renewal now preserves the selected method’s scopes, including an omitted scope. I also fixed the same issue for an omitted resource and added coverage for selecting the second method and repeated renewals.

sauldhernandez (approved):

PR #1226, opened 2026-06-08
ci(release): sign and notarize macOS releases

jsummerfield:
## Intent

Start shipping macOS releases as Developer ID-signed and Apple-notarized artifacts without breaking existing install paths during the transition.

## Changes

- Adds a protected macOS signing/notarization release job that signs the built `coral` binary, submits the macOS ZIP to Apple with `notarytool`, and uploads signed macOS ZIP artifacts.
- Keeps signed macOS tarballs in the release output temporarily so the current Homebrew tap keeps working until it is updated to ZIPs.
- Updates the install script to prefer macOS ZIPs and fall back to legacy macOS tarballs for current/latest releases.
- Updates installation docs and README language for notarized macOS ZIP artifacts and macOS direct installs.

## Verification

- `sh -n scripts/install.sh`
- `bash -n scripts/install.sh`
- Parsed release and validate workflow YAML with Python/PyYAML
- `git diff --check`
- `make docs-check`
- Live Linux temp install against latest release
- Simulated macOS arm64 install against latest release, including ZIP 404 fallback to legacy tarball

## Follow-up

Update `withcoral/homebrew-tap` to point macOS at ZIP artifacts and remove its quarantine-stripping `postflight` once this release workflow is in place.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -223,8 +223,10 @@ jobs:
           set -euo pipefail
 
           certificate_path="$RUNNER_TEMP/developer_id_application.p12"
+          developer_id_ca_path="$RUNNER_TEMP/DeveloperIDG2CA.cer"
           key_path="$RUNNER_TEMP/AuthKey_${APP_STORE_CONNECT_API_KEY_ID}.p8"
           keychain_path="$RUNNER_TEMP/coral-signing.keychain-db"
+          original_keychains_path="$RUNNER_TEMP/original-keychains.txt"
 
           decode_base64_to_file() {
             value="$1"
@@ -239,6 +241,17 @@ jobs:
 
           decode_base64_to_file "$APPLE_DEVELOPER_ID_CERTIFICATE_BASE64" "$certificate_path"
           decode_base64_to_file "$APP_STORE_CONNECT_API_KEY_P8_BASE64" "$key_path"
+          security list-keychains -d user \
+            | sed -e 's/^ *//' -e 's/^"//' -e 's/"$//' > "$original_keychains_path"
+          curl --fail --location --retry 3 --output "$developer_id_ca_path" \
+            https://www.apple.com/certificateauthority/DeveloperIDG2CA.cer
+          if ! security import "$developer_id_ca_path" \
+            -k "$HOME/Library/Keychains/login.keychain-db"; then
+            security find-certificate -Z \
+              -c "Developer ID Certification Authority" \
+              "$HOME/Library/Keychains/login.keychain-db" \
+              | grep -q "SHA-256 hash: F16CD3C54C7F83CEA4BF1A3E6A0819C8AAA8E4A1528FD144715F350643D2DF3A"
+          fi
 
           security create-keychain -p "$APPLE_KEYCHAIN_PASSWORD" "$keychain_path"
           security set-keychain-settings -lut 21600 "$keychain_path"
@@ -246,11 +259,16 @@ jobs:
           security import "$certificate_path" \
             -P "$APPLE_DEVELOPER_ID_CERTIFICATE_PASSWORD" \
             -A \
-            -t cert \
             -f pkcs12 \
+            -T /usr/bin/codesign \
             -k "$keychain_path"
-          security set-key-partition-list -S apple-tool:,apple: -k "$APPLE_KEYCHAIN_PASSWORD" "$keychain_path"
-          security list-keychains -d user -s "$keychain_path"
+          security set-key-partition-list -S apple-tool:,apple:,codesign: -s -k "$APPLE_KEYCHAIN_PASSWORD" "$keychain_path"
+
+          keychains=("$keychain_path")
+          while IFS= read -r existing_keychain; do
+            [ -n "$existing_keychain" ] && keychains+=("$existing_keychain")
+          done < "$original_keychains_path"
+          security list-keychains -d user -s "${keychains[@]}"
           security find-identity -p codesigning -v "$keychain_path"
 
       - name: Sign macOS binary
@@ -334,8 +352,22 @@ jobs:
         if: always()
         shell: bash
         run: |
+          original_keychains_path="$RUNNER_TEMP/original-keychains.txt"
+          if [ -s "$original_keychains_path" ]; then
+            keychains=()
+            while IFS= read -r existing_keychain; do
+              [ -n "$existing_keychain" ] && keychains+=("$existing_keychain")
+            done < "$original_keychains_path"
+            if [ "${#keychains[@]}" -gt 0 ]; then
+              security list-keychains -d user -s "${keychains[@]}" || true
+            fi
+          fi
           security delete-keychain "$RUNNER_TEMP/coral-signing.keychain-db" || true
-          rm -f "$RUNNER_TEMP/developer_id_application.p12" "$RUNNER_TEMP"/AuthKey_*.p8
+          rm -f \
+            "$RUNNER_TEMP/developer_id_application.p12" \
+            "$RUNNER_TEMP/DeveloperIDG2CA.cer" \
+            "$RUNNER_TEMP/original-keychains.txt" \
+            "$RUNNER_TEMP"/AuthKey_*.p8
 
   checksums:
     needs:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -141,8 +141,8 @@ jobs:
           & $coral --version
           & $coral source discover
 
-      - name: Package Unix
-        if: runner.os != 'Windows'
+      - name: Package Linux
+        if: runner.os == 'Linux'
         shell: bash
         env:
           TARGET: ${{ matrix.target }}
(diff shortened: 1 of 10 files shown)


.github/workflows/release.yml
```diff
@@ -166,15 +177,202 @@ jobs:
...
+      - name: Install Apple signing assets
+        shell: bash
+        env:
+          APP_STORE_CONNECT_API_KEY_ID: ${{ secrets.APP_STORE_CONNECT_API_KEY_ID }}
+          APP_STORE_CONNECT_API_KEY_P8_BASE64: ${{ secrets.APP_STORE_CONNECT_API_KEY_P8_BASE64 }}
+          APPLE_DEVELOPER_ID_CERTIFICATE_BASE64: ${{ secrets.APPLE_DEVELOPER_ID_CERTIFICATE_BASE64 }}
+          APPLE_DEVELOPER_ID_CERTIFICATE_PASSWORD: ${{ secrets.APPLE_DEVELOPER_ID_CERTIFICATE_PASSWORD }}
+          APPLE_KEYCHAIN_PASSWORD: ${{ secrets.APPLE_KEYCHAIN_PASSWORD }}
+        run: |
+          set -euo pipefail
```
simonwhitaker: I remember a wise man once pushing back when I tried to land a GHA workflow with non-trivial inline shell script strings in it. He suggested extracting them into standalone shell scripts instead. That was good advice.

.github/workflows/release.yml
```diff
@@ -166,15 +177,202 @@ jobs:
...
+          path: unsigned-${{ matrix.target }}
+
+      - name: Install Apple signing assets
+        shell: bash
+        env:
+          APP_STORE_CONNECT_API_KEY_ID: ${{ secrets.APP_STORE_CONNECT_API_KEY_ID }}
+          APP_STORE_CONNECT_API_KEY_P8_BASE64: ${{ secrets.APP_STORE_CONNECT_API_KEY_P8_BASE64 }}
+          APPLE_DEVELOPER_ID_CERTIFICATE_BASE64: ${{ secrets.APPLE_DEVELOPER_ID_CERTIFICATE_BASE64 }}
+          APPLE_DEVELOPER_ID_CERTIFICATE_PASSWORD: ${{ secrets.APPLE_DEVELOPER_ID_CERTIFICATE_PASSWORD }}
+          APPLE_KEYCHAIN_PASSWORD: ${{ secrets.APPLE_KEYCHAIN_PASSWORD }}
```
simonwhitaker: Does this need to be a repo secret? You create the keychain, passing this password as an arg, then read the keychain, using the same password. It could just be a temporary string generated by the script, I think.

docs/getting-started/installation.mdx
```diff
@@ -6,9 +6,9 @@ description: "Install the Coral CLI"
 Coral ships as a single local CLI. Once installed, the same binary is used for source management, SQL queries, and the MCP stdio server.
 
 <Note>
-  Coral publishes macOS, Linux, and x86_64 Windows artifacts. Windows support
-  currently targets native Windows 10/11 on `x86_64-pc-windows-msvc`; ARM64
-  Windows is not released.
+  Coral publishes Developer ID-signed and notarized macOS ZIPs, Linux tarballs,
+  and x86_64 Windows ZIPs. Windows support currently targets native Windows
+  10/11 on `x86_64-pc-windows-msvc`; ARM64 Windows is not released.
```
AlbertQM: The previous version was way more human friendly, I'd drop these changes altogether 😄

.github/workflows/release.yml
```diff
@@ -166,15 +177,202 @@ jobs:
...
+              "$HOME/Library/Keychains/login.keychain-db" \
+              | grep -q "SHA-256 hash: F16CD3C54C7F83CEA4BF1A3E6A0819C8AAA8E4A1528FD144715F350643D2DF3A"
+          fi
+
+          security create-keychain -p "$APPLE_KEYCHAIN_PASSWORD" "$keychain_path"
+          security set-keychain-settings -lut 21600 "$keychain_path"
+          security unlock-keychain -p "$APPLE_KEYCHAIN_PASSWORD" "$keychain_path"
+          security import "$certificate_path" \
+            -P "$APPLE_DEVELOPER_ID_CERTIFICATE_PASSWORD" \
+            -A \
```
simonwhitaker: The docs for this flag say:

```
     -A  Allow any application to access the imported key without warning (insecure, not recommended!)
```

Do we need it?

.github/workflows/release.yml
```diff
@@ -166,15 +177,202 @@ jobs:
...
+      - name: Sign macOS binary
+        shell: bash
+        env:
+          APPLE_CODESIGN_IDENTITY: ${{ secrets.APPLE_CODESIGN_IDENTITY }}
+        run: |
+          set -euo pipefail
+
+          binary="unsigned-${TARGET}/coral"
+          chmod +x "$binary"
+          codesign \
```
simonwhitaker: High confidence: this signs whatever macOS artifact was built from `github.event.release.tag_name || inputs.tag`, but the workflow never proves that the value being signed is the actual trusted release ref. The `release` environment only gates the workflow run’s `GITHUB_REF`; for `workflow_dispatch`, that can be `main` while `inputs.tag` is arbitrary. Before any Apple signing material is available, add a guard that requires the run to be on the intended tag/ref, validates the tag pattern, and verifies the tag resolves to an expected/reachable release commit. The safer option is to remove free-form signing dispatch or add required reviewer protection.

.github/workflows/release.yml
```diff
@@ -166,15 +177,202 @@ jobs:
...
+                --issuer "$APP_STORE_CONNECT_API_ISSUER_ID" \
+                "$notary_log" || true
+              cat "$notary_log" || true
+            fi
+            exit "$notary_status"
+          fi
+
+          spctl --assess --type exec --verbose=4 "$binary"
+
+      - name: Upload signed macOS artifact
```
simonwhitaker: High confidence: this upload action runs while the decoded Developer ID P12, App Store Connect P8, and unlocked signing keychain still exist. `actions/upload-artifact` does not need Apple credentials, but it executes as the same runner user and can read `$RUNNER_TEMP` / use the unlocked keychain. Delete the P12/P8 and lock/delete the keychain immediately after signing/notarization, keep a final `if: always()` cleanup for failures, and preferably upload from a follow-up job that never receives Apple secrets.

.github/workflows/release.yml
```diff
@@ -166,15 +177,202 @@ jobs:
...
+            --issuer "$APP_STORE_CONNECT_API_ISSUER_ID" \
+            --wait \
+            --output-format json > "$notary_result"
+          notary_status=$?
+          set -e
+
+          cat "$notary_result"
+          submission_id="$(plutil -extract id raw -o - "$notary_result" 2>/dev/null || true)"
+
+          if [ "$notary_status" -ne 0 ]; then
```
simonwhitaker: Medium confidence: this only checks the `notarytool` process exit status. Since the command already captures JSON, explicitly parse and require `.status == "Accepted"` before uploading artifacts, and fetch the notary log for anything else. `spctl` on the local signed binary is useful, but it is not proof that Apple accepted the submitted ZIP.

simonwhitaker (commented):
Security review comments added inline. I’m leaving this as comments rather than an approval/request-changes action so you can choose the review disposition.

.github/workflows/release.yml
```diff
@@ -166,15 +177,202 @@ jobs:
...
+          legacy_archive="coral-${TARGET}.tar.gz"
+          staging="$RUNNER_TEMP/package-${TARGET}"
+          key_path="$RUNNER_TEMP/AuthKey_${APP_STORE_CONNECT_API_KEY_ID}.p8"
+          notary_result="$RUNNER_TEMP/notary-${TARGET}.json"
+          notary_log="$RUNNER_TEMP/notary-${TARGET}-log.json"
+
+          mkdir -p "$staging"
+          cp "$binary" "$staging/coral"
+          chmod +x "$staging/coral"
+          (cd "$staging" && ditto -c -k --keepParent coral "$GITHUB_WORKSPACE/$archive")
```
simonwhitaker: Using `ditto --keepParent` embeds the parent directory name, so the resulting zip contains `package-${TARGET}/coral` rather than `coral` at the top level. When `scripts/install.sh` extracts the zip with `unzip` and then runs `mv coral "$INSTALL_DIR/coral"`, the move fails because `coral` is nested inside a subdirectory, breaking the macOS direct-install path that the README now promotes. The legacy `tar.gz` built on the next line packages `coral` at the top level correctly, so only the zip is affected. Drop `--keepParent` so the binary sits at the archive root.

```suggestion
          (cd "$staging" && ditto -c -k coral "$GITHUB_WORKSPACE/$archive")
```

xtask/src/main.rs
```diff
@@ -25,7 +27,9 @@ use clap::{Parser, Subcommand};
 
 mod detect;
 mod docs;
+mod env;
```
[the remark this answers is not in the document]
jsummerfield: That's fine for xtask only

README.md
```diff
@@ -132,13 +132,13 @@ of Coral to your first SQL query. If you prefer an interactive wizard, you can r
...
 
-On macOS:
+On macOS, use Homebrew:
 
 ```bash
 brew install withcoral/tap/coral
 ```
 
-Or on Linux:
+Or use the direct installer on macOS or Linux:
```
AlbertQM: This also looks like a forced change made because in agents.md we say "you must update the docs", WDYT?

Hoping that once #1235 lands, we'll stop getting these

AlbertQM (commented):
Thanks for this! Will leave it to Simon to see if all his security comments were addressed 👀

simonwhitaker (approved):

PR #1237, opened 2026-06-08
feat(engine): add response.require_array_rows for fail-closed HTTP sources

### LINKED ISSUE #1064: HTTP source: ok_path cannot model string/numeric success flags (e.g. Etherscan status "1"/"0")
### Problem

For HTTP sources, `response.ok_path` is the only way to classify a provider success/failure that arrives as **HTTP 200**. But the success value is evaluated strictly as a JSON boolean:

https://github.com/withcoral/coral/blob/main/crates/coral-engine/src/backends/http/fetch.rs#L177-L180

```rust
let ok = get_path_value(&payload, &target.response().ok_path)
    .and_then(Value::as_bool)
    .unwrap_or(false);
```

`Value::as_bool` only matches a JSON `true`/`false`. Any string or number at `ok_path` is treated as failure.

### Impact

Many widely-used APIs signal errors in an HTTP 200 body with a **string or integer** status, not a boolean. Etherscan is a common example:

- success: `{"status":"1","message":"OK","result":[ ... ]}`
- failure: `{"status":"0","message":"NOTOK","result":"Invalid API Key"}`

Because `status` is the string `"1"`/`"0"`:

- With no `ok_path`, `rows_path: [result]` + `row_strategy: direct` turns a failure body into **one bogus row** (null data fields) — an invalid key can pass `coral source test`.
- With `ok_path: [status]`, a **successful** response (`"1"`) is also classified as failure (verified locally), so the source is unusable.

So there is currently no way to model this provider class safely. Context: withcoral/coral#947 (Etherscan community source) is blocked on exactly this.

### Proposed solutions (any one)

1. **`ok_equals` / `ok_value`** — compare the value at `ok_path` to an authored value, e.g.
   ```yaml
   response:
     ok_path: [status]
     ok_equals: "1"        # string/number/bool compare
     error_path: [result]
   ```
2. **Coerce common truthy/falsey scalars** at `ok_path` (`"1"`/`"true"`/`1` → true; `"0"`/`"false"`/`0` → false) in addition to JSON booleans.

(1) is the most explicit and avoids surprises. Happy to contribute a PR if the approach is agreed.

OoJae:
## Summary

Adds an opt-in `response.require_array_rows` so HTTP sources can **fail closed** when a provider returns HTTP 200 with an error scalar where a data array is expected. This resolves #1064 and unblocks the Etherscan community source (#947).

## Motivation

Some providers signal failures as HTTP 200 with an error string at the same JSON path that holds the data array on success. A boolean `ok_path` can't model this. Etherscan is the canonical case — I verified the live API:

| Response | `status` | `result` |
|---|---|---|
| Has rows | `"1"` | array |
| Valid query, **no results** | `"0"` | **empty array `[]`** |
| Invalid key / rate-limit / bad chain | `"0"` | **string** (error text) |

Note `status` is `"0"` for **both** the empty-but-valid case and real errors, so the commonly-suggested `ok_path:[status]` + `ok_equals:"1"` approach would wrongly reject legitimate empty results. The reliable discriminator is **`result` is an array (success/empty) vs a string (error)**.

## Change

`response.require_array_rows: bool` (default `false`). When `true`, a non-array value at `rows_path` is raised as a provider error (detail from `error_path`) instead of being coerced into a row. Backward compatible — existing sources (e.g. boolean `ok_path:[ok]`) are untouched.

- `coral-spec`: `ResponseSpec.require_array_rows` + JSON schema entry
- HTTP backend ([`http/fetch.rs`](crates/coral-engine/src/backends/http/fetch.rs)): array check after the existing `ok_path` check; the `error_path`→detail logic is factored into a shared `response_error_detail` helper and reused by both branches
- Tests: array → rows, empty array → 0 rows, scalar → fails closed
- Docs: response-fields table + an Etherscan example

## Verification

`cargo fmt`, `cargo clippy --all-targets -- -D warnings`, and `cargo test -p coral-engine -p coral-spec` (all green; 3 new tests). End-to-end against the live Etherscan V2 API with the `require_array_rows` manifest:

- valid key → `coral source test` passes
- empty block range (`status:"0"`, `result:[]`) → **0 rows, no error**
- invalid key (`status:"0"`, `result:"Invalid API Key"`) → query **fails** with `Invalid API Key` surfaced (previously became a bogus all-null row)

## Scope note

Implemented in the HTTP backend (where the need is). The field lives on the shared `ResponseSpec`; MCP-backend parity can follow if useful, but is out of scope here to keep the change focused.

### DIFF
--- a/crates/coral-engine/src/backends/http/fetch.rs
+++ b/crates/coral-engine/src/backends/http/fetch.rs
@@ -174,31 +174,39 @@ pub(super) async fn fetch_rows(
             break;
         };
 
+        let provider_failure = |detail: String| {
+            DataFusionError::External(Box::new(ProviderQueryError::ApiRequest {
+                source_schema: client.source_schema.clone(),
+                table: target.name().to_string(),
+                status: None,
+                method: None,
+                url: None,
+                filters: filter_values.clone(),
+                detail,
+            }))
+        };
+
         if !target.response().ok_path.is_empty() {
             let ok = get_path_value(&payload, &target.response().ok_path)
                 .and_then(Value::as_bool)
                 .unwrap_or(false);
             if !ok {
-                let err = if target.response().error_path.is_empty() {
-                    "unknown source API error".to_string()
-                } else {
-                    get_path_value(&payload, &target.response().error_path)
-                        .and_then(Value::as_str)
-                        .unwrap_or("unknown source API error")
-                        .to_string()
-                };
-                return Err(DataFusionError::External(Box::new(
-                    ProviderQueryError::ApiRequest {
-                        source_schema: client.source_schema.clone(),
-                        table: target.name().to_string(),
-                        status: None,
-                        method: None,
-                        url: None,
-                        filters: filter_values.clone(),
-                        detail: err,
-                    },
+                return Err(provider_failure(response_error_detail(
+                    &target.response().error_path,
+                    &payload,
                 )));
             }
+        }
+
+        // Fail closed when the provider returns HTTP 200 with a non-array value
+        // at `rows_path` (e.g. an error string where a data array is expected).
+        if target.response().require_array_rows
+            && !get_path_value(&payload, &target.response().rows_path).is_some_and(Value::is_array)
+        {
+            return Err(provider_failure(response_error_detail(
+                &target.response().error_path,
+                &payload,
+            )));
         }
 
         let mut rows = extract_rows(target.response(), &payload);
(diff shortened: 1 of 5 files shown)


antonmry:
Hi @OoJae. Thanks for the contribution. We've been discussing this internally. We need to find a way to support APIs with this type of unconventional behaviour, but at the same type, the solution needs to be more general, and not so focused on a particular case. We don't have an answer yet about how to do that, but we are working on GraphQL and MCP support that could be aligned with a potential solution. Let's discuss the approach in https://github.com/withcoral/coral/issues/1064 first and align it with the current Coral roadmap. I'm closing this PR in the meantime, we can always re-open it once it's clear how to solve this particular limitation.

PR #1244, opened 2026-06-10
fix(engine/http): reject unconsumed filters backed only by echo columns

iaptsiauri:
When the selected request route does not consume an equality filter, the engine enforces it locally through the same-named column. If that column is a from_filter echo, it always reflects the filter value, so every row "matches" and the result set is silently mislabeled (e.g. github.issues WHERE owner='x' returned the whole viewer feed stamped owner='x').

Fail the scan with a structured FILTER_NOT_APPLICABLE error instead. Filters no route ever consumes keep the echo behavior: they are deliberate client-side annotations with no request semantics to miss.


crates/coral-engine/src/backends/http/provider.rs
```diff
@@ -293,14 +294,51 @@ impl TableProvider for HttpSourceTableProvider {
...
+                if !routable_filters.contains(filter_name) {
+                    continue;
+                }
+                let echoes_filter = self
+                    .table
+                    .columns()
+                    .iter()
+                    .find(|column| column.name == *filter_name)
+                    .is_some_and(|column| {
+                        matches!(column.resolved_expr(), ExprSpec::FromFilter { .. })
```
sauldhernandez: This is too narrow. The bug is not specific to a column whose expression is exactly `ExprSpec::FromFilter`; the same mislabeling still happens for valid expressions that render from the filter without using response-row data, for example `template: "{{filter.owner}}"`, `coalesce([from_filter owner, ...])`, or `replace(from_filter owner, ...)`. In those cases a partially routed `WHERE owner = 'octocat'` query can still fetch the broad default request, evaluate `owner` from the active filter for every row, pass local filtering, and return mislabeled rows. The check needs recursive expression analysis that rejects local enforcement unless the filtered column is proven to carry response data.

crates/coral-engine/tests/engine/http_tests.rs
```diff
@@ -662,6 +662,78 @@ async fn local_route_filter_can_use_request_filter_values_in_template_column() {
...
+
+    let error = CoralQuery::execute_sql(
+        &[source],
+        test_runtime(),
+        "SELECT title, owner FROM http_echo_filter.issues WHERE owner = 'octocat'",
+    )
+    .await
+    .expect_err("filter the route does not consume and whose column merely echoes it must fail");
+
+    let rendered = error.to_string();
```
sauldhernandez: This regression test only verifies rendered prose. The change introduces a structured error contract (`FILTER_NOT_APPLICABLE`, failed-precondition status, and filter metadata), and CLI/MCP clients and telemetry branch on that structured payload rather than the display string. Please assert the returned `CoreError::QueryFailure`/`StructuredQueryError` fields, or add unit coverage for `ProviderQueryError::UnenforceableFilter::to_structured`, so a future reason/status/metadata regression cannot pass while still containing `cannot be applied`.

sauldhernandez (changes requested):
Requesting changes: the direct `FromFilter` guard does not close the full silent-mislabeling bug class, and the new structured error contract needs assertion coverage.

sauldhernandez (approved):

crates/coral-engine/src/backends/http/provider.rs
```diff
@@ -293,14 +295,56 @@ impl TableProvider for HttpSourceTableProvider {
...
+            let mut routable_filters = self.backend.request_filter_names(&self.table.request);
+            for route in &self.table.requests {
+                routable_filters.extend(self.backend.request_filter_names(&route.request));
+                routable_filters.extend(route.when_filters.iter().cloned());
+            }
+            for filter_name in local_filter_values.keys() {
+                if !routable_filters.contains(filter_name) {
+                    continue;
+                }
+                let echoes_filter = self
```
jsummerfield: I think this guard is still checking the wrong invariant. The problem is not only “does this column echo the same filter?”; it is “can this unconsumed filter be proven from response-row data?”

As written, a routable-but-unconsumed filter can still be enforced locally against a column backed by a literal, arg, default, or another filter-derived expression. That still lets rows pass without the selected request or the response proving the predicate. I would make this guard reject local enforcement unless the target column expression depends on response data, while continuing to allow filters already consumed by the HTTP request.

PR #1199, opened 2026-06-04
chore(xtask): benchmark MCP model surface tokens

Bradley-Butcher:
## Intent

Add a repeatable live MCP token benchmark for Coral's model-visible MCP surface. The command launches a real `coral mcp-stdio` server, calls real MCP methods, and counts both model-surface tokens and full protocol-result tokens with `tiktoken-rs`.

This now measures more than `tools/call` result text:

- `initialize`
- `tools/list`
- `resources/list`
- `resources/read` for `coral://guide`
- `resources/read` for `coral://tables`
- representative `tools/call` cases for discovery, SQL, empty results, and errors

## Benchmark command

```shell
cargo run --locked -p xtask -- mcp-token-bench --coral-bin target/debug/coral
```

By default it uses the real live Coral config. Pass `--config-dir` only for explicit fixture work.

## Real live run used for the stack

Binaries measured:

```text
/tmp/coral-token-bins/coral-main
/tmp/coral-token-bins/coral-pr1
/tmp/coral-token-bins/coral-pr2
target/debug/coral          # PR4 top
```

Config: default live Coral config. Transport: real `coral mcp-stdio`. Tokenizer: `gpt-5` / `o200k_base`.

Incremental stack results:

| PR slice | Comparison | Model-surface before | Model-surface after | Saved vs parent | Protocol before | Protocol after | Saved vs parent |
|---|---|---:|---:|---:|---:|---:|---:|
| PR1 compact JSON tool text | `main` -> PR1 | 173224 | 154598 | 18626 / 10.8% | 237895 | 210775 | 27120 / 11.4% |
| PR2 compact catalog page text | PR1 -> PR2 | 154598 | 140254 | 14344 / 9.3% | 210775 | 195393 | 15382 / 7.3% |
| PR3 benchmark harness | PR2 -> PR3 | 140254 | 140254 | 0 / 0.0% | 195393 | 195393 | 0 / 0.0% |
| PR4 compact `coral://tables` JSON | PR3 -> PR4 | 140254 | 117362 | 22892 / 16.3% | 195393 | 163386 | 32007 / 16.4% |

Cumulative after PR4: model-surface tokens drop from `173224` to `117362`, saving `55862 / 32.2%`; protocol-result tokens drop from `237895` to `163386`, saving `74509 / 31.3%`.

Context surface rows after PR2 showed the next hotspot clearly:

| case | surface | PR2 model-surface tokens | PR4 model-surface tokens | PR4 saved |
|---|---|---:|---:|---:|
[...]

PR #1246, opened 2026-06-10
feat(mcp): add CCR result handles

iaptsiauri:
## Summary

- store large MCP `sql` results in an in-memory, process-local `ResultStore` and return a compact envelope: `result_id`, row/column counts, schema summary, a 20-row preview, and a suggested `next_call`
- add a `result_get(result_id, offset?, limit?, columns?)` tool that pages stored results and projects columns by name, slicing Arrow batches before JSON conversion
- keep small results inline: `<= 20` rows always, and `<= 100` rows when the compact `{"rows": [...]}` body is `<= 8 KiB`, so tiny/empty results pay no envelope overhead
- bound memory with a 30-minute TTL, LRU eviction (monotonic access sequence), and 64-result / 128 MiB caps; results larger than the store cap return a preview-only shape with a warning
- map handle errors by variant: unknown/expired handles and invalid column projections are MCP invalid-params; store unavailability stays an internal error
- tool and resource descriptions keep the dynamic connected-sources/visible-counts behavior from #1218

## Why

Large query results dominate MCP tokens even when the agent only needs a preview, the schema, or a small slice. Returning a handle plus preview lets agents fetch only the rows and columns they need instead of paying for the full result or re-running broad SQL.

## Actual live benchmark

Command (using the `mcp-token-bench` harness from #1199, cherry-picked locally for measurement and intentionally not included in this PR):

`cargo run --locked -p xtask -- mcp-token-bench --coral-bin target/debug/coral --json`

Config: default live Coral config. Comparison: `main` (f00eb29) -> this PR. Two CCR cases were added to the harness: a config-independent 300-row `VALUES` query (`ccr.sql_many_rows`) and its suggested `result_get` follow-up page (`ccr.result_get_page`).

| Case | Model before | Model after | Model saved | Protocol before | Protocol after | Protocol saved |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| Full 22-case benchmark | 75,891 | 68,071 | 7,820 / 10.3% | 111,800 | 95,063 | 16,737 / 15.0% |
| `ccr.sql_many_rows` (300 rows) | 6,228 | 553 | 5,675 / 91.1% | 13,076 | 1,169 | 11,907 / 91.1% |
| `sql.github_pulls_columns` (100 rows) | 3,431 | 826 | 2,605 / 75.9% | 7,003 | 1,712 | 5,291 / 75.6% |
| `sql.wide_coral_tables` (20 rows) | 1,372 | 1,372 | 0 | 2,861 | 2,861 | 0 |
| `sql.narrow_many_tables` (50 rows) | 609 | 609 | 0 | 1,338 | 1,338 | 0 |
| `sql.empty_result` | 4 | 4 | 0 | 28 | 28 | 0 |
| `context.tools_list` | 2,761 | 3,132 | -371 | 2,761 | 3,132 | -371 |
| `context.resource_guide` | 1,582 | 1,671 | -89 | 1,681 | 1,771 | -89 |

New row with no `main` counterpart: `ccr.result_get_page` (50 rows) costs 1,119 model tokens. The realistic agent flow for the 300-row query — handled `sql` plus one `result_get` page — costs 1,672 model tokens vs 6,228 on `main` (-73%), with 70 rows of data and the full schema seen.

Notes:

- small and empty SQL results are byte-identical to `main`: the adaptive inline threshold avoids an envelope tax on the common tiny-result path
- `context.tools_list` grows by 371 tokens for the `result_get` tool schema, and the guide grows by 89 tokens for handle guidance; one large query recoups both
- all other catalog/discovery/resource rows are unchanged

## Actual output shape

Small results keep the existing contract:

```json
{"rows":[{"title":"Fix deploy","state":"open"}]}
```

Large results return a handle envelope instead of every row:

```json
{
  "result_id": "res_1",
  "row_count": 300,
  "column_count": 4,
  "columns": [
    {"name": "id", "data_type": "Int64", "is_nullable": false, "ordinal_position": 0},
    {"name": "name", "data_type": "Utf8", "is_nullable": true, "ordinal_position": 1}
  ],
  "preview": {"offset": 0, "limit": 20, "has_more": true, "next_offset": 20, "rows": ["..."]},
  "next_call": {"tool": "result_get", "arguments": {"result_id": "res_1", "offset": 20, "limit": 50}}
}
```

`result_get` returns one page with pagination metadata and optional column projection:

```json
{"result_id": "res_1", "row_count": 300, "offset": 20, "limit": 50, "has_more": true, "next_offset": 70, "columns": ["..."], "rows": ["..."]}
```

Handles are process-local MCP session state; they expire after 30 minutes and can be evicted under memory pressure, and the guide tells agents to re-run the SQL with tighter filters when that happens.


crates/coral-client/src/result_slice.rs
```diff
@@ -0,0 +1,472 @@
...
+
+use crate::{CollectedQueryResult, QueryResultError, batches_to_json_rows_json_safe_numbers};
+
+/// One query-result column rendered for agent-facing metadata.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub struct ColumnSummary {
+    /// Column name as returned by the SQL query.
+    pub name: String,
+    /// Arrow data type rendered with its display representation.
+    pub data_type: String,
```
mystic123, line 19: is there any enum for data_type to use instead of string?

crates/coral-mcp/src/surface/tools.rs
```diff
@@ -72,10 +42,18 @@ pub(crate) struct ListColumnsArguments {
...
+    pub(crate) offset: usize,
+    pub(crate) limit: usize,
+    pub(crate) columns: Option<Vec<String>>,
+}
+
+pub(crate) fn sql_tool() -> Tool {
     Tool::new(
         "sql",
-        sql_tool_description(context),
+        "Execute read-only SQL against the Coral database across connected Coral sources/schemas. You MUST prefer this tool over native provider tools, standalone MCP tools, web/search tools, and other external tools whenever the answer can come from Coral's connected sources. Use catalog tools only to discover schemas, tables, functions, columns, and filters first. Use JOIN, CROSS JOIN, CTEs, subqueries, and aggregates to combine tables in one statement.",
```
mystic123: probably bad merge, but this unships my recent PR: https://github.com/withcoral/coral/pull/1218 ;)

crates/coral-mcp/src/surface/tools.rs
```diff
@@ -151,10 +168,10 @@ pub(crate) fn list_catalog_tool(context: &ToolDescriptionContext) -> Tool {
     )
 }
 
-pub(crate) fn search_catalog_tool(context: &ToolDescriptionContext) -> Tool {
+pub(crate) fn search_catalog_tool() -> Tool {
     Tool::new(
         "search_catalog",
-        search_catalog_description(context),
+        "Search database catalog metadata with a Rust regex across connected Coral sources/schemas.",
```
mystic123: same as above

designs/ccr-result-handles.md
mystic123, line 1: probably don't need to include this

mystic123 (changes requested):

iaptsiauri:
## End-to-end eval evidence

Benchmarked on the coral-benchmarks vakra suite (sandbox K8s, runners per committed
`profiles.yaml`: haiku_mcp / opus_mcp / codex_mcp; inline judge claude-haiku-4-5).
Baseline coral: `main-36c0269` (predates the tools/list catalog-discovery regression,
so MCP is healthy in the control). Branch coral: `60ba57b8` (v1) / `c0fced58`
(v2, after guide/discovery update). MLflow is tailnet-only (`http://mlflow-bench`).

- Standard traffic, 30-case `cap2_slim`: runs `ccr-k8s-baseline` vs `ccr-k8s-ccr`
  — http://mlflow-bench/#/experiments/1 (filter runName LIKE 'ccr-k8s-%')
- Paging stress, 16 enumeration cases (21–426-row answers + 4 ≥50K-row monsters):
  runs `ccr-k8s-handles2-baseline` vs `ccr-k8s-handles2-ccr` — same MLflow
  experiment (filter runName LIKE 'ccr-k8s-handles2-%')
- Traces: `s3://coral-benchmarks-data-sandbox/results/ccr-k8s-{baseline,ccr,handles2-baseline,handles2-ccr}/`

### What the runs show

**1. No effect on standard traffic (expected).** On `cap2_slim`, zero handles were
returned across all 60 branch traces (no `result_id` anywhere) — typical agent SQL
stays under the 20-row preview. Accuracy/tokens/latency deltas all within noise
(8 discordant verdicts across 132 paired case-runs split 6/2, sign-test p ≈ 0.14).

**2. Handles engage and pay off where results are large.** On the 16-case paging
set, agents called `result_get` for real (counted from `tool_use` events in
traces, v2 build): haiku 47 calls / opus 22 / codex 3, uniformly `limit: 500`,
with column projection used (19/47 haiku calls). Versus baseline (full inline
row dumps):

| runner | judge accuracy | median tokens | p50 latency |
|---|---|---|---|
| haiku | same | 245K → 188K (**−23%**) | −5% |
| opus  | **+6 pts** | 217K → 209K (−4%) | **−15%** |
| codex | same | 245K → 125K (**−49%**) | −23% |

The opus gain is verified in the traces: it produced the complete 82-row answer
on `cap3_image_and_language_0006` only by paging through a handle (baseline
returned 79 of 82 rows). No case in either arm failed by answering from an
unpaged preview.

**3. Known cost: pathological enumerations.** On a ~52K-row answer
(`movie_platform_0009`) opus pages 18× (≈9K rows) and carries +12% total tokens
for the arm. Baseline burns comparable tokens receiving those rows inline; no
design answers a 52K-row enumeration well.

### Repro

- Branch arm: `bench coral-build --ref iaptsiauri/ccr-result-handles` →
  `bench deploy-branch vakra --coral-builder-tag <tag> --cases <16-case list> --run-id <id>`
- Baseline arm: `bench deploy vakra --image-tag main-36c0269-<bench-sha> --cases <same> --run-id <id>`
- 16-case list: vakra_cap3_restaurant_0005, cap2_olympics_0027, cap3_books_0011,
  cap3_image_and_language_0006, cap2_soccer_2016_0004, cap2_authors_0013,
  cap2_movie_platform_0022, cap2_hockey_0015, cap3_world_development_indicators_0016,
  cap2_olympics_0012, cap2_beer_factory_0029, cap2_mondial_geo_0007,
  cap2_authors_0015, cap2_movie_platform_0011, cap2_movie_platform_0009,
  cap2_authors_0016

mystic123 (approved):

AlbertQM (approved):

crates/coral-mcp/src/result_handles.rs
```diff
@@ -0,0 +1,302 @@
...
+
+    pub(crate) fn sql_value(&self, result: CollectedQueryResult) -> Result<Value, tonic::Status> {
+        if let Some(value) = inline_sql_value_if_small(&result)? {
+            return Ok(value);
+        }
+
+        // JSON object rows collapse duplicate column names and result_get
+        // projects columns by name, so duplicate-named results cannot be
+        // paged through a handle. Keep the legacy inline shape for them.
+        if has_duplicate_column_names(&result) {
```
sauldhernandez, line 122: This path bypasses the new result-handle policy for duplicate-column results and serializes every row inline. That defeats the large-result/token control for common joins like `SELECT a.id, b.id ...`, and it still loses data because JSON object rows collapse duplicate keys. The fallback should be bounded and explicit instead: require aliases, represent duplicate columns without object-key collision, or return a byte-bounded preview-only/error shape rather than dumping the full result.

crates/coral-mcp/src/result_handles.rs
```diff
@@ -0,0 +1,302 @@
...
+    }
+    let rows = batches_to_json_rows_json_safe_numbers(result.batches())
+        .map_err(|error| query_result_status(&error))?;
+    let rows_len = rows.len();
+    let value = serialize_tool_value(SqlRowsValue { rows })?;
+    let compact_len = serde_json::to_string(&value)
+        .map_err(|error| tonic::Status::internal(error.to_string()))?
+        .len();
+    let preview_would_include_full_result = result.row_count() <= SQL_PREVIEW_ROWS;
+    if preview_would_include_full_result || compact_len <= SQL_INLINE_MAX_BYTES {
```
sauldhernandez: This still allows oversized MCP responses. Any result with `<= SQL_PREVIEW_ROWS` rows is returned inline regardless of compact byte size, and the preview path below always serializes 20 full rows before the store-size fallback can help. One huge text/blob cell in the first page can produce a giant response and memory spike despite the `ResultStore` cap. Inline and preview serialization need their own byte bound/truncation policy.

crates/coral-mcp/src/surface/tools.rs
```diff
@@ -833,6 +942,79 @@ fn optional_bool_argument(
...
+    })?;
+    if values.is_empty() {
+        return Err(ErrorData::invalid_params(
+            format!("argument '{key}' must contain at least one column name"),
+            None,
+        ));
+    }
+    let mut out = Vec::with_capacity(values.len());
+    for value in values {
+        let column = value
```
sauldhernandez: High confidence: trimming projected column names changes the result contract. Arrow field names are matched exactly in `slice_result`, and quoted SQL aliases can legally contain leading/trailing spaces; `schema_summary` exposes those exact names. `columns: [" id"]` gets rewritten to `id`, which can select the wrong column or fail when only the spaced alias exists. Preserve exact string values and only reject strings that are entirely whitespace.

crates/coral-client/src/result_slice.rs
```diff
@@ -0,0 +1,529 @@
...
+        }
+    }
+    out
+}
+
+fn next_offset(row_count: usize, offset: usize, limit: usize) -> Option<usize> {
+    if offset >= row_count {
+        return None;
+    }
+    if limit == 0 {
```
sauldhernandez: Moderate confidence: for `limit == 0`, `next_offset` is the same offset, so `has_more` advertises a non-advancing pagination cursor. A client following the normal `has_more`/`next_offset` contract can keep requesting the identical metadata-only empty page forever. Either make metadata-only pages a distinct contract or avoid returning a normal next page token for `limit: 0`.

docs/guides/use-coral-over-mcp.mdx
```diff
@@ -240,7 +241,33 @@ Some sources expose native search endpoints (for example, GitHub issue search) a
...
 ## Query result format
 
-The `sql` tool returns rows as a JSON array of objects. To preserve exact values across JSON parsers, Coral encodes precision-sensitive numeric types as JSON **strings** rather than JSON numbers:
+For small results, the `sql` tool returns rows inline as a JSON array of objects:
+
+```json
+{"rows":[{"title":"Fix deploy","state":"open"}]}
+```
+
+For larger results, `sql` may return an in-memory result handle instead of every row. The response includes `result_id`, `row_count`, column metadata, a preview page, and a suggested `next_call`:
```
sauldhernandez, line 257: High confidence: this section says the handle response includes column metadata, but the JSON example omits the `columns` array. It also does not document the `preview_only` shape returned when the in-memory store rejects a too-large result. Since this is a new public MCP contract, the docs and tests should cover every response shape clients need to branch on.

crates/coral-mcp/src/result_store.rs
```diff
@@ -0,0 +1,420 @@
...
+
+use coral_client::{CollectedQueryResult, result_slice::result_estimated_bytes};
+
+const RESULT_ID_PREFIX: &str = "res_";
+const DEFAULT_RESULT_TTL: Duration = Duration::from_mins(30);
+const DEFAULT_MAX_RESULTS: usize = 64;
+const DEFAULT_MAX_BYTES: usize = 128 * 1024 * 1024;
+
+#[derive(Clone)]
+pub(crate) struct ResultStore {
```
sauldhernandez, line 24: Design comment: there is no real seam here for externalizing the cache store. `ResultHandles::new()` constructs a concrete in-memory `ResultStore`, the limits/constructors are private or test-only, and stored values are `Arc<CollectedQueryResult>` rather than serialized cache entries. If memory pressure or an external cache is needed in derived products, this needs an injected storage trait or an app-level serialized result contract; as written, swapping the store requires editing MCP internals.

sauldhernandez (changes requested):

crates/coral-mcp/src/result_store.rs
iaptsiauri: This is intentional as for now ResultStore is MCP local only, making this app-level requires much bigger changes like cross-client eviction rules and etc. I want to try this in one env only and see whether or not it is useful and if it is we can generalize it later

jamesaud:
Some interesting observations:

### There's 4 cases in error (this may effect other aggregate metrics)

http://mlflow-bench/#/experiments/1/runs/20d5db3654394caf851248fd9cd579c0/evaluations

### This case went on a loop increasing the offest

http://mlflow-bench/#/experiments/1/runs/20d5db3654394caf851248fd9cd579c0/evaluations?selectedEvaluationId=tr-59cd14d1551b8932a3a6ad5ef9549068
image

<br>

### This case Haiku started JQ-ing the file

http://mlflow-bench/#/experiments/1/runs/84964ee22ce74e75b6e3ef3013c783ff/evaluations?selectedEvaluationId=tr-e814cddeb8f0e19a3ddd8f375627a81c

image



<br>
<br>

### There's verbose instructions (guidance and result_preview: Error) I wonder if it is too specific in driving behavior? 

image

image



<br>

### The auto-researcher loop i ran discovered a similar feature, if there's anything useful to glean from the meta result  fields it decided to use. 
image

PR #1247, opened 2026-06-11
feat(sources/community/youtrack): add youtrack community source

ravindhar1108:
## Summary

Adds a new community source for YouTrack, enabling users to query issues, projects, users, comments, agile boards, sprints, and related metadata directly through Coral SQL.

### Tables Included

The source exposes 11 relational tables mapped from the YouTrack REST API:

* users
* projects
* issues
* comments
* tags
* agile_boards
* sprints
* groups
* roles
* saved_searches
* custom_fields

### Pagination

Implemented `offset_limit` pagination using YouTrack's native `$skip` and `$top` parameters, allowing Coral to retrieve datasets beyond the default API page size.

### Filter Pushdown

The `issues` table supports native YouTrack search query pushdown through the `query` parameter.

Example:

```sql
SELECT *
FROM youtrack.issues
WHERE query = '#Unresolved';
```

### Authentication

* Bearer token authentication
* Configurable base URL
* Supports both JetBrains-hosted and self-hosted YouTrack instances

### Compatibility

Migrated supported endpoints to the current YouTrack API where available (e.g. `/api/users` instead of legacy Hub endpoints) to maintain compatibility with YouTrack 2026.1+.

### Setup

```bash
export YOUTRACK_BASE_URL="https://yourinstance.myjetbrains.com/youtrack"
export YOUTRACK_TOKEN="perm:your_token_here"

coral source add youtrack --file sources/community/youtrack/manifest.yaml
```

#### Users

```sql
SELECT id, login, name, email
FROM youtrack.users
LIMIT 5;
```

```text
+-----+--------------+----------------+--------------------+
| id  | login        | name           | email              |
+-----+--------------+----------------+--------------------+
| 2-1 | <user_login> | <display_name> | <email>            |
| 2-0 | guest        | guest          |                    |
| 2-2 | <user_login> | <display_name> |                    |
+-----+--------------+----------------+--------------------+
```

#### Issues

```sql
SELECT id, readable_id, summary, created_at
FROM youtrack.issues
LIMIT 5;
```

```text
+------+----------------+-----------------------------------------------------+---------------+
| id   | readable_id    | summary                                             | created_at    |
+------+----------------+-----------------------------------------------------+---------------+
| 3-22 | demo_project-5 | Docs: Update README with setup instructions         | 1780228465000 |
| 3-21 | demo_project-4 | Security: Upgrade Jackson databind library          | 1780228464000 |
| 3-20 | demo_project-3 | Feature Request: Add caching to InventoryController | 1780228463000 |
| 3-19 | demo_project-2 | Bug: NullPointerException in PaymentController      | 1780228462000 |
+------+----------------+-----------------------------------------------------+---------------+
```

#### Saved Searches

```sql
SELECT id, name, query
FROM youtrack.saved_searches
LIMIT 5;
```

```text
+------+-----------------+-----------------------+
| id   | name            | query                 |
+------+-----------------+-----------------------+
| 11-0 | Assigned to me  | for: me               |
| 11-1 | Commented by me | commenter: me         |
| 11-2 | Reported by me  | by: me                |
| 11-3 | All issues      | document type: Issue  |
| 11-4 | All tickets     | document type: Ticket |
+------+-----------------+-----------------------+
```

These results were captured from a live YouTrack instance using Coral and validate successful authentication, data retrieval, pagination support, and table mappings.

## Notes

* `YOUTRACK_BASE_URL` should not include `/api`.
* `YOUTRACK_BASE_URL` should not end with a trailing slash.
* Authentication requires a permanent token generated from the YouTrack user profile.


jsummerfield (changes requested):
Thanks for the YouTrack source. The source is in the right community-source scope and the local static checks pass, but I found several current-head issues that need changes before approval.

Findings

- High: `sources/community/youtrack/README.md:61`-`79` and the PR body show example commands, but no captured successful `coral source add`, `coral source test`, or representative `coral sql` output from a real YouTrack instance. Mandatory live Coral evidence is blocking for community sources.

- Medium: `sources/community/youtrack/manifest.yaml:290`-`298` models `tags` through `/api/issueTags`. JetBrains marks `/api/issueTags` as deprecated and points callers to the current `/api/tags` resource. Please use `/tags` for this table and align the fields with `Tag`, not deprecated `IssueTag`. Docs: https://www.jetbrains.com/help/youtrack/devportal/resource-api-issueTags.html and https://www.jetbrains.com/help/youtrack/devportal/resource-api-tags.html

- Medium: `sources/community/youtrack/manifest.yaml:144`-`169` makes `youtrack.issues` an unfiltered, fully pageable table with no `fetch_limit_default`. YouTrack documents `/api/issues` as returning all matching issues when `query` is absent, sorted by updated desc and bounded only by server/export limits. For an issue tracker table, please add a conservative `fetch_limit_default` and make the guide push users toward `WHERE query = ...`. Docs: https://www.jetbrains.com/help/youtrack/devportal/resource-api-issues.html

- Medium: `sources/community/youtrack/README.md:21` says the token needs “read scopes,” but YouTrack permanent tokens are constrained by the user’s permissions, and this source exposes privileged access-management tables like `groups` and `roles`. Please name the practical permission/version constraints, especially that users/groups/roles are YouTrack REST API 2026.1+ and roles require permissions such as `Read Role`. Docs: https://www.jetbrains.com/help/youtrack/devportal/authentication-with-permanent-token.html, https://www.jetbrains.com/help/youtrack/devportal/api-users-yt-vs-hub.html, https://www.jetbrains.com/help/youtrack/devportal/resource-api-groups.html, and https://www.jetbrains.com/help/youtrack/devportal/resource-api-roles.html

The table set is promising, but the README needs to set a clearer first-success path and permission expectation.

ravindhar1108:
Thanks for the review @jsummerfield 

Addressed the review feedback:

* Added live Coral validation evidence to the README and PR description, including actual query results from a YouTrack instance.
* Updated the Token Setup section to document permission requirements and version constraints instead of referring to OAuth-style scopes.
* Migrated the `tags` table from the deprecated `/issueTags` endpoint to `/tags` and aligned the schema with the current Tag resource.
* Added a conservative `fetch_limit_default: 1000` to the `issues` table and updated documentation/examples to encourage use of the `query` filter for targeted issue retrieval.

jsummerfield (changes requested):
Findings:

- Medium: the new README live-output block still includes account-specific user data at `sources/community/youtrack/README.md:131`-`:138`, including a real-looking login/name row. Please sanitize captured output before this becomes checked-in product documentation, the same way we avoid committing real account, customer, or internal fixture data. Placeholder rows like `<user_login>`, `<display_name>`, and `<email>` are fine as long as the row counts and command output shape remain representative.

The previous blockers are otherwise moving in the right direction: the README/PR now include captured Coral output, the tag endpoint moved away from deprecated `/api/issueTags`, and `coral source lint sources/community/youtrack/manifest.yaml` passes on this head.

I did not approve the paused workflows because the source still needs the README output sanitized first.

ravindhar1108:
Thanks for the review. I've sanitized the captured README output to remove account-specific user information while preserving the representative output structure and row counts.

sources/community/youtrack/manifest.yaml
```diff
@@ -0,0 +1,610 @@
...
+  # ──────────────────────────────────────────────────────────────────────────
+  - name: custom_fields
+    description: Custom field settings defined globally in the instance.
+    request:
+      method: GET
+      path: /admin/customFieldSettings/customFields
+      query:
+        - name: fields
+          from: literal
+          value: "id,name,emptyFieldText"
```
jsummerfield: Could you fix this field selection? `/api/admin/customFieldSettings/customFields` returns `CustomField` objects, and `emptyFieldText` is not a top-level `CustomField` attribute. JetBrains documents that value under `fieldDefaults(emptyFieldText)` or the `/customFields/{fieldID}/fieldDefaults` subresource. As written, this column will either be empty or fail instead of exposing the intended value. Please request `fieldDefaults(emptyFieldText)` and read `fieldDefaults.emptyFieldText`, or drop this column if it is not needed.

Docs:
- https://www.jetbrains.com/help/youtrack/devportal/operations-api-admin-customFieldSettings-customFields.html
- https://www.jetbrains.com/help/youtrack/devportal/resource-api-admin-customFieldSettings-customFields-fieldID-fieldDefaults.html

sources/community/youtrack/README.md
```diff
@@ -0,0 +1,168 @@
...
+coral source add --file sources/community/youtrack/manifest.yaml
+```
+
+To rotate or update your token, run the same command again.
+
+### Token Setup
+
+Generate a permanent token in your YouTrack instance under **Profile** -> **Account Security** -> **Tokens**.
+
+Permanent tokens are constrained by the permissions of the user who created them rather than OAuth scopes. The user account generating the token must have access to the resources being queried.
```
jsummerfield: Could you make this setup guidance more precise? JetBrains documents permanent tokens as constrained by both selected token scope and the user's account permissions, but this currently says they are constrained by account permissions rather than scopes. The README should tell users which token scope is needed for the `/api/...` calls and should be concrete about the access-management tables. For example, `groups` requires access to the `Read Groups` feature, Update Project, or Low-level Admin Read plus group visibility; the current `roles` docs list Read permission in at least one scope, Update Organization, or Admin Read App, not a `Read Role` permission.

Docs:
- https://www.jetbrains.com/help/youtrack/devportal/authentication-with-permanent-token.html
- https://www.jetbrains.com/help/youtrack/devportal/Manage-Permanent-Token.html
- https://www.jetbrains.com/help/youtrack/devportal/resource-api-groups.html
- https://www.jetbrains.com/help/youtrack/devportal/resource-api-roles.html

sources/community/youtrack/manifest.yaml
```diff
@@ -0,0 +1,610 @@
...
+            - description
+
+  # ──────────────────────────────────────────────────────────────────────────
+  # issues
+  # ──────────────────────────────────────────────────────────────────────────
+  - name: issues
+    description: Issues in the YouTrack instance.
+    guide: |
+      By default, querying issues without a filter fetches a large, unconstrained list.
+      To avoid exhausting API limits and pulling excessive data, strongly prefer using the `query` filter with YouTrack's native search syntax (e.g., `WHERE query = '#Unresolved'`).
```
jsummerfield: Please wrap this guide line below the source lint limit. `make lint-sources` currently fails with `line too long (181 > 150 characters)` here.

jsummerfield (changes requested):
Thanks for the updates here. This is much closer: the live evidence is now present, the issues table has safer query guidance, and the tags endpoint moved to the documented `/api/tags` resource.

I found two remaining blockers on the current head before this can be approved: the custom-fields table is requesting a field from the wrong response shape, and the token/permission setup docs are still misleading for access-management tables. There is also one mechanical source-lint failure to fix.

ravindhar1108:
Thanks for the review! I've pushed updates addressing all of this:

- Mapped emptyFieldText correctly via fieldDefaults in the custom_fields table.
- Clarified the required token scopes and specific access requirements in the README.
- Wrapped the long lines to pass the lint-sources limit.

jsummerfield (changes requested):
Thanks for the quick follow-up here. The `custom_fields` shape is fixed, the source lint issue is fixed, and the manifest/table review is otherwise clean on this head.

I found one remaining setup-doc blocker:

Medium: `sources/community/youtrack/README.md` lines 23-25 still say the token must be granted the **YouTrack** scope. That is too narrow for the source surface being advertised here. JetBrains documents permanent tokens as limited by both selected token scope and account permissions, and built-in Hub installations expose separate YouTrack and YouTrack Administration scopes. This source includes core YouTrack tables plus admin/access-management surfaces like `/api/admin/projects`, `/api/admin/customFieldSettings/customFields`, `/api/groups`, and `/api/roles`, so the setup docs should tell users to select the scope(s) needed for the tables they plan to query and to use an account with the matching project/access-management permissions.

Please update the Token Setup section to avoid implying that the YouTrack scope alone is sufficient for every table. A good fix would explicitly distinguish standard issue/tag/comment/agile access from the admin/access-management tables, and mention that saved searches are only those visible to the authenticated user.

Docs I checked:
- https://www.jetbrains.com/help/youtrack/devportal/Manage-Permanent-Token.html
- https://www.jetbrains.com/help/youtrack/devportal/authentication-with-permanent-token.html
- https://www.jetbrains.com/help/youtrack/devportal/api-users-yt-vs-hub.html
- https://www.jetbrains.com/help/youtrack/devportal/resource-api-savedQueries.html

Thanks again. Once that wording is corrected, this looks close to approval.

ravindhar1108:
Thanks! I've pushed an update to fix the token scope docs. 
I rewrote the Token Setup section to explicitly distinguish between the standard YouTrack scope (for issues, comments, etc.) and the YouTrack Administration scope (for projects, users, etc.). 
I also added a note that saved_queries are limited to the authenticated user's visibility. 

Ready for another look!

jsummerfield (changes requested):
Findings:

- Medium: `sources/community/youtrack/manifest.yaml:33`-`:37` and `sources/community/youtrack/README.md:84`-`:122` still make first success depend on admin/access-management access. The token setup now correctly says users should select the scopes and permissions for the tables they plan to query, but `coral source test youtrack` always runs `users`, `projects`, and `groups`. JetBrains documents permanent tokens as limited by selected scope and account permissions, and the groups/roles resources are separately permission-gated ([permanent tokens](https://www.jetbrains.com/help/youtrack/devportal/Manage-Permanent-Token.html), [groups](https://www.jetbrains.com/help/youtrack/devportal/resource-api-groups.html), [roles](https://www.jetbrains.com/help/youtrack/devportal/resource-api-roles.html)). A user with a valid standard token for `issues`, `comments`, `tags`, or `agile_boards` can follow the quick start and still get a failing smoke test. Please either keep `test_queries` to standard-scope tables, or explicitly document that `coral source test` validates the broad/admin path and requires access-management permissions, then add a separate standard-scope first-success query path.

- Low: `sources/community/youtrack/manifest.yaml:203`, `:211`, `:276`, `:283`, `:428`, and `:436` expose YouTrack millisecond epoch values as raw `Int64` columns named `created_at`, `updated_at`, `start_date`, and `finish_date`. Coral has `format_timestamp` with `input: milliseconds`, and these names read like real temporal columns. Please expose them as `Timestamp` via `format_timestamp`, or rename them to make the raw millisecond unit explicit.

- Low: `sources/community/youtrack/manifest.yaml:233` says `comments.issue_id` must be the database ID, not the readable ID. YouTrack documents `/api/issues/{issueID}` as accepting either the database ID or the project-readable issue ID, and the comments resource uses that same path segment ([specific issue](https://www.jetbrains.com/help/youtrack/devportal/operations-api-issues.html), [issue comments](https://www.jetbrains.com/help/youtrack/devportal/resource-api-issues-issueID-comments.html)). If readable IDs work here, documenting that would make the table much easier to use; if comments is stricter than the parent issue resource, please call that out.

The scope docs are materially clearer now, `saved_searches` is consistent, live Coral evidence is present, and local static checks pass. I did not approve the paused workflows because the first-success path still needs the change above.

ravindhar1108:
Thanks for the review! I've updated the test queries to use standard-scope tables, converted the millisecond timestamps to use Timestamp with format_timestamp, and clarified the issue_id documentation.

jsummerfield (approved):
Thanks for the follow-up here. The current head resolves the remaining blockers from the previous review: the default test queries now use standard-scope tables, timestamp fields are exposed as Timestamp values, and comments document/use the readable issue ID path correctly.

I re-ran the source checks and a local mock YouTrack runtime pass on this exact head. The mock pass covered `coral source test youtrack`, timestamp conversion for issues/sprints/comments, `comments` with `issue_id = 'DEMO-1'`, `custom_fields.fieldDefaults.emptyFieldText`, and the expected bearer-auth YouTrack request paths. Validate and PR-title workflows are also green on this head.

The remaining README polish around putting standard-scope examples before admin examples is non-blocking from my side.

PR #1255, opened 2026-06-12
perf(episode): O(1) write-side index + per-workspace cap (BENCH-496)

jamesaud:
## BENCH-496 — PR-H1: hardening the episode store

First of the post-merge hardening PRs. Closes the deferred cubic **P2** on #1214 (the O(N) write-path scan) and adds the interim growth backstop the always-registered `OpenEpisode` route needs.

### The problem
`open_episode`'s idempotency/conflict check was an **O(N) scan + parse of the whole per-workspace log**, held under the **global** `state_lock`. So:
- populating *N* episodes is **O(N²)**, and
- opens **serialize across workspaces** (the lock is global; only the file path is per-workspace).

### The fix
Front the log with a **per-workspace in-memory index** (`episode_id -> { intent, parent }`), hydrated once from disk:
- idempotency/conflict check and the size cap are now **O(1)**;
- the per-workspace index lock makes check-then-append atomic **without blocking other workspaces**;
- the cross-process `state_lock` is held only **briefly around the append** — no scan.

Cloning the store shares the index (it's cloned per request); a cold store re-hydrates from the log; a poisoned lock self-heals (re-hydrate on next open) rather than bricking the store.

### Growth backstop
`MAX_EPISODES_PER_WORKSPACE = 100_000` → `CapacityExceeded` → `RESOURCE_EXHAUSTED`. The `OpenEpisode` route is always registered, so the raw log needs a ceiling until JSONL→Parquet compaction + retention land (PR-P1). Generous enough never to bite a real workspace during the experimental phase; becomes configurable under `[episodes]` then.

### Tests
- a **cold store hydrates** existing episodes from disk (idempotent reopen, conflict detection, no duplicate append) — proves the index reflects the persisted log, not just in-process writes;
- the **cap** rejects new ids past the ceiling while still allowing idempotent reopens;
- the cap is **per-workspace**;
- existing torn-write / idempotency / parent-lineage tests unchanged and green.

`make rust-checks` locally: fmt ✅ · clippy `-D warnings` ✅ · tests ✅ · rustdoc `-D warnings` ✅.

### DIFF
--- a/crates/coral-app/src/episode/service.rs
+++ b/crates/coral-app/src/episode/service.rs
@@ -83,12 +83,14 @@ fn non_blank(field: &str, value: &str) -> Result<String, Status> {
 }
 
 /// Maps an [`EpisodeStoreError`] to a gRPC [`Status`]. A reopen with a different
-/// intent/parent is a client-visible precondition failure; IO/serialization
-/// faults are internal and logged rather than surfaced.
+/// intent/parent is a client-visible precondition failure; a full workspace is
+/// resource-exhausted; IO/serialization faults are internal and logged rather than
+/// surfaced.
 fn open_episode_status(error: &EpisodeStoreError) -> Status {
     match error {
         EpisodeStoreError::Conflict { .. } => Status::failed_precondition(error.to_string()),
         EpisodeStoreError::InvalidIntent { .. } => Status::invalid_argument(error.to_string()),
+        EpisodeStoreError::CapacityExceeded { .. } => Status::resource_exhausted(error.to_string()),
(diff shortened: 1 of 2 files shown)


jamesaud:
Superseded by #1261. We decided to keep the episode store **disk-only** for now — the in-memory write-side index here was premature, since the read-side index/representation should be co-designed with retrieval (it may end up compacted on disk rather than in RAM). #1261 instead bounds the log by a byte ceiling with oldest-out eviction (a byte-bounded FIFO), which is what we actually need at this stage. The O(N²)-scan fix returns later alongside the indexing work.

PR #1257, opened 2026-06-12
feat(engine): plan source functions as logical nodes

Bradley-Butcher:
## Why

Source-scoped table functions (`github.pulls(owner => 'x')`) are planned today by rewriting the call into a hidden, name-mangled DataFusion UDTF registered in the flat UDTF namespace. That design has two structural problems:

1. **Wrong phase for parameters.** `TableFunctionImpl::call` runs while DataFusion converts SQL into a logical plan — before `with_param_values` substitutes `$name` placeholders. Parameterized source-function calls (`owner => $owner`) are therefore impossible. The upcoming recipe work needs exactly that (a recipe's declared arguments bind into its stored SQL); without this change, recipes would have to own their own SQL-rewriting layer to splice values in as literals.
2. **A shadow callable surface.** Every function has a hidden globally-registered twin (`__coral_udtf_<hex>_<hex>`) that exists only as a rewrite target, yet is technically callable by anyone who discovers the name.

## What

- The relation planner still intercepts and validates the call, but now parks it as a `SourceFunctionNode` logical-plan extension carrying the argument expressions (positional, manifest order, NULL = absent) and the manifest result schema.
- A new analyzer rule (`coral_source_functions`) resolves each node into an ordinary provider `TableScan` before the optimizer runs — so plan shape, projection/limit pushdown, and execution are exactly what the UDTF path produced. `SourceFunctionRegistry::install()` registers planner and rule as a pair.
- Backends implement a Coral-owned `SourceFunctionProviderFactory` trait (previously DataFusion's `TableFunctionImpl`); the factory travels inside `RegisteredTableFunction`, and the hidden-UDTF machinery — name mangling, `register_udtf`, the separate factory map — is deleted.

## Behavior

Unchanged for every existing query. Fully-literal calls validate eagerly at relation-planning time, so all argument errors (unknown function/argument, duplicates, wildcards, SETTINGS, missing required, allowed-values) keep their existing messages **and** their existing phase. The pre-existing engine suite passes unmodified, except three tests that invoked the mangled internal name directly; one of those is repurposed into a negative test pinning that the internal surface no longer exists.

Deliberate, visible changes:
- A stray `$name` placeholder in a call argument now fails with `github.pulls argument 'q' is bound to parameter $q, but no value was provided for it` (previously a confusing `must be a literal`). No surface can supply parameter values yet; the binder arrives with the recipe stack, where the node's `expressions()` / `with_exprs_and_inputs()` contract lets `with_param_values` reach the arguments.
- Unoptimized EXPLAIN shows `SourceFunction: github.pulls`; optimized and physical plans are unchanged in shape, with table references reading `github.pulls` instead of the mangled hex name.
- Mangled internal names are no longer callable.
- Unsupported relation modifiers on function calls (`WITH ORDINALITY`, `TABLESAMPLE`, table/index hints, time-travel, PARTITION, JSON path) now fail with clear errors instead of executing while silently ignoring the modifier — main inherits DataFusion's silent-ignore for these (verified empirically; plain tables still behave that way engine-wide). The rejection destructures every `TableFactor::Table` field without `..`, so a sqlparser upgrade that adds a modifier fails compilation here and forces a decision.
- Qualified column references on function results now resolve like table columns: `SELECT github.pulls.id FROM github.pulls(...)` and `SELECT pulls.id FROM ...` both work. Previously the scan was named by the mangled identifier, so no user-writable qualifier could ever match (the unknown-column hint suggested quoting the mangled name). The scan now carries `TableReference::partial(schema, function)`.
- `DESCRIBE SELECT * FROM github.pulls(owner => $owner)` with an unbound placeholder now returns the result schema instead of failing at planning. The parked node carries the manifest-declared schema, so plan-time schema introspection works without argument values — the same operation recipe schema inference will perform.

Coverage note: the analyzer resolves nodes through subquery plans (`transform_up_with_subqueries`) — source-function calls inside scalar/EXISTS/IN subqueries are pinned by dedicated tests, since subqueries live in expressions and a plain plan walk misses them.

## Validation

- `make rust-checks` passes (1181 tests).
- `make perf-check` passes: `coral.tables` mean 196 ms against the 750 ms threshold (source registration no longer touches the UDTF registry).
- Exercised against a real workspace (github + linear sources): catalog listing, literal function calls returning live API results, alias usage, and the new unbound-placeholder error.

### DIFF
--- a/crates/coral-engine/src/backends/mod.rs
+++ b/crates/coral-engine/src/backends/mod.rs
@@ -11,7 +11,7 @@
 //! |---|---|---|
 //! | `mod.rs` | Module entry. `CompiledBackendSource` impl, `compile_source` / `compile_manifest`, internal module declarations. | always |
 //! | `provider.rs` | `DataFusion` `TableProvider` implementation. | if the backend exposes tables |
-//! | `function.rs` | `DataFusion` `TableFunctionImpl` for source-scoped UDTFs. | only if the backend exposes table functions |
+//! | `function.rs` | Source-function provider factory and `build_registered_table_function` wiring. | only if the backend exposes table functions |
 //! | `client.rs` | Configured stateful wrapper (the value the rest of the backend talks to) and any transport-abstracting trait. | if the backend has a per-source client; skip if config is per-table (file backend) |
 //! | `transport.rs` | Per-instance transport impls (HTTP requests, stdio child spawn, object-store wiring, ...). | if there are multiple transports or transport code is non-trivial |
 //! | `response.rs` | Decode one response from the backend into the JSON payload that `shared/response_rows::extract_rows` consumes. | if response decoding is non-trivial |
@@ -81,8 +81,8 @@ mod composite;
 pub(crate) use common::{
     BackendCompileRequest, BackendRegistration, BackendRegistrationContext,
     BackendSchemaRegistration, CompiledBackendSource, RegisteredInput, RegisteredSource,
-    RegisteredTable, RegisteredTableFunction, SourceTableFunctions, build_registered_inputs,
-    build_registered_table, build_registered_table_function, internal_table_function_name,
+    RegisteredTable, RegisteredTableFunction, SourceFunctionProviderFactory,
+    build_registered_inputs, build_registered_table, build_registered_table_function,
     registered_columns_from_schema, registered_columns_from_specs, required_filter_names,
     schema_from_columns, validate_lookup_key_filter_backend_support,
 };
--- a/crates/coral-engine/src/backends/common.rs
+++ b/crates/coral-engine/src/backends/common.rs
@@ -1,7 +1,6 @@
 //! Shared internal backend contracts and registry-visible metadata.
 
 use std::collections::{BTreeMap, BTreeSet, HashMap};
-use std::fmt::Write;
 use std::sync::{Arc, OnceLock};
 
 use crate::{QueryRuntimeContext, QuerySource, RequestAuthenticator, SourceInputResolver};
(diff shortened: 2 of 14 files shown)


Bradley-Butcher:
* **#1257** <a href="https://app.graphite.com/github/pr/withcoral/coral/1257?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1263](https://github.com/withcoral/coral/pull/1263) <a href="https://app.graphite.com/github/pr/withcoral/coral/1263?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1298](https://github.com/withcoral/coral/pull/1298) <a href="https://app.graphite.com/github/pr/withcoral/coral/1298?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1257?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-engine/src/backends/mod.rs
```diff
@@ -80,10 +80,10 @@ pub(crate) mod common;
 mod composite;
 pub(crate) use common::{
     BackendCompileRequest, BackendRegistration, BackendRegistrationContext, CompiledBackendSource,
-    RegisteredSource, RegisteredTable, RegisteredTableFunction, SourceTableFunctions,
+    RegisteredSource, RegisteredTable, RegisteredTableFunction, SourceFunctionProviderFactory,
     build_registered_inputs, build_registered_table, build_registered_table_function,
```
sauldhernandez, line 85: This export makes `SourceFunctionProviderFactory` the backend-facing source-function abstraction, but the module-layout contract at the top of this file still tells backend authors that `function.rs` should implement DataFusion `TableFunctionImpl` source-scoped UDTFs. That guidance now points at the deleted hidden-UDTF design and will steer the next backend toward the wrong ownership boundary. Please update the table to describe the Coral provider factory / `build_registered_table_function` path instead.

sauldhernandez (approved):
Minor doc update required, otherwise LGTM

Bradley-Butcher:
Removed big jim as a reviewer as runtime files are no longer touched (merge issue)

crates/coral-engine/src/backends/mod.rs
Bradley-Butcher: Done. The backend module-layout table now describes `function.rs` as the source-function provider factory plus `build_registered_table_function` wiring, so it points backend authors at the current Coral provider-factory path instead of the old hidden-UDTF design.

PR #1261, opened 2026-06-15
feat(episode): bound the per-workspace log by bytes with oldest-out eviction (BENCH-496)

jamesaud:
## BENCH-496 — bound the raw episode log

The `OpenEpisode` route is always-registered (and callable today via the generated `EpisodeServiceClient`), so the per-workspace `episodes.jsonl` can grow without bound. This bounds it.

### Design: byte ceiling + oldest-out eviction (byte-bounded FIFO)
- A per-workspace **byte ceiling** (default **256 MiB**). When a new record would push the log over, the **oldest records are evicted** to make room — the log keeps the most recent episodes within budget.
- The **newest record is always kept** (even if it alone exceeds the ceiling); the store **never rejects** an open.
- **No time-based retention, no background job.** The idempotency/conflict scan already reads the whole log on every open, so eviction folds into the same pass.

### Why this shape (decisions captured in-thread)
- **Disk-only, no in-memory index.** The write-side index (the O(N²)-scan fix) was deliberately dropped: the read-side index/representation should be co-designed with retrieval (it may live compacted on disk, not in RAM), so committing to an in-memory shape now is premature. At alpha volume the O(N) scan is fine.
- **Bytes, not time.** Time-based age-out would drop episode history by clock; a byte-FIFO keeps the most recent N MiB regardless of age, which is what "bound disk, keep recent" wants.
- **Shares the right layer.** Reuses `storage_fs` (`write_atomic` for the eviction rewrite, `append_file_private` for the common append, `FileLock`) — the same low-level layer telemetry/feedback use. The trace store's higher-level `RollingJsonlWriter`/time-prune is intentionally *not* reused: episodes are read-modify-write + byte-FIFO + single-file, vs traces' write-only + time-pruned + rolling. (That machinery is a natural extraction later for the write-only `trajectory_steps.jsonl` capture.)

### Cost
The common under-budget open is a cheap append. Once a workspace sits at the ceiling, each new open rewrites the surviving log (O(N)) — acceptable at alpha volume; the real fix is compaction (PR-P1).

### Tests
- evicts oldest to stay under the ceiling; newest kept even when it alone exceeds; an evicted id is reusable (not a conflict); surviving records keep idempotency/conflict semantics. Existing torn-write / parent-lineage / idempotency tests unchanged.

`make rust-checks` locally: fmt ✅ · clippy `-D warnings` ✅ · tests ✅ · rustdoc `-D warnings` ✅.

### DIFF
--- a/crates/coral-app/src/episode/store.rs
+++ b/crates/coral-app/src/episode/store.rs
@@ -233,9 +233,15 @@ fn append_within_budget(
 ) -> Result<(), EpisodeStoreError> {
     let line_len = serde_json::to_vec(&record)?.len() as u64 + 1;
     // Fast path: it fits on top of what's already on disk. `file_len` counts physical
-    // bytes (so any torn-record bytes are included conservatively), avoiding a rewrite
-    // in the common case.
-    if file_len(path)?.saturating_add(line_len) <= max_bytes {
+    // bytes (so any torn-record bytes are included conservatively); add the separator
+    // newline `append_record` will prepend when the file lacks a trailing one, so a
+    // torn-tailed file at the boundary doesn't slip one byte over the ceiling.
+    let leading = u64::from(file_needs_leading_newline(path)?);
+    if file_len(path)?
+        .saturating_add(line_len)
+        .saturating_add(leading)
+        <= max_bytes
+    {
         return append_record(path, &record);
     }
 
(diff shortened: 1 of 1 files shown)


crates/coral-app/src/episode/store.rs
```diff
@@ -156,46 +184,130 @@ pub(crate) fn now_unix_nanos() -> u128 {
...
+    path: &Path,
+    existing: Vec<PersistedEpisode>,
+    record: PersistedEpisode,
+    max_bytes: u64,
+) -> Result<(), EpisodeStoreError> {
+    let line_len = serde_json::to_vec(&record)?.len() as u64 + 1;
+    // Fast path: it fits on top of what's already on disk. `file_len` counts physical
+    // bytes (so any torn-record bytes are included conservatively), avoiding a rewrite
+    // in the common case.
+    if file_len(path)?.saturating_add(line_len) <= max_bytes {
```
jamesaud: This fast path undercounts the bytes that `append_record` may write when the existing file has no trailing newline. In that case `append_record` prepends a separator newline before the new JSONL record, so a file where `file_len(path)? + line_len == max_bytes` still takes this branch and ends up one byte over the configured ceiling. Torn tails are the case this code is already preserving, so the budget check should include `file_needs_leading_newline(path)? as u64` (or send newline-less files through the rewrite path).

jamesaud: Good catch — fixed in `72940f3f`. The fast path now adds `u64::from(file_needs_leading_newline(path)?)` to the budget, so a torn-tailed file sitting at `file_len + line_len == max_bytes` is routed to the rewrite path (which produces a clean, newline-terminated log) instead of appending one byte over. Added a regression test (`byte_ceiling_holds_when_the_log_has_a_torn_tail`) that seeds a no-trailing-newline record exactly at the boundary and asserts the ceiling holds — it fails on the old accounting and passes now.

crates/coral-app/src/episode/store.rs
```diff
@@ -134,18 +167,13 @@ impl EpisodeStore {
...
-        // skippable instead of merging with (and corrupting) this record.
-        if file_needs_leading_newline(&path)? {
-            bytes.push(b'\n');
-        }
-        bytes.extend_from_slice(&record);
-        bytes.push(b'\n');
-        storage_fs::append_file_private(&path, &bytes)?;
-        Ok(())
+
+        append_within_budget(
```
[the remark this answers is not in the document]
jamesaud: Intentional tradeoff — documented in `7ce18fb1` rather than changed.

This is the inherent tension between the BENCH-496 conflict contract and bounding the log: to reject a reused id *after* its record is gone we'd have to remember every id forever (an ever-growing id-set), which defeats the byte ceiling this PR adds. So eviction necessarily forgets the oldest ids.

It's safe in practice, and the guarantee still holds where it matters:
- The conflict check is authoritative for every **retained** record. Eviction only drops the *oldest* records at the 256 MiB ceiling; a client retry — the case the conflict check exists for — always carries a recent id that is never the one evicted.
- Episode ids are **client-minted and unique per attempt** (ULID/UUID), so "reopen an evicted id with different metadata" doesn't occur for real callers.

I scoped the `open_episode` doc to say the conflict guarantee covers retained records and why eviction is safe given id uniqueness. Happy to revisit if we ever move to non-unique/caller-supplied ids.

sauldhernandez (approved):

jamesaud:
### Merge activity

* **Jun 17, 2:57 PM UTC**: @jamesaud merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1261).

PR #1262, opened 2026-06-15
feat(mcp): add open_episode surface

jamesaud:
This introduces a flag-gated MCP tool for Episodes, a building block for the learning layer work. An example of a feature being built on this is [trajectory memory](withcoral.dsp.so/KsbA34bE-rfc-trajectory-memory-visual).

## Summary

- Adds feature-gated MCP `open_episode` tool backed by `EpisodeService.OpenEpisode`.
- Lets agents carry `episode_id` explicitly on follow-up Coral MCP tools; MCP forwards it as `coral-episode-id` metadata and records `episode.id` on MCP call spans.
- Wires episodes feature resolution through `coral mcp-stdio` and adds client `EpisodeClient` support.

## Validation

- `cargo fmt --check`
- `cargo test -p coral-mcp`
- `cargo test -p coral-cli --features cli-test-server --test mcp`
- `cargo clippy -p coral-mcp --all-targets --all-features --locked -- -D warnings`


crates/coral-mcp/src/surface/tools.rs
```diff
@@ -210,24 +236,27 @@ pub(crate) fn search_catalog_tool(context: &ToolDescriptionContext) -> Tool {
     )
 }
 
-pub(crate) fn describe_table_tool() -> Tool {
+pub(crate) fn describe_table_tool(episodes_enabled: bool) -> Tool {
```
jamesaud: **Inconsistent signature for one flag.** `describe_table_tool`/`list_columns_tool` take a bare `episodes_enabled: bool`, while `sql_tool`/`list_catalog_tool`/`search_catalog_tool` take `&ToolDescriptionContext` — which already carries `episodes_enabled` (set via `with_episodes_enabled`). The split surfaces at the `list_tools` call site:

```rust
sql_tool(&tool_context),
list_catalog_tool(&tool_context),
search_catalog_tool(&tool_context),
describe_table_tool(self.options.episodes_enabled),  // reaches past the context
list_columns_tool(self.options.episodes_enabled),
```

Giving these two a `&ToolDescriptionContext` param too (even though they ignore the counts) lets all five share one shape and keeps the flag's source of truth in the context.

crates/coral-mcp/src/surface/tools.rs
```diff
@@ -440,6 +554,65 @@ fn connected_source_names_text(source_names: &[String]) -> Option<String> {
...
 }
 
+fn tool_input_schema(mut value: Value, episodes_enabled: bool) -> Arc<Map<String, Value>> {
+    if episodes_enabled {
+        add_episode_id_property(&mut value);
+    }
+    json_object_schema(&value)
+}
+
+fn add_episode_id_property(value: &mut Value) {
```
jamesaud: **Duplicated schema fragment.** The nullable episode-id shape —

```json
{ "anyOf": [ { "type": "string", "maxLength": CORAL_EPISODE_ID_MAX_LEN }, { "type": "null" } ] }
```

— is repeated three times: here in `add_episode_id_property`, in `create_episode_tool`'s `parent_episode_id`, and in `create_episode_output_schema`'s `parent_episode_id`. A small helper like `fn nullable_episode_id_schema() -> Value` would keep the shape and the max-length constraint in one place and prevent drift if either changes.

crates/coral-mcp/src/server.rs
```diff
@@ -342,31 +430,53 @@ impl CoralMcpServer {
...
                     .await
             }
             "list_columns" => {
-                self.list_columns_tool_result(request.arguments.as_ref())
+                self.list_columns_tool_result(request.arguments.as_ref(), context)
                     .await
             }
+            "create_episode" if self.options.episodes_enabled => {
+                let arguments =
+                    crate::surface::create_episode_arguments(request.arguments.as_ref())?;
```
jamesaud: Minor: `create_episode_arguments` (here) and `create_episode_tool` (in `list_tools`) are referenced fully-qualified as `crate::surface::...`, but they're already re-exported from `surface` and every sibling (`required_string_argument`, `optional_episode_id_argument`, `feedback_tool`, …) is pulled into the `use crate::surface::{…}` block at the top of the file. Adding these two there keeps the imports consistent.

jamesaud (commented):
.

crates/coral-mcp/src/server.rs
```diff
@@ -76,12 +91,58 @@ impl ToolCallOutcome {
     }
 }
 
+#[derive(Debug, Default)]
+struct ToolCallContext {
```
jamesaud: **Naming — the struct over-promises slightly.** `ToolCallContext` currently holds a single field, the episode tag. The generic "context" name reads like it carries broader per-call state than it does. If more per-call data is genuinely coming (the trajectory-memory roadmap suggests it is), the name is appropriately forward-looking and this is fine as-is. If episodes stays the only thing that needs per-call context, something like `EpisodeTag` would describe it more honestly and avoid wrapping a generic abstraction around one `Option`. Just flagging the judgment call — not a blocker.

crates/coral-mcp/src/server.rs
```diff
@@ -76,12 +91,58 @@ impl ToolCallOutcome {
...
+            return Ok(Self::default());
+        }
+        let episode_id = optional_episode_id_argument(arguments, "episode_id")?;
+        if let Some(episode_id) = episode_id.as_deref() {
+            telemetry::record_episode_id(span, episode_id);
+        }
+        let episode_id_metadata = episode_id
+            .as_deref()
+            .map(|episode_id| {
+                episode_id.parse().map_err(|error| {
```
jamesaud, line 111: **This error branch is effectively unreachable.** `optional_episode_id_argument` (called just above) already rejects anything that isn't ASCII-graphic and within `CORAL_EPISODE_ID_MAX_LEN`, and ASCII-graphic (`0x21..=0x7E`) is a *stricter* set than what `MetadataValue<Ascii>::parse` accepts (which also permits space, `0x20`). So by the time we reach `.parse()` it can't fail, and the `"is not valid metadata"` arm is dead in practice. It's harmless defensiveness, but the same string is now validated twice with the second check unreachable. If you want to DRY it, you could treat the `MetadataValue` parse as the single source of truth and slim down `optional_episode_id_argument` accordingly — though keeping both is a defensible belt-and-suspenders choice too.

crates/coral-mcp/src/server.rs
```diff
@@ -76,12 +91,58 @@ impl ToolCallOutcome {
...
+        options: &McpOptions,
+        arguments: Option<&Map<String, Value>>,
+        span: &tracing::Span,
+    ) -> Result<Self, ErrorData> {
+        if !options.episodes_enabled {
+            return Ok(Self::default());
+        }
+        let episode_id = optional_episode_id_argument(arguments, "episode_id")?;
+        if let Some(episode_id) = episode_id.as_deref() {
+            telemetry::record_episode_id(span, episode_id);
```
jamesaud: **Minor: hidden side effect in a `from_*` constructor.** `from_tool_request` both builds the context and writes `episode.id` to the span. A constructor that mutates a span is a little surprising to someone scanning for where telemetry gets recorded. Since `call_tool` already has both the parsed context and the span in hand, you could record there and keep this function pure. Worth weighing against the fact that the symmetric write now also lives in `create_episode_value` — so moving both to their call sites would be the consistent version of this. Stylistic, not blocking.

jamesaud (commented):
.

crates/coral-mcp/src/server.rs
jamesaud: Done in a391db90: `create_episode_arguments` and `create_episode_tool` are now imported through the existing `crate::surface::{...}` list and the call sites use the unqualified names. This is still present at the current PR head 25fc6ad6.

crates/coral-mcp/src/server.rs
jamesaud: Done in 25fc6ad6: renamed `ToolCallContext` to `EpisodeTag`, since the type is intentionally episode-specific right now and only owns the optional episode id plus pre-parsed metadata value.

crates/coral-mcp/src/server.rs
jamesaud: Done in 25fc6ad6: `EpisodeTag::from_tool_request` is now pure. Existing episode ids are recorded in `call_tool`, and the newly created episode id is recorded in the `create_episode` dispatch path after persistence succeeds.

crates/coral-mcp/src/server.rs
jamesaud: Left this as a defensive conversion rather than relaxing validation to gRPC metadata rules. The MCP surface deliberately rejects spaces/control bytes via `optional_episode_id_argument`, even though `MetadataValue<Ascii>` would accept some broader input. The parse branch is still useful as a typed boundary check if those validation rules drift later.

crates/coral-mcp/src/surface/tools.rs
jamesaud: Done in a391db90: the repeated nullable episode-id schema is now centralized behind `nullable_episode_id_schema(...)`, and the input/output schema sites reuse it with optional descriptions where needed.

crates/coral-mcp/src/surface/tools.rs
jamesaud: Done in a391db90: `describe_table_tool` and `list_columns_tool` now take `&ToolDescriptionContext`, and `list_tools` passes the same `tool_context` to all five catalog/query tool builders.

jamesaud: Done in b669c89c: removed the episode-related `.mdx` additions from this draft PR. `git diff origin/main...HEAD -- docs` is now empty, so the PR keeps the MCP surface/code/tests but does not add episodes documentation yet.

mystic123 (approved):

crates/coral-mcp/src/surface/tools.rs
```diff
@@ -440,6 +545,65 @@ fn connected_source_names_text(source_names: &[String]) -> Option<String> {
...
+            nullable_episode_id_schema(Some(EPISODE_ID_ARGUMENT_DESCRIPTION)),
+        );
+}
+
+fn nullable_episode_id_schema(description: Option<&str>) -> Value {
+    let mut schema = json!({
+        "anyOf": [
+            {
+                "type": "string",
+                "maxLength": CORAL_EPISODE_ID_MAX_LEN
```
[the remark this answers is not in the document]
jamesaud: Fixed in f29fcb05. Cubic found that the episode-id JSON schema did not advertise the same constraints enforced by runtime validation; `episode_id_string_schema()` now includes `minLength: 1`, `maxLength: CORAL_EPISODE_ID_MAX_LEN`, and `pattern: "^[!-~]+$"`, and the nullable episode-id schemas reuse it. I also added test assertions for advertised `episode_id` and `parent_episode_id` schemas. Validation: `cargo fmt --check`, `cargo test -p coral-mcp`, `cargo clippy -p coral-mcp --all-targets --all-features --locked -- -D warnings`.

crates/coral-mcp/src/server.rs
```diff
@@ -342,31 +435,56 @@ impl CoralMcpServer {
...
+                    Err(status) if status.code() == tonic::Code::InvalidArgument => {
+                        Err(status_to_error_data(&status))
+                    }
+                    Err(status) => Ok(ToolCallOutcome::ToolError {
+                        operation: "Episode opening",
+                        status,
+                    }),
+                }
+            }
             "feedback" if self.options.feedback_enabled => {
```
jsummerfield, line 476: Is it intentional that `feedback` is excluded from episode attribution? I expect it would yield valuable information at the end of a failed episode.

jamesaud: That's fair, added in 09cfd343: the feedback tool now receives the episode-aware tool context, advertises optional `episode_id` when episodes are enabled, and submits feedback through `EpisodeTag::request(...)` so `coral-episode-id` metadata is forwarded like the other episode-attributed MCP tools.

Validated with:

- `cargo fmt --check`
- `cargo test -p coral-mcp`
- `cargo test -p coral-cli --features cli-test-server --test mcp`
- `cargo clippy -p coral-mcp --all-targets --all-features --locked -- -D warnings`

jsummerfield:
The MCP side forwards coral-episode-id on all four catalog tools (list_catalog/search_catalog/describe_table/list_columns), but the backend drops it on every one of them. How come?

jamesaud:
Thank you for identifying that. I had manually tested end to end but just looked at the `execute_sql` tool. It is now fixed and have added a regression test.   
  
\--

The backend catalog path now extracts coral-episode-id in CatalogService, threads it through CatalogDiscovery into QueryManager, and stamps [episode.id](http://episode.id) on the backend coral.query spans for the catalog-backed operations.

Added regression coverage: query::manager::tests::catalog_service_stamps_episode_id_on_query_spans, which calls all four catalog gRPC methods with episode metadata and asserts the backend query spans carry [episode.id](http://episode.id).

jsummerfield:
The feature is correct and well-tested. But episode attribution is now threaded as an explicit parameter through **42 functions across 9 files (19 of them pure forwards-only signatures)**, and the recent _"preserve catalog attribution"_ commit shows it still spreading (a `&QueryAttribution` param added through 5 more `CatalogDiscovery` methods, plus `episode_id_from_metadata` copy-pasted at more service edges). Worth containing now — before a third request-scoped concern (tenant, auth principal, request id) repeats the same spread.

**Core observation:** the `coral-episode-id` metadata header _is_ the propagation channel. Between extraction at the MCP edge and use at the gRPC edge, the id already lives in the metadata map — `EpisodeTag`, `QueryAttribution`, and the `&`\-threaded params are re-carrying a typed copy of it. We already handle the _same class_ of concern (request-scoped context → spans + across the service boundary) cleanly for **trace context**, and episodes should ride the same rails rather than inventing a parallel hand-threaded channel:

- **Outbound:** `TraceContextInterceptor` (`coral-client/src/propagation.rs:41`) injects `traceparent` on every call from ambient `Span::current()` — zero threading.
- **Inbound:** `GrpcMethodAnnotatedService` (`coral-app/src/transport.rs:66`) extracts request info into `request.extensions_mut()` once (`:93`), read back via `grpc_method()` (`:170`).

Three independent follow-ups, in increasing risk:

### 1\. App-inbound extraction → one tower layer (lowest risk; do first)

`episode_id_from_metadata(request.metadata())` is copy-pasted at **6 service entry points** (`catalog/service.rs:47,90,139,169` + `query/service.rs:42,72`). Extend the existing `GrpcMethodAnnotatedService` to parse the header into an `Option<EpisodeId>` extension once:

```rust
// in GrpcMethodAnnotatedService::call, next to the GrpcServerMethod insert (transport.rs:92-93)
if let Some(ep) = request.headers()
    .get(CORAL_EPISODE_ID_METADATA_KEY)
    .and_then(|v| v.to_str().ok())
    .and_then(|s| EpisodeId::parse(s).ok())
{ request.extensions_mut().insert(ep); }
```

Each handler then reads `QueryAttribution::from_extensions(req.extensions())` instead of re-extracting. The best-effort semantics (missing → `None`, malformed → ignored) move into the layer unchanged. This is a copy of an idiom already in the file, and it directly undoes the catalog-fix churn.

### 2\. Schema decoration → collection decorator (low risk)

`episodes_enabled` is threaded through `ToolDescriptionContext` (`tools.rs:20`, `with_episodes_enabled:41`) into all six `*_tool()` builders (`:108,165,227,259,316,352`). The mutation is uniform, so decorate the built list once at `list_tools` instead of passing the flag into each builder:

```rust
let mut tools = vec![sql_tool(&ctx), list_catalog_tool(&ctx), /* … */];
if self.options.episodes_enabled {
    tools = tools.into_iter().map(with_episode_id_argument).collect();
    tools.push(open_episode_tool());   // minted, intentionally NOT decorated
}
```

Drops the bool from the context + all six builders, and makes the `open_episode` exclusion explicit instead of implicit-in-which-builder-takes-the-flag. One gotcha: `Tool.input_schema` is `Arc<Map<…>>`, so `with_episode_id_argument` must `Arc::make_mut`/rebuild (reuse the body of `add_episode_id_property`). Bonus: it becomes unit-testable on a single `Tool`.

### 3\. MCP-outbound threading → task-local + interceptor (medium risk; the headline win)

`EpisodeTag` is threaded as `&EpisodeTag` through 7 methods (`dispatch_tool`, `load_table_description`, `submit_feedback_value`, and the four `*_tool_result` methods) purely to call `EpisodeTag::request` (`server.rs:132`). Mirror `TraceContextInterceptor`: bind the id once where it's already extracted (`call_tool`, `server.rs:625`) and inject via a sibling interceptor:

```rust
tokio::task_local! { static EPISODE_ID: Option<MetadataValue<Ascii>>; } // best-effort: None across an interior spawn
// call_tool: EPISODE_ID.scope(tag.into_metadata(), dispatch_tool(request, &span)).await   ← episode param gone
// EpisodeInterceptor::call: EPISODE_ID.try_with(|ep| ep.map(|v| req.metadata_mut().insert(KEY, v.clone())));
```

Deletes all 7 forwards-only params and `EpisodeTag::request`. Two real caveats:

- tonic 0.14 doesn't compose two `Interceptor`s in one `InterceptedService` (`client.rs:30`) — fold both into a single combined interceptor in `grpc_service` (`:122`), don't nest.
- A task-local goes silently `None` across an interior `tokio::spawn`. The dispatch path is a single linear task today, so it's fine; flag it on the declaration since it's an attribution-loss (not a safety) failure mode.

The interceptor removes the _threading_ but can't _originate_ the value — entry extraction stays at `call_tool`, which is correct and is exactly one site.

### Out of scope (intentionally keep as-is)

The `&QueryAttribution` through `CatalogDiscovery` → `QueryManager` is legitimate plumbing to the one consumer that stamps the span; `attribution.rs` already documents that boundary. Don't hoist it to a `CatalogDiscovery` field unless that struct is constructed per-request — otherwise it's cross-request contamination.

### Summary

| # | Change | Template already in repo | Risk | Removes |
| --- | --- | --- | --- | --- |
| 1 | inbound extraction → tower layer | `GrpcMethodAnnotatedService` | lowest | 6 copy-pasted extractions |
| 2 | schema decoration → collection decorator | — | low | `episodes_enabled` through 6 builders |
| 3 | outbound threading → task-local + interceptor | `TraceContextInterceptor` | medium | `EpisodeTag::request` + 7 `&EpisodeTag` params |

1 and 2 are cheap and independent — they kill ~9 of the forwards-only sites on their own. 3 lands the `&EpisodeTag` deletion that motivated this.

**Bigger picture:** episodes is the _second_ request-scoped concern after trace context, and it re-spread because there's no shared seam. A small `RequestContext` extension — inbound layer lifts known headers in, outbound interceptor re-injects from ambient — would make the _next_ such concern a one-field add instead of a 42-function thread.

jamesaud:
Addressed in `65bdda3f`.

TLDR: I contained episode attribution at the request-context boundaries instead of forwarding it through individual tool/service helpers.

- App inbound: `GrpcMethodAnnotatedService` now parses `coral-episode-id` once into request extensions; query/catalog handlers read `QueryAttribution::from_extensions(...)`.
- MCP schema: tool builders no longer take `episodes_enabled`; `list_tools` decorates the ordinary tools once with `with_episode_id_argument`, while `open_episode` stays intentionally undecorated.
- MCP outbound: removed `EpisodeTag::request` and the forward-only `&EpisodeTag` params; `coral-client` now uses one combined request-context interceptor for trace context + task-local episode metadata.
- Kept `QueryAttribution` through `CatalogDiscovery -> QueryManager` as-is, since that is the legitimate boundary to the span-stamping consumer.

jsummerfield (approved):
Thanks, much nicer.

PR #1320, opened 2026-06-17
feat(ui): Use dialog to confirm souce removal

AlbertQM:
## Summary

This is our pattern, not sure how the old version landed in there.

## Test plan

Trying to remove a source, see the dialog

image

Before this PR: 

image

### DIFF
--- a/ui/src/views/sources/source-detail.css.ts
+++ b/ui/src/views/sources/source-detail.css.ts
@@ -96,24 +96,6 @@ export const alertError = style({
   paddingInline: 12,
 })
 
-export const removeConfirm = style({
-  alignItems: 'center',
-  background: theme.pill.red.background,
-  border: `1px solid ${theme.pill.red.stroke}`,
-  borderRadius: 8,
-  display: 'flex',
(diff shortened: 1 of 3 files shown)


martinrajdl:
Oh this was my call because I saw the double dialog and thought it was hallucinated and wrong, I've generally understood dialogs on top of dialogs to be a big ux no no. 

Wdyt?

martinrajdl (approved):

AlbertQM:
> Oh this was my call because I saw the double dialog and thought it was hallucinated and wrong, I've generally understood dialogs on top of dialogs to be a big ux no no.
> 
> Wdyt?

I remember having this conversation (pointing out this was the pattern we settled on), must've been reverted back at some point! Thanks for the reminder 🙈 

What's wrong with this UX?
The current version (before this PR) has layout shift, buttons jump around, and I thought it looked very odd 🤔 

Stacked dialogs are fine IMO, Pavel was OK with it and many products use them, e.g. see Base UI https://base-ui.com/react/components/dialog#nested-dialogs

PR #1355, opened 2026-06-18
feat: add explicit pagination overlays for DSL v4 sources

### LINKED ISSUE #1233: Add explicit pagination overlays
V4 pagination should be modeled explicitly instead of inferred from OpenAPI parameter names.

Current V4 OpenAPI import only detects pagination when an operation has query params named exactly `page` and `per_page`. Everything else materializes as `mode: none`, and V4 manifests currently have no authored pagination override surface. That means paginated APIs can silently install as first-page-only sources.

This is the wrong contract. OpenAPI specs usually expose parameters and response schemas, but not enough control-flow semantics to safely derive pagination behavior. Wrong pagination is worse than no pagination because it can silently return partial, duplicated, or skipped data.

Examples of pagination shapes that need explicit modeling:

- Stripe list endpoints: `limit`, `starting_after`, `has_more`, and last returned object id as next cursor.
- Stripe search endpoints: request `page`, response `next_page`.
- GitHub REST endpoints: follow the `Link` header `rel="next"` URL.
- Slack methods: cursor pagination for many methods, older method-specific schemes for others.
- Microsoft Graph/OData: follow the full `@odata.nextLink` URL.

Expected direction:

- Add V4 pagination overlays that can target operations.
- Support reusable source/provider-level overlay blocks to avoid repeating the same config for many operations.
- Store pagination provenance in materialized artifacts: `authored`, `profile_generated`, `unsupported`, or `none`.
- Emit diagnostics when a likely paginated operation has no overlay.
- Extend the shared pagination model/runtime for missing common shapes, especially cursor-from-last-row and response-body next-link URLs.
- Remove or de-emphasize implicit `page`/`per_page` detection so V4 does not silently pretend pagination is correct.

The key requirement is an explicit pagination contract. Any future generator or provider profile should produce overlays, not hidden importer behavior.

simonwhitaker:
Closes #1233

## Why

DSL v4 needs pagination to be an authored part of the source spec, not behavior guessed by an importer after it has normalized operations. Relying on generated operation names is especially brittle because manifest authors do not know those names until after import, and name-based heuristics can turn ordinary `cursor`, `limit`, or `page` parameters into runtime behavior the author did not ask for.

This PR moves pagination into explicit surface overlays. Importers can still warn when an operation looks paginated but no overlay matched, but they no longer create pagination behavior from heuristics.

## What changed

- Added `surfaces[].pagination` for DSL v4 OpenAPI and MCP surfaces.
- Added two authoring levels:
  - `profiles`: reusable match rules that apply one pagination contract to many operations/tools.
  - `operations`: exact per-operation overlays or exceptions; these win over profiles.
- OpenAPI overlays target raw source-native identifiers: either `operation_id` or `method` + `path`, but not both.
- MCP overlays target raw tool names with `target.tool`.
- Added `unsupported` overlays for operations that look pagination-like but should intentionally not paginate.
- Threaded pagination provenance through imported IR, generated projections, and materialized runtime packages so authored/profile/unsupported/no-pagination cases remain distinguishable.
- Extended HTTP runtime pagination with response-body next URLs and cursor-from-last-row support.
- Updated the preview `github_v4` and `stripe_v4` manifests to use explicit v4 pagination profiles.

## Authoring Examples

A source author puts pagination under the surface that owns the operations.

For OpenAPI APIs that use link headers, one profile can cover all matching GET operations with a `per_page` query parameter:

```yaml
surfaces:
  - id: rest
    type: openapi
    url: https://example.com/openapi.yaml
    pagination:
      profiles:
        - name: rest_link_header
          match:
            methods: [get]
            query_params: [per_page]
          pagination:
            mode: link_header
            page_size:
              default: 30
              max: 100
              query_param: per_page
```

For Stripe-style list endpoints, the author can express the real contract directly: request the next page with `starting_after`, read `has_more` from the response envelope, and take the next cursor from the last returned row's `id`:

```yaml
pagination:
  profiles:
    - name: stripe_rest_cursor_lists
      match:
        methods: [get]
        query_params: [limit, starting_after]
        response_cursor_path: [has_more]
      pagination:
        mode: cursor_query
        cursor_param: starting_after
        cursor_from_last_row_path: [id]
        has_more_path: [has_more]
        page_size:
          default: 100
          max: 100
          query_param: limit
```

For OpenAPI operation-specific behavior, authors target the raw OpenAPI contract instead of Coral's imported operation name. Use either `operation_id`:

```yaml
pagination:
  operations:
    - target:
        operation_id: searchIssues
      pagination:
        mode: link_header
        page_size:
          default: 30
          max: 100
          query_param: per_page
```

Or use `method` + `path` when the operation ID is absent, unstable, or not the best source-native identifier:

```yaml
pagination:
  operations:
    - target:
        method: get
        path: /search/issues
      pagination:
        mode: link_header
        page_size:
          default: 30
          max: 100
          query_param: per_page
```

For MCP tools, profiles match tool contracts rather than OpenAPI operations. A cursor-based MCP tool can be declared like this:

```yaml
surfaces:
  - id: mcp
    type: mcp
    pagination:
      profiles:
        - name: cursor_tools
          match:
            tool_args: [cursor]
            response_cursor_path: [meta, nextCursor]
          pagination:
            type: cursor
            cursor_arg: cursor
            response_cursor_path: [meta, nextCursor]
```

And offset/limit MCP tools can use an offset profile:

```yaml
pagination:
  profiles:
    - name: offset_tools
      match:
        tool_args: [limit, offset]
        offset_args: true
      pagination:
        type: offset
        limit_arg: limit
        default_limit: 50
        max_limit: 200
        offset_arg: offset
```

When something has pagination-looking names but is not actually pageable, authors can say so explicitly. This suppresses missing-overlay warnings without adding runtime pagination:

```yaml
pagination:
  operations:
    - target:
        operation_id: items/list
      unsupported:
        reason: The provider exposes a cursor-looking filter that is not a pagination cursor.
```

## Validation

- `git diff --check`
- `RUSTC_WRAPPER= make schema-check`
- `RUSTC_WRAPPER= make rust-checks`
- `RUSTC_WRAPPER= cargo run --locked -p coral-cli -- source lint sources/core-v4/github_v4/manifest.yaml`
- `RUSTC_WRAPPER= cargo run --locked -p coral-cli -- source lint sources/core-v4/stripe_v4/manifest.yaml`


crates/coral-engine/src/backends/http/fetch.rs
```diff
@@ -101,14 +101,18 @@ pub(super) async fn fetch_rows(
         );
         let base_url = render_template(&client.base_url, &render_context)?;
         let base_url = normalize_base_url(&base_url);
-        let following_link_header = matches!(
+        let following_pagination_url = matches!(
             pagination.mode,
-            ValidatedPaginationMode::LinkHeader | ValidatedPaginationMode::Auto
+            ValidatedPaginationMode::LinkHeader
+                | ValidatedPaginationMode::Auto
```
simonwhitaker, line 107: TODO: we shouldn't be doing any kind of automatic pagination detection. This mode should go.

crates/coral-spec/src/v4/schema.rs
```diff
@@ -83,6 +88,272 @@ enum V4InputSpecSchema {
...
+
+#[derive(Debug, Serialize, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
+struct V4UnsupportedPaginationSchema {
+    reason: String,
+}
+
+#[derive(Debug, Serialize, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
+struct V4HttpPaginationSpecSchema {
```
simonwhitaker, line 278: This massive struct means that all keys are available, regardless of mode, but all are optional. So you could specify mode=cursor, then only define the `page_param` property, and none of the cursor-related properties.

Instead, this should be an enum of mode types, which will (presumably!) translate into a `oneOf` construct in the JSON schema.

PR #1359, opened 2026-06-18
feat(sources/community/dodopayments): add Dodo Payments community source

guglxni:
## Summary

Add a [Dodo Payments](https://dodopayments.com) community source so Coral can query payments, customers, subscriptions, refunds, disputes, and payouts through SQL.

Dodo Payments is an all-in-one payments, billing, and merchant-of-record platform designed for SaaS and AI products. It supports 190+ countries, 80+ currencies, and 30+ payment methods with built-in support for subscriptions, usage-based billing, and credit-based billing.

## Source scope

- **Tables (6):** payments, customers, refunds, disputes, payouts, subscriptions
- **Auth:** Bearer token via DODO_API_KEY (kind: secret)
- **Base URL:** Configurable via DODO_BASE_URL (kind: variable), defaulting to https://live.dodopayments.com
- **Pagination:** Page-based (page_number starting at 0, page_size max 100)
- **Read-only:** All write operations (payment creation, refunds, subscription management, etc.) are intentionally out of scope

## Provider docs

- API reference: https://docs.dodopayments.com/api-reference/introduction
- Dashboard: https://app.dodopayments.com
- Test mode: https://docs.dodopayments.com/miscellaneous/test-mode-vs-live-mode

## Validation

- [x] `coral source lint` passes
- [x] `coral source add --file` with test credentials — connected successfully, 6/6 test queries passed
- [x] `coral source test dodopayments` passes — 6/6 test queries passed
- [x] Live query proof — all 6 tables returned valid empty results (fresh test account, auth confirmed)

> A fresh test account returns 0 rows until test data is created. The 0-row responses with HTTP 200 confirm successful authentication and correct schema mapping.

## Cross-source JOIN opportunities

- **Stripe:** Compare revenue for customers on both platforms via customer_email
- **HubSpot:** Revenue by customer lifecycle stage
- **SendGrid:** Revenue by customers who opened recent emails

## Notes

- Community source — install via `coral source add --file sources/community/dodopayments/manifest.yaml`
- Test mode keys (`dodo_test_*`) can be used at `https://test.dodopayments.com`
- Rate limits: 40 req/s burst, 240 req/min sustained (Tier 0)
- Test mode accounts may return empty results for some tables until test data is created


jsummerfield (changes requested):
Thanks for adding this source. I reviewed the current head (`500701f02133b5b90b1750638e7436de09576384`) against the Dodo API docs and the existing Coral source surfaces.



I’m requesting changes because the current source can pass its live checks against an empty test account while still exposing/documenting fields and examples that do not hold for real rows.

1. Several tested columns are not in the documented list response rows. The live validation only returns `0 rows`, but the test queries select `refunds.status`, `disputes.dispute_status`, and `payouts.status` / `payouts.currency` (`README.md` lines 470-477). Dodo’s list response docs for [refunds](https://docs.dodopayments.com/api-reference/refunds/get-refunds), [disputes](https://docs.dodopayments.com/api-reference/disputes/get-disputes), and [payouts](https://docs.dodopayments.com/api-reference/payouts/get-payouts) do not show those fields on returned items. Some of those names are valid filters, but that is not enough to expose them as response columns. Please either remove/adjust the unsupported columns, or include provider docs / non-empty live evidence proving those fields are actually present on the list endpoints and set nullability accordingly.

2. The customer join guidance is wrong for refunds and disputes. `README.md` lines 150-151 and `manifest.yaml` lines 257-258 say `customers` joins directly to `refunds` and `disputes` via `customer_id`, but those tables do not expose `customer_id`. As implemented, users need to go through `payments.payment_id`, or the source needs to expose a proven `customer_id` column there.

3. The cross-source examples currently do not run against the checked-in source schemas. For example, `stripe.payment_intents.amount_total`, `hubspot.contacts.lifecycle_stage`, and `sendgrid.messages` are not the existing Coral columns/tables those examples target. The timestamp comparisons also use `NOW() - INTERVAL ...` against `Utf8` timestamp columns. Please update these examples to use real source surfaces and valid types, or remove them until they can be tested.

4. Cubic’s three README findings are valid: payments has 7 non-pagination filters, not 8; the payout docs mention chargeback/refund aggregations but the manifest only exposes `tax`; and the customer join guidance issue above also applies in the manifest description.

Once the row-level schema and README examples line up with real list responses, this should be straightforward to re-review. The auth, base URL, endpoint paths, `items` row paths, and pagination shape otherwise look consistent with Dodo’s docs.

guglxni:
@cubic-dev-ai review this pull request

jsummerfield (changes requested):
Thanks for updating this. The `/refunds`, `/disputes`, and `/payouts` `rows_path: [items]` change addresses the response-envelope issue from my last pass, but I still need to request changes before this becomes user-facing.

- `sources/community/dodopayments/manifest.yaml:163` / `sources/community/dodopayments/README.md:132`: `payments.updated_at` is exposed and documented, but Dodo's List Payments response does not include `updated_at` in the item shape, and their generated SDK model also omits it. Please remove this column from `payments`, or add live/docs evidence showing it is actually returned by the list endpoint.
  - Dodo docs: https://docs.dodopayments.com/api-reference/payments/get-payments
  - SDK model: https://github.com/dodopayments/dodopayments-python/blob/main/src/dodopayments/types/payment_list_response.py
- `sources/community/dodopayments/README.md:150` / `sources/community/dodopayments/manifest.yaml:257`: the customer guidance says `customers` joins to `refunds` and `disputes` via `customer_id`, but those tables expose `customer_id` only as a filter, not as a result column. Either expose and validate that column, or rewrite the guidance to use the supported path through `payments.payment_id` / explicit filtering.
- `sources/community/dodopayments/README.md:367`: the cross-source examples still are not executable against the current Coral catalog. `stripe.payment_intents` uses `amount`, not `amount_total`; `hubspot.contacts` uses `lifecyclestage`, not `lifecycle_stage`; `sendgrid.messages` does not exist. The examples also compare Dodo `Utf8` timestamp strings directly to SQL interval expressions. Please update these to queries that actually run, or remove them.
- `sources/community/dodopayments/README.md:346`: the README says `payouts` includes deprecated chargeback/refund aggregations, but the manifest omits `chargebacks` and `refunds` even though Dodo's List Payouts response includes them. Please either add and validate those columns, or correct the README claim.
  - Dodo docs: https://docs.dodopayments.com/api-reference/payouts/get-payouts
- `sources/community/dodopayments/README.md:483`: the 0-row validation output proves authentication and query execution, but not “correct schema mapping.” A 0-row result does not exercise row paths, column paths, or nullability. Please reword that claim, or include live non-empty evidence for the mapped tables/columns.

Local checks I ran on this head: `coral source lint sources/community/dodopayments/manifest.yaml`, `make lint-sources`, and `git diff --check` all passed.

jsummerfield (changes requested):
One remaining blocker before this is ready:

- `sources/community/dodopayments/manifest.yaml` exposes `payments.refund_status`, but the payment row still omits Dodo’s payment-level `dispute_status`; `sources/community/dodopayments/README.md` mirrors that omission in the payments column list. Dodo’s generated payment list model includes `dispute_status` as the most recent dispute status for the payment, alongside `refund_status` and `invoice_url` ([SDK type](https://github.com/dodopayments/dodopayments-typescript/blob/02b33c66384c629f5e315907ec11c38197e2d83d/src/resources/payments.ts#L643-L706), [OpenAPI schema](https://github.com/dodopayments/dodo-docs/blob/96683396562c09685a68bfc56417fe9a1ed8d8bc/openapi/openapi.documented.yml#L30585-L30680)). For a billing source, dispute state is core payment context; users should not need to scan/join the separate disputes table just to see whether each payment is currently disputed. Please add `dispute_status` to `dodopayments.payments` and document it in the README. I would also either add `invoice_url` or soften the README claim that the column definitions follow the OpenAPI list-response schemas so it is explicit that this is a selected subset.

I rechecked the earlier blockers on this head and they look resolved: `payments.updated_at` is gone, refund/dispute `customer_id` is clearly filter-only, the cross-source examples now use valid source schemas, payout refund/chargeback fields are present and documented as deprecated, and the live-output section no longer overclaims row mapping for the 0-row tables. Local validation also passes: `coral source lint sources/community/dodopayments/manifest.yaml`, `make lint-sources`, and `git diff --check`.

Thanks for the cleanup so far; this is close.

jsummerfield (changes requested):
Thanks for the quick follow-up. The previous payment-level `dispute_status` blocker is resolved on this head, and `invoice_url` is now exposed/documented too.

I found one remaining schema issue before this is ready:

- `sources/community/dodopayments/manifest.yaml:856` / `sources/community/dodopayments/README.md:313` expose `subscriptions.brand_id` as a returned column, but Dodo’s `GET /subscriptions` surface documents `brand_id` as a query filter, not as a field on each `SubscriptionListResponseItem`. The generated TypeScript list response model also omits `brand_id`. As written, real subscription rows will return null for that column even though the README presents it as row data. Please either remove the `subscriptions.brand_id` column, or make it explicit/virtual filter echo state if that is the intended SQL behavior.

Provider refs:
- https://docs.dodopayments.com/api-reference/subscriptions/get-subscriptions-1
- https://github.com/dodopayments/dodo-docs/blob/d9dd775272288ea87ebe3d754751ceae48e7e949/openapi/openapi.documented.yml#L22155-L22236
- https://github.com/dodopayments/dodo-docs/blob/d9dd775272288ea87ebe3d754751ceae48e7e949/openapi/openapi.documented.yml#L34490-L34634
- https://github.com/dodopayments/dodopayments-typescript/blob/02b33c66384c629f5e315907ec11c38197e2d83d/src/resources/subscriptions.ts#L826-L966

Local checks on this head still pass: `coral source lint sources/community/dodopayments/manifest.yaml`, `make lint-sources`, and `git diff --check`.

jsummerfield (approved):
Thanks for the update. I re-reviewed the current head and the previous `subscriptions.brand_id` blocker is fixed: Dodo exposes it as a list-subscriptions query parameter, and this now models it as a virtual filter echo instead of a response field.

I also checked the six table shapes against Dodo's current docs/OpenAPI-derived SDK, verified the source lint locally, and the fork workflows are green. The timestamp columns could be tightened to Coral `Timestamp` types in a future follow-up, but I do not think that should block this source.

PR #1363, opened 2026-06-21
feat(spec): add OpenAPI 3.1 support for DSL v4 importer

### LINKED ISSUE #1321: Add OpenAPI 3.1 support

guglxni:
## Summary

- Accept OpenAPI **3.0.x** and **3.1.x** documents in the DSL v4 OpenAPI surface importer; reject 3.2+ for now.
- Handle OpenAPI 3.1 JSON Schema patterns used by the importer: nullable `type` arrays (e.g. `["string", "null"]`), object/array type arrays, and schemas without an explicit `type` when `properties` are present.
- Add unit/integration coverage in `coral-spec` and an end-to-end materialization test in `coral-app`.

Closes #1321.

## Test plan

- [x] `cargo test -p coral-spec v4::openapi --all-features` (21 tests)
- [x] `cargo test -p coral-app --lib sources::manager::tests::import_v4_openapi_31_source_writes_materialized_artifacts --all-features`
- [x] `cargo clippy -p coral-spec -p coral-app --all-targets --all-features --locked -- -D warnings`
- [x] `make docs-check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Features
+
+* **spec:** import OpenAPI 3.1 documents in DSL v4 ([#1321](https://github.com/withcoral/coral/issues/1321))
+
 ## [0.4.3](https://github.com/withcoral/coral/compare/v0.4.2...v0.4.3) (2026-06-15)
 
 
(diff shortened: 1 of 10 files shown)


guglxni:
@cubic-dev-ai review this pull request

simonwhitaker:
Thanks for this, but we're still discussing internally how we want to handle OpenAPI 3.1, and we'll pick up the coding task once we've decided.

PR #1365, opened 2026-06-22
fix(app): skip poisoned v4 sources during broad load

iaptsiauri:
Skips DSL v4 sources with missing or incompatible artifacts during source loading, preserves a warning with source name and error detail.

### DIFF
--- a/crates/coral-app/src/query/manager.rs
+++ b/crates/coral-app/src/query/manager.rs
@@ -283,10 +283,14 @@ impl QueryManager {
         for source in config.workspace_sources(workspace_name) {
             match self.load_query_source(workspace_name, &source) {
                 Ok((query_source, _version)) => query_sources.push(query_source),
-                Err(
-                    error @ (AppError::Credentials(CredentialsError::Unavailable(_))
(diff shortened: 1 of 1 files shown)


sauldhernandez (approved):

crates/coral-app/src/query/manager.rs
```diff
@@ -283,10 +283,14 @@ impl QueryManager {
...
-                    | AppError::MissingOrIncompatibleV4Materialization { .. }),
-                ) => {
+                Err(error @ AppError::MissingOrIncompatibleV4Materialization { .. }) => {
+                    tracing::warn!(
+                        source = %source.name,
+                        detail = %error,
+                        "skipping source with incompatible DSL v4 materialization"
+                    );
+                }
+                Err(error @ AppError::Credentials(CredentialsError::Unavailable(_))) => {
```
simonwhitaker, line 293: Don't we want the same behaviour for `CredentialsError::Unavailable`? i.e. log it and continue. Otherwise that error can also poison the whole source load, right?

PR #1368, opened 2026-06-23
feat(reef): add schema explorer

martinrajdl:
## Summary
- Add a Schema route and navigation entry with a schema/table explorer, table search, lazy column loading, required-filter indicators, and shared tooltip/table UI primitives.

- Make catalog metadata loading avoid keychain reads for passive browsing while keeping runtime query credential loading strict. -  this is the change to manager.rs - I'm not 100% sure this is the right separation of concerns there, but it doesn't sound directionally incorrect, would love any thoughts/suggestions. 

- Keep the imported explorer focused on schemas, tables, and selected table columns by removing the copied guide/query/results/add-source/sidebar-column UI.

Screenshot 2026-06-23 at 15 18 29


ui/src/lib/router.ts
```diff
@@ -14,6 +14,10 @@ function parseHash(): ParsedLocation {
     return { route: { kind: 'sources' } }
   }
 
+  if (segments[0] === 'schema' || segments[0] === 'schema-explorer') {
```
lazybean: Do we ever set `schema-explorer` in the URL? Could not find.

ui/src/views/schema-explorer.tsx
```diff
@@ -0,0 +1,447 @@
...
+              />
+              {search ? (
+                <button
+                  aria-label="Clear filter"
+                  className={styles.clearButton}
+                  onClick={() => setSearch('')}
+                  type="button"
+                >
+                  <Icon color="tertiary" name="X" size="14" />
+                </button>
```
lazybean: :wrench: :bee: There must be a WAX button!

ui/src/lib/schema-explorer.ts
```diff
@@ -0,0 +1,163 @@
...
+  )
+
+  return response.items.flatMap((item) => (item.item.case === 'table' ? [item.item.value] : []))
+}
+
+async function listColumnsForTable(schemaName: string, tableName: string): Promise<ColumnDef[]> {
+  const columns: ColumnDef[] = []
+  let offset = 0
+
+  while (true) {
```
lazybean: stripe payment_intents is an example that needs this pagination loop.

ui/src/views/schema-explorer.tsx
```diff
@@ -0,0 +1,447 @@
...
+                      {selectedTable.table.description}
+                    </Typography.BodySmall>
+                  ) : null}
+                </div>
+                {selectedTable.table.requiredFilters.length > 0 ? (
+                  <div className={styles.requiredFilterGroup}>
+                    {selectedTable.table.requiredFilters.map((filter) => (
+                      <span className={styles.requiredFilterPill} key={filter}>
+                        {filter}
+                      </span>
```
lazybean: :wrench: Worth retrieving the wax/Pill from coral-cloud/reef or phoebe

ui/src/views/schema-explorer.tsx
```diff
@@ -0,0 +1,447 @@
...
+                    {selectedTable.schemaName}.{selectedTable.table.name}
+                  </Typography.HeadingSmall>
+                  {selectedTable.table.description ? (
+                    <Typography.BodySmall as="p" className={styles.description}>
+                      {selectedTable.table.description}
+                    </Typography.BodySmall>
+                  ) : null}
+                </div>
+                {selectedTable.table.requiredFilters.length > 0 ? (
+                  <div className={styles.requiredFilterGroup}>
```
lazybean: :thinking_face: It is not immediately cleare what they are.  




  
  
  
 Adding a tooltip saying 'The follwing filters are required: ....' would help understand what those pills are for.

ui/src/wax/components/table/table.tsx
```diff
@@ -0,0 +1,108 @@
+import classNames from 'classnames'
+import { createContext, useContext } from 'react'
+
+import * as styles from './table.css'
+
+type TableStyle = 'compact' | 'default'
+
+const TableStyleContext = createContext<TableStyle>('compact')
```
lazybean: :no_entry: CSS does not need react context.  
Please copy the table from the wax reef which is context free..

ui/vite.config.ts
```diff
@@ -15,6 +15,7 @@ export default defineConfig(({ mode }) => {
       },
     },
     server: {
+      host: '127.0.0.1',
```
lazybean: Is that change needed at all for this?

lazybean (commented):
Thank you for bringing that back.  
A few early comments. I am now overheating and will review more tomorrow.

ui/src/lib/schema-explorer.ts
```diff
@@ -0,0 +1,163 @@
...
+    const response = await catalogClient.listColumns(
```
lazybean: Instead of doing serial request and stopping when `hasMore === false` can we use `totalCount` and query all pages at once in parallel?

ui/src/App.tsx
```diff
@@ -12,7 +13,13 @@ export function App() {
...
   return (
     <Shell>
-      {location.route.kind === 'sources' ? <SourcesIndex /> : <TracesPage />}
+      {location.route.kind === 'sources' ? (
+        <SourcesIndex />
+      ) : location.route.kind === 'schema' ? (
+        <SchemaExplorer />
+      ) : (
+        <TracesPage />
+      )}
```
AlbertQM: @lazybean will we migrate to React Router?

ui/src/wax/components/table/table.tsx
AlbertQM: Just a tiny heads-up that there's a PR to update the styles of that table to match the latest Figma with the OSS font

ui/src/views/schema-explorer.css.ts
```diff
@@ -0,0 +1,288 @@
...
+import { theme } from '@/wax/theme/theme.css'
+
+const spin = keyframes({
+  from: { transform: 'rotate(0deg)' },
+  to: { transform: 'rotate(360deg)' },
+})
+
+export const spinAnimation = style({
+  animation: `${spin} 1s linear infinite`,
+})
```
AlbertQM: This animation is duplicated in a bunch of files, can we extract it and reuse? We already have wax `animations` for that

ui/src/views/schema-explorer.css.ts
```diff
@@ -0,0 +1,288 @@
...
+  borderRadius: 4,
+  color: theme.content.primary,
+  cursor: 'pointer',
+  display: 'flex',
+  gap: 6,
+  minWidth: 0,
+  paddingBlock: 4,
+  paddingInline: 8,
+  textAlign: 'left',
+  transition: 'background-color 150ms ease',
```
AlbertQM: we should use animation.colorTransition no? To keep all animations the same 👀

ui/src/lib/schema-explorer.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  return columns
+}
+
+function optional(value: string): string | undefined {
+  const trimmed = value.trim()
+  return trimmed ? trimmed : undefined
+}
+
+function withTimeout<T>(promise: Promise<T>, timeoutMs: number, message: string): Promise<T> {
```
AlbertQM: Which behaviour did you obseve that granted the need for this timeout logic?

reef/app/views/schema-explorer.tsx
```diff
@@ -0,0 +1,447 @@
...
+            tables: connector.tables.map((table) =>
+              table.name === tableName ? { ...table, columns, columnsLoaded: true } : table,
+            ),
+          }
+        : connector,
+    ),
+  }
+}
+
+export function SchemaExplorer() {
```
AlbertQM, line 107: The SchemaExplorer is one of the last things we polished with Pavel 😢 -- code is in adp repo, could we port that? it has few nice touches 😄 e.g. (Virtual and _italic_ for virtual columns), actually showing the columns on the sidebar

Feel free to push back on this though if you think the current UI is neater

AlbertQM (commented):
Thank you! This looks like the right first step, thanks for porting it and removing the unnecessary bits :smile:

reef/app/views/schema-explorer.tsx
martinrajdl: I removed it quite intentionally. You need to wait for the data to load and just have duplicit information in the UI, you're already opening the table to view the columns on the left. Seems quite bad to show it twice.

ui/src/App.tsx
martinrajdl: From what I heard last, we'll move this to coral-cloud so if that has router we'll likely use that.

lazybean: That should not block this PR.

AlbertQM: Definitely not blocking, was just curious to see where we're headed. Thanks both!

crates/coral-app/src/query/manager.rs
lazybean, line 1: We need @sauldhernandez for this.

PR #1372, opened 2026-06-24
feat(search): add universal search contracts

mystic123:
## Summary

This is the bottom PR for the SQLite-only Universal Search stack. It adds the public contract and inert app/client wiring, without adding any storage backend, indexing, retrieval, MCP tool, or CLI entrypoint yet.

- Adds `SearchService` proto/messages with typed result payloads, provider status, coverage, and truncation metadata.
- Wires an app-level `SearchManager`/`SearchService` shell into the gRPC server and client.
- Keeps the inert shell behavior directly in `SearchManager`; there is no separate provider-orchestration engine layer until a later PR has real provider orchestration to own it.
- Reports explicit `NotEnabled` shell statuses for every declared provider: catalog metadata, observed values, and native fanout.
- Carries the public `Skipped` provider state through the app-domain enum and mapper for future request-level provider gating.
- Leaves provider coverage absent/null for `NotEnabled` and `Skipped` providers; enabled providers can return structured coverage.
- Exposes existing table-function `kind` and `search_limits` metadata through catalog proto responses, so `kind: search` functions can be returned as catalog metadata.
- Fails loudly if app/engine catalog metadata contains malformed `search_limits_json`, instead of silently omitting corrupted search-limit metadata.
- Documents the new `SearchService` contract and its current inert `NotEnabled` behavior in the architecture docs.
- Adds regression coverage for the shell response, disabled/skipped provider coverage absence, empty-query validation, table-function search metadata transport, and malformed search-limit metadata.

## Boundary

This PR declares the typed result payload contract but intentionally keeps the app shell result list empty. Catalog metadata and column hint payload mapping are exercised in #1423. `kind: search` functions are represented as catalog metadata table-function items, not as a separate `native_search_path` payload. Observed-value payload mapping will be covered with the observed-value provider PR, and active native fanout execution will get its own later provider/result contract.

## Why

This keeps the contract reviewable before the SQLite catalog/observed-value providers arrive. The service currently returns explicit not-wired-yet provider statuses rather than pretending search is implemented.

## Validation

- `cargo fmt --check`
- `cargo check -p coral-app --all-targets`
- `cargo test -p coral-app search:: --lib`

Earlier stack validation for this PR also included:

- `CARGO_TARGET_DIR=/Users/pakapica/src/withcoral/.cargo-target/universal-search-sqlite-pr1 cargo check --locked -p coral-app -p coral-client -p coral-api`
- `CARGO_TARGET_DIR=/Users/pakapica/src/withcoral/.cargo-target/universal-search-sqlite-pr1 cargo test --locked -p coral-app search::service::tests::skipped_provider_status_maps_without_coverage -- --nocapture`
- `CARGO_TARGET_DIR=/Users/pakapica/src/withcoral/.cargo-target/universal-search-sqlite-pr1 cargo test --locked -p coral-app transport::tests::table_function_to_proto_rejects_malformed_search_limits -- --nocapture`
- `CARGO_TARGET_DIR=/Users/pakapica/src/withcoral/.cargo-target/universal-search-sqlite-pr1 cargo test --locked -p coral-app search_tests -- --nocapture`
- `CARGO_TARGET_DIR=/Users/pakapica/src/withcoral/.cargo-target/universal-search-sqlite-pr1 cargo clippy --locked -p coral-app --all-targets -- -D warnings`

### DIFF
--- a/apps/docs/project/architecture.mdx
+++ b/apps/docs/project/architecture.mdx
@@ -68,6 +68,7 @@ The local server and core orchestrator.
 | Feedback persistence | Workspace-scoped blocked-agent feedback reports (`feedback/`) |
 | Query orchestration | Loads sources, builds the engine, executes SQL (`query/`) |
 | Catalog discovery | Lists/searches/describes query-visible tables and columns (`catalog/`) |
+| Universal Search | Owns the `SearchService` shell and app-domain search result contract (`search/`) |
 
 This is the largest crate. It ties `coral-spec` and `coral-engine` together and manages all persistent local state.
 
@@ -117,7 +118,7 @@ The shared gRPC contract.
 | Protobuf definitions | `proto/coral/v1/` transport contracts for Coral services and messages |
 | Generated bindings   | Tonic-generated Rust types and service traits                                          |
 
-All inter-crate communication between `coral-client` and `coral-app` goes through these generated types.
+All inter-crate communication between `coral-client` and `coral-app` goes through these generated types. The active service contracts include source management, query execution, catalog discovery, workspaces, feedback, episodes, traces, and Universal Search.
 
 ## Dependency graph
 
--- a/crates/coral-api/proto/coral/v1/search.proto
+++ b/crates/coral-api/proto/coral/v1/search.proto
@@ -1,0 +1,228 @@
+syntax = "proto3";
+
+package coral.v1;
+
+import "coral/v1/catalog.proto";
+import "coral/v1/resources.proto";
+
+// Universal Search request for clue routing over Coral metadata and local memory.
+message SearchRequest {
+  // Workspace whose current search context should be inspected.
+  Workspace workspace = 1;
+  // User or agent clue to route to likely Coral surfaces. The server trims
+  // surrounding whitespace, rejects empty queries, and rejects queries longer
+  // than 512 bytes.
+  string query = 2;
+  // Optional maximum results. Zero applies the API default of 10. Non-zero
+  // values must be between 1 and 50 inclusive; larger values are rejected
+  // rather than clamped.
+  uint32 limit = 3;
+}
+
+// Universal Search response with typed, bounded routing hints.
+message SearchResponse {
+  // Ordered search hints.
+  repeated SearchResult results = 1;
+  // Provider coverage and error disclosure.
+  repeated SearchProviderStatus provider_statuses = 2;
+  // Result truncation metadata.
+  SearchResultTruncation truncation = 3;
+}
+
+// Result truncation metadata.
+message SearchResultTruncation {
+  // Whether more results were available than returned.
+  bool truncated = 1;
+  // Number of results returned in this response.
+  uint32 returned_count = 2;
+  // Applied maximum result count.
+  uint32 max_results = 3;
+  // Human-readable truncation note.
+  string note = 4;
+}
+
+// Backing provider status for a search response.
+message SearchProviderStatus {
+  // Provider name.
+  SearchProvider provider = 1;
+  // Provider state for this request.
+  SearchProviderState state = 2;
+  // Short coverage or error note.
+  string note = 3;
+  // Structured provider-local coverage for this request. Absent when the provider is not enabled or skipped.
+  SearchProviderCoverage coverage = 4;
+}
+
+// Structured provider-local coverage for one request.
+message SearchProviderCoverage {
+  // Units the provider considered eligible for this request.
+  uint32 eligible_units = 1;
+  // Units the provider actually searched for this request.
+  uint32 searched_units = 2;
+  // Units that failed while this provider ran.
+  uint32 failed_units = 3;
+  // Candidates returned by the provider before global fusion/truncation.
+  uint32 returned_count = 4;
+  // Whether the provider had more candidates than its local retrieval window.
+  bool has_more = 5;
+  // Whether the provider stopped because its maintenance/search budget expired.
+  bool budget_exhausted = 6;
+  // Whether the provider timed out.
+  bool timed_out = 7;
+  // Whether the provider served from stale or partially refreshed local state.
+  bool stale_index = 8;
+}
+
+// Universal Search backing providers.
+enum SearchProvider {
+  // Unknown provider.
+  SEARCH_PROVIDER_UNSPECIFIED = 0;
+  // Catalog metadata provider.
+  SEARCH_PROVIDER_CATALOG_METADATA = 1;
+  // Local observed-value provider.
+  SEARCH_PROVIDER_OBSERVED_VALUES = 2;
+  // Active provider-native fanout provider.
+  SEARCH_PROVIDER_NATIVE_FANOUT = 3;
+}
+
+// Provider state for one request.
+enum SearchProviderState {
+  // Unknown provider state.
+  SEARCH_PROVIDER_STATE_UNSPECIFIED = 0;
+  // Provider returned one or more results.
+  SEARCH_PROVIDER_STATE_RESULTS_FOUND = 1;
+  // Provider ran and returned no results.
+  SEARCH_PROVIDER_STATE_EMPTY = 2;
+  // Provider is not enabled.
+  SEARCH_PROVIDER_STATE_NOT_ENABLED = 3;
+  // Provider was skipped for this request.
+  SEARCH_PROVIDER_STATE_SKIPPED = 4;
+  // Provider returned partial results.
+  SEARCH_PROVIDER_STATE_PARTIAL = 5;
+  // Provider failed for this request.
+  SEARCH_PROVIDER_STATE_ERROR = 6;
+}
+
+// One typed Universal Search result.
+message SearchResult {
+  // Provider that produced this result.
+  SearchProvider provider = 1;
+  // Typed result payload.
+  oneof payload {
+    // Table or table-function metadata hit with search context.
+    CatalogMetadata catalog_metadata = 10;
+    // Column, filter, argument, or result-column hint.
+    ColumnHint column_hint = 11;
+    // Local observed-value hit. Reserved for the observed-value provider.
+    ObservedValue observed_value = 12;
+  }
+}
+
+// Table or table-function metadata hit with search-specific context.
+message CatalogMetadata {
+  // Matched catalog item.
+  CatalogItem item = 1;
+  // Metadata fields that matched the query.
+  repeated string matched_fields = 2;
+  // Compact table column hints for ordinary table hits.
+  SearchTableColumnPreview table_column_preview = 3;
+}
+
+// Compact column hints for an ordinary table metadata hit.
+message SearchTableColumnPreview {
+  // Total column count for the table before preview capping.
+  uint32 column_count = 1;
+  // Capped columns selected for query-writing orientation.
+  repeated SearchTableColumnPreviewColumn columns = 2;
+  // Number of table columns not included in the capped preview.
+  uint32 omitted_column_count = 3;
+}
+
+// One compact column hint in a table metadata hit preview.
+message SearchTableColumnPreviewColumn {
+  // Column name as exposed by the query engine.
+  string name = 1;
+  // The DataFusion/Arrow data type rendered as text.
+  string data_type = 2;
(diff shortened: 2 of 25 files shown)


mystic123:
* **#1472** <a href="https://app.graphite.com/github/pr/withcoral/coral/1472?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1573** <a href="https://app.graphite.com/github/pr/withcoral/coral/1573?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1372** <a href="https://app.graphite.com/github/pr/withcoral/coral/1372?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1372?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

AlbertQM (approved):

apps/docs/project/architecture.mdx
```diff
@@ -117,7 +118,9 @@ The shared gRPC contract.
 | Protobuf definitions | `proto/coral/v1/` transport contracts for Coral services and messages |
 | Generated bindings   | Tonic-generated Rust types and service traits                                          |
 
-All inter-crate communication between `coral-client` and `coral-app` goes through these generated types.
+All inter-crate communication between `coral-client` and `coral-app` goes through these generated types. The active service contracts include source management, query execution, catalog discovery, workspaces, feedback, episodes, traces, and Universal Search.
+
+`SearchService` is the Universal Search contract. It returns typed, bounded routing hints plus provider status, coverage, and truncation metadata. In the current shell implementation the service is intentionally inert: concrete providers are not wired yet, so the app returns no results and reports each declared provider as `NotEnabled`.
```
jsummerfield: just kill these - low value adds

crates/coral-api/proto/coral/v1/search.proto
```diff
@@ -0,0 +1,228 @@
...
+  // Source schema that owns the surface.
+  string schema_name = 2;
+  // Table or table-function name.
+  string surface_name = 3;
+  // Whether the owning surface is a table or table function.
+  SearchSurfaceKind surface_kind = 4;
+  // Column, required filter, argument, or result-column name.
+  string name = 5;
+  // Data type when known.
+  string data_type = 6;
```
jsummerfield: enum?

crates/coral-api/proto/coral/v1/search.proto
```diff
@@ -0,0 +1,228 @@
...
+  // Number of table columns not included in the capped preview.
+  uint32 omitted_column_count = 3;
+}
+
+// One compact column hint in a table metadata hit preview.
+message SearchTableColumnPreviewColumn {
+  // Column name as exposed by the query engine.
+  string name = 1;
+  // The DataFusion/Arrow data type rendered as text.
+  string data_type = 2;
```
jsummerfield, line 146: enum?

crates/coral-api/proto/coral/v1/search.proto
```diff
@@ -0,0 +1,228 @@
...
+    // Source, table, or table-function metadata hit with search context.
+    CatalogMetadata catalog_metadata = 10;
+    // Column, filter, argument, or result-column hint.
+    ColumnHint column_hint = 11;
+    // Local observed-value hit. Reserved for the observed-value provider.
+    ObservedValue observed_value = 12;
+  }
+}
+
+// Source, table, or table-function metadata hit with search-specific context.
```
jsummerfield: **Contract gap: this promises source hits that `CatalogItem` can't represent.**

This comment (and the oneof arm comment on `catalog_metadata = 10`) says "Source, table, or table-function metadata hit", but `CatalogMetadata.item` is a `CatalogItem`, whose oneof only has `table` and `table_function` variants. A source-*name* match can surface indirectly as per-surface hits with `matched_fields=["schema_name"]` — which is what #1423 actually does — but a source-level hit as its own result (a match on the Source manifest's `description`, or any hit for a source with no query-visible tables) has no legal payload.

Suggest narrowing to "Table or table-function metadata hit" so the contract says only what the type can express. A `SourceMetadata` payload can be added additively later — as a new `SearchResult.payload` oneof variant, not a widened `CatalogItem` (that message is shared with the catalog RPCs, where a source isn't a queryable item). If narrowing, the new `SEARCH_RESPONSE_MAX_MESSAGE_SIZE` comment in `crates/coral-api/src/lib.rs` makes the same "source ... descriptions" promise and should be updated in the same commit.

crates/coral-app/src/search/service.rs
```diff
@@ -0,0 +1,148 @@
...
+impl SearchServiceApi for SearchService {
+    async fn search(
+        &self,
+        request: Request<ProtoSearchRequest>,
+    ) -> Result<Response<ProtoSearchResponse>, Status> {
+        let span = grpc_span(&request);
+        let search = self.search.clone();
+        instrument_grpc(span, async move {
+            let request = request.into_inner();
+            let workspace_name = workspace_name_from_proto(request.workspace.as_ref())?;
```
jsummerfield, line 42: **A nonexistent workspace returns OK here, unlike every other workspace-scoped RPC.**

`workspace_name_from_proto` validates presence and name format only, and `SearchManager` never checks the workspace exists — so `Search` against an unknown workspace returns an OK shell response. Sources, catalog, and query all route through `AppConfig::require_workspace` and return `NotFound` with the structured `WORKSPACE_NOT_FOUND` reason (pinned by `source_requests_reject_unknown_workspace` in `tests/grpc/workspace_lifecycle_tests.rs`).

Since this PR registers the service live, ships `SearchClient`, and pins the shell's observable behavior in gRPC tests, deferring this pins the wrong behavior — the contract would silently flip OK→NotFound when providers land. The plumbing is already here: give `SearchManager` a `config_store` (the `QueryManager`/`SourceManager` pattern), make `search()` fallible with `require_workspace` (mirroring `query/manager.rs`), and the existing `SearchManagerError::App` → `app_status` path already maps `WorkspaceNotFound` → `NotFound`. A gRPC test mirroring `source_requests_reject_unknown_workspace` would pin it.

crates/coral-engine/src/contracts/catalog.rs
```diff
@@ -91,4 +93,8 @@ pub struct TableFunctionInfo {
     pub arguments: Vec<TableFunctionArgumentInfo>,
     /// Columns returned by the function.
     pub result_columns: Vec<TableFunctionResultColumnInfo>,
+    /// Function role. Search functions perform provider-native retrieval.
+    pub kind: SourceTableFunctionKind,
+    /// JSON-encoded provider search limit metadata, when declared by the source.
+    pub search_limits_json: Option<String>,
```
jsummerfield: **Consider carrying the typed `SearchLimitsSpec` here instead of the JSON string.**

The engine serializes the already-validated spec to JSON at registration (`serialize_search_limits`), this contract carries the string, and coral-app's `search_limits_json_to_proto` immediately re-parses and re-validates it. That makes `table_function_to_proto` fallible, rippling `Result` through `catalog_item_to_proto`, `catalog_search_result_to_proto`, `validate_source_response_to_proto`, `catalog/service.rs`, and `sources/service.rs` — plus two tests for `Status::internal` paths that are unreachable from validated in-process data.

This PR already applies the right pattern to the sibling field: `kind` went `String` → `SourceTableFunctionKind`, with `as_str()` only at the Arrow row. Same move here: hold `search_limits: Option<SearchLimitsSpec>` in `RegisteredTable`/`RegisteredTableFunction` and `TableFunctionInfo`, and serialize to JSON only when building the `search_limits_json` Utf8 column in `runtime/catalog.rs` — which is already how `arguments_json`/`result_columns_json` are produced there. That keeps the transport mapper infallible and deletes the parse/validate/error machinery. No new dependency: this struct already exposes `coral_spec::SourceTableFunctionKind`, and after manifest validation `max_top_k ≤ 1000` / `max_calls_per_query ≤ 100`, so the usize→u32 conversion can't overflow either.

crates/coral-api/proto/coral/v1/search.proto
```diff
@@ -0,0 +1,228 @@
...
+  string description = 8;
+  // Metadata fields that matched the query.
+  repeated string matched_fields = 9;
+  // How the field should be used when querying the owning surface.
+  SearchFieldRole field_role = 10;
+}
+
+// Local observed-value result.
+message ObservedValue {
+  // Matched value or redacted display value.
```
jsummerfield: **Clients can't tell raw from redacted.**

"Matched value or redacted display value" commits the contract to two states, but no field distinguishes them — a consumer can't know whether `value` is safe to reuse (e.g. as a SQL literal) or display-only. Adding a discriminator later is wire-compatible but unsafe in practice: clients compiled against today's proto would silently read the zero value ("raw") for a later-added flag.

Two clean options: add an explicit enum now (`OBSERVED_VALUE_REDACTION_{UNSPECIFIED, RAW, REDACTED}` — matching this PR's UNSPECIFIED-zeroed enum convention, and avoiding the `bool` false-default trap), or drop "or redacted display value" from the comment so the contract only says what this PR commits to, and introduce redaction semantics with the observed-value provider PR.

jsummerfield (commented):
Design review pass on the contract. Four inline findings below, each verified against the diff, the existing service conventions, and the upstack provider PR (#1423) before posting. The workspace-validation one is the one I'd fix in this PR; the rest are contract/seam tightening that is cheapest at the bottom of the stack.

crates/coral-api/proto/coral/v1/search.proto
mystic123: The public Catalog API also exposes datafusion data types as simple strings and we don't have proto message for datatypes anywhere.  
It would be good to introduce it and refactor everywhere it fits, but it's a side quest. I will create a linear ticket for this.  
  
(https://github.com/withcoral/coral/blob/main/crates/coral-api/proto/coral/v1/catalog.proto#L35)

crates/coral-api/proto/coral/v1/search.proto
mystic123: good point, removing "redacted" for now, I will reintroduce it later in the stack

jsummerfield (approved):

crates/coral-api/proto/coral/v1/search.proto
```diff
@@ -0,0 +1,228 @@
...
+  // Table-function argument.
+  SEARCH_FIELD_ROLE_TABLE_FUNCTION_ARGUMENT = 3;
+  // Table-function result column.
+  SEARCH_FIELD_ROLE_TABLE_FUNCTION_RESULT_COLUMN = 4;
+}
+
+// A field-level hint that may hold or accept the query clue.
+message ColumnHint {
+  // Workspace that scopes this hint.
+  Workspace workspace = 1;
```
sauldhernandez: Wouldn't searches already be scope-bound?

crates/coral-app/src/bootstrap/server.rs
```diff
@@ -421,10 +424,12 @@ async fn start_server(
         service: trace_service,
         local_trace_store_dir,
     } = trace_components;
+    let search_manager = SearchManager::new(query_manager.config_store());
```
sauldhernandez: I'd suggest this is instantiated outside (like the other managers). That way you also remove the exposuore of query manager's config store.

Bradley-Butcher (approved):
This is a nit/style thing, but will prob make maintenance easier:

`crates/coral-app/src/search/service.rs`: Rust already has explicit machinery for this kind of conversion. I’d prefer we use `From/TryFrom` impls in a small `search/proto.rs` module instead of growing a set of free `*_to_proto` helpers in `service.rs`. Use `TryFrom` where conversion can fail, like proto request parsing or response payload mapping, and `From` where the mapping is total, like provider enums, statuses, coverage, and truncation.

mystic123:
### Merge activity

* **Jul 8, 8:20 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1372).
* **Jul 8, 8:20 AM UTC**: @mystic123 merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1372).

PR #1393, opened 2026-06-26
feat(mcp)!: support SQL query batches

jsummerfield:
The MCP `sql` tool could only accept one top-level SQL string, so agents that needed several independent Coral queries had to issue repeated tool calls, preserve ordering themselves, and recover from partial failures outside Coral. This PR moves that orchestration into the MCP adapter with a bounded `queries[]` batch contract and deterministic ordered results.

This is also a contract cleanup. MCP SQL results now use structured output as the authoritative machine-readable surface, and mixed-success batches use the same canonical tool-error envelope as other Coral MCP errors.

**Breaking MCP change:** existing MCP clients must update calls from `sql: "SELECT ..."` to `queries: ["SELECT ..."]`, and raw MCP clients should read results from `structuredContent`.

## What changed

- Adds a required `queries[]` input for the MCP `sql` tool, bounded from 1 to 10 non-blank SQL strings.
- Executes independent SQL queries concurrently while returning results in input order.
- Returns successful batches as the SQL batch object directly.
- Returns mixed-success batches as a canonical structured tool error with top-level `error` and the ordered SQL batch under `data`.
- Preserves episode metadata across every parallel backend SQL request without adding a new dependency.
- Keeps per-query errors on the existing `ToolError` vocabulary and shares the same schema helper between top-level and per-query errors.
- Updates the MCP guide for the new `queries[]`, `structuredContent`, and partial-failure contracts.

## Review follow-up

- Replaced the generic tool payload abstraction with an explicit SQL batch outcome in the MCP server.
- Added the `{ error, data }` partial-failure result shape instead of treating the batch itself as the top-level successful payload.
- Moved reusable JSON-schema object handling into a shared surface helper.
- Added input-schema `pattern` validation to match runtime non-blank SQL validation.
- Added coverage proving a batch `episode_id` reaches every backend SQL request.
- Avoided adding `futures`; each spawned query is wrapped in the existing episode-metadata scope.

## How to verify

```bash
cargo test -p coral-mcp
cargo test -p coral-cli --features cli-test-server --test mcp
make docs-check
make rust-checks
```

These passed locally before pushing the branch.

### DIFF
--- a/crates/coral-mcp/src/server.rs
+++ b/crates/coral-mcp/src/server.rs
@@ -38,7 +38,7 @@ use crate::{
         optional_episode_id_argument, required_string_argument, search_catalog_arguments,
         search_catalog_tool, search_catalog_value, sql_arguments, sql_tool, status_to_error_data,
         tables_resource, tables_resource_content, tool_error_from_status, tool_error_result,
-        tool_error_with_data_result, with_episode_id_argument,
+        with_episode_id_argument,
     },
     telemetry,
 };
@@ -782,14 +782,17 @@ fn finish_tool_call(
                 Ok(value) => value,
                 Err(status) => {
                     telemetry::record_tonic_status(span, &status);
-                    return Ok(tool_error_result(tool_error_from_status("Query", &status)));
+                    return Ok(tool_error_result(
+                        tool_error_from_status("Query", &status),
+                        None,
+                    ));
                 }
             };
             if batch.has_errors() {
                 telemetry::record_sql_batch_partial_failure(span);
-                Ok(tool_error_with_data_result(
+                Ok(tool_error_result(
                     batch.partial_failure_error(),
-                    serialized,
+                    Some(serialized),
                 ))
             } else {
                 telemetry::record_success(span);
@@ -798,9 +801,10 @@ fn finish_tool_call(
         }
         Ok(ToolCallOutcome::ToolError { operation, status }) => {
             telemetry::record_tonic_status(span, &status);
-            Ok(tool_error_result(tool_error_from_status(
-                operation, &status,
-            )))
+            Ok(tool_error_result(
+                tool_error_from_status(operation, &status),
+                None,
+            ))
         }
         Err(error) => {
             telemetry::record_protocol_error(span, &error);
--- a/crates/coral-mcp/src/server.rs
+++ b/crates/coral-mcp/src/server.rs
@@ -1,10 +1,5 @@
 //! RMCP server implementation for Coral's stdio MCP surface.
 
-use std::sync::{
-    Arc,
-    atomic::{AtomicUsize, Ordering},
-};
-
 use coral_api::v1::{
     CatalogItemKind as ProtoCatalogItemKind, DescribeTableRequest, DescribeTableResponse,
     ExecuteSqlRequest, ListCatalogRequest, ListCatalogResponse, ListColumnsRequest,
@@ -358,27 +353,16 @@ impl CoralMcpServer {
         queries: Vec<String>,
     ) -> Result<SqlBatchValue, tonic::Status> {
         let mut tasks = tokio::task::JoinSet::new();
-        let next_to_submit = Arc::new(AtomicUsize::new(0));
-        let submission_notify = Arc::new(tokio::sync::Notify::new());
         for (index, sql) in queries.into_iter().enumerate() {
             let server = self.clone();
-            let next_to_submit = Arc::clone(&next_to_submit);
-            let submission_notify = Arc::clone(&submission_notify);
-            tasks.spawn(async move {
-                while next_to_submit.load(Ordering::Acquire) != index {
-                    submission_notify.notified().await;
-                }
-                next_to_submit.fetch_add(1, Ordering::AcqRel);
-                submission_notify.notify_waiters();
-                server.execute_one_sql_query(index, sql).await
-            });
+            tasks.spawn(async move { server.execute_one_sql_query(index, sql).await });
         }
 
         let mut results = Vec::new();
         while let Some(joined) = tasks.join_next().await {
             results.push(joined.map_err(|error| tonic::Status::internal(error.to_string()))?);
         }
-        SqlBatchValue::from_unordered(results)
+        Ok(SqlBatchValue::from_unordered(results))
     }
 
     async fn open_episode(
@@ -809,11 +793,7 @@ fn finish_tool_call(
                 Ok(build_tool_result(payload.value))
             }
             ToolPayloadStatus::Error => {
-                telemetry::record_tool_payload_error(
-                    span,
-                    "sql_batch_partial_failure",
-                    "One or more SQL queries failed",
-                );
+                telemetry::record_sql_batch_partial_failure(span);
                 let mut result = CallToolResult::structured_error(payload.value);
                 result.content = Vec::new();
                 Ok(result)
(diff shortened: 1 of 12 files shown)


docs/guides/use-coral-over-mcp.mdx
```diff
@@ -234,13 +234,66 @@ Agents should treat the discovery tools as catalog helpers, not as replacement p
...
+  "queries": [
+    "SELECT table_name FROM coral.tables WHERE schema_name = 'github' ORDER BY table_name",
+    "SELECT column_name, data_type FROM coral.columns WHERE schema_name = 'github' AND table_name = 'pulls' ORDER BY ordinal_position"
+  ]
+}
+```
+
+Each entry must be one independent read-only SQL statement. Coral starts entries concurrently, so a later entry must not depend on rows, errors, or side effects from an earlier entry. If the work is dependent, combine it into one SQL statement or make separate tool calls after inspecting each result.
+
+`queries` must contain 1 to 10 non-blank strings. Missing `queries`, an empty array, more than 10 entries, non-string entries, and blank strings are MCP argument errors and do not dispatch any SQL.
```
AlbertQM, line 252: Thanks for clarifying this!

docs/guides/use-coral-over-mcp.mdx
```diff
@@ -234,13 +234,66 @@ Agents should treat the discovery tools as catalog helpers, not as replacement p
...
+
+<Note>
+This is a breaking MCP contract change: update existing calls from `sql: "SELECT ..."` to `queries: ["SELECT ..."]`.
+</Note>
```
AlbertQM: This statement should be a BREAKING CHANGE in the changelog (and it will be thanks to our machinery), rather than sit in this part of the docs 👀 

```suggestion
```

docs/reference/cli-reference.mdx
```diff
@@ -246,7 +246,7 @@ By default, this server exposes:
 
 | Type     | Name             | Description                                             |
 | -------- | ---------------- | ------------------------------------------------------- |
-| Tool     | `sql`            | Execute read-only SQL against the Coral database        |
+| Tool     | `sql`            | Execute 1 to 10 independent read-only SQL queries against the Coral database |
```
AlbertQM: These changes weren't here before, are we also updating the `sql` cli command?

docs/guides/use-coral-over-mcp.mdx
```diff
@@ -234,13 +234,66 @@ Agents should treat the discovery tools as catalog helpers, not as replacement p
...
+
 ### Searching a provider's data
 
 Some sources expose native search endpoints (for example, GitHub issue search) as table functions with `kind = 'search'`. Prefer these over scanning a regular table when the agent needs ranked, query-driven results. They return provider-ranked candidates, so agents should preserve any returned rank columns and use the returned ids plus catalog-described tables to fetch full detail rows when the search result itself is incomplete.
 
 ## Query result format
 
-The `sql` tool returns rows as a JSON array of objects. To preserve exact values across JSON parsers, Coral encodes precision-sensitive numeric types as JSON **strings** rather than JSON numbers:
+MCP tool results use `structured_content` as the only payload channel. The `content` array is empty for successful tool results and structured tool errors, so clients should not parse text content for rows or errors.
```
AlbertQM: What's a payload channel?

docs/guides/use-coral-over-mcp.mdx
```diff
@@ -234,13 +234,66 @@ Agents should treat the discovery tools as catalog helpers, not as replacement p
...
+        "grpc_code": "InvalidArgument",
+        "retryable": false,
+        "metadata": {}
+      }
+    }
+  ]
+}
+```
+
+`results` is ordered to match `queries`, and each item includes its zero-based `index`. Successful items have `status: "success"` and `rows`. Failed items have `status: "error"` and an `error` object. A batch with any failed item sets the MCP tool result error indicator while still returning the full structured batch for partial-success inspection.
```
AlbertQM: This sounds so LLM-y -- literally enumerating things we can see from the snippet above. Is it necessary?

crates/coral-mcp/src/server.rs
```diff
@@ -319,13 +334,35 @@ impl CoralMcpServer {
...
-    ) -> Result<Value, tonic::Status> {
-        serialize_tool_value(SqlRowsValue {
-            rows: self.query_rows(request).await?,
-        })
+        queries: Vec<String>,
+    ) -> Result<SqlBatchValue, tonic::Status> {
+        let mut tasks = tokio::task::JoinSet::new();
+        for (index, sql) in queries.into_iter().enumerate() {
+            let server = self.clone();
+            tasks.spawn(async move { server.execute_one_sql_query(index, sql).await });
```
Bradley-Butcher: Batched SQL is dropping episode attribution here.

`call_tool` wraps `dispatch_tool` in `with_episode_metadata(...)`, but `execute_sql_batch` then runs each query inside `JoinSet::spawn`. `with_episode_metadata` is task-local; `coral-client/src/propagation.rs` documents that it does not cross an interior `tokio::spawn` unless the caller scopes it again.

I verified this end-to-end on this branch with OTLP export:

- `coral.mcp.call_tool` for `sql` had `episode.id=ep_46d8a824684d43cf88491659171b3fb3`
- both downstream `coral.query` spans for `SELECT 1 AS one` and `SELECT 2 AS two` had empty `episode.id`

So the MCP layer sees the episode id, but the backend `ExecuteSql` requests do not. Please pass the episode metadata into each spawned batch task and call `with_episode_metadata(...)` inside the task, or use a concurrency shape that stays within the existing task-local scope. A regression test should assert that every query in a batch carries `coral-episode-id`.

Bradley-Butcher (commented):
Two contract comments from local smoke testing.

docs/guides/use-coral-over-mcp.mdx
```diff
@@ -234,13 +234,64 @@ Agents should treat the discovery tools as catalog helpers, not as replacement p
...
+
+
 ### Searching a provider's data
 
 Some sources expose native search endpoints (for example, GitHub issue search) as table functions with `kind = 'search'`. Prefer these over scanning a regular table when the agent needs ranked, query-driven results. They return provider-ranked candidates, so agents should preserve any returned rank columns and use the returned ids plus catalog-described tables to fetch full detail rows when the search result itself is incomplete.
 
 ## Query result format
 
-The `sql` tool returns rows as a JSON array of objects. To preserve exact values across JSON parsers, Coral encodes precision-sensitive numeric types as JSON **strings** rather than JSON numbers:
+MCP tool results use `structured_content` as the only payload channel. The `content` array is empty for successful tool results and structured tool errors, so clients should not parse text content for rows or errors.
```
Bradley-Butcher: This guide is describing the MCP JSON-RPC wire response, so this should use `structuredContent`, not `structured_content`.

I hit this in the smoke check too: the actual `tools/call` response had `structuredContent`. `structured_content` is the Rust SDK field name, and this line is part of the breaking client migration guidance.

Bradley-Butcher (approved):
LGTM. The episode attribution fix now re-scopes the metadata inside each spawned batch query task, and the MCP guide now uses the wire field name structuredContent. CI is green; I’m treating the remaining CLI-reference wording nit around queries[] discoverability as non-blocking.

crates/coral-mcp/src/surface/sql.rs
```diff
@@ -0,0 +1,214 @@
...
+                ));
+            }
+            Ok(query.to_string())
+        })
+        .collect::<Result<Vec<_>, _>>()?;
+    Ok(SqlArguments { queries })
+}
+
+pub(crate) fn sql_input_schema() -> Arc<Map<String, Value>> {
+    json_object_schema(&json!({
```
simonwhitaker, line 132: We should try to wean ourselves off these hard-coded `json!` blobs. Can you generate this from the underlying structs and enums instead, using `schemars`?

simonwhitaker (approved):

crates/coral-mcp/src/surface/sql.rs
jsummerfield: Ack, good shout. Will do in a follow up now.

PR #1399, opened 2026-06-26
feat(sources/core-v4/x): add X v4 source

kyracheng:
## Summary

- add a preview DSL v4 source spec for X/Twitter under `sources/core-v4/twitter_v4`
- point the OpenAPI surface at X API v2's published OpenAPI document
- configure bearer-token auth, X rate-limit headers, and smoke test queries for recent search, user posts, and mentions

## Validation

- `coral source lint sources/core-v4/twitter_v4/manifest.yaml`
- `CORAL_CONFIG_DIR=/private/tmp/coral-twitter-v4-test-3 X_ACCESS_TOKEN=fake coral source add --file sources/core-v4/twitter_v4/manifest.yaml`
  - source add/materialization succeeded against the live OpenAPI spec
  - generated `twitter_v4` with 81 tables/functions
  - declared query tests resolved to generated projections, then failed with expected 401s because the validation token was fake
- `python3 -c 'import yaml; yaml.safe_load(open("sources/core-v4/twitter_v4/manifest.yaml")); print("yaml ok")'`
- `git diff --check`

## Notes

- The source uses `X_ACCESS_TOKEN` rather than `X_BEARER_TOKEN` because X API v2 mixes app-only bearer-token endpoints with OAuth user-context endpoints. Endpoint access depends on the token type and X API tier.
- Current DSL v4 projection generation exposes some X endpoints with rough generated names, especially around path-collision cases. The smoke tests use names produced by materialization today.


sources/core-v4/twitter_v4/manifest.yaml
```diff
@@ -0,0 +1,35 @@
+name: twitter_v4
+dsl_version: 4
+description: Query posts, users, search results, timelines, mentions, and engagement from X (formerly Twitter).
+surfaces:
+  - id: rest
+    type: openapi
+    url: https://api.twitter.com/2/openapi.json
+    base_url: https://api.x.com
```
simonwhitaker: Please remove, this is already defined in the OpenAPI spec

sources/core-v4/twitter_v4/manifest.yaml
```diff
@@ -0,0 +1,35 @@
+name: twitter_v4
+dsl_version: 4
+description: Query posts, users, search results, timelines, mentions, and engagement from X (formerly Twitter).
+surfaces:
+  - id: rest
+    type: openapi
+    url: https://api.twitter.com/2/openapi.json
```
simonwhitaker:
```suggestion
    url: https://api.x.com/2/openapi.json
```

sources/core-v4/twitter_v4/manifest.yaml
```diff
@@ -0,0 +1,35 @@
...
+        from: literal
+        value: coral-dsl-v4
+    rate_limit:
+      remaining_header: x-rate-limit-remaining
+      reset_header: x-rate-limit-reset
+
+test_queries:
+  - SELECT id, text FROM twitter_v4.search_recents(query => 'from:XDevelopers', max_results => 10) LIMIT 5
+  - SELECT id, text FROM twitter_v4.path_2_users_tweets__b240539d WHERE id = '2244994945' AND max_results = 10 LIMIT 5
+  - SELECT id, text FROM twitter_v4.path_2_users_tweets__e4416c7d WHERE id = '2244994945' AND max_results = 10 LIMIT 5
```
simonwhitaker: I'd suggest removing this final test query, it doesn't add anything over the query on L34.

You might also consider removing the query on L34, too, and just keep the `search_recents(...)` query on L33.

sources/core-v4/twitter_v4/manifest.yaml
```diff
@@ -0,0 +1,35 @@
+name: twitter_v4
```
simonwhitaker: Suggest you rename this to `x_v4` rather than use the old name. (That means renaming the folder, too.)

simonwhitaker (changes requested):

simonwhitaker (approved):

PR #1403, opened 2026-06-26
feat(app): persist workspace state

sauldhernandez:
## Intent

Introduce the app-owned workspace persistence foundation without changing the CLI or MCP surface yet.

## What it enables

Adds explicit workspace records, workspace-scoped installed source catalog state, and a workspace manager that can create and remove workspace-owned config, artifacts, and credentials.

## Usage examples

Downstack PRs expose this through commands such as `coral workspace create work`, while app code can use `WorkspaceManager::create_workspace` and `WorkspaceManager::delete_workspace`.

### DIFF
--- a/crates/coral-app/src/state/config.rs
+++ b/crates/coral-app/src/state/config.rs
@@ -1,6 +1,8 @@
 //! Persists the installed source catalog in top-level `config.toml`.
 
-use std::collections::BTreeMap;
+#![expect(dead_code, reason = "used in next stack PR")]
+
+use std::collections::{BTreeMap, BTreeSet};
 
 use coral_engine::{DependentJoinConfig, DependentJoinSourceConfig, MemorySize, QueryMemoryConfig};
 use serde::{Deserialize, Serialize};
@@ -14,11 +16,13 @@ use crate::sources::model::{InstalledSource, SourceOrigin};
 use crate::state::AppStateLayout;
 use crate::storage::fs::{self as storage_fs, FileLock};
 use crate::workspaces::WorkspaceName;
+use crate::workspaces::{WorkspaceRecord, WorkspaceStore};
 
 #[derive(Debug, Clone)]
 pub(crate) struct AppConfig {
     version: u32,
     engine: PersistedEngineConfig,
+    workspaces: WorkspaceCatalog,
     catalog: SourceCatalog,
 }
 
@@ -27,12 +31,21 @@ impl Default for AppConfig {
         Self {
             version: default_config_version(),
             engine: PersistedEngineConfig::default(),
+            workspaces: WorkspaceCatalog::default(),
             catalog: SourceCatalog::default(),
         }
     }
 }
 
 impl AppConfig {
+    pub(crate) fn workspaces(&self) -> Vec<WorkspaceRecord> {
+        self.workspaces.list()
+    }
+
+    pub(crate) fn has_workspace(&self, workspace_name: &WorkspaceName) -> bool {
+        self.workspaces.contains(workspace_name)
+    }
+
     pub(crate) fn workspace_sources(&self, workspace_name: &WorkspaceName) -> Vec<InstalledSource> {
         self.catalog.workspace_sources(workspace_name)
     }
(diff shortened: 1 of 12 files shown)


sauldhernandez:
* **#1403** <a href="https://app.graphite.com/github/pr/withcoral/coral/1403?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1403?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `graphite-base/1403`


<h2></h2>

crates/coral-app/src/state/config.rs
```diff
@@ -469,20 +570,68 @@ impl ConfigStore {
...
+    config_store: ConfigStore,
+}
+
+impl ConfigWorkspaceStore {
+    pub(crate) fn new(config_store: ConfigStore) -> Self {
+        Self { config_store }
+    }
+}
+
+impl WorkspaceStore for ConfigWorkspaceStore {
```
jsummerfield: `ConfigWorkspaceStore` is a pure pass-through (four verbatim forwards to `ConfigStore`), so a reader chasing `delete_workspace` reads three identical signatures before the logic. The narrowed trait boundary is worth keeping, but you can `impl WorkspaceStore for ConfigStore` directly and delete the wrapper + its four bodies. (And with a single impl and no test fake, even `Arc<dyn WorkspaceStore>` could be concrete unless a second impl is genuinely coming.)

crates/coral-app/src/workspaces/manager.rs
```diff
@@ -0,0 +1,193 @@
...
+    }
+}
+
+struct WorkspaceCredentialSnapshot {
+    credential_set_id: CredentialSetId,
+    snapshot: CredentialMaterialSnapshot,
+}
+
+fn workspace_delete_backup_path(workspace_dir: &Path, workspace_name: &WorkspaceName) -> PathBuf {
+    workspace_dir.with_file_name(format!(
```
jsummerfield: This now exists in two managers and a test string. Extract one narrow `fs` helper (a `DirectoryBackupGuard` / `rename_with_rollback` owning the `.delete.rollback.{uuid}` name, rename, restore-on-failure, commit-on-success); leave the per-manager orchestration (credentials, store calls) alone. The two copies have already diverged on parent-dir cleanup, which is the failure mode of duplicating it.

jsummerfield:
Part of a software-design pass over the workspaces stack (#1403–#1414). The foundation is solid — `WorkspaceName` validation centralized, workspace as a first-class scoping key in config. A few structural notes, mostly small.

### 🟡 `ConfigWorkspaceStore` is a pass-through layer

`ConfigWorkspaceStore` (`state/config.rs:575–611`) forwards all four methods verbatim to identically-signed `ConfigStore` methods (`:478–518`), and it's the only `impl WorkspaceStore` (no test fake substitutes it). So `WorkspaceManager` traces `Arc<dyn WorkspaceStore>` → `ConfigWorkspaceStore` → `ConfigStore` for zero behavioral difference, and a reader chasing `delete_workspace` reads three identical signatures before the real logic. The trait usefully *narrows* `ConfigStore`'s broad surface to the four ops the manager needs — but the wrapper struct isn't what delivers that. → `impl WorkspaceStore for ConfigStore` directly and delete the wrapper + its four bodies. (With one impl and no fake, even `Arc<dyn WorkspaceStore>` could be concrete unless a second impl is genuinely coming.)

### 🟡 The `.delete.rollback.{uuid}` backup convention is duplicated across managers

The sibling-backup naming + rename-to-backup / restore-on-failure / commit-on-success dance (`workspace_delete_backup_path` + inline logic in `delete_workspace_state` + `restore_workspace_dir`, `workspaces/manager.rs:95–127, 238–258`) is a second hand-rolled copy of the same primitive already in `sources/manager.rs` (~`:509`), and the format string is re-encoded a third time in a test. The two copies have **already drifted** — the source-manager version cleans up empty parent dirs (`cleanup_empty_parent`), the workspace one doesn't — which is the classic failure mode of duplicating a crash-safety primitive. → extract one narrow fs helper (a `DirectoryBackupGuard` / `rename_with_rollback` owning the naming + rename + restore-on-failure + commit-on-success); leave the per-manager orchestration alone. (This feeds the broader deletion-architecture note on #1413 — the rollback primitive wants one owner.)

### 🟢 Hand-rolled credential rollback at each fallible step

`remove_workspace_credentials` (`manager.rs:144–180`) repeats `restore_workspace_credentials(...); return Err(error)` at every early return, and the `remove_material` arm has to push the just-removed snapshot *before* restoring — an ordering that's easy to get wrong as the loop grows. → fold push+restore+bail into one helper (a plain RAII `Drop` won't fit, since the rollback is itself fallible and order-sensitive). Low priority; correct today.

### 🟢 `!has_workspace → WorkspaceNotFound` guard duplicated ~6×

The guard is hand-rolled in four `config.rs` accessors (`list_workspace_sources`/`get_source`/`upsert_source`/`remove_source`) and twice more in `query/manager.rs`. It matters because the underlying accessors **fail open** — `SourceCatalog::workspace_sources` ends in `.unwrap_or_default()` (empty vec for a missing workspace), `get_source` returns `Option` → `SourceNotFound` — so a future accessor that forgets the guard silently treats an unknown workspace as empty/not-found instead of erroring. → one `AppConfig::require_workspace(name) -> Result<(), AppError>` the scoped methods call, so the "unknown workspace → WorkspaceNotFound" rule has one owner.

jsummerfield (approved):
Design-review (APoSD) re-review: the workspace-deletion rework resolved all prior findings cleanly — single commit-point + best-effort GC, shared `DirectoryBackup`/`WorkspacePaths`, `is_default()`/`require_workspace`/`WORKSPACE_SPAN_ATTRIBUTE` single owners, and whole-trace pruning with a production-shaped test. LGTM.

PR #1412, opened 2026-06-26
fix(app): add workspace trace pruning store

sauldhernandez:
## Intent

Add the local trace-store primitive needed to delete workspace-attributed trace history.

## What it enables

The local JSONL trace store can coordinate active writers, identify workspace spans and descendants, preserve other workspace spans in shared traces, ignore malformed records, and roll back file rewrites on in-process failures.

## Usage examples

App code can call `delete_workspace_traces(trace_store_dir, &workspace_name)` to remove local trace records attributed to a workspace.

PR #1411, opened 2026-06-26
refactor(app): narrow workspace manager artifact dependency

sauldhernandez:
## Intent

Narrow the workspace manager's dependency on app state layout before adding trace cleanup wiring.

## What it enables

The workspace manager now receives the specific workspace artifact root it needs, reducing constructor coupling and making later trace-store wiring more explicit.

## Usage examples

Server bootstrap passes `layout.workspaces_root()` to `WorkspaceManager` instead of handing over the broader layout object.

PR #1409, opened 2026-06-26
fix(app): harden workspace deletion rollback

sauldhernandez:
## Intent

Harden workspace deletion rollback before layering trace-history cleanup on top.

## What it enables

Workspace deletion now has stronger tests around rollback behavior, missing workspace handling, default workspace protection, and preservation of unrelated workspace state.

## Usage examples

If `coral workspace remove work` cannot finish deleting workspace-owned artifacts, the workspace config and credential material are restored instead of leaving a partial delete.

PR #1413, opened 2026-06-26
fix(app): prune traces during workspace remove

sauldhernandez:
## Intent

Wire local trace-history cleanup into workspace removal.

## What it enables

`workspace remove` now removes workspace-attributed local trace history when trace history is enabled, and it restores workspace config, artifact directories, and credentials if trace cleanup fails.

## Usage examples

`coral workspace remove work` deletes workspace `work` and prunes local trace records attributed to `work`; if trace files cannot be rewritten, the command fails and leaves `work` configured.

### DIFF
--- a/crates/coral-app/src/bootstrap/server.rs
+++ b/crates/coral-app/src/bootstrap/server.rs
@@ -57,7 +57,7 @@ use crate::state::ConfigStore;
 use crate::telemetry::TelemetryConfig;
 use crate::telemetry::service::TraceService;
 use crate::transport::GrpcMethodAnnotatedService;
-use crate::workspaces::{WorkspaceManager, WorkspaceService};
+use crate::workspaces::{WorkspaceLifecycleLock, WorkspaceManager, WorkspaceService};
 
 /// A static asset (e.g., a built SPA file) served on the same port as
 /// gRPC-Web.
@@ -262,20 +262,27 @@ impl ServerBuilder {
             self.config.enable_stderr_logs,
             internal_trace_store_dir.clone(),
         )?;
+        let installed_trace_store_dir = installed_trace_store
+            .as_ref()
+            .map(|store| store.dir.clone());
         let config_store = ConfigStore::new(layout.clone());
         let credential_config = CredentialStorageConfig::load(&layout)?;
         let credential_store =
             CredentialStore::with_preference(layout.clone(), credential_config.storage);
         let credential_manager = CredentialManager::new(credential_store);
+        let workspace_lifecycle_lock = WorkspaceLifecycleLock::default();
         let source_manager = SourceManager::new(
             config_store.clone(),
             credential_manager.clone(),
             layout.clone(),
+            workspace_lifecycle_lock.clone(),
         );
         let workspace_manager = WorkspaceManager::new(
             config_store.clone(),
             credential_manager.clone(),
             layout.clone(),
+            installed_trace_store_dir,
+            workspace_lifecycle_lock,
         );
         let feedback_manager =
             FeedbackManager::with_publisher(layout.clone(), self.config.feedback_publisher);
@@ -780,17 +787,18 @@ enabled = false
         let config_store = ConfigStore::new(layout.clone());
         let credential_store = CredentialStore::new(layout.clone());
         let credential_manager = CredentialManager::new(credential_store);
-        let source_manager = SourceManager::new(
+        let source_manager = SourceManager::new_for_tests(
             config_store.clone(),
             credential_manager.clone(),
             layout.clone(),
         );
         let feedback_manager = FeedbackManager::new(layout.clone());
         let episode_store = EpisodeStore::new(layout.clone());
-        let workspace_manager = WorkspaceManager::new(
+        let workspace_manager = WorkspaceManager::new_for_tests(
             config_store.clone(),
             credential_manager.clone(),
             layout.clone(),
+            None,
         );
         let query_manager = QueryManager::new(
             config_store,
@@ -1166,17 +1174,18 @@ tables:
         let config_store = ConfigStore::new(layout.clone());
         let credential_store = CredentialStore::new(layout.clone());
         let credential_manager = CredentialManager::new(credential_store);
-        let source_manager = SourceManager::new(
+        let source_manager = SourceManager::new_for_tests(
             config_store.clone(),
             credential_manager.clone(),
             layout.clone(),
         );
         let feedback_manager = FeedbackManager::new(layout.clone());
         let episode_store = EpisodeStore::new(layout.clone());
-        let workspace_manager = WorkspaceManager::new(
+        let workspace_manager = WorkspaceManager::new_for_tests(
             config_store.clone(),
             credential_manager.clone(),
             layout.clone(),
+            None,
         );
         let query_manager = QueryManager::new(
             config_store,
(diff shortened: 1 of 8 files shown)


sauldhernandez:
* **#1421** <a href="https://app.graphite.com/github/pr/withcoral/coral/1421?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1420** <a href="https://app.graphite.com/github/pr/withcoral/coral/1420?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1419** <a href="https://app.graphite.com/github/pr/withcoral/coral/1419?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1414** <a href="https://app.graphite.com/github/pr/withcoral/coral/1414?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1413** <a href="https://app.graphite.com/github/pr/withcoral/coral/1413?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1413?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1412** <a href="https://app.graphite.com/github/pr/withcoral/coral/1412?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1411** <a href="https://app.graphite.com/github/pr/withcoral/coral/1411?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1410** <a href="https://app.graphite.com/github/pr/withcoral/coral/1410?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1409** <a href="https://app.graphite.com/github/pr/withcoral/coral/1409?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1408** <a href="https://app.graphite.com/github/pr/withcoral/coral/1408?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1407** <a href="https://app.graphite.com/github/pr/withcoral/coral/1407?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1406** <a href="https://app.graphite.com/github/pr/withcoral/coral/1406?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1405** <a href="https://app.graphite.com/github/pr/withcoral/coral/1405?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1404** <a href="https://app.graphite.com/github/pr/withcoral/coral/1404?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1461** <a href="https://app.graphite.com/github/pr/withcoral/coral/1461?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

jsummerfield:
Reviewed the whole stack (#1403–#1414) for software-design/complexity. Overall it's a well-designed feature: workspace identity is a validated newtype owned in one place, config treats the workspace as a first-class scoping key, and the delete path has real rollback semantics. The notes below are about the **deletion architecture**, which is the one area I'd want to firm up before the stack settles — this PR is where the orchestration lands, so I'm putting the main comment here.

### 🟠 The deletion story has no single owner of the atomicity contract

A single `delete_workspace` touches four subsystems, and the *rollback* knowledge is duplicated into three of them independently:

| Participant | Work | Rollback |
|---|---|---|
| Traces | `telemetry::delete_workspace_traces` → ~6 fns in `local_store.rs` | internal to the call, **not exposed** — the manager can't undo it |
| Credentials | `remove_workspace_credentials` (`manager.rs:140`) | `restore_workspace_credentials` (`manager.rs:183`) |
| Workspace dir | inline in `delete_workspace_state` (`manager.rs:94–108`) | free fns `workspace_delete_backup_path` + `restore_workspace_dir` (`manager.rs:238–258`) |
| Config entry (authority) | `ConfigStore::delete_workspace` (`state/config.rs:497`) | none — its failure is what *triggers* the others' rollback |

So to answer "is delete atomic?" a reader has to hold all four in their head plus the orchestration order, and notice that one participant (traces) silently has no undo. That diffusion is the root cause of the two concrete issues below — they aren't independent bugs, they're symptoms of the deletion concept not having a home.

**What I'd aim for** — *not* a transaction framework (over-kill for a single-user local app), but one commit point with everything else falling out of it:

1. Validate + guard **once**.
2. The **config-catalog entry is the source of truth** — removing it is *the* commit.
3. Everything else (dir, credentials, traces) is cleanup of an already-gone workspace → run it **after** the commit, **best-effort, log-on-failure**.

That collapses the flow to `guard → remove config entry → best-effort GC`, and most of the rollback machinery (`restore_workspace_dir`, `restore_workspace_credentials`, the credential snapshot dance, the backup-rename) **disappears** — so it's *less* code, and `delete_workspace` gets a one-sentence contract: *"removes the workspace; on success it's gone, on failure nothing changed; leftover artifacts are cleaned opportunistically."* Both issues below evaporate as a side effect. (If "no orphaned artifacts ever" is a hard requirement, the alternative is an explicit compensating-action scope where traces structurally *can't* register an undo — which makes "irreversible last" enforced rather than remembered. I'd default to the best-effort version.)

### 🔴 Most acute symptom: ordering makes rollback incoherent

`delete_workspace` (`manager.rs:68–80`) runs the **irreversible** step first:

```rust
self.delete_workspace_traces(workspace_name).await?;   // l.78 — no restore counterpart
self.delete_workspace_state(workspace_name)            // l.79 — fully rolled back on failure
```

`delete_workspace_state` is carefully transactional (snapshots/restores credentials, renames the dir to a `.delete.rollback.<uuid>` backup it restores at `:104–136`). But `delete_workspace_traces` has no restore the manager can call — the snapshots inside `rewrite_trace_files_excluding_span_keys` only cover a mid-rewrite failure and are dropped on `Ok`. So if `store.delete_workspace` then fails, credentials + dir are restored and **the workspace survives with its trace history already destroyed** — exactly the partial-failure case the rollback machinery exists to prevent. Under the reframing above this is free; as a minimal fix, prune traces *last* (after the state delete commits), best-effort.

Two minor companions in the same spot: trace deletion runs **outside** `lifecycle_lock` (only taken inside `delete_workspace_state` at `:86`, unlike `create_workspace`); and the load-bearing ordering has no explanatory comment.

### 🟢 Smaller items (symptoms of the same diffusion)

- **Default-workspace guard is triplicated, and one copy is dead.** The rule lives at `manager.rs:72`, `manager.rs:87`, and `config.rs:502`, in two idioms (`.as_str() == DEFAULT_WORKSPACE_ID` vs `== &WorkspaceName::default()`). `delete_workspace_state` is private and only reached past the `:72` check, so the `:87` guard is unreachable. → drop the inner guard; give `WorkspaceName` a single `is_default()`.
- **`DeletedWorkspaceRecord` is a single-field wrapper** (`model.rs:11`) introduced here by widening the store signature from `Option<WorkspaceRecord>`. It's constructed once and unwrapped immediately (`manager.rs:112`), never reaches proto, and its doc ("metadata captured before deletion") overstates — it holds only the name. → return `Option<WorkspaceRecord>`, or actually capture the discarded removed-trace count so it earns its keep.
- **Constructor naming is backwards:** production uses `new_with_lifecycle_lock` while bare `new` is `#[cfg(test)]` (same in `SourceManager`). A reader scanning for "how is this built in the app?" lands on `new` and finds it test-only. → let production own `new`; rename the test ctor.
- **`delete_workspace_state` is a vague name** — "state" is a catch-all for credentials + dir + config entry. A one-line doc enumerating what it removes (and that it's the rollback-covered half) would do it.

Happy to pair on the reframing if useful.

sauldhernandez:
@jsummerfield comments addressed, ready for re-review

jsummerfield (approved):
Design-review (APoSD) re-review: the workspace-deletion rework resolved all prior findings cleanly — single commit-point + best-effort GC, shared `DirectoryBackup`/`WorkspacePaths`, `is_default()`/`require_workspace`/`WORKSPACE_SPAN_ATTRIBUTE` single owners, and whole-trace pruning with a production-shaped test. LGTM.

sauldhernandez:
### Merge activity

* **Jul 1, 9:30 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1413).
* **Jul 1, 10:36 AM UTC**: [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1413) rebased this pull request as part of a merge.
* **Jul 1, 10:41 AM UTC**: @sauldhernandez merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1413).

PR #1431, opened 2026-06-30
refactor(mcp): align tool contracts with surface owners

jsummerfield:
## Summary

- Generate MCP tool input/output schemas from typed surface structs and enums instead of hard-coded JSON blobs.
- Move SQL, catalog, episode, feedback, tool-name, and shared argument contracts into the surface modules that enforce them.
- Share pagination defaults, episode-id validation, and structured tool-error envelopes so schemas and runtime behavior stay aligned.
- Fix follow-up schema drift and argument parsing issues found during review, including preserving exact MCP string arguments.

## Validation

- `cargo test -p coral-mcp --lib`
- `cargo test -p coral-cli --features cli-test-server --test mcp mcp_stdio_raw_tools_list_advertises_client_compatible_schemas`
- `cargo clippy -p coral-mcp --all-targets --all-features -- -D warnings`
- `make rust-checks`

### DIFF
--- a/crates/coral-mcp/src/surface/catalog.rs
+++ b/crates/coral-mcp/src/surface/catalog.rs
@@ -1,17 +1,26 @@
 use coral_api::v1::{
     ColumnSearchResult, DescribeTableResponse, ListCatalogResponse, ListColumnsResponse,
-    SearchCatalogResponse, Table as ProtoTable, TableFunction as ProtoTableFunction,
-    TableFunctionArgument as ProtoTableFunctionArgument,
+    PaginationResponse, SearchCatalogResponse, Table as ProtoTable,
+    TableFunction as ProtoTableFunction, TableFunctionArgument as ProtoTableFunctionArgument,
     TableFunctionResultColumn as ProtoTableFunctionResultColumn, TableSummary as ProtoTableSummary,
     catalog_item,
 };
-use serde::Serialize;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
 use serde_json::{Map, Value};
+use std::sync::Arc;
 
+use super::schema::tool_output_schema;
 use super::values::{
-    format_schema_table_equivalent, format_sql_identifier, insert_pagination_fields,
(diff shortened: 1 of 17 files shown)


crates/coral-mcp/src/surface/catalog.rs
```diff
@@ -1,94 +1,376 @@
...
+    pub(crate) kind: Option<CatalogToolKind>,
+    #[schemars(flatten, with = "DefaultPaginationInput")]
+    pub(crate) pagination: Pagination,
+}
+
+#[derive(JsonSchema)]
+pub(crate) struct SearchCatalogArguments {
+    #[schemars(
+        length(min = 1),
+        pattern(r"\S"),
```
simonwhitaker: This looks wrong; if I'm understanding correctly, this will match any value with at least one non-space character in it. So "!", "hello world", "🐘🐟" would all validate. Only strings consisting entirely of whitespace characters will fail to validate.

If you want to enforce "no whitespace allowed", you could use `r"^\S*$"`. Or you might want to be more specific, e.g. `r"^[a-zA-Z0-9_-]+"`.

There are 10 more of these, I won't comment on the rest but the same applies to all.

simonwhitaker (approved):

crates/coral-mcp/src/surface/catalog.rs
simonwhitaker: (I actually wrote this comment against `DescribeTableArguments.schema`, then moved it here because this was the first instance of `r"\S"`. But for this regex argument, `\S` *might* be intentional? Maybe you don't want to be in the business of writing a regex to validate regexes. 😱 But the general point holds for many of the other uses of this pattern (e.g. `DescribeTableArguments.schema`), and you *might* want to change this one, too.)

jsummerfield: This is naff, thanks for spotting. Will remove, since schemars already does what we need inferring from stype String + min_length = 1

PR #1468, opened 2026-07-01
feat(engine): support named query parameters

Bradley-Butcher:
## Summary
- Add engine support for named SQL parameters such as `$author`.
- Introduce `QueryParameters` plus Coral-owned `QueryParameterValue` values.
- Convert Coral parameter values to DataFusion values inside `coral-engine`, during planning, explain, and execution.
- Extend source-scoped table-function argument binding so table-function calls can receive parameter-backed values.

## Why
Saved functions need stable parameter binding before the higher-level function lifecycle lands. This PR adds the lower-level engine capability first: callers provide typed Coral parameter values, and the engine binds those values through DataFusion.

## Behavior
Existing SQL without parameters should behave the same.

New supported shape:

```sql
select *
from github.pulls
where user__login = $author
```

The caller supplies `$author` separately through `QueryParameters`.

Coral accepts the v1 scalar subset we support end to end: strings, integers, floats, booleans, and typed NULLs for those families. Unsupported DataFusion scalar variants are not part of the public app-to-engine contract.

## Assumptions
- Parameters are value parameters only. They do not replace identifiers, table names, schemas, SQL fragments, or operators.
- `QueryParameterValue` is Coral-owned. DataFusion conversion stays behind the engine boundary.
- This PR only adds engine support. It does not add saved functions, app lifecycle APIs, CLI commands, or MCP presentation.

## Validation
- `cargo test -p coral-engine --test engine query_parameters --all-features --locked`
- From the top of the stack: `cargo test -p coral-engine --test engine udf --all-features --locked`
- From the top of the stack: `cargo test -p coral-app functions --all-features --locked`
- From the top of the stack: installed `coral`, added a SQL function, listed it, and queried live GitHub data through `functions.open_coral_prs_by_author(author => 'Bradley-Butcher')`.

### DIFF
--- a/crates/coral-engine/src/contracts/query.rs
+++ b/crates/coral-engine/src/contracts/query.rs
@@ -371,26 +371,145 @@ impl QueryRuntimeContext {
 
 /// Named SQL query parameters, keyed by parameter name without the `$`
 /// prefix: binding `owner` supplies `$owner` in the statement.
-pub type QueryParameters = BTreeMap<String, QueryParameterValue>;
-
-/// Value bound to one named SQL query parameter (`$name`).
 ///
-/// Values are data, never SQL text: they bind into the planned statement
-/// through `DataFusion` parameter substitution, so no quoting or escaping
-/// rules apply anywhere in the request path.
+/// Values are typed Coral scalar values. Callers should treat values as data,
+/// never SQL text.
+#[derive(Debug, Clone, Default, PartialEq)]
+pub struct QueryParameters {
+    values: BTreeMap<String, QueryParameterValue>,
+}
+
+impl QueryParameters {
+    /// Builds an empty parameter set.
+    #[must_use]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Returns true when no parameters are present.
+    #[must_use]
+    pub fn is_empty(&self) -> bool {
+        self.values.is_empty()
+    }
+
+    /// Inserts one named parameter value and returns the previous value.
+    pub fn insert(
+        &mut self,
+        name: impl Into<String>,
+        value: QueryParameterValue,
+    ) -> Option<QueryParameterValue> {
+        self.values.insert(name.into(), value)
+    }
+
+    /// Iterates over parameter names and values in deterministic order.
+    pub fn iter(&self) -> impl Iterator<Item = (&String, &QueryParameterValue)> {
+        self.values.iter()
+    }
+
+    /// Iterates over parameter names in deterministic order.
+    pub fn keys(&self) -> impl Iterator<Item = &String> {
+        self.values.keys()
+    }
+}
+
+impl From<BTreeMap<String, QueryParameterValue>> for QueryParameters {
+    fn from(values: BTreeMap<String, QueryParameterValue>) -> Self {
+        Self { values }
+    }
+}
+
+impl<const N: usize> From<[(String, QueryParameterValue); N]> for QueryParameters {
+    fn from(values: [(String, QueryParameterValue); N]) -> Self {
+        Self {
+            values: BTreeMap::from(values),
+        }
+    }
+}
+
+impl FromIterator<(String, QueryParameterValue)> for QueryParameters {
+    fn from_iter<T: IntoIterator<Item = (String, QueryParameterValue)>>(iter: T) -> Self {
+        Self {
+            values: iter.into_iter().collect(),
+        }
+    }
+}
+
+/// Scalar types supported by named SQL query parameters.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum QueryParameterType {
+    /// UTF-8 string.
+    String,
+    /// 64-bit signed integer.
+    Integer,
+    /// 64-bit floating point value.
+    Float,
+    /// Boolean.
+    Boolean,
+}
+
+/// One typed SQL query parameter value.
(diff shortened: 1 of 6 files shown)


Bradley-Butcher:
* **#1535** <a href="https://app.graphite.com/github/pr/withcoral/coral/1535?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1534** <a href="https://app.graphite.com/github/pr/withcoral/coral/1534?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1533** <a href="https://app.graphite.com/github/pr/withcoral/coral/1533?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1532** <a href="https://app.graphite.com/github/pr/withcoral/coral/1532?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1531** <a href="https://app.graphite.com/github/pr/withcoral/coral/1531?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1530** <a href="https://app.graphite.com/github/pr/withcoral/coral/1530?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1529** <a href="https://app.graphite.com/github/pr/withcoral/coral/1529?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1501** <a href="https://app.graphite.com/github/pr/withcoral/coral/1501?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1490** <a href="https://app.graphite.com/github/pr/withcoral/coral/1490?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1473** <a href="https://app.graphite.com/github/pr/withcoral/coral/1473?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1468** <a href="https://app.graphite.com/github/pr/withcoral/coral/1468?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1468?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-engine/src/runtime/query.rs
```diff
@@ -605,8 +622,139 @@ impl QueryRuntimeAdapter {
...
+fn validate_query_parameter_value(name: &str, value: &ScalarValue) -> Result<(), DataFusionError> {
+    match value {
+        ScalarValue::Utf8(_)
+        | ScalarValue::LargeUtf8(_)
+        | ScalarValue::Utf8View(_)
+        | ScalarValue::Int8(_)
+        | ScalarValue::Int16(_)
+        | ScalarValue::Int32(_)
+        | ScalarValue::Int64(_)
+        | ScalarValue::UInt8(_)
```
jsummerfield: I think this validation list needs to match every surface that can consume a bound parameter. Right now this accepts `UInt*`, `Int8`, `Int16`, and `Utf8View`, but HTTP source-function args still stringify through `backends/shared/filter_expr.rs::literal_to_string`, which does not handle those variants. That means an accepted value can work in ordinary SQL but fail later for `some_func(id => $id)` with “argument must be a literal.” Can we either extend the shared literal conversion to the same subset, or narrow this validator to what all bound surfaces actually support?

crates/coral-engine/src/contracts/query.rs
```diff
@@ -369,6 +370,67 @@ impl QueryRuntimeContext {
...
 }
 
+/// Named SQL query parameters, keyed by parameter name without the `$`
+/// prefix: binding `owner` supplies `$owner` in the statement.
+///
+/// Values are native `DataFusion` scalar values. Coral validates the supported
+/// subset before planning; callers should treat values as data, never SQL text.
+#[derive(Debug, Clone, Default, PartialEq)]
+pub struct QueryParameters {
+    values: BTreeMap<String, ScalarValue>,
```
jsummerfield: This makes `datafusion::ScalarValue` part of the public app-to-engine parameter contract. I would keep the DataFusion conversion behind the engine boundary and expose a Coral-owned value enum or constructors instead. Otherwise the saved-function/app layer has to learn DataFusion’s scalar variants and typed-null model just to pass user values, and this API becomes harder to change if we later want a transport-shaped parameter contract.

Bradley-Butcher: I was gaslit by Codex

Bradley-Butcher: I changed this so QueryParameters now carries a Coral-owned QueryParameterValue enum instead of exposing datafusion::ScalarValue.
The app/saved-function layer now passes Coral values only: string, integer, float, boolean, or typed nulls for those families. coral-engine converts those to DataFusion ScalarValue internally at the binding point. That keeps DataFusion’s wider scalar model and typed-null details behind the engine boundary.

crates/coral-engine/src/runtime/query.rs
Bradley-Butcher: I narrowed the public parameter contract to the subset Coral supports end to end: string, integer, float, boolean, and typed nulls for those families.
That removes caller-provided UInt*, narrow integer, Float32, Utf8View, etc. from the parameter surface. UDF inference may still observe DataFusion string representations like Utf8View, but those are mapped to Coral’s string kind internally; callers still provide QueryParameterValue::String.

jsummerfield (approved):

PR #1470, opened 2026-07-01
feat(search): add source scan observation hook

mystic123:
## Summary

- Adds `EngineExtensions.source_observation_publishers` and a minimal `SourceScanObservation` contract for non-blocking source-scan observation.
- Wires HTTP and MCP table/function scans through shared `JsonExec` so publishers see typed, mapped source batches before DataFusion projection.
- Preserves app layering: app code installs publishers through `EngineExtensionsProvider`; MCP/CLI adapters stay thin and only pass through engine metadata.
- Adds regression tests proving projected SQL output can be narrower than the observed source-scan batch for both HTTP and MCP sources.

## Deliberately out of scope

- No observed-value storage, SQLite projection, queue, retrieval, retention, governance, or shutdown drainage.
- No adapter-owned observation logic beyond passing surface identity and publishers into shared execution.
- No observation-stage/scope classification; later ingestion indexes observed positive values rather than treating events as completeness evidence.
- Fallback runtime retries may re-observe a source scan if the primary runtime partially executed first; the later ingestion/indexer PR must be idempotent.

## Verification

- `cargo clippy -p coral-engine --all-targets --all-features --locked -- -D warnings`
- `cargo test -p coral-engine source_scan_observation --lib`
- `cargo test -p coral-app query::extensions::tests::provider_lists_merge_source_observation_publishers_in_call_order --lib`
- `cargo check -p coral-app --lib`

## Feature flag

Defines the canonical default-off `observed_values_search` runtime feature (`coral features enable observed_values_search`, with `--enable-observed-values-search` and `--disable-observed-values-search` process overrides). The generic observation hook remains available, but it is inert until an enabled app lifecycle installs observed-value publishers.

PR #1464, opened 2026-07-01
feat(search): add sqlite search store

mystic123:
## Summary

This PR adds the local SQLite storage layer for Universal Search. It sits between #1372 and #1423.

It is intentionally a substrate PR: it creates the workspace-scoped SQLite store and catalog index primitives, but it does not wire search into CLI, MCP, or the catalog provider yet.

## What This Adds

- Adds one per-workspace SQLite search database at `workspaces/<workspace>/search/search.sqlite3`.
- Keeps workspace identity bound to `SqliteSearchStore`, so catalog callers use operation-shaped methods instead of passing raw SQLite connections plus a separate workspace name.
- Creates the SQLite file through Coral's private-file helper before opening it with `rusqlite`.
- Adds `SqliteSearchStore` for directory creation, connection setup, SQLite pragmas, FTS5/trigram capability checks, schema versioning, and catalog projection operations.
- Adds `SqliteCatalogIndex` for the SQLite-specific catalog metadata projection queries.
- Keeps SQLite DTOs separate from catalog provider/result types.
- Constrains stored catalog vocabularies (`doc_kind`, `surface_kind`, `field_role`) at the schema boundary and rejects unknown values on decode.
- Uses FTS5 trigram search for normal text matching.
- Uses exact/prefix fallback for short identifiers like `id` and `q`.
- Escapes `%`, `_`, and `\` in prefix fallback queries so user text is treated literally.
- Normalizes search terms before retrieval and ignores empty terms.
- Reports matched catalog fields, including `field_name` and `surface_name`.
- Probes one row past each retrieval limit so `retrieval_limited` means real truncation.
- Rejects newer SQLite schema versions instead of silently downgrading them.

## Important Boundary

SQLite only returns a bounded candidate window. It does not own final product ranking.

Provider-specific ranking, including exact-vs-prefix relevance decisions, lives in #1423. This avoids turning the storage layer into a second search product with hidden heuristics.

This PR is still metadata-only. Observed-values capture/search is intentionally not enabled here; that layer needs its own credential/account scoping, retention, and read-governance contract before raw source values are collected.

## Not In This PR

- No catalog provider wiring.
- No CLI or MCP search surface.
- No observed-values storage or queueing.
- No native fanout.

## Validation

- `cargo fmt --check`
- `cargo test -p coral-app search:: --lib`
- `cargo clippy -p coral-app --all-targets --all-features --locked -- -D warnings`

PR #1462, opened 2026-07-01
feat(search): expose metadata search in CLI and MCP

mystic123:
## Summary

- add a shared Universal Search response renderer in `coral-client` for thin local adapters
- add `coral search [--json] [--limit] <query...>` backed by `SearchService.Search`
- add an MCP `search` tool that calls `SearchService.Search` and returns typed results, provider statuses, and truncation metadata
- update docs, README, the MCP guide, and the maintained Coral skill for the new MCP `search` tool and removed `search_catalog` tool

## Stack

- follows #1423 (`pawel/sqlite-search-catalog`)

## Tests

- `cargo fmt --check`
- `cargo run -p coral-cli -- search --help`
- `make docs-check`
- `cargo test -p coral-mcp`
- `cargo test -p coral-cli --features cli-test-server search`

PR #1423, opened 2026-06-30
feat(search): add sqlite catalog provider

mystic123:
## Summary

This is the catalog-provider PR in the SQLite-only Universal Search stack. It now sits on top of #1464, which owns the SQLite store and pure catalog index primitives.

- Builds `CatalogSearchSnapshot` from the existing app/engine catalog view. It does not load source manifests directly or call providers.
- Converts that snapshot into the storage DTO expected by `SqliteCatalogIndex`.
- Adds `CatalogMetadataProvider`, which opens `SqliteSearchStore` for the request workspace, uses a request-scoped SQLite connection, refreshes the catalog projection, searches it, and maps hits into typed Universal Search results.
- Wires catalog metadata into `SearchManager`/`UniversalSearchEngine`.
- Returns typed `catalog_metadata` and `column_hint` results. Native provider search functions are discovered as ordinary `CatalogMetadata(TableFunction(kind = SEARCH))` hits, not executed and not emitted as a separate native-search-path payload.

## Boundaries

This PR deliberately does not add CLI/MCP surfaces, observed-value ingestion, queues, drainage, or native fanout execution. Observed values and native fanout still report explicit disabled/not-enabled provider statuses. Active native fanout execution will use a later provider/result contract.

## Validation

- `cargo fmt`
- `cargo test -p coral-app search::catalog`
- `cargo test -p coral-app --test grpc search`

PR #1472, opened 2026-07-01
feat(search): enqueue observed values in sqlite

mystic123:
## Summary

- adds the durable SQLite observed-value substrate: migration, source/workspace generations, canonical rows, FTS rows, queue jobs, and enqueue generation checks
- applies bounded collection plus best-effort structured secret suppression for JSON, URL-query, and form-encoded values before persistence; mandatory sensitive-name/value heuristics remain defense in depth rather than a DLP guarantee
- publishes source-scan observations through separate writer, publisher, and source-scope modules while treating runtime-contract and credential identities as opaque app-supplied inputs
- removes the hand-maintained semantic runtime-contract mirror from this storage PR
- repairs migrations first, then resets only irreparable observed-specific objects while preserving the shared SQLite file and catalog state
- keeps this PR inert: no source-authored indexing policy, QueryManager wiring, projection drainage, provider registration, ranking, or user-visible search results

## Commit layout

1. `feat(search): add observed queue storage` owns the migration, queue/store primitives, generations, and storage tests
2. `feat(search): collect governed observed values` owns collection budgets and sensitive-column/value suppression
3. `feat(search): publish source scan observations` owns writer lifecycle, publisher integration, and opaque source-scope derivation
4. Follow-up review commits add structured sanitization, defer semantic scope identity to the app runtime boundary, and recover from irreparable observed duplicates without deleting catalog state

The previous monolithic `source_scan.rs` is split into `observed/writer.rs`, `observed/publisher.rs`, `observed/source_scope.rs`, and focused collection helpers so lifecycle, publication, storage identity, and privacy behavior can be reviewed independently.

## Split boundary

Draft PR #1761 owns source-authored `do_not_index` policy. PR #1480 owns the app runtime-contract fingerprint, credential revision, and lifecycle wiring. Projection, storage-ceiling governance, retrieval policy, provider registration, and public maintenance remain in later stack slices.

## Validation

- focused observed collector, source-scope, queue, migration, and store tests
- strict Clippy and rustdoc through top-of-stack `make rust-checks`
- isolated real-source observation/search smoke through the final stack

## Feature flag

The observed SQLite substrate and migrations may exist while `observed_values_search` is disabled, but no writer is constructed and no observations are enqueued. This PR remains behaviorally dormant until the gated lifecycle wiring in #1480.

### DIFF
--- a/crates/coral-app/src/search/observed/collector.rs
+++ b/crates/coral-app/src/search/observed/collector.rs
@@ -16,9 +16,7 @@ use arrow::record_batch::RecordBatch;
 use arrow::util::display::array_value_to_string;
 
 use crate::hash::sha256_hex;
-use crate::search::observed::sensitive::{
-    exclude_sensitive_json_fields, is_sensitive_column, is_sensitive_value,
-};
+use crate::search::observed::sensitive::{is_sensitive_column, sanitize_observed_value};
 use crate::search::observed::sqlite_queue::{ObservedValueCandidate, ObservedValuesQueuePayload};
 
 #[derive(Debug, Clone, Default)]
@@ -54,12 +52,11 @@ impl ObservedValuesCollector {
                 ) else {
                     continue;
                 };
-                let Some(display_value) = exclude_sensitive_json_fields(display_value) else {
+                let Some(display_value) =
+                    sanitize_observed_value(display_value, self.budget.value_bytes_limit)
+                else {
                     continue;
                 };
-                if is_sensitive_value(&display_value) {
-                    continue;
-                }
                 let search_text = normalize_search_text(&display_value);
                 if search_text.is_empty() {
                     continue;
@@ -251,7 +248,7 @@ mod tests {
     }
 
     #[test]
-    fn suppresses_embedded_sensitive_values() {
+    fn suppresses_embedded_sensitive_values_and_preserves_safe_url_content() {
         let batch = batch(
             ["note"],
             [vec![
@@ -267,7 +264,10 @@ mod tests {
             .map(|candidate| candidate.display_value.as_str())
             .collect::<Vec<_>>();
 
-        assert_eq!(values, ["Ticket SK-101 is public"]);
+        assert_eq!(
+            values,
+            ["https://example.test/callback", "Ticket SK-101 is public"]
+        );
     }
 
     #[test]
--- a/crates/coral-app/src/search/observed/collector.rs
+++ b/crates/coral-app/src/search/observed/collector.rs
@@ -12,22 +12,8 @@ use arrow::record_batch::RecordBatch;
 use arrow::util::display::array_value_to_string;
 
 use crate::hash::sha256_hex;
+use crate::search::observed::sensitive::{is_sensitive_column, is_sensitive_value};
 use crate::search::observed::sqlite_queue::{ObservedValueCandidate, ObservedValuesQueuePayload};
-
-const SENSITIVE_COLUMN_NAMES: &[&str] = &[
-    "apikey",
-    "authorization",
-    "authtoken",
-    "cookie",
-    "credential",
-    "password",
-    "passwd",
-    "privatekey",
-    "refreshtoken",
-    "secret",
-    "session",
-    "token",
-];
 
 #[derive(Debug, Clone, Default)]
 pub(crate) struct ObservedValuesCollector {
@@ -106,13 +92,18 @@ pub(crate) struct ObservedValuesCollectionBudget {
     pub(crate) job_bytes_limit: usize,
 }
 
+const DEFAULT_CANDIDATE_LIMIT: usize = 10_000;
+const DEFAULT_CANDIDATE_BYTES_LIMIT: usize = 8 * 1024 * 1024;
+const DEFAULT_VALUE_BYTES_LIMIT: usize = 4 * 1024;
+const DEFAULT_JOB_BYTES_LIMIT: usize = 1024 * 1024;
+
 impl Default for ObservedValuesCollectionBudget {
     fn default() -> Self {
         Self {
-            candidate_limit: 10_000,
-            candidate_bytes_limit: 8 * 1024 * 1024,
-            value_bytes_limit: 4 * 1024,
-            job_bytes_limit: 1024 * 1024,
+            candidate_limit: DEFAULT_CANDIDATE_LIMIT,
+            candidate_bytes_limit: DEFAULT_CANDIDATE_BYTES_LIMIT,
+            value_bytes_limit: DEFAULT_VALUE_BYTES_LIMIT,
+            job_bytes_limit: DEFAULT_JOB_BYTES_LIMIT,
         }
     }
 }
@@ -146,82 +137,6 @@ fn normalize_search_text(value: &str) -> String {
         .collect::<Vec<_>>()
         .join(" ")
         .to_lowercase()
-}
-
-fn is_sensitive_column(column_name: &str) -> bool {
-    let normalized = column_name
-        .chars()
-        .filter(char::is_ascii_alphanumeric)
-        .collect::<String>()
-        .to_ascii_lowercase();
-    SENSITIVE_COLUMN_NAMES
-        .iter()
-        .any(|name| normalized.contains(name))
-}
-
-fn is_sensitive_value(value: &str) -> bool {
-    let trimmed = value.trim();
-    if trimmed.contains("-----BEGIN ") || trimmed.contains(" PRIVATE KEY-----") {
-        return true;
-    }
-    is_sensitive_token(trimmed)
-        || contains_sensitive_token(trimmed)
-        || contains_sensitive_key_value_pair(trimmed)
-}
-
-fn is_sensitive_token(value: &str) -> bool {
-    let lower = value.to_ascii_lowercase();
-    if (lower.starts_with("sk-") && value.len() >= 20)
-        || lower.starts_with("ghp_")
-        || lower.starts_with("github_pat_")
-        || lower.starts_with("xoxb-")
-        || lower.starts_with("xoxp-")
-        || lower.starts_with("xoxa-")
-        || lower.starts_with("ya29.")
-    {
-        return true;
-    }
-    looks_like_jwt(value)
-}
-
-fn contains_sensitive_token(value: &str) -> bool {
-    value
-        .split(|ch: char| !(ch.is_ascii_alphanumeric() || matches!(ch, '-' | '_' | '.')))
-        .any(is_sensitive_token)
-}
-
-fn contains_sensitive_key_value_pair(value: &str) -> bool {
-    value
-        .split(|ch: char| {
-            matches!(
-                ch,
-                '?' | '&' | ';' | ',' | '"' | '\'' | '{' | '}' | '[' | ']' | ' ' | '\t' | '\n'
-            )
-        })
-        .filter_map(|segment| segment.split_once('=').or_else(|| segment.split_once(':')))
-        .any(|(key, value)| !value.trim().is_empty() && is_sensitive_column(key))
-}
-
-fn looks_like_jwt(value: &str) -> bool {
-    let mut parts = value.split('.');
-    let Some(header) = parts.next() else {
-        return false;
-    };
-    let Some(payload) = parts.next() else {
-        return false;
-    };
-    let Some(signature) = parts.next() else {
-        return false;
-    };
-    if parts.next().is_some() {
-        return false;
-    }
-    [header, payload, signature].iter().all(|part| {
-        part.len() >= 8
-            && part
-                .chars()
-                .all(|ch| ch.is_ascii_alphanumeric() || ch == '-' || ch == '_')
-    })
 }
 
 #[cfg(test)]
--- a/crates/coral-app/src/search/observed/collector.rs
+++ b/crates/coral-app/src/search/observed/collector.rs
@@ -7,7 +7,11 @@
 
 use std::collections::HashSet;
 
-use arrow::array::Array;
+use arrow::array::{
+    Array, BinaryArray, BinaryViewArray, LargeBinaryArray, LargeStringArray, StringArray,
+    StringViewArray,
+};
+use arrow::datatypes::DataType;
 use arrow::record_batch::RecordBatch;
 use arrow::util::display::array_value_to_string;
 
@@ -26,6 +30,7 @@ impl ObservedValuesCollector {
     pub(crate) fn collect_batch(&self, batch: &RecordBatch) -> ObservedValuesQueuePayload {
         let mut values = Vec::new();
         let mut observed_bytes = 0_usize;
+        let mut inspected_cells = 0_usize;
         let mut seen = HashSet::new();
         let schema = batch.schema();
         let eligible_column_indices = (0..batch.num_columns())
@@ -33,9 +38,12 @@ impl ObservedValuesCollector {
             .collect::<Vec<_>>();
         for row_index in 0..batch.num_rows() {
             for &column_index in &eligible_column_indices {
-                if values.len() >= self.budget.candidate_limit {
+                if values.len() >= self.budget.candidate_limit
+                    || inspected_cells >= self.budget.inspected_cell_limit
+                {
                     return ObservedValuesQueuePayload { values };
                 }
+                inspected_cells += 1;
                 let field = schema.field(column_index);
                 let column_name = field.name();
                 let column = batch.column(column_index);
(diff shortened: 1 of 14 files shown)


mystic123:
* **#1472** <a href="https://app.graphite.com/github/pr/withcoral/coral/1472?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1472?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1573** <a href="https://app.graphite.com/github/pr/withcoral/coral/1573?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1372** <a href="https://app.graphite.com/github/pr/withcoral/coral/1372?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

mystic123:
## Review guide

*Current guide for the observed-values queue and collection slice. This PR is stacked directly on #1573 and remains intentionally inert until the later app-wiring and projection PRs.*

**What this PR owns:** migration 0002 for observed-value epochs, canonical/FTS storage, and the durable queue; bounded row-major collection; best-effort sensitive-value suppression and structured sanitization; source-scan publication; a bounded background-writer handoff; stale-write protection; and observed-schema repair. It does **not** activate collection in `QueryManager`, drain the queue into `observed_values`, register a search provider, or expose retrieval or maintenance behavior.

**How it works:** `SearchObservationHandle` builds per-surface registrations from selected sources. Each queue job records both the installed source owner used for lifecycle clears and the runtime component source used by SQL and search. Scope IDs hash the component surface plus opaque runtime-contract and credential inputs; this PR deliberately uses a pre-activation seed, and #1480 supplies the real app-owned identities. The publisher reserves one of 128 writer slots **before** collecting or serializing, samples cells row-major under candidate, inspected-cell, candidate-byte, value-byte, and serialized-job limits, removes recognized sensitive JSON fields and URL/form pairs while retaining safe siblings, then hands serialized JSON to the writer. The durable queue is capped per workspace and coalesces a pending job with the same owner, scope, surface, and epoch by replacing its payload.

**Race and recovery model:** a publisher captures a workspace/source epoch. Enqueue and clear both use `IMMEDIATE` transactions. Enqueue accepts a job only if the captured epoch still matches; clear atomically deletes matching queue, canonical, and FTS rows and advances the relevant epoch. If enqueue commits first, clear removes it; if clear commits first, the later enqueue is stale. On open, SQLite first replays the idempotent migrations. If the observed schema remains irreparable while the catalog schema is valid, only observed-specific objects are reset and recreated, preserving catalog rows and shared metadata. If the catalog schema is also invalid, Coral rebuilds the disposable search index.

**Where to focus review:**
- `observed/sqlite_store.rs`: both transaction orderings, workspace-versus-source invalidation, queue-cap enforcement, last-write-wins coalescing, and the post-upsert job lookup.
- `observed/collector.rs` and `observed/sensitive.rs`: row-major fairness, hard bounds before expensive value rendering where raw lengths are available, and the deliberate false-positive/false-negative trade-off of name/token heuristics and structured sanitization.
- `observed/source_scope.rs` and `observed/publisher.rs`: owner identity versus runtime component identity, opaque scope inputs, and duplicate `SurfaceKey` behavior (`HashMap::insert` is last-write-wins).
- `search/sqlite_store.rs`: repair-first behavior and the boundary between an observed-only reset and a full disposable search-index rebuild.

**Known best-effort boundaries:** writer saturation, durable-queue saturation, stale epochs, serialization failures, and store failures drop observations and log at debug level. Sensitive-data filtering is defense in depth, not DLP: unrecognized sensitive values under innocuous field names can still be persisted as plaintext JSON in the local search database. Source-authored `do_not_index` is intentionally owned by #1761.

**Tests:** coverage includes row-major ordering and inspected-cell, candidate, byte, value, and job limits; deduplication; field/value suppression; nested JSON and URL/form sanitization; scope determinism and runtime-contract/credential partitioning; owner-versus-component identity; publisher-to-SQLite persistence without recognized secrets; queue durability, coalescing, cap, and stale-epoch behavior; real lock-contention tests for both clear/enqueue commit orders; rerunnable migrations; repair without data loss; and observed-only reset after irreparable shape or duplicate-row failures. Intentional gaps in this slice: the production queue cap is represented by a smaller test-only cap, and `observed_values`/FTS population and querying land later.

**Stack boundary:** #1761 adds source-authored `do_not_index`; #1480 supplies runtime-contract and credential identities and wires collection into app lifecycle; #1591 drains and projects queue jobs; #1612 adds storage ceilings; #1593 enables policy-aware retrieval; #1594 exposes maintenance; and #1756 owns the user-facing guide.

crates/coral-app/src/search/observed/collector.rs
```diff
@@ -0,0 +1,372 @@
...
+use std::collections::HashSet;
+
+use arrow::array::Array;
+use arrow::record_batch::RecordBatch;
+use arrow::util::display::array_value_to_string;
+
+use crate::hash::sha256_hex;
+use crate::search::observed::sqlite_queue::{ObservedValueCandidate, ObservedValuesQueuePayload};
+
+const SENSITIVE_COLUMN_NAMES: &[&str] = &[
```
sauldhernandez: Should this be configurable?

crates/coral-app/src/search/observed/collector.rs
```diff
@@ -0,0 +1,372 @@
...
+        return true;
+    }
+    is_sensitive_token(trimmed)
+        || contains_sensitive_token(trimmed)
+        || contains_sensitive_key_value_pair(trimmed)
+}
+
+fn is_sensitive_token(value: &str) -> bool {
+    let lower = value.to_ascii_lowercase();
+    if (lower.starts_with("sk-") && value.len() >= 20)
```
sauldhernandez: I get this is an initial approach - but how does the final picture of this look like? is it going to be a configurable/customizable filter? is it going to be a self-learning filter via flagging?  
  
My worry here is that the current approach may provide a false sense of security: we're filtering for specific patterns, not "filtering sensitive information" in general, which requires something more complex.

crates/coral-app/src/search/observed/publisher.rs
```diff
@@ -0,0 +1,447 @@
...
+            return;
+        };
+
+        let payload = self.collector.collect_batch(observation.batch);
+        if payload.is_empty() {
+            return;
+        }
+        match self.writer.try_enqueue(ObservedValuesWrite {
+            workspace_name: self.workspace_name.clone(),
+            source_name: scope.observed_source_name.clone(),
```
sauldhernandez: **P1: Preserve the query-visible schema separately**

This matches observations with the component schema (`observation.source_name`) but writes `scope.observed_source_name`, which is the logical installed source. Those differ for multi-surface v4 packages: an observation from `github_v4_mcp.list_issues` is persisted as `github_v4.list_issues`, so the eventual API cannot return the queryable schema and same-named surfaces are ambiguous.

Please persist both identities: logical owner for lifecycle clearing and generations, component schema for search results and SQL references. Add a multi-surface v4 publisher-to-storage test.

Confidence: high.

crates/coral-app/src/search/observed/collector.rs
```diff
@@ -0,0 +1,372 @@
...
+    looks_like_jwt(value)
+}
+
+fn contains_sensitive_token(value: &str) -> bool {
+    value
+        .split(|ch: char| !(ch.is_ascii_alphanumeric() || matches!(ch, '-' | '_' | '.')))
+        .any(is_sensitive_token)
+}
+
+fn contains_sensitive_key_value_pair(value: &str) -> bool {
```
sauldhernandez: **P1: Parse structured values before secret classification**

This splitter removes quotes and braces before applying `split_once`. A realistic JSON cell such as `{"api_key":"literal-secret"}` becomes separate `api_key`, `:`, and `literal-secret` segments, so no sensitive key/value pair is detected and the plaintext value is persisted when the outer column is named something like `metadata`. Coral emits manifest `Json` columns as JSON text, so this path is reachable.

The publisher test also never places its configured `literal-secret` into the batch, so it does not prove that echoed credentials are removed. Please classify decoded JSON, URL, and form structures recursively and compare candidates against resolved source secrets in memory without persisting those secrets.

Confidence: high.

crates/coral-app/src/search/observed/publisher.rs
```diff
@@ -0,0 +1,447 @@
...
+            tracing::debug!(
+                workspace = %self.workspace_name.as_str(),
+                source = %observation.source_name,
+                surface = %observation.surface_name,
+                "observed-values source-scan observation did not match a known source surface"
+            );
+            return;
+        };
+
+        let payload = self.collector.collect_batch(observation.batch);
```
sauldhernandez, line 182: **P1: Apply admission and work limits before collecting**

`SourceObservationPublisher` requires prompt, non-blocking publication, but this synchronously scans and formats the batch before checking whether the writer channel has capacity. The collector caps accepted candidates rather than inspected cells, so duplicate, null, oversized, or sensitive batches can still scan the entire matrix. It can also place 8 MiB payloads into a 128-entry channel while the 1 MiB durable limit is applied only after dequeue, allowing more than 1 GiB plus allocation overhead to accumulate. Once the channel is full, later batches still pay the full collection cost before being dropped.

Please reserve admission before collection, byte-bound the in-memory queue at the durable budget, add inspected-cell or time limits, and reject oversized string/binary cells before `array_value_to_string` copies them.

Confidence: high.

crates/coral-app/src/search/observed/source_scope.rs
```diff
@@ -0,0 +1,694 @@
...
+fn surface_scope(
+    source: &QuerySource,
+    component_source_name: &str,
+    surface_kind: ObservedValuesSurfaceKind,
+    surface_name: &str,
+    component_shape: &Value,
+    surface_shape: &Value,
+    generation: ObservedValuesGeneration,
+) -> ObservedSourceSurfaceScope {
+    let scope = json!({
```
sauldhernandez: **P1: Partition observations by credential/account identity**

This scope includes the manifest and non-secret variables but no stable credential or account identity. Two installations with the same source contract connected to different accounts therefore produce the same scope ID. If credentials are replaced or OAuth is re-authorized without a source-generation clear, values observed from account A remain eligible under account B.

Please add an app-owned, non-secret credential identity or revision that changes on account replacement but stays stable across ordinary token refresh. Do not hash raw credentials. Add tests proving account replacements produce distinct scopes and that no credential material reaches SQLite.

Confidence: high.

crates/coral-app/src/search/observed/source_scope.rs
```diff
@@ -0,0 +1,694 @@
...
+        match self {
+            ValueSourceSpec::Template { template } => json!({
+                "from": "template",
+                "template": template_shape(template),
+            }),
+            ValueSourceSpec::OneOf { values } => json!({
+                "from": "one_of",
+                "values": values.iter().map(value_source_shape).collect::<Vec<_>>(),
+            }),
+            ValueSourceSpec::Literal { value } => json!({
```
sauldhernandez: **P2: Use a complete versioned runtime-contract fingerprint**

Reducing a literal to its JSON type makes behaviorally different requests hash identically. The same problem exists for filter and argument defaults, which retain only whether a default exists, plus other manually projected fields such as custom-auth config and S3 session-token binding. A bundled manifest can change these values across a binary upgrade without a source lifecycle clear, leaving observations from the previous runtime contract live under the same scope.

This 694-line hand-maintained mirror will keep drifting as manifest types evolve. Please move a versioned canonical fingerprint to the spec or runtime-package boundary, cover every behavior-affecting value deterministically, and handle credential identity separately. Mutation tests should prove relevant contract changes alter the scope.

Confidence: high.

crates/coral-app/src/search/sqlite_store.rs
```diff
@@ -397,10 +403,12 @@ fn migrate_if_needed(connection: &mut Connection) -> Result<(), SqliteSearchErro
...
     }
 
+    // Repair mode deliberately reruns every current migration when the version
+    // stamp says "current" but required schema objects are missing. Keep every
+    // migration idempotent: use IF NOT EXISTS and conflict-safe metadata writes.
     let repair_current_version = user_version == SEARCH_SQLITE_SCHEMA_VERSION;
     for migration in SEARCH_SQLITE_MIGRATIONS {
-        let should_apply = migration.version > user_version
-            || (repair_current_version && migration.version == user_version);
+        let should_apply = migration.version > user_version || repair_current_version;
```
sauldhernandez: **P2: Repair historical objects during version upgrades**

When a stamped v1 database is missing a v1 object, this skips migration 1 because the database is not at the current version. If `search_meta` is missing, migration 2 then fails while stamping metadata, rolls back, and every later open repeats the same permanent failure. If `catalog_documents` is missing instead, migration 2 can commit and `open` returns success with an incomplete schema because nothing revalidates it before returning.

Please replay the now-idempotent historical migrations whenever schema validation fails during an upgrade, then revalidate the complete schema before returning. Add regression tests for stamped-v1 databases missing `search_meta`, a catalog table, and the catalog FTS table.

Confidence: high.

crates/coral-app/src/search/observed/collector.rs
```diff
@@ -0,0 +1,372 @@
...
+    budget: ObservedValuesCollectionBudget,
+}
+
+impl ObservedValuesCollector {
+    pub(crate) fn collect_batch(&self, batch: &RecordBatch) -> ObservedValuesQueuePayload {
+        let mut values = Vec::new();
+        let mut observed_bytes = 0_usize;
+        let mut seen = HashSet::new();
+        let schema = batch.schema();
+        for column_index in 0..batch.num_columns() {
```
sauldhernandez: **P2: Sample fairly across columns before truncation**

Candidates are produced column-major, and the writer later keeps only the largest serialized prefix that fits 1 MiB. With a normal 8,192-row batch and a unique leftmost ID column, that prefix can be consumed entirely by IDs before title, name, or description columns are visited. Those later columns then remain systematically unsearchable even without load or errors.

Please enforce round-robin or per-column quotas under the final serialized budget and add a wide-batch regression test proving that useful late columns survive.

Confidence: high.

sauldhernandez (changes requested):

sauldhernandez:
Would a simpler approach be possible around the generations guard? It's fairly complex.

mystic123:
@sauldhernandez Re [the generations-guard question](https://github.com/withcoral/coral/pull/1472#issuecomment-4966866056): I simplified this to a small epoch protocol in `b38f0cff5`.

A publisher captures the workspace/source epoch. Enqueue and clear both use an `IMMEDIATE` transaction: enqueue inserts only if the captured epoch is still current, while clear atomically deletes existing rows and advances the relevant epoch. That makes both orderings safe: if enqueue commits first, clear deletes it; if clear commits first, enqueue observes a stale epoch and drops the write.

The workspace and source counters remain separate so a workspace clear can invalidate every in-flight source without enumerating them, while a source clear does not invalidate unrelated sources. Both transaction orderings now have explicit race tests.

mystic123:
@sauldhernandez Heads-up: I’ve put the entire remaining observed-values search stack behind a default-off runtime feature flag.

Coral’s existing feature convention is `observed_values_search` (`Feature::ObservedValuesSearch`), enabled via `[features] observed_values_search = true`, `coral features enable observed_values_search`, or the transient `--enable-observed-values-search` flag.

For this PR specifically, the SQLite schema and migrations may exist while the feature is off, but no observation writer is constructed and nothing is enqueued. The downstream collection, projection/drainage, retrieval, and observed maintenance paths are also inert through #1756. Catalog search remains available, and explicit observed-value clear remains available only for cleanup.

The full open stack was restacked and republished with the flag defaulting off; all current non-draft Validate workflows passed.

sauldhernandez (approved):

mystic123:
### Merge activity

* **Jul 17, 7:09 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1472).
* **Jul 17, 7:10 AM UTC**: [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1472) couldn't merge this pull request because a downstack PR #1573 failed to merge.
* **Jul 17, 8:40 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1472).
* **Jul 17, 8:41 AM UTC**: [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1472) couldn't merge this PR because it had merge conflicts.

PR #1471, opened 2026-07-01
refactor(spec): remove pagination from v4 projections

simonwhitaker:
## Summary

- Remove pagination from the DSL v4 `Projection` model and generated `projections.yaml` artifact shape.
- Keep pagination as an execution detail on semantic IR attachments and copy REST pagination into runtime HTTP specs from the matched operation.
- Bump the projection generator version to invalidate stale materialized projection catalogs.

## Why

Pagination was duplicated into projections even though it is not part of the SQL-facing projection schema. That made the projection catalog carry engine execution details and created two possible sources of truth.

## Impact

Existing materialized DSL v4 sources with old projection catalogs will need to be re-added so `projections.yaml` is regenerated without pagination. Runtime pagination behavior is preserved through semantic IR.

## Validation

- `cargo test -p coral-spec`
- `cargo test -p coral-app sources::runtime_package`
- `make rust-checks`

PR #1481, opened 2026-07-02
feat(spec): support parameter metadata overrides

### LINKED ISSUE #1233: Add explicit pagination overlays

simonwhitaker:
## Summary

- Add DSL v4 `parameter_metadata.yaml` pagination overrides at `workspaces/{workspace}/sources/{source}/overrides/surfaces/{surface}/parameter_metadata.yaml`.
- Support ordered surface-wide pagination strategies plus per-operation overrides, with matching by operation id, HTTP method, path pattern, and required query params.
- Allow overrides for query-param pagination and non-query pagination metadata, including body cursor paths, response cursor paths, page-size body paths, link-header mode, and max page limits.
- Apply overrides when loading materialized v4 sources without rewriting generated artifacts, then sync projection input exposure to the effective pagination contract.

Closes #[**1233**](https://github.com/withcoral/coral/issues/1233)

## Behavior Notes

- Surface-level strategies are evaluated in order; the first matching strategy applies.
- `operation_overrides` take precedence for inconsistent individual operations.
- Generated projections recompute REST input exposure so stale inferred pagination params are restored/hidden correctly.
- Explicit `overrides/projections.yaml` files keep their custom exposure choices, while current pagination-owned query inputs remain internal.

## Example Override File

Save this as `workspaces/{workspace}/sources/{source}/overrides/surfaces/rest/parameter_metadata.yaml`:

```yaml
# Ordered surface-wide strategies. The first matching strategy wins.
pagination:
  # Match operations by explicit operation id and override page pagination params.
  - name: page_number_params
    match:
      operation_ids: [widgets/list]
    mode: page
    page_param: page_number
    page_start: 1
    page_step: 1
    page_size:
      default: 50
      max: 100
      query_param: per_page
    max_pages: 20

  # Match by HTTP method, path glob, and required query params; use offset pagination.
  - name: offset_limit_params
    match:
      methods: [GET]
      path_patterns: [/v1/events*]
      required_query_params: [offset, limit]
    mode: offset
    offset_param: offset
    offset_start: 0
    offset_step: 100
    page_size:
      default: 100
      max: 500
      query_param: limit

  # Match any operation that has the provider's cursor query params.
  - name: cursor_query_params
    match:
      required_query_params: [cursor, limit]
    mode: cursor_query
    cursor_param: cursor
    response_cursor_path: [meta, next_cursor]
    page_size:
      default: 100
      max: 100
      query_param: limit

  # Link-header pagination has no cursor query param, but can still own page size.
  - name: github_link_header
    match:
      methods: [GET]
      path_patterns: [/repos/*/issues]
    mode: link_header
    link_header_require_results: true
    page_size:
      default: 100
      max: 100
      query_param: per_page

# Per-operation overrides win over any matching surface strategy.
operation_overrides:
  reports/search:
    pagination:
      mode: cursor_body
      cursor_body_path: [page, cursor]
      response_cursor_path: [pagination, next]
      page_size:
        default: 25
        max: 100
        body_path: [page, size]
      max_pages: 10

  audit/export:
    pagination:
      mode: none
```

## Validation

- `cargo test -p coral-spec parameter_metadata --all-features`
- `cargo test -p coral-app parameter_metadata --all-features`
- `cargo clippy -p coral-spec --all-targets --all-features --locked -- -D warnings`
- `cargo clippy -p coral-app --all-targets --all-features --locked -- -D warnings`
- `make rust-checks`

### DIFF
--- a/crates/coral-app/src/sources/materialization.rs
+++ b/crates/coral-app/src/sources/materialization.rs
@@ -183,7 +183,6 @@ pub(crate) fn load_v4_materialization(
     let diagnostics: Vec<Diagnostic> =
         read_artifact_yaml(source_name, "diagnostics", &diagnostics_path)?;
     let mut surfaces = Vec::new();
-    let mut semantic_irs = Vec::new();
     for fingerprint_surface in &fingerprint.surfaces {
         let surface = manifest
             .surface(&fingerprint_surface.surface_id)
@@ -212,7 +211,6 @@ pub(crate) fn load_v4_materialization(
             &surface.id,
             &mut semantic_ir,
         )?;
-        semantic_irs.push(semantic_ir.clone());
         surfaces.push(MaterializedSurface {
             surface_id: surface.id.clone(),
             semantic_ir,
@@ -226,7 +224,11 @@ pub(crate) fn load_v4_materialization(
     } else {
         ProjectionPaginationInputSyncMode::RecomputeRestInputExposure
     };
-    sync_projection_pagination_inputs(&semantic_irs, &mut projections, projection_sync_mode);
+    sync_projection_pagination_inputs(
+        surfaces.iter().map(|surface| &surface.semantic_ir),
+        &mut projections,
+        projection_sync_mode,
+    );
     let materialized = V4MaterializedSource {
         fingerprint,
         surfaces,
--- a/crates/coral-app/src/sources/materialization.rs
+++ b/crates/coral-app/src/sources/materialization.rs
@@ -8,11 +8,13 @@ use std::time::Duration;
 use coral_spec::v4::{
     Diagnostic, DiagnosticSeverity, Fingerprint, FingerprintSurface, MCP_IMPORTER_VERSION,
     MaterializedSurface, McpToolCatalog, OPENAPI_IMPORTER_VERSION, PROJECTION_GENERATOR_VERSION,
-    ProjectionCatalog, SURFACE_IMPORTER_VERSION, SemanticIr, SurfaceType,
-    V4_ARTIFACT_SCHEMA_VERSION, V4MaterializedSource, V4SourceManifest,
-    generate_projection_catalog, import_mcp_surface, import_openapi_surface,
-    normalize_mcp_tool_catalog, normalize_source_document, openapi_document_metadata,
-    validate_materialized_source, validate_openapi_base_url_template,
+    ProjectionCatalog, ProjectionPaginationInputSyncMode, SURFACE_IMPORTER_VERSION, SemanticIr,
+    SurfaceType, V4_ARTIFACT_SCHEMA_VERSION, V4MaterializedSource, V4SourceManifest,
+    apply_parameter_metadata_overrides, generate_projection_catalog, import_mcp_surface,
+    import_openapi_surface, normalize_mcp_tool_catalog, normalize_source_document,
+    openapi_document_metadata, parse_parameter_metadata_overrides_yaml,
+    sync_projection_pagination_inputs, validate_materialized_source,
+    validate_openapi_base_url_template,
 };
 use coral_spec::{
     ManifestCredentialMethod, ManifestCredentialMethodKind, ManifestInputKind, ManifestInputSpec,
@@ -35,6 +37,7 @@ const DESCRIPTOR_USER_AGENT: &str = "coral-dsl-v4-materializer";
 pub(crate) const PROJECTIONS_FILENAME: &str = "projections.yaml";
 pub(crate) const FINGERPRINT_FILENAME: &str = "fingerprint.yaml";
 pub(crate) const DIAGNOSTICS_FILENAME: &str = "diagnostics.yaml";
+pub(crate) const PARAMETER_METADATA_OVERRIDE_FILENAME: &str = "parameter_metadata.yaml";
 
 #[derive(Debug)]
 pub(crate) struct MaterializationBuild {
@@ -155,6 +158,8 @@ pub(crate) fn load_v4_materialization(
 ) -> Result<V4MaterializedSource, AppError> {
     let fingerprint_path = layout.v4_fingerprint_file(workspace_name, source_name);
     let projections_path = layout.v4_projections_file(workspace_name, source_name);
+    let projections_override_path =
+        layout.v4_projections_override_file(workspace_name, source_name);
     let diagnostics_path = layout.v4_diagnostics_file(workspace_name, source_name);
     if !fingerprint_path.exists() || !projections_path.exists() || !diagnostics_path.exists() {
         return Err(incompatible_materialization_error(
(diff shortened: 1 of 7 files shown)


simonwhitaker:
* **#1481** <a href="https://app.graphite.com/github/pr/withcoral/coral/1481?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1481?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1471** <a href="https://app.graphite.com/github/pr/withcoral/coral/1471?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

simonwhitaker:
There was an interesting wrinkle that Codex spotted during the authoring of this PR:

> Current shape: OpenAPI import stores one inferred PaginationSpec on each REST operation, then projection generation hides whatever query inputs that spec owns. That means the override likely has to be applied before projection generation and before writing semantic-ir.yaml, not only at runtime, or SQL will expose/hide the wrong parameters.

I pushed back on that, and said the override should be applied at runtime, and that for now, users can write their own projections if their pagination override means the generated projection is now exposing or hiding the wrong parameters. That's what this PR implements.

I'm not 100% sure that's the right decision though. What Codex was arguing for basically becomes the means to define pagination overrides _at source add time_. This ends up looking something like:

```
coral source add --file some-api.yaml --pagination-overrides some-api-pagination.yaml
```

The more I think about this, the more undecided I become about which option I prefer. For example, if I `coral source add` some source, then discover it mostly worked fine, but it hid the `page` parameter on every API response, assuming it was for pagination control, when in fact it has some semantic meaning in my model and should be a regular column. At that point, I have to:

1. Fix pagination for my source with a pagination override file
2. Write my own set of projections

It might be much nicer to:

1. Fix pagination for my source with a pagination override file
2. **Generate** a new set of projections using my updated IR. (The pagination override file is really an IR override.)

I think we can do this in a subsequent PR if it's something we decide we want. And we can discuss whether step 2 in that last list looks like running `coral source add` again, or running some new command (e.g.`coral source refresh-projections`), or something else.

Interested to hear thoughts on this.

crates/coral-spec/src/v4/parameter_metadata.rs
```diff
@@ -0,0 +1,1066 @@
...
+    Ok(())
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ProjectionPaginationInputSyncMode {
+    RecomputeRestInputExposure,
+    PreserveExistingExposure,
+}
+
+pub fn sync_projection_pagination_inputs(
```
iaptsiauri: I am not sure if this is relevant and there is a chance of this happening, but a projection generated pre-override as a TableFunction stays a TableFunction even after the override makes those params pagination-owned

crates/coral-spec/src/v4/parameter_metadata.rs
```diff
@@ -0,0 +1,1066 @@
...
+        IrInputLocation::Path | IrInputLocation::Query | IrInputLocation::ToolArg => {
+            default_exposure
+        }
+        IrInputLocation::Header | IrInputLocation::Cookie | IrInputLocation::Body => {
+            SqlInputExposure::Internal
+        }
+    }
+}
+
+fn pagination_query_param_names(pagination: &PaginationSpec) -> BTreeSet<String> {
```
iaptsiauri: This function is same as `crates/coral-spec/src/v4/projections/pagination.rs::pagination_query_param_names` the only diff is the  `BTreeSet<String>`

crates/coral-app/src/sources/materialization.rs
```diff
@@ -197,8 +203,16 @@ pub(crate) fn load_v4_materialization(
...
+        let mut semantic_ir: SemanticIr =
             read_artifact_yaml(source_name, "semantic IR", &semantic_ir_path)?;
+        apply_parameter_metadata_override_file(
+            layout,
+            workspace_name,
+            source_name,
+            &surface.id,
+            &mut semantic_ir,
+        )?;
+        semantic_irs.push(semantic_ir.clone());
```
iaptsiauri: deep clones complete `semantic_ir`  to modify a slice later and then passes `semantic_ir` to the `MaterializedSurface`

crates/coral-spec/src/v4/parameter_metadata.rs
```diff
@@ -0,0 +1,963 @@
...
+        }
+    }
+
+    for (operation_id, operation_override) in &overrides.operation_overrides {
+        let operation = ir
+            .operations
+            .iter_mut()
+            .find(|operation| operation_matches_identifier(operation, operation_id))
+            .ok_or_else(|| {
+                ManifestError::validation(format!(
```
iaptsiauri: I think this error is is unreachable and redundant we already did validation in `validate_for_surface`

crates/coral-spec/src/v4/parameter_metadata.rs
```diff
@@ -0,0 +1,963 @@
...
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct NamedPaginationOverride {
+    pub name: String,
+    #[serde(rename = "match")]
+    pub match_rules: PaginationStrategyMatch,
+    #[serde(flatten)]
+    pagination: PaginationSpec,
```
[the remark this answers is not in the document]
simonwhitaker: The concern is valid, but the right fix is not to introduce a new, identical type with one additional serde annotation. We should probably mark PaginationSpec with `#[serde(deny_unknown_fields)]`, but I'll defer that to a separate PR.

crates/coral-spec/src/v4/parameter_metadata.rs
simonwhitaker: Ah nice catch, thanks! Will fix.

crates/coral-spec/src/v4/parameter_metadata.rs
simonwhitaker: Yeah it's valid. Have a look at my comment on the PR (i.e. not an inline comment) from Friday. I'm undecided on the right way to handle this. I *think* the right approach is to leave it as-is for now: adding a parameter override might require projections to change, in which case the user is on the hook to provide their own projection overrides too. That's going to add usability friction, but it's the simplest thing that works for now, and we can improve it in a subsequent PR. But I'm interested to hear your thoughts on this.

iaptsiauri (approved):

PR #1482, opened 2026-07-02
feat(engine): infer function SQL signatures

Bradley-Butcher:
## Intent

Infer a Coral SQL UDF's argument and result contract by planning its SQL body. Nothing is registered or executed.

For example, DataFusion can infer `$min_id` from `events.id >= $min_id`. If a body only projects `SELECT $value`, the type is unclear: inference returns `InvalidInput` and asks the author to write an explicit cast such as `CAST($value AS VARCHAR)`.

## Contract

- DataFusion's planned parameter fields are the source of truth for ordinary SQL expressions.
- Direct source-function arguments use the function's declared `ManifestDataType`; this preserves semantic types such as `Json` and `Timestamp`.
- Explicit `CAST` and `TRY_CAST` expressions provide a type when the planner cannot.
- An unresolved parameter fails with concrete cast guidance.
- Contradictory uses fail instead of applying Coral-specific widening or precedence rules; the author must cast uses consistently.
- Result columns retain the Arrow types and nullability from the planned output schema.

This deliberately does not implement a second type-evidence resolver in Coral. The later lifecycle PR treats inference failure as an add failure.

## What changed

- Added the engine-side SQL definition and inferred-signature contracts.
- Added `CoralQuery::infer_udf_signature`.
- Added planner-led extraction for parameters and result columns.
- Added five contract tests: source declarations, explicit casts, ordinary planner inference, unresolved parameters, and conflicting parameters.

## Review order

1. `contracts/udfs.rs` — the engine contract.
2. `runtime/query.rs` — planner-led signature extraction.
3. `runtime/udfs.rs` — conversion into the public signature.
4. `tests/engine/udf_tests.rs` — the five supported/failure paths.

## Not in this PR

- Runtime argument binding.
- Table-function registration or execution.
- Storage, lifecycle, API, app, or CLI changes.
- Numeric widening, occurrence-order recovery, positional-placeholder policy, or other custom inference heuristics.

## Validation

- `cargo test -p coral-engine --test engine 'udf_tests::infer_udf_signature_'`
- `cargo clippy -p coral-engine --all-targets --all-features --locked -- -D warnings`
- `make rust-checks` from the top of the restacked series: 1,537 tests passed
- `make perf-check`: 274.7 ms mean against the 750 ms limit

### DIFF
--- a/crates/coral-engine/src/runtime/query.rs
+++ b/crates/coral-engine/src/runtime/query.rs
@@ -3,6 +3,8 @@
 use std::collections::{BTreeSet, HashMap};
 use std::sync::Arc;
 
+use arrow::datatypes::{Field, FieldRef};
+use coral_spec::ManifestDataType;
 use datafusion::common::tree_node::{TreeNode, TreeNodeRecursion};
 use datafusion::common::{ScalarValue, TableReference};
 use datafusion::dataframe::DataFrame;
@@ -52,6 +54,12 @@ pub(crate) struct QueryRuntimeAdapter {
     failures: Vec<SourceRegistrationFailure>,
     schema_to_source: HashMap<String, String>,
     query_result_observers: Vec<Arc<dyn QueryResultObserver>>,
+}
+
+pub(crate) struct InferredSqlSignature {
+    pub(crate) parameter_fields: HashMap<String, Option<FieldRef>>,
+    pub(crate) declared_parameter_types: HashMap<String, ManifestDataType>,
+    pub(crate) planned_schema: Arc<arrow::datatypes::Schema>,
 }
 
 struct FallbackRuntime {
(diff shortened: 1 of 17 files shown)


Bradley-Butcher:
* **#1501** <a href="https://app.graphite.com/github/pr/withcoral/coral/1501?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1529](https://github.com/withcoral/coral/pull/1529) <a href="https://app.graphite.com/github/pr/withcoral/coral/1529?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1530](https://github.com/withcoral/coral/pull/1530) <a href="https://app.graphite.com/github/pr/withcoral/coral/1530?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1482** <a href="https://app.graphite.com/github/pr/withcoral/coral/1482?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1482?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Bradley-Butcher:
@simonwhitaker have assigned you as a reviewer as my buddy is away. 

LMK if the PR is too big to absorb, you need more deets, or higher level overview.

Thank you for your service 🫡 .

crates/coral-engine/src/runtime/query.rs
```diff
@@ -634,6 +680,113 @@ impl QueryRuntimeAdapter {
...
+}
+
+fn set_parameter_field(
+    parameter_fields: &mut HashMap<String, Option<FieldRef>>,
+    parameter: &str,
+    field: FieldRef,
+) -> Result<(), DataFusionError> {
+    match parameter_fields.get(parameter) {
+        Some(Some(existing))
+            if existing.data_type() != field.data_type()
```
jsummerfield: Could we preserve `Timestamp`'s two supported representations when merging these claims? A `Timestamp` source argument is intentionally carried as `Utf8`, while an explicit `CAST($since AS TIMESTAMP)` or timestamp-column comparison supplies an Arrow timestamp. When one placeholder is used in both contexts, this check rejects them as conflicting—even though the error recommends the cast already present. I reproduced that through the public inference API. The immediate-base `types::manifest_claim_accepts_arrow` helper encoded and tested this exact compatibility before this PR removes it. Could we reconcile the declared `ManifestDataType` with the Arrow claim here and add the mixed-use case as a regression test?

jsummerfield (commented):
Verified against the current PR head and reproduced through the public engine API.

jsummerfield (approved):

crates/coral-engine/src/runtime/query.rs
Bradley-Butcher: Good catch. I fixed the underlying split representation rather than preserving a generic Timestamp/Utf8 compatibility exception. Manifest Timestamp now lowers to Coral’s canonical Arrow timestamp, and query parameters carry timestamps as typed values. HTTP and MCP serialize them to RFC3339 only at the source request boundary. `infer_udf_signature_accepts_timestamp_source_argument_and_cast` covers the mixed source-argument plus explicit-cast case through the public inference API, while cross-type conflicts still fail.

Bradley-Butcher:
### Merge activity

* **Jul 14, 3:07 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1482).
* **Jul 14, 3:07 PM UTC**: @Bradley-Butcher merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1482).

PR #1484, opened 2026-07-02
fix(spec): use OpenAPI tags for v4 operation names

### LINKED ISSUE #1474: [Bug]:  v4 OpenAPI generation ignores tags and publishes collision-derived operation/table/function names
### What happened?

`github_v4` is generated from GitHub’s OpenAPI description, but the generated surface does not use OpenAPI `tags` as the primary grouping mechanism. Instead, v4 derives names from response entity names and then resolves collisions by prepending REST path context or appending hash suffixes.

That makes the public catalog look like REST path debris instead of an explicit operation catalog.

This is the wrong first cut. For now, we should prefer explicit grouped operation names like:

```text
github.billing.get_github_billing_ai_credit_usage_report_org
```

We do not need to be clever about deduping words yet. Repetition is acceptable if the result is explicit, stable, and traceable to the upstream OpenAPI operation.

**Evidence**

The OpenAPI importer currently reads and normalizes `operationId`, but does not preserve `tags` / category metadata as operation grouping identity: [operations.rs](crates/coral-spec/src/v4/surfaces/openapi/operations.rs:31).

Projection names are then derived mostly from response entity/cardinality: [names.rs](coral-spec/src/v4/projections/names.rs:253).

When collisions happen, v4 prepends path context or appends a stable hash: [names.rs](crates/coral-spec/src/v4/projections/names.rs:119).

The generated `github_v4` catalog currently has examples like:

| Current | Desired initial grouped operation |
|---|---|
| `organizations_settings_billing_ai_credit_usages` | `github.billing.get_github_billing_ai_credit_usage_report_org` |
| `users_settings_billing_ai_credit_usages` | `github.billing.get_github_billing_ai_credit_usage_report_user` |
| `users_repositories` | `github.repos.list_for_user` |
| `users_repositories__a7622f19` | `github.activity.list_repos_watched_by_user` |
| `projects_v2_item_with_contents` | `github.projects.list_items_for_org` |
| `users_projectsv2_projects_v2_item_with_contents` | `github.projects.list_items_for_user` |
| `orgs_projectsv2_views_projects_v2_item_with_contents` | `github.projects.list_view_items_for_org` |
| `users_projectsv2_views_projects_v2_item_with_contents` | `github.projects.list_view_items_for_user` |

**Desired behavior**

Use provider grouping metadata first:

1. Group by OpenAPI `tags`.
2. Use the operationId leaf after the slash as the operation name.
3. Normalize to a stable dotted path.

Example:

```yaml
tags:
  - billing
operationId: billing/get-github-billing-ai-credit-usage-report-org
```

Should produce:

```text
github.billing.get_github_billing_ai_credit_usage_report_org
```

Not:

```text
github_v4.organizations_settings_billing_ai_credit_usages
```

**Non-goals**

Do not try to merge org/user variants into one semantic SQL relation in this issue.

Do not try to aggressively remove repeated words yet. `github.billing.get_github_billing_ai_credit_usage_report_org` is acceptable for the first pass.

Do not treat path-context collision recovery as the primary naming strategy.

**Acceptance criteria**

- v4 semantic IR preserves OpenAPI operation grouping metadata.
- GitHub v4 generated operations can be addressed/displayed using tag-grouped names.
- No public operation name requires a hash suffix like `__a7622f19`.
- Collision recovery remains an internal fallback, not the generated user-facing name.
- Tests cover the concrete before/after cases above.

### Steps to reproduce

.

### Expected behavior

.

### Environment

.

### Logs, output, or evidence

```shell

```

simonwhitaker:
## Summary

- Adds generic `IrOperationNaming` metadata to DSL v4 semantic IR.
- Populates OpenAPI naming hints from the first non-empty `tags` entry and the `operationId` leaf.
- Uses those hints for generated REST projection names before falling back to entity/path/hash collision naming.
- Updates the GitHub v4 preview test queries to the new tag-grouped operation names.

Fixes #1474.

## Discussion

I generated fresh GitHub v4 projections in a temp config dir after this change and compared them with the existing materialized GitHub v4 `projections.yaml` from local Coral state, checking how many projections needed the addiition of a suffix to make the name unique.

Results for REST projections:

| Catalog | Generator | REST projections | Stable suffix names |
|---|---:|---:|---:|
| Before, materialized local state | `derive-read-v6` | 1194 | 136 |
| After, temp generation | `derive-read-v7` | 1194 | 0 |

The old 136 stable-suffix names broke down as:

| Visibility | Count |
|---|---:|
| published | 46 |
| hidden | 90 |

Other useful observations:

- REST operation IDs matched exactly between old and new: 1194 common, 0 missing, 0 added.
- Old generated catalog had 500 `PROJECTION_NAME_COLLISION_RESOLVED` diagnostics.
- New generated catalog had 0 collision diagnostics.
- Of the old 500 collision resolutions, 136 needed hash suffixes and 364 were path-context-only collision names.
- 886 of 1194 REST projection names changed.
- Published REST shape stayed the same size: 403 tables and 196 table functions before and after.

Concrete issue examples now line up:

```text
organizations_settings_billing_ai_credit_usages -> billing_get_github_billing_ai_credit_usage_report_org
users_settings_billing_ai_credit_usages         -> billing_get_github_billing_ai_credit_usage_report_user
users_repositories                              -> repos_list_for_user
users_repositories__a7622f19                    -> activity_list_repos_watched_by_user
projects_v2_item_with_contents                  -> projects_list_items_for_org
users_projectsv2_projects_v2_item_with_contents -> projects_list_items_for_user
```

The temp generation used `GITHUB_TOKEN=dummy`, so MCP discovery/test queries failed and the generated comparison was intentionally REST-only. That does not affect the OpenAPI projection-name comparison.

## Validation

- `cargo test -p coral-spec v4`
- `make rust-checks`


crates/coral-spec/src/v4/openapi_tests.rs
```diff
@@ -48,6 +48,131 @@ paths: {}
...
+        - {name: org, in: path, required: true, schema: {type: string}}
+      responses:
+        '200':
+          content:
+            application/json:
+              schema: {$ref: '#/components/schemas/Usage'}
+  /quotes:
+    get:
+      tags: ['forex', 'finance', 'quotes']
+      operationId: quotes/list
```
jsummerfield: Worth adding a test for the 1Forge example: tags exists, but operationId doesn't.

iiuc, we'll fall back to the old entity/path naming, where we could be [tag]_[derived fallback], e.g. `forex_get_quotes`

jsummerfield (approved):

PR #1493, opened 2026-07-02
fix(v4): preserve scalar row projection types

jsummerfield:
Fixes DSL v4 projection generation for OpenAPI endpoints whose row type is a top-level scalar value.

Before this change, scalar row responses were always exposed as a single `value` column with `Json` type. That erased useful type information for endpoints that return arrays of strings, integers, numbers, booleans, or timestamps.

This now preserves the scalar type when generating the synthetic `value` column, using the same type mapping already used for scalar object fields.

## Before / After

Given an OpenAPI response like:

```yaml
/strings:
  get:
    operationId: list_strings
    responses:
      '200':
        content:
          application/json:
            schema:
              type: array
              items:
                type: string
```

Before:

```sql
SELECT column_name, data_type
FROM coral.columns
WHERE schema_name = 'scalar_rows'
  AND table_name = 'list_strings';
```

```text
column_name | data_type
------------+----------
value       | Json
```

After:

```text
column_name | data_type
------------+----------
value       | Utf8
```

The same now applies across the scalar shapes imported by v4:

```text
OpenAPI row item                 Before   After
-------------------------------  -------  ---------
type: string                     Json     Utf8
type: integer                    Json     Int64
type: number                     Json     Float64
type: boolean                    Json     Boolean
type: string, format: date-time  Json     Timestamp
```

That means generated SQL surfaces can expose scalar rows as typed values directly:

```sql
SELECT value
FROM scalar_rows.list_integers;
```

Before, `value` was a JSON column even though each row was an integer. After, `value` is `Int64`, so downstream queries can use it as a normal numeric column.

```sql
SELECT value
FROM scalar_rows.list_timestamps
WHERE value >= timestamp '2026-01-01T00:00:00Z';
```

Before, this required JSON handling or casts because the generated column type was `Json`. After, the column is generated as `Timestamp`.

## Implementation Notes

- Adds a shared projection type mapper for v4 projection columns.
- Uses that mapper for both object fields and top-level scalar row projections.
- Keeps non-scalar row shapes as `Json`.
- Adds coverage for string, integer, number, boolean, and timestamp scalar rows.

PR #1492, opened 2026-07-02
fix(v4): make required REST inputs table functions

jsummerfield:
## Summary

DSL v4 preview projections should not expose provider-required request inputs as ordinary table filters. If a REST endpoint cannot be called without a public input such as a path parameter, the SQL surface should make that requirement explicit by generating a table function.

This updates v4 projection derivation so REST operations with required public inputs become `kind: table` table functions. Pagination-owned query parameters stay internal, search projections keep their existing search-function behavior, and runtime registration remains unchanged.

## Before / After

Before, required request inputs could appear as `WHERE` filters on a plain table:

```sql
SELECT id, number, title, state
FROM github_v4.issues_list_for_repo
WHERE owner = 'octocat'
  AND repo = 'Hello-World'
LIMIT 5;
```

After, the same endpoint is represented as a table function with explicit arguments:

```sql
SELECT id, number, title, state
FROM github_v4.issues_list_for_repo(owner => 'octocat', repo => 'Hello-World')
LIMIT 5;
```

For GitHub pull requests, the generated projection now has the same shape:

```sql
SELECT id, number, title, state
FROM github_v4.pull_requests(owner => 'withcoral', repo => 'coral', state => 'open')
LIMIT 20;
```

Pagination parameters such as `page` and `per_page` remain Coral-managed internal inputs, not public function arguments.

## Validation

- `cargo test -p coral-spec v4::projection_tests --locked`
- `cargo test -p coral-spec v4::openapi_tests --locked`
- `cargo test -p coral-app sources::runtime_package --locked`
- `cargo run --locked -p coral-cli -- source lint sources/core-v4/github_v4/manifest.yaml`
- `git diff --check`
- `make rust-checks`

PR #1494, opened 2026-07-02
fix(v4): keep OpenAPI object responses as rows

jsummerfield:
## Summary

- Keep OpenAPI `type: object` responses as root rows.
- Remove implicit unwrapping of array-valued object properties such as `items`, `data`, `results`, and `rows`.
- Preserve nested object and array fields as JSON columns instead of silently replacing the parent object with nested rows.
- Update the GitHub v4 smoke query to match the preserved search-response envelope.

## Why

The importer currently guesses that a likely-looking array property inside an object contains the operation's real rows. It also treats the sole non-metadata array property as a row collection.

That guess is ambiguous and can select the wrong entity. A normal resource such as a project with `fields: [...]`, a bundle with `items: [...]`, or a GitLab user with `custom_attributes: [...]` can be misclassified as a list of those nested values. Coral then generates the wrong row path and SQL columns, discarding the declared parent object.

OpenAPI describes the response as an object, so this change keeps that object as the row schema and leaves `row_path` at the response root. Only an explicitly declared top-level array imports its `items` schema as list rows.

## Behavior change

For a response shaped like:

```json
{
  "total_count": 2,
  "items": [
    {"id": 1, "title": "A"},
    {"id": 2, "title": "B"}
  ]
}
```

Coral previously generated two rows with `id` and `title` columns by implicitly selecting `items`. After this change it generates one response-envelope row with `total_count` and `items` (JSON) columns.

The HTTP runtime already normalizes a selected root object into one row and a selected root array into multiple rows. As a result, a defective descriptor that declares an object but returns a root array of matching objects can still produce rows without inventing a nested path.

## Tradeoffs and follow-up work

- Genuine wrapper responses such as `{ "data": [...] }` are no longer flattened automatically. Explicit response-row mapping or separately generated nested-collection relations should own that behavior; property-name guessing should not.
- This PR fixes row-schema and row-path selection, but it does not fix pagination inference for descriptors that declare singleton objects while returning lists. That remains part of #1579.
- Installed DSL v4 sources retain their materialized artifacts until the user explicitly re-adds them; this change does not introduce query-time rematerialization.

## Validation

- `make rust-checks` — 1,689 tests passed, 2 skipped; formatting, clippy, and rustdoc passed.
- `make lint-sources`
- `git diff --check`

### DIFF
--- a/crates/coral-spec/src/v4/openapi_tests.rs
+++ b/crates/coral-spec/src/v4/openapi_tests.rs
@@ -1,7 +1,9 @@
 use std::collections::BTreeMap;
 
 use super::*;
-use crate::{PaginationMode, SourceTableFunctionKind, parse_source_manifest_yaml};
+use crate::{
+    ManifestDataType, PaginationMode, SourceTableFunctionKind, parse_source_manifest_yaml,
+};
 
 #[test]
 fn extracts_openapi_document_metadata() {
@@ -182,7 +184,7 @@ components:
 }
 
 #[test]
-fn importer_recognizes_common_wrapped_list_response_fields() {
+fn importer_keeps_common_envelope_objects_as_singletons() {
     let manifest = parse_source_manifest_yaml(
         r"
 name: statusgator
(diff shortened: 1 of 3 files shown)


jsummerfield:
* **#1494** <a href="https://app.graphite.com/github/pr/withcoral/coral/1494?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1494?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1493** <a href="https://app.graphite.com/github/pr/withcoral/coral/1493?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1492** <a href="https://app.graphite.com/github/pr/withcoral/coral/1492?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

simonwhitaker (approved):
Thank you 🙏

PR #1501, opened 2026-07-03
feat(engine): execute UDF table functions

Bradley-Butcher:
## Intent

Execute a validated Coral SQL UDF as a DataFusion table function.

This PR is the execution slice: it takes the inferred and validated definitions from the preceding PRs and installs them into one query runtime. Catalog discovery remains in #1530.

## Execution lifecycle

1. `QueryRuntimeConfig` receives validated UDF definitions.
2. Runtime setup normalizes published names, rejects duplicates and source-function collisions, validates declared results, and plans each body.
3. A relation planner parks UDF calls as extension nodes.
4. DataFusion binds outer query parameters.
5. An analyzer validates call arguments and expands each parked node into the planned UDF body.
6. The existing source-function analyzer resolves source calls inside that expanded body.

This ordering is the core invariant: body parameters are never substituted as SQL text, and source functions inside a UDF use the same planner/analyzer path as ordinary queries.

## What changed

- Added `QueryRuntimeConfig::with_udfs` and runtime UDF registration.
- Added the UDF relation planner and expansion analyzer.
- Added canonical identifier handling, duplicate/collision checks, and result-schema validation.
- Added scenario tests for execution, query parameters, normalization, limits, unsupported modifiers, collisions, and declared result mismatches.
- Kept the catalog-app adjustment at the gRPC service boundary: the larger runtime makes four catalog futures exceed the workspace `large_futures` lint threshold, so only those service calls are boxed. Catalog behavior is unchanged.

## Boundaries

- `coral-engine` owns compilation and execution of generic runtime definitions.
- This PR supports table-function publication only.
- UDF bodies are Coral SQL and may call installed source functions.
- No UDF definitions means the existing source-only runtime path is unchanged.

## Review-size note

Most of the diff is one cohesive planner/analyzer module plus scenario coverage: `runtime/udf_calls.rs` is about 480 lines and the engine test additions are about 630 lines. There is no smaller independently useful execution slice between “park the call” and “expand a typed body.”

## Review order

1. `contracts/query.rs` — runtime config input.
2. `runtime/udf_calls.rs` — park, bind, and expand.
3. `runtime/query.rs` — registration and analyzer order.
4. `runtime/scoped_table_functions.rs` and `runtime/source_functions.rs` — shared name and source seams.
5. `coral-app/src/catalog/service.rs` — boundary-only future boxing.
6. `tests/udf_tests.rs` — execution and failure cases.

## Not in this PR

- Catalog discovery; that is #1530.
- Storage, lifecycle, API, app loading, or CLI commands.
- Nested UDF calls.

## Validation

- `cargo test -p coral-engine --test engine --all-features --locked udf_`
- `cargo clippy -p coral-engine -p coral-app --all-targets --all-features --locked -- -D warnings`
- `make rust-checks` from the top of the restacked series

### DIFF
--- a/crates/coral-engine/src/runtime/udf_calls.rs
+++ b/crates/coral-engine/src/runtime/udf_calls.rs
@@ -71,13 +71,11 @@ impl UdfCallRegistry {
         Ok(registry)
     }
 
-    /// Installs this relation planner together with the analyzer rule that
-    /// expands the nodes it parks. The two are a pair: any session that can
-    /// plan UDF calls must also be able to expand them.
+    /// Installs the relation planner that parks UDF calls. Runtime bootstrap
+    /// installs the expansion rule before `DataFusion`'s default analyzer rules
+    /// so the expanded body receives normal analysis.
     pub(crate) fn install(self, ctx: &SessionContext) -> Result<()> {
-        ctx.register_relation_planner(Arc::new(self))?;
-        ctx.add_analyzer_rule(Arc::new(UdfCallAnalyzerRule));
-        Ok(())
+        ctx.register_relation_planner(Arc::new(self))
     }
 
     fn insert_function(
--- a/crates/coral-engine/src/runtime/udf_calls.rs
+++ b/crates/coral-engine/src/runtime/udf_calls.rs
@@ -1,0 +1,480 @@
+//! UDF SQL call planning.
+//!
+//! UDFs publish as ordinary scoped SQL table functions, but their body is
+//! Coral SQL. The planner parks a UDF call until query parameters have been
+//! bound, then expands the call into the udf body plan with udf arguments
+//! supplied as `DataFusion` parameter values.
+
+use std::cmp::Ordering;
+use std::collections::{BTreeSet, HashMap, HashSet};
+use std::fmt;
+use std::hash::{Hash, Hasher};
+use std::sync::Arc;
+
+use arrow::datatypes::{DataType, Schema};
+use datafusion::common::config::ConfigOptions;
+use datafusion::common::tree_node::Transformed;
+use datafusion::common::{DFSchema, DFSchemaRef, TableReference};
+use datafusion::error::{DataFusionError, Result};
+use datafusion::logical_expr::planner::{
+    PlannedRelation, RelationPlanner, RelationPlannerContext, RelationPlanning,
+};
+use datafusion::logical_expr::sqlparser::ast::TableFactor;
+use datafusion::logical_expr::{
+    Expr, Extension, LogicalPlan, LogicalPlanBuilder, UserDefinedLogicalNodeCore,
+};
+use datafusion::optimizer::AnalyzerRule;
+use datafusion::prelude::SessionContext;
+
+use crate::runtime::query::{read_only_sql_options, reject_unknown_parameters};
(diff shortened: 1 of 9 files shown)


Bradley-Butcher:
* **#1501** <a href="https://app.graphite.com/github/pr/withcoral/coral/1501?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1529](https://github.com/withcoral/coral/pull/1529) <a href="https://app.graphite.com/github/pr/withcoral/coral/1529?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1530](https://github.com/withcoral/coral/pull/1530) <a href="https://app.graphite.com/github/pr/withcoral/coral/1530?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1501?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-engine/src/runtime/udf_calls.rs
```diff
@@ -0,0 +1,480 @@
...
+
+        Ok(registry)
+    }
+
+    /// Installs this relation planner together with the analyzer rule that
+    /// expands the nodes it parks. The two are a pair: any session that can
+    /// plan UDF calls must also be able to expand them.
+    pub(crate) fn install(self, ctx: &SessionContext) -> Result<()> {
+        ctx.register_relation_planner(Arc::new(self))?;
+        ctx.add_analyzer_rule(Arc::new(UdfCallAnalyzerRule));
```
jsummerfield: Could we arrange for DataFusion's normal analysis to see the expanded body? `create_logical_plan` caches the body before analysis, and `add_analyzer_rule` appends this expander after the default `TypeCoercion` pass. I reproduced this with an Int64 argument and body `SELECT concat($id, '-x')`: the equivalent direct SQL is valid, but the function fails internally because the scalar was not cast to string. `build_session_context` already centralizes `SessionStateBuilder`, so could we order expansion ahead of the defaults (or otherwise run the defaults afterward) and keep this as a regression test?

jsummerfield (commented):
Verified against the current PR head; reproduced as a public execution failure.

jsummerfield (approved):

crates/coral-engine/src/runtime/udf_calls.rs
Bradley-Butcher: Good catch — fixed in 4f2ad9235. The UDF expander is now installed ahead of DataFusion's default analyzer rules, so the expanded body receives normal grouping and type coercion before source-function binding. The regression uses an Int64 argument with `concat($id, '-x')` and also asserts that the response schema matches the physical batch schema; the same path now passes through the real CLI/IPC boundary.

Bradley-Butcher:
### Merge activity

* **Jul 14, 4:40 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1501).
* **Jul 14, 4:40 PM UTC**: @Bradley-Butcher merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1501).

PR #1519, opened 2026-07-03
feat(desktop): publish mac app with updates

martinrajdl:
Ships the macOS desktop app lifecycle: a downloadable installer on every release, and automatic update -> download -> install for running apps.

## Download (installer on the internet)
- Every release builds a **signed + notarized universal** (arm64 + x86_64) macOS app — sidecar lipo'd from both targets — and uploads `coral-desktop-mac-universal.dmg` + `.zip` (+ blockmaps, `latest-mac.yml`) to the same GitHub Release as the CLI, included in `checksums.sha256`.
- The artifact name is version-less, so the stable link always serves the newest release:
  `https://github.com/withcoral/coral/releases/latest/download/coral-desktop-mac-universal.dmg` (documented in installation.mdx).
- CI verifies `codesign`/`stapler`, `latest-mac.yml` presence, and exact artifact counts before upload; a failed desktop build fails the whole release, so update metadata can never silently go missing.

## Update flow (new release -> download -> install)
- `electron-updater` (GitHub provider): checks 5s after startup **and every 4h thereafter** (long-running sessions pick up releases without a restart), auto-downloads, notifies, installs on quit. Manual **Coral -> Check for Updates...** menu item with result dialogs.
- `release-please` keeps the desktop version (package.json + lockfile) in lockstep with the release tag, so version comparison is exact and each release carries the matching sidecar.

## Verified locally
- Unsigned zip+dmg build produces correct `latest-mac.yml` (zip-first + sha512) and embedded `app-update.yml` (github/withcoral/coral).
- Launched the packaged build: updater fires, resolves the latest GitHub release (`v0.5.1`), requests `latest-mac.yml` — 404 as expected since no release ships desktop artifacts yet. The first release off this PR closes that loop.
- Repo tags (`vX.Y.Z`) match what electron-updater's GitHub provider parses; Apple secrets already exist under the names the CLI notarize job uses.

## Fixed during review
- `@electron/notarize` expects `APPLE_API_KEY` to be a **path** to the .p8 key, not base64 contents — the workflow now decodes the secret to `RUNNER_TEMP` first. This would have failed the first real release run.
- Validate now has a `Desktop macOS package` job for desktop, Reef, UI, Rust, and release-workflow inputs. It runs the release-shaped unsigned macOS package path and asserts the DMG, ZIP, blockmaps, and `latest-mac.yml`; `AGENTS.md` now tells contributors and agents to use that package path for desktop distribution changes instead of relying only on type-checks.

## Not verifiable before a real release
Signing + notarization themselves (secrets live in the `release` environment). Prerequisite: `APPLE_TEAM_ID` set as a release environment variable, or present in `APPLE_CODESIGN_IDENTITY` (workflow handles both, fails loudly otherwise).

https://claude.ai/code/session_01LHpYSWV9NFgBWsgRfF3nkA

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -318,9 +318,92 @@ jobs:
             coral-${{ matrix.target }}.zip
             coral-${{ matrix.target }}.tar.gz
 
+  build-desktop-macos:
+    needs:
+      - validate-release-ref
+    environment: release
+    runs-on: macos-latest
+    env:
+      TAG_NAME: ${{ needs.validate-release-ref.outputs.tag-name }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          ref: ${{ needs.validate-release-ref.outputs.commit-sha }}
+          persist-credentials: false
+
+      - name: Install Rust toolchain
+        run: rustup show
+
+      - name: Set up Node
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
+        with:
+          node-version-file: apps/desktop/.node-version
+          package-manager-cache: false
+
+      - name: Install desktop dependencies
+        run: npm ci --prefix apps/desktop
+
+      - name: Build, sign, and notarize desktop app
+        shell: bash
+        env:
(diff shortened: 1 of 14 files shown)


jsummerfield:
@martinrajdl close? looks like you split it into smaller PRs

PR #1523, opened 2026-07-06
fix(desktop): load Sources & Traces from the sidecar in dev

martinrajdl:
two changes 
- add missing isMinimized prop to sidebar button component 
- routing the app's backend calls through the Vite dev server to the sidecar (same-origin), instead of failing on a blocked cross-origin request. (this was working for me previously because of a local setup override)


apps/reef/app/wax/components/sidebar-button/sidebar-button.tsx
```diff
@@ -68,7 +68,9 @@ export function SidebarButton<T extends ElementType = 'button'>(
   return (
     <Component {...componentProps}>
       <Icon className={iconStyles({ variant })} color="inherit" name={icon} size="18" />
-      {children && <span className={textStyles({ variant })}>{children}</span>}
+      {!isMinimized && children && (
```
martinrajdl: 😢

AlbertQM (approved):
:dog-science:

PR #1525, opened 2026-07-06
refactor(reef): remove unused icon sizes

AlbertQM:
## Summary

We shouldn't add icon sizes willy-nilly. We already enough sizes to cover all our use cases. Less icon sizes means more consistent UI


lazybean (approved):
thank you

PR #1526, opened 2026-07-06
refactor(reef): remove legacy theme breakpoint shim

AlbertQM:
## Summary

We already have a single source of truth for breakpoints, let's reuse that and not introduce more


apps/reef/app/styles/theme.ts
```diff
@@ -1,4 +0,0 @@
-export const breakpoints = {
-  sidebarCollapse: '963px',
```
lazybean, line 2: That one does not exist in the existing breakpoint file.  
  
But it seems it is not used anyway.

lazybean (approved):

PR #1527, opened 2026-07-06
refactor(reef): use card list for sources

AlbertQM:
## Summary

Moving the polished CardList that was living in the other `reef` to this one. We can carefully extend this one if needed.
Dropped the "Configured" at the bottom, as Cards are already grouped at the top in the "Configured" group.

## Test plan

image

Dialog still works

image

### DIFF
--- a/apps/reef/app/wax/components/card/card-list.css.ts
+++ b/apps/reef/app/wax/components/card/card-list.css.ts
@@ -3,7 +3,7 @@ import { style } from '@vanilla-extract/css'
 export const grid = style({
   display: 'grid',
   gap: '16px',
-  gridTemplateColumns: 'repeat(auto-fit, minmax(min(300px, 100%), 1fr))',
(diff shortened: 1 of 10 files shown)


apps/reef/app/wax/components/card/card-list.css.ts
```diff
@@ -3,7 +3,7 @@ import { style } from '@vanilla-extract/css'
 export const grid = style({
   display: 'grid',
   gap: '16px',
-  gridTemplateColumns: 'repeat(auto-fit, minmax(min(300px, 100%), 1fr))',
+  gridTemplateColumns: 'repeat(auto-fill, minmax(min(300px, 100%), 1fr))',
```
lazybean, line 6: List with 1 item:   
before  




  
after:

AlbertQM: Exactly! The top one looked awful IMO 🙈 (I built it that way, so I can diss it 😂)

lazybean (approved):
Thank you

PR #1532, opened 2026-07-06
feat(app): persist installed function artifacts

Bradley-Butcher:
## Why

Give `coral-app` ownership of installed function state. Each workspace stores a small inventory record plus the original SQL artifact; runtime planning remains a later concern.

## What changed

- Adds a validated `FunctionName` and workspace-scoped inventory entries.
- Adds function paths to `AppStateLayout` and persists the original `.sql` artifact with atomic filesystem writes.
- Runs function install and remove under the shared `WorkspaceLifecycleLock`, then the app state lock, matching the established workspace/source lifecycle order.
- Keeps artifact and inventory updates consistent: if the second update fails, the prior artifact is restored.
- Surfaces both failures when an inventory update and its artifact rollback fail, rather than reporting a failed install while leaving a new body active.
- Retains the app-internal `FunctionArtifactStore` abstraction behind `Arc<dyn FunctionArtifactStore>`. This is the intended storage and failure-injection seam; it does not widen the public API.

## Boundaries

- `coral-app` owns storage, workspace layout, and lifecycle consistency.
- The production caller derives the runtime definition by validating the same SQL passed to install; this manager checks the declared name before persistence.
- This PR does not plan or execute function SQL and adds no transport or CLI surface.

## Validation

- Install writes inventory and artifact.
- Install waits for the workspace lifecycle lock before mutating state.
- Inventory failure restores the previous artifact.
- Inventory plus rollback failure returns a compound `FailedPrecondition` and is covered with the artifact-store failure seam.
- Remove deletes both records and reports typed not-found errors.
- `make rust-checks` (1,686 tests, clippy, formatting, and rustdoc)

PR #1531, opened 2026-07-06
feat(api): add function lifecycle contract

Bradley-Butcher:
## Why

Define the transport contract for installing and managing user-authored SQL functions. This PR is contract-only: it gives later app and CLI changes a typed lifecycle API without moving storage or planning into `coral-api`.

## Contract

- `AddFunction`, `ListFunctions`, and `DeleteFunction` are explicitly workspace-scoped.
- An installed function has one runtime state: `ready` or `invalid`.
- `ready` carries inferred arguments, result columns, and the single SQL table-function publication target.
- `invalid` carries an actionable reason. Invalid installed functions remain visible so callers can inspect or delete them.

The generated client stays thin; this PR adds only the `FunctionClient` accessor.

## Not in this PR

- Persistence or workspace layout.
- Validation, planning, or execution.
- gRPC service implementation.
- CLI or MCP presentation.

## Validation

- `cargo check -p coral-api -p coral-client --all-features --locked`
- `make rust-checks` on the completed stack

PR #1530, opened 2026-07-06
feat(engine): catalog UDF table functions

Bradley-Butcher:
## Intent

Make catalog discovery report the same UDF table functions that the runtime can execute.

This is catalog parity, not another execution path. Runtime setup derives catalog rows from the same normalized, validated UDF definitions used for registration, then supplies one snapshot to both `coral.table_functions` and app catalog discovery.

## What changed

- Added catalog-only table-function metadata for published UDFs.
- Merged those rows with existing provider-backed source-function rows.
- Included the merged snapshot in `coral.table_functions` and `collect_table_functions`.
- Reused canonical published names and the same duplicate/source-collision validation as execution.
- Added parity tests for catalog rows, normalization, arguments, result columns, and collisions.

## Invariants

- A UDF cannot be executable but absent from discovery in the same runtime snapshot.
- Name collisions fail before either registration or catalog publication.
- Source table functions remain provider-backed; UDF rows are catalog metadata and do not add a second factory.
- Existing source catalog rows and execution behavior are unchanged.

## Review order

1. `runtime/udfs.rs` — validated UDF-to-catalog mapping.
2. `runtime/catalog.rs` — source and catalog-only row merge.
3. `runtime/query.rs` — creation and reuse of one runtime snapshot.
4. `contracts/catalog.rs` — generic table-function discovery result.
5. `tests/udf_tests.rs` — execution/discovery parity.

## Not in this PR

- UDF execution semantics; those are in #1501.
- Storage, lifecycle, API, app loading, MCP, or CLI changes.

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p coral-engine --test engine --all-features --locked udf_table_function`
- `cargo clippy -p coral-engine --all-targets --all-features --locked -- -D warnings`
- `make rust-checks` from the top of the restacked series

PR #1529, opened 2026-07-06
feat(engine): compose UDFs with source table functions

Bradley-Butcher:

PR #1533, opened 2026-07-06
feat(app): load installed functions into query runtime

Bradley-Butcher:
## Why

Connect app-owned installed functions to the query runtime for the selected workspace. This is the assembly layer: `coral-app` reads authored artifacts and active sources, while `coral-engine` remains responsible for SQL planning and execution.

## What changed

- Parses installed artifacts into engine `UdfRuntimeDefinition`s.
- Prepares one request-scoped engine runtime for the selected sources, then uses that same runtime for signature inference, UDF installation, and the catalog/query operation.
- Validates SQL publication collisions against source schemas and other installed functions.
- Keeps invalid or stale installed functions in inventory with an error status, but does not publish them into the runtime.
- Passes the bootstrap-owned `WorkspaceLifecycleLock` through `QueryManager` to `FunctionManager`, so function mutation shares the established workspace/source lifecycle order.

## Parameter contract

DataFusion inference is authoritative. Source declarations supply schema information for source relations; they do not guess unresolved function parameter types. If a parameter remains untyped or has conflicting uses, installation fails and the author must add an explicit `CAST`.

## Boundaries

- `coral-app` owns artifact interpretation, runtime package assembly, and lifecycle-lock composition.
- `coral-engine` owns the request-scoped prepared runtime, signature inference, UDF installation, planning, and execution.
- The prepared runtime is not persisted or shared across requests.
- No gRPC or CLI surface is added here.

## Validation

- Manager tests cover ready/invalid inventory, runtime publication, lifecycle lock ordering, rollback failures, and one source-decorator preparation per function-enabled query.
- Validation tests cover source and installed-function publication collisions.
- Query integration executes an installed function against an installed source.
- Engine tests cover late UDF installation through analyzer coercion, source table-function calls, catalog publication, and dependent-join fallback.
- `make rust-checks` (1,699 tests, clippy, formatting, and rustdoc)
- `make perf-check` (199 ms mean against the 750 ms limit)

### DIFF
--- a/crates/coral-app/src/query/manager.rs
+++ b/crates/coral-app/src/query/manager.rs
@@ -1,6 +1,6 @@
 //! Query-time loading, validation, and execution over installed sources.
 
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::BTreeMap;
 use std::future::Future;
 use std::sync::Arc;
 use std::time::Instant;
@@ -11,7 +11,7 @@ use coral_engine::{
     RuntimeSourcePackage, SourceInputResolver, SourceValidationReport, StatusCode, TableInfo,
     UdfRuntimeDefinition,
 };
-use coral_spec::{ManifestInputKind, ManifestInputSpec, parse_function_sql};
+use coral_spec::{ManifestInputKind, ManifestInputSpec};
 use opentelemetry::trace::Status as OtelStatus;
 use serde_json::json;
 use tracing::Instrument as _;
@@ -554,6 +554,13 @@ impl QueryManager {
         Ok(runtime)
     }
 
+    #[cfg_attr(
+        not(test),
+        expect(
+            dead_code,
+            reason = "function lifecycle service callers land upstack in the split stack"
+        )
+    )]
     pub(crate) async fn list_functions(
         &self,
         workspace_name: &WorkspaceName,
@@ -571,10 +578,21 @@ impl QueryManager {
             .map_err(QueryManagerError::App)
     }
 
+    #[expect(
+        dead_code,
+        reason = "function lifecycle service callers land upstack in the split stack"
+    )]
     pub(crate) fn function_manager(&self) -> FunctionManager {
         self.function_manager.clone()
     }
 
+    #[cfg_attr(
+        not(test),
+        expect(
+            dead_code,
+            reason = "function lifecycle service callers land upstack in the split stack"
+        )
+    )]
     pub(crate) async fn validate_udf_sql(
         &self,
         workspace_name: &WorkspaceName,
@@ -593,7 +611,7 @@ impl QueryManager {
                 .load_config_unlocked()
                 .map_err(QueryManagerError::App)?;
             let loaded_sources = self
-                .load_function_validation_sources(workspace_name, &config, raw_sql)
+                .load_query_sources_from_config(workspace_name, &config)
                 .map_err(QueryManagerError::App)?;
             (loaded_sources, config)
         };
@@ -607,28 +625,6 @@ impl QueryManager {
             )
             .await
             .map_err(QueryManagerError::App)
-    }
-
-    fn load_function_validation_sources(
-        &self,
-        workspace_name: &WorkspaceName,
-        config: &AppConfig,
-        raw_sql: &str,
-    ) -> Result<Vec<LoadedQuerySource>, AppError> {
-        let workspace_sources = config.workspace_sources(workspace_name);
-        let source_names = function_validation_source_names(raw_sql, &workspace_sources)?;
-        if source_names.is_empty() {
-            return self.load_query_sources_from_config(workspace_name, config);
-        }
-
-        let mut loaded_sources = Vec::new();
-        for source in workspace_sources {
-            if source_names.contains(&source.name) {
-                let (loaded_source, _version) = self.load_query_source(workspace_name, &source)?;
-                loaded_sources.push(loaded_source);
-            }
-        }
-        Ok(loaded_sources)
     }
 
     async fn runtime_config_with_udfs(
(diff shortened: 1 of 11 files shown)


Bradley-Butcher:
* **#1535** <a href="https://app.graphite.com/github/pr/withcoral/coral/1535?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1534** <a href="https://app.graphite.com/github/pr/withcoral/coral/1534?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1533** <a href="https://app.graphite.com/github/pr/withcoral/coral/1533?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1533?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1532** <a href="https://app.graphite.com/github/pr/withcoral/coral/1532?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1531** <a href="https://app.graphite.com/github/pr/withcoral/coral/1531?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1530** <a href="https://app.graphite.com/github/pr/withcoral/coral/1530?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1501** <a href="https://app.graphite.com/github/pr/withcoral/coral/1501?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1529](https://github.com/withcoral/coral/pull/1529) <a href="https://app.graphite.com/github/pr/withcoral/coral/1529?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

crates/coral-app/src/query/manager.rs
```diff
@@ -457,6 +479,95 @@ impl QueryManager {
...
+        expect(
+            dead_code,
+            reason = "function lifecycle service callers land upstack in the split stack"
+        )
+    )]
+    pub(crate) async fn list_functions(
+        &self,
+        workspace_name: &WorkspaceName,
+    ) -> Result<Vec<FunctionListing>, QueryManagerError> {
+        let (loaded_sources, config) = self
```
jsummerfield: Could this inventory path avoid the query fail-closed loader? `load_query_sources` returns immediately for an unavailable keychain or missing v4 materialization, so one broken source makes `ListFunctions` return no installed functions—even an independent `select 1` function. The API contract says installed functions remain visible when invalid, and `FunctionManager::evaluate_function_listings` already follows that pattern for artifact and per-function planning failures. Could we preserve the inventory identities and turn shared preparation failures into invalid statuses for pending functions, while keeping actual query execution fail-closed? The existing unavailable-keychain fixture plus a constant function would make a focused regression test.

crates/coral-app/src/query/manager.rs
```diff
@@ -457,6 +479,95 @@ impl QueryManager {
...
+        &self,
+        workspace_name: &WorkspaceName,
+        selected_sources: &[LoadedQuerySource],
+        config: &AppConfig,
+    ) -> Result<QueryRuntimeConfig, QueryManagerError> {
+        let runtime = self
+            .runtime_config(workspace_name, selected_sources, config)
+            .map_err(QueryManagerError::App)?;
+        let query_sources = query_sources_from_loaded(selected_sources);
+        let functions = self
```
jsummerfield, line 671: Could we retain the runtime built for signature inference instead of discarding it here? `load_runtime_udfs` calls `infer_udf_signatures`, which builds and registers every source, and the ensuing `CoralQuery` operation immediately builds and registers them again. With any valid installed function, every query and catalog call therefore performs source registration and file-schema discovery twice. The batch inference API already shares one adapter across all functions, and the engine treats source decorators as one-shot registration hooks. Could that preparation boundary be extended so inference and the operation share one registered runtime? A decorator `prepare` counter would provide a focused regression test.

jsummerfield (commented):
Verified against the current PR head; traced both paths through the engine runtime builder.

jsummerfield (approved):

crates/coral-app/src/query/manager.rs
Bradley-Butcher: Fixed. `ListFunctions` now catches only the shared source-preparation failures that already make query loading fail closed (unavailable credentials, missing/incompatible v4 materialization, and invalid v4 projection overrides). It still loads the installed function inventory, preserves artifact-specific errors, and marks otherwise-pending functions invalid with the shared preparation failure. Query execution remains fail closed. Added the suggested unavailable-keychain regression with an installed constant function.

crates/coral-app/src/query/manager.rs
Bradley-Butcher: Done in 66bf60ff0. `CoralQuery::prepare` now builds one request-scoped registered runtime; function signature inference uses that runtime, then installs the validated UDFs into the same adapter before the catalog/query operation. This does not add a persisted artifact or cross-request cache.

I added the suggested decorator counter regression: after function validation, a function-enabled query invokes `SourceDecorator::prepare` exactly once. I also covered late-installed UDFs through the dependent-join fallback and kept the existing coercion, source table-function call, and catalog/execution regressions on the new path.

Validation: `make rust-checks` (1,699 tests), plus `make perf-check` at 199 ms against the 750 ms threshold.

Bradley-Butcher:
### Merge activity

* **Jul 15, 11:10 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1533).
* **Jul 15, 11:11 AM UTC**: @Bradley-Butcher merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1533).

PR #1543, opened 2026-07-07
feat(reef): navigation-progress-bar

lazybean:
Add a root-level progress bar that waits briefly before showing during React Router loading states, with a scoped Storybook scenario to exercise the fixed-position bar under theme comparison.

Constraint: React Router useNavigation requires router context and Storybook theme comparison renders two themed panels in one iframe.
Rejected: Render the bar from app-shell only | root ownership keeps it available for all route layouts.
Rejected: Override production CSS for Storybook | a transformed story viewport preserves fixed production behavior while containing preview rendering.
Confidence: high
Scope-risk: narrow
Directive: Keep NavigationProgressBar at the route root unless a future global shell replaces root app chrome.
Tested: npm run check --prefix apps/reef; npm run typecheck --prefix apps/reef; npm test --prefix apps/reef; npm run build --prefix apps/reef; npm run build-storybook --prefix apps/reef; playwright-cli check of themeComparison URL.
Not-tested: Manual desktop Electron navigation.

PR #1386, opened 2026-06-25
feat(reef): wire sources route actions

lazybean:
Add the `sources` route, with loader and actions.  
The view shows nothing yet, but the data is there.  
  
\-----  
Constraint: Keep this branch as route/data plumbing plus an empty page shell so the UI implementation can review separately.

Rejected: Include source cards and dialogs here | That would make the lower PR too large and mix routing/data with UI implementation.

Confidence: high

Scope-risk: moderate

Directive: Keep full source-page rendering, OAuth UI, provider assets, and toast polish in the upstack implementation branch.

Tested: npm run check --prefix reef; npm run typecheck --prefix reef; npm test --prefix reef -- --run --project unit app/lib/source-form.test.ts; npm test --prefix reef -- --run --project architecture

PR #1385, opened 2026-06-25
feat(reef): support 22px Wax icons

lazybean:
Add 22px width to wax icons  
  
\----   
  
Constraint: Source dialog fallback icons need to match the existing UI header sizing.

Rejected: Use nearest 20px token | It leaves Reef source dialogs visually smaller than the UI reference.

Confidence: high

Scope-risk: narrow

Directive: Keep this token addition limited to Wax Icon sizing unless another consumer needs broader scale work.

Tested: npm run check --prefix reef; npm run typecheck --prefix reef; npm test --prefix reef -- --run --project unit app/lib/source-form.test.ts

Not-tested: Full Reef suite still has pre-existing Storybook coverage architecture failures.

PR #1552, opened 2026-07-07
feat(reef): make Wax cards polymorphic

lazybean:
## Summary

- Replace Card-owned React Router props (`to`, `prefetch`, `preventScrollReset`) with a polymorphic `as` prop, matching other WAX component patterns.
- Add `CardList as` and `getCardProps` so list call sites can render cards as router links, buttons, anchors, or other components without putting route-specific fields on `CardItem`.
- Keep WAX `Card` router-agnostic and presentational while preserving interactive card styling for `as`/`onClick` usage.
- Add Storybook examples for rendering both `Card` and `CardList` as React Router `Link` components.

## Why

Reviewer feedback correctly pointed out that WAX primitives should not import or encode React Router behavior. The source page still needs card-shaped navigation, but that should be composed at the app call site instead of built into the design-system card.

## Validation

- `npm run check --prefix apps/reef`
- `npm run typecheck --prefix apps/reef`
- `npm test --prefix apps/reef`
- `npm run build --prefix apps/reef`
- `npm run build-storybook --prefix apps/reef`

### DIFF
--- a/apps/reef/app/wax/components/card/card.tsx
+++ b/apps/reef/app/wax/components/card/card.tsx
@@ -1,7 +1,6 @@
 import classNames from 'classnames'
-import { ReactNode } from 'react'
+import React, { ElementType, ReactNode } from 'react'
 
-import { Button } from '@/wax/components'
 import { Pill, type PillProps } from '@/wax/components/pill'
 import { Typography } from '@/wax/components/typography'
 
@@ -11,15 +10,29 @@ export type CardHeaderPill = Omit<PillProps, 'children' | 'size'> & {
   label: string
 }
 
-export interface CardProps {
+interface CardBaseProps {
+  className?: string
   description: string
   headerPill?: CardHeaderPill
   icon?: ReactNode
-  onSelect?: () => void
+  interactive?: boolean
   title: string
 }
 
-export function Card({ description, headerPill, icon, onSelect, title }: CardProps) {
+export type CardProps<T extends ElementType = 'div'> = CardBaseProps &
+  Omit<React.ComponentPropsWithoutRef<T>, 'as' | keyof CardBaseProps> & {
+    as?: T
+  }
+
+export function Card<T extends ElementType = 'div'>(
+  props: CardProps<T> & { ref?: React.Ref<HTMLElement> },
+) {
+  const { as, className, description, headerPill, icon, interactive, ref, title, ...rest } = props
+
+  const Component = (as ?? 'div') as ElementType
+  const type = 'type' in props ? props.type! : 'button'
+  const isInteractive = interactive ?? (as !== undefined || 'onClick' in props)
+
   const content = (
     <>
       <span className={styles.header}>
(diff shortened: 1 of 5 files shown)


lazybean:
* **#1386** <a href="https://app.graphite.com/github/pr/withcoral/coral/1386?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1385](https://github.com/withcoral/coral/pull/1385) <a href="https://app.graphite.com/github/pr/withcoral/coral/1385?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1543](https://github.com/withcoral/coral/pull/1543) <a href="https://app.graphite.com/github/pr/withcoral/coral/1543?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1552** <a href="https://app.graphite.com/github/pr/withcoral/coral/1552?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1552?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/reef/app/wax/components/card/card.tsx
```diff
@@ -16,10 +17,22 @@ export interface CardProps {
   headerPill?: CardHeaderPill
   icon?: ReactNode
   onSelect?: () => void
+  prefetch?: LinkProps['prefetch']
+  preventScrollReset?: LinkProps['preventScrollReset']
   title: string
+  to?: LinkProps['to']
```
AlbertQM: Would we ever pass a `prefetch` but no `to`? Ie if these 3 new props are related, maybe we could group them? `link?: {to:...`

apps/reef/app/wax/components/card/card.tsx
```diff
@@ -16,10 +17,22 @@ export interface CardProps {
...
 }
 
-export function Card({ description, headerPill, icon, onSelect, title }: CardProps) {
+export function Card({
```
AlbertQM: I think WAX components should stay purely presentational -- we're now adding some behaviour. I'd create a `SourceCard` or similar which uses a `Card` under the hood. WDYT?
This would make it easier, amongst other things, to reuse the base Card in our future mobile app, which wouldn't use this `Link` 🤔

lazybean: Or give a 'render' or 'as' prop?  
  
  
The issue is that card-list renders the card, so there is no way to reuse most of the code.

AlbertQM: `as` works (I thought I have suggested that too, my bad). This stays consistent with some of our other components which already support `as`! 😄

AlbertQM: The alternative is updating CardList to make it a bit more generic, so instead of rendering Card, it's composable (ie you pass whatever element as card). I think this might be the neater solution. WDYT?

AlbertQM: This should help #1585

lazybean: I think I prefer the 'as'  
Your last alternative means you actually have a (Anything) List, not a CardList

lazybean: I think  i like 1585

AlbertQM (approved):
Thank you!

apps/reef/app/wax/components/card/card.tsx
lazybean: How would I make my LinkCard List with 1585 though?

AlbertQM: With your `as` and the composable 1585, it would look like this:

```
 <Card.List>
          {cards.map(({ id, ...card }) => (
            <Card.Item key={id}>
              <Card.Card {...card} as={Link} prefetch="intent" to={`/sources/${id}`} />
            </Card.Item>
          ))}
        </Card.List>
```

though I'll change this a bit to be CardList.Container, CardList.Card and just Card in the middle

PR #1542, opened 2026-07-07
feat(reef): add schema explorer

martinrajdl:
Rebased reconstruction of #1368 onto current main.

## What this adds
- **Schema route + nav entry** in Reef with a schema/table explorer: table search, lazy column loading, and required-filter indicators, built on the shared table/pill/tooltip wax primitives.
- **Catalog metadata loading avoids keychain reads** for passive browsing while runtime query credential loading stays strict (`crates/coral-app/src/query/manager.rs`).

## Why a reconstruction, not a literal rebase
The original #1368 branch predated two things now on main:
- **#1463** (Sources & Traces UI) — which #1368 duplicated wholesale, so ~8000 of its lines are already upstream.
- **#1491** (`reef/` → `apps/reef/`, `desktop/` → `apps/desktop/` move) — so every modified file was a modify/delete conflict.

A literal `git rebase` was therefore all conflicts and already-merged noise. Instead I re-applied only the genuinely net-new schema surface on top of main:
- 5 new files (schema explorer view/css/lib/route + spin animation) at `apps/reef` paths.
- `table.*` / `pill.*` taken wholesale (identical to base on main; the `style`→`variant` rename only touches the story, no production usage).
- `coral-clients.ts` adapted to main's memoized `getCoralTransport()` (the original used a now-removed `runtime.url`).
- `routes.ts` / `sidebar.tsx` / `architecture.test.ts` merged (main already had sources/traces; added schema).
- `manager.rs` 3-way merged (clean, zero conflicts); `resilience_tests.rs` taken wholesale.

## Verification (all green locally)
- Reef: typecheck, oxlint, oxfmt, production build, architecture test (6 passed)
- Rust: `cargo check`, `cargo clippy --all-targets`, manager unit tests (15) + resilience integration test

Supersedes #1368.

https://claude.ai/code/session_01LHpYSWV9NFgBWsgRfF3nkA

PR #1564, opened 2026-07-07
feat(coral-app): browse catalog metadata without keychain reads

martinrajdl:
Backend half of the schema explorer

The schema explorer needs to browse source metadata passively: schemas, tables, columns, filters, and inputs.
Before this change, that metadata path reused normal query-source loading, which reads real credential material. That can trigger keychain access/prompts or fail when credentials are unavailable, even though the explorer is not actually querying data.
This change adds a catalog/metadata loading mode: for browsing, keychain-backed secrets are represented by in-memory placeholders so sources can register their static metadata without reading real credentials. Actual SQL execution and source validation still use real credential loading and fail strictly if credentials are missing.

followed up by #1565 and originally splti from #1542

### DIFF
--- a/crates/coral-app/src/query/manager.rs
+++ b/crates/coral-app/src/query/manager.rs
@@ -490,19 +490,33 @@ impl QueryManager {
                     })
                 }
                 Some(CredentialStorageKind::Keychain) => {
+                    // Keychain values aren't read for passive browsing, but still
+                    // validate that every required secret is configured so a
+                    // misconfigured source fails closed here instead of silently
+                    // dropping out of the catalog during engine registration.
                     let placeholder_secrets = catalog_placeholder_secrets(
                         source_spec,
                         source.secrets.iter().map(String::as_str),
                     );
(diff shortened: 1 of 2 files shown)


martinrajdl:
Closing as redundant.

The schema explorer (#1565) only needs the existing \`CatalogService.ListCatalog\` gRPC endpoint, which is already on main and is what the CLI and MCP use. So #1565 works standalone against main's catalog path — browsing reads the keychain the same way the CLI does.

Two reasons this backend change isn't worth carrying:
- It's a pure enhancement ("browse without reading the keychain"), not required for the feature.
- Main has since evolved this exact area independently (added \`CredentialResolutionMode::StoredOnly\` + skip-failed-sources), so this PR's \`SourceCredentialMode\`/placeholder mechanism now overlaps with main's own approach.

If avoiding keychain reads during passive browse is still wanted, it should be a follow-up built on main's \`CredentialResolutionMode\` (extend it to also skip the load-time read), rather than the separate mechanism here.

PR #1565, opened 2026-07-07
feat(reef): add schema explorer

martinrajdl:
Adds a Schema explorer to Reef — browse a workspace's schemas, tables, and columns.

## What's here
- **Routes.** `/schema` is a layout (tree + `<Outlet>`); `/schema/:schemaName/:tableName` is a child route whose `clientLoader` fetches that table's columns. Both loaders are **deferred** (return the promise, rendered via `Suspense`/`Await`) so loading/error states work in SPA mode — the desktop build forbids `HydrateFallback` on non-root routes. Table selection is URL-driven (linkable, back/forward aware).
- **Tree.** Schema rows + table links as wax `Button.Container` (bare), with a hover highlight and active-row highlight. Collapsed by default (a search auto-expands matching schemas) so the first paint is one row per schema rather than one per table.
- **Header.** Reuses the shared query-stream `PageHeader` (title + search in the actions slot).
- **Columns.** Rendered as a compact data grid — dense monospace rows, zebra shading, single header rule, sticky header (name / type / nullable / description).
human note: not using the wax table, it doesn't look good at all, I think a dataTable component for sql results would be nice long term valid component to have which is different to the table 
Screenshot 2026-07-09 at 0 35 41


apps/reef/app/routes/schema.tsx
```diff
@@ -0,0 +1,20 @@
+import { fetchSchemaFromCoral } from '@/lib/schema-explorer'
+import { SchemaExplorer } from '@/views/schema-explorer'
+
+import type { Route } from './+types/schema'
+
+// Desktop is SPA-only (no server) and the catalog is fetched over gRPC-web via
```
lazybean: will that work for non desktop version

apps/reef/app/routes/schema.tsx
```diff
@@ -0,0 +1,20 @@
...
+// the browser transport + desktop bridge, so this must be a clientLoader. The
+// schema promise is deferred (not awaited) so the route renders immediately and
```
lazybean: We are reintroducing the navigation-progress-bar  (https://app.graphite.com/github/pr/withcoral/coral/1543/feat(reef)-navigation-progress-bar) , do you actually want a separate loading indication?

apps/reef/app/views/schema-explorer.tsx
```diff
@@ -0,0 +1,421 @@
...
+      return next
+    })
+  }
+
+  return (
+    <section aria-label="Schema explorer" className={styles.root}>
+      <PageHeader
+        title={
+          <>
+            <Typography.BodyStrong as="span" variant="secondary">
```
lazybean: span is the default.

```suggestion
            <Typography.BodyStrong variant="secondary">
```

apps/reef/app/views/schema-explorer.tsx
```diff
@@ -0,0 +1,421 @@
...
+  )
+}
+
+// Deferred loader data: stream in the skeleton, then the resolved tree. SPA mode
+// forbids HydrateFallback on non-root routes, so loading is handled with
+// Suspense/Await rather than a route-level fallback.
+export function SchemaExplorer({ schema }: { schema: Promise<SchemaResponse> }) {
+  return (
+    <Suspense fallback={<SchemaExplorerSkeleton />}>
+      <Await errorElement={<SchemaLoadError />} resolve={schema}>
```
lazybean: :wrench: the schema is critical data, and if it errors the whole page errors.  
I would put ShemaLoadError in the route [ErrorBoundary](https://reactrouter.com/start/framework/route-module#errorboundary) directly.  
Additonnal benefit is that  if we actually do not want to defer that data the error is still handled the same way.

Extra benefit: no need for 'Await', you can pass the 'schema' promise in the child component and '[use](https://reactrouter.com/start/framework/route-module#errorboundary)' it

apps/reef/app/views/schema-explorer.tsx
lazybean, line 1: There is a mismatch between this file and the routes.  
We have 3 schema explorer related routes, but they all import from this big file.  
I think there would be benefit on splitting it the same way (in a views/schema-explorer/ folder

apps/reef/app/routes/schema-table.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const table = findSchemaTable(schema, params.schemaName, params.tableName)
+
+  return (
+    <TableDetailLayout schemaName={params.schemaName} tableName={params.tableName} table={table}>
+      <Suspense fallback={<ColumnsPending />}>
+        <Await errorElement={<ColumnsLoadError />} resolve={loaderData.columns}>
+          {(columns) => <ColumnsTable columns={columns} />}
+        </Await>
+      </Suspense>
+    </TableDetailLayout>
```
lazybean: :wrench: I like when the route is mostly a re-export from a component in views.  
Could you move that in views/schema-explorer/schema-table.tsx (or similar)

lazybean (commented):
Thank you!  
  
If you do not mind waiting for my stack and use loader instead of clientLoader I would be grateful.  
  
Also I think it is worth thinking of what data is critical to the page and what is not.  
We  (will) have the nav progress bar as global pending UI. I think it works well and is consistent.  
We should not make user wait (global pending UI), then wait again (skeleton) without anything useful to see.  
  
For nested nav (switching tables) I think a skeleton is OK (and even better) as we can keep part of the UI.  
But for main pages (sources <-> traces <-> schema) I belive the global progress bar is good consistent pattern

apps/reef/app/routes/schema.tsx
martinrajdl: this is great, love it

martinrajdl:
@lazybean I've rebased on top of your changes and fixed some of the other things you highlighted, should be good to go now, tested again a built version locally - the loader particularly is lovely, everything working as before

apps/reef/app/routes/schema-table.tsx
```diff
@@ -0,0 +1,27 @@
+import { fetchTableColumnsFromCoral } from '@/lib/schema-explorer'
+import { catalogClientForRequest } from '@/lib/coral-request.server'
+import { SchemaTableError, SchemaTableView } from '@/views/schema-explorer/schema-table'
+
+import type { Route } from './+types/schema-table'
+
+// Columns are deferred (the promise is returned un-awaited) so nested
+// table-to-table navigation keeps the layout and shows a local pending state
+// instead of blocking on the global progress bar. The abort signal matters
```
lazybean: :thinking_face: This seems actually wrong.

Please check how the table to table navigation look.

I checkout your code and only the first time you go to the schema-table view there will be the fallback.

[The Suspense already resolved once, by resuspending it , it is cleaver and show the old thing instead.](https://react.dev/reference/react/Suspense#caveats)  
  
\> If Suspense was displaying content for the tree, but then it suspended again, the fallback will be shown again **unless the update causing it was caused by startTransition** or useDeferredValue.

One possible fix, is to give a 'key' to Suspend element.

Or actually await in the loader and rely on the global nav.  
The local pending UI is not that nice in my opinion

apps/reef/app/views/schema-explorer/schema-table.tsx
```diff
@@ -0,0 +1,152 @@
...
+    </TableDetailLayout>
+  )
+}
+
+// Deferred columns stream in under Suspense; `use()` propagates a rejection to
+// the route ErrorBoundary above.
+export function SchemaTableView({ columns }: { columns: Promise<ColumnDef[]> }) {
+  return (
+    <TableDetailLayout>
+      <Suspense fallback={<ColumnsPending />}>
```
lazybean: (See my other bigger comment first)  
  
:wrench: You need a key (with shemaName and tableName) here.

Or give up on deferring this data (I think it is preferable)

apps/reef/app/views/schema-explorer/schema-table.tsx
```diff
@@ -0,0 +1,152 @@
...
+    <table className={styles.dataGrid}>
+      <thead className={styles.dataGridHead}>
+        <tr>
+          <th className={styles.dataGridHeadCell}>name</th>
+          <th className={styles.dataGridHeadCell}>type</th>
+          <th className={styles.dataGridHeadCell}>nullable</th>
+          <th className={styles.dataGridHeadCell}>description</th>
+        </tr>
+      </thead>
+      <tbody>
```
lazybean: :wrench: There is WAX/table

lazybean:
Thank you.   
I think defering the data is unnecessary, add complexity, and does not yield what you expect (local suspense broken when switching table). Please see inline comment.

apps/reef/app/views/schema-explorer/schema-table.tsx
martinrajdl: Yes, have you seen in? It look absolutely horrible used in this case. I'm giving up on this but please the fact that some component exists doesn't mean it was designed for this usecase or even looks ok.

martinrajdl:
@lazybean i addressed the loading. Please take a look when you get a chance.

apps/reef/app/views/schema-explorer/schema-table.tsx
lazybean: Thank you for trying.  
  
  
I actually checked it in storybook, where it looked fine; I have not try to put it in there, though.  
  
What makes it horrible? Is that fixable?  
It is sad that we have a reuseable component that is no resuable, no?

martinrajdl: Pavel designed this component for Phoebe to show data in the chat ui which was very generous with space, as chat uis are compared to something that is more like an sql viewer. It has too much spacing everywhere and isn't showing the data compactly at all, it looks very out of place imo.

martinrajdl: Although I'd rather have a terrible looking table than round of reviews at this point.

lazybean: The last round of review found actual issue (inconsistent Suspense).

lazybean (approved):

apps/reef/app/views/schema-explorer/schema-table.tsx
lazybean: And your opinion about that table is valuable and make sense. We do not have chat ui, so the wax/table as it is now needs to be fixed.

PR #1566, opened 2026-07-07
feat: database source mvp

iaptsiauri:
NOTE: For now manifest is still using inputs, this is for completeness purposes. So that manifest and connection is usable. As soon as identities will land manifest shape will be changed
 
## Summary

Adds the DSL v4 database-source MVP.

- Adds v4 `database` surfaces with provider-specific `postgres`, `mysql`, and `sqlite` connection specs.
- Database-only v4 sources run without materialized artifacts.
- `coral-engine` engine registers database backend through `datafusion-table-providers`.
- Exposes database tables as source-owned catalogs with inner namespaces, addressable as `schema.namespace.table`.
- Adds `namespace` to table/catalog metadata and updates catalog RPC, MCP helper, runtime catalog, and query-error paths for namespaced database tables.
## Tests

- `cargo test -p coral-app catalog::discovery::tests --lib`
- `cargo test -p coral-spec v4::manifest_tests --lib`
- `cargo test -p coral-engine backends::database::tests --lib`

### DIFF
--- a/crates/coral-api/proto/coral/v1/catalog.proto
+++ b/crates/coral-api/proto/coral/v1/catalog.proto
@@ -61,6 +61,9 @@ message Table {
   repeated string required_filters = 6;
   // User-facing query guidance.
   string guide = 7;
+  // Inner namespace for database sources (the remote schema); such tables
+  // are addressed as schema_name.namespace.name. Empty for two-part tables.
+  string namespace = 8;
 }
 
 // Table metadata without full column definitions.
@@ -78,6 +81,9 @@ message TableSummary {
   repeated string required_filters = 5;
   // User-facing query guidance.
   string guide = 6;
+  // Inner namespace for database sources (the remote schema); such tables
+  // are addressed as schema_name.namespace.name. Empty for two-part tables.
+  string namespace = 7;
 }
 
 // An argument accepted by a query-visible table function.
--- a/crates/coral-api/proto/coral/v1/catalog.proto
+++ b/crates/coral-api/proto/coral/v1/catalog.proto
@@ -61,9 +61,9 @@ message Table {
   repeated string required_filters = 6;
   // User-facing query guidance.
   string guide = 7;
-  // Inner namespace for database sources (the remote schema); such tables
-  // are addressed as schema_name.namespace.name. Empty for two-part tables.
-  string namespace = 8;
+  // SQL catalog containing the schema. Empty for two-part tables; otherwise
+  // the table is addressed as catalog_name.schema_name.name.
+  string catalog_name = 8;
 }
 
 // Table metadata without full column definitions.
@@ -81,9 +81,9 @@ message TableSummary {
   repeated string required_filters = 5;
   // User-facing query guidance.
   string guide = 6;
-  // Inner namespace for database sources (the remote schema); such tables
-  // are addressed as schema_name.namespace.name. Empty for two-part tables.
-  string namespace = 7;
+  // SQL catalog containing the schema. Empty for two-part tables; otherwise
+  // the table is addressed as catalog_name.schema_name.name.
+  string catalog_name = 7;
 }
 
 // An argument accepted by a query-visible table function.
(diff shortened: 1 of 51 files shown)


crates/coral-api/proto/coral/v1/catalog.proto
```diff
@@ -61,6 +61,9 @@ message Table {
   repeated string required_filters = 6;
   // User-facing query guidance.
   string guide = 7;
+  // Inner namespace for database sources (the remote schema); such tables
+  // are addressed as schema_name.namespace.name. Empty for two-part tables.
+  string namespace = 8;
```
antonmry: I wonder if we should namespace for a broader functionality, not only database sources. Also, I'm bit worried about potential conflicts with the namespace concept already used in Coral

crates/coral-engine/src/backends/database.rs
```diff
@@ -0,0 +1,436 @@
...
+            sqlite_catalog(connection, context).await
+        }
+        (provider, _) => Err(DataFusionError::Execution(format!(
+            "database provider '{}' does not match connection configuration",
+            provider.as_str()
+        ))),
+    }
+}
+
+async fn postgres_catalog(
```
antonmry, line 116: Note for myself: Would it be possible to generalize this? Maybe with a strategy pattern?

crates/coral-api/proto/coral/v1/catalog.proto
iaptsiauri: Yeah this is a major concern for me as well. Of course to solve the coral namespace conflict we can rename this field but that is not a real solution. Another approach I am considering right now is to actually dropping the namespace field and just combine `schema_name_namespace.name` but this also has it's drawbacks as we would have to have some special characters to keep in mind

crates/coral-engine/src/backends/database.rs
iaptsiauri: ahh this is good catch, I am pretty sure we can do something like that indeed.

crates/coral-engine/src/runtime/registry.rs
```diff
@@ -237,19 +252,65 @@ async fn register_source(
...
 fn register_backend_registration(
+    ctx: &SessionContext,
     catalog: &dyn datafusion::catalog::CatalogProvider,
     source_decorators: &mut [Box<dyn SourceDecorator>],
     query_source: &QuerySource,
     source_name: &str,
     registration: BackendRegistration,
     result: &mut SourceRegistrationResult,
 ) -> std::result::Result<(), CoreError> {
+    // Source decorators wrap table providers at registration time, but catalog
```
antonmry, line 288: Note for myself: what happens if a table is added to the DB after the source has been registered? Do the user needs to remove and readd the source?

iaptsiauri: I have it in the next PR, where I add cache and TTL. After certain configured ttl we re run catalog discovery to detect changes

crates/coral-engine/Cargo.toml
```diff
@@ -19,6 +19,11 @@ datafusion.workspace = true
 datafusion-datasource.workspace = true
 datafusion-datasource-json.workspace = true
 datafusion-functions-json.workspace = true
+datafusion-table-providers = { workspace = true, features = [
+  "mysql",
+  "postgres",
+  "sqlite-bundled",
+] }
```
antonmry, line 26: Note for myself: coral binary is growing a lot, and it will keep growing with any new backend. We should see how to address this problem (Cargo features? Dynamic loading? Both?)

crates/coral-spec/src/v4/schema.rs
```diff
@@ -88,6 +97,106 @@ struct V4McpSurfaceSchema {
     server: McpServerSpec,
 }
 
+#[derive(Debug, Serialize, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
+struct V4PostgresDatabaseSurfaceSchema {
```
antonmry, line 102: Note for myself: explore if it would be possible to have a single schema for all databases

PR #1570, opened 2026-07-07
feat(reef): add reusable empty page

AlbertQM:
## Summary

Bringing back the empty page component, and using it in our screens for consistency

## Test plan

https://github.com/user-attachments/assets/4e69d02b-a5d5-494d-8787-4e5134dea775


apps/reef/app/wax/components/scroll-area/scroll-area.tsx
```diff
@@ -60,6 +63,14 @@ export function Container({
   width = '100%',
   ...rest
 }: ScrollAreaProps) {
+  const contentStyle: React.CSSProperties | undefined =
```
AlbertQM, line 66: In #1571, I'm changing the layout and I won't be using `fillContent` anymore. However, I thought this was a good change (better than the initial solution Codex came up with, of just adding a className to the .Content), so leaving it in. Just like `contrainWidth`, it's useful to sometimes wanting to fill the whole vertical viewport 😄

martinrajdl (approved):
🙏🙏🙏🙏🙏

PR #1550, opened 2026-07-07
feat(search): add local catalog snapshot loader

mystic123:
## Summary

- add an app-owned `CatalogSnapshotLoader` that builds catalog metadata from installed local source state and DSL v4 materialized projection artifacts
- avoid query-runtime provider registration, credential refresh, MCP startup, HTTP calls, and file schema inference while preparing catalog snapshots
- add SQLite catalog source/workspace clear APIs and best-effort source lifecycle invalidation for add/re-add/delete flows

## Validation

- `cargo test -p coral-app search::catalog:: --lib`
- `cargo test -p coral-app --lib`
- `cargo clippy -p coral-app --all-targets --all-features -- -D warnings`
- `git diff --check`

PR #1480, opened 2026-07-02
feat(search): wire observed values into app lifecycle

mystic123:
## Summary

- creates one `SearchObservationHandle` during server bootstrap and passes narrow handles into the source/query lifecycle paths
- loads a canonical `LoadedRuntimeSource` containing the exact `QuerySource` plus a versioned, non-secret runtime-contract fingerprint
- fingerprints manifest bytes, deterministic non-secret variables, and materialized v4 runtime artifacts; paths and resolved secrets are excluded
- persists a non-secret credential revision that rotates on new/replaced credentials or reauthentication, remains stable for stored-credential reimport, and does not rotate during ordinary token refresh
- teaches `QueryManager` to install source-scan observed-values publishers with the canonical runtime fingerprint and credential revision for query execution over selected sources
- keeps metadata, catalog, explain, describe, and validation paths from installing observed-values publishers
- clears/increments observed-values source state from source lifecycle operations; cleanup is best effort and does not depend on a pre-existing SQLite search DB
- drains the observed-values background writer during server shutdown without blocking the async runtime, including when the server task returns an error
- keeps projection, retrieval, queue-drain processing, retention, and adapter behavior out of scope; observed values are collected into the durable queue but are still not returned by search

## Split boundary

PR #1472 owns the opaque SQLite observed-values substrate. PR #1761 owns source-authored `do_not_index` and is this PR's direct parent. This PR owns final write-side runtime identity and app lifecycle wiring. PR #1593 later reuses the same canonical loader for live read-side scope checks.

MCP and CLI remain thin: neither adapter opens SQLite, drains observed-values queues, ranks observed values, or owns observed-values lifecycle behavior.

## Validation

- credential revision lifecycle and runtime-fingerprint mutation tests
- writer/live scope identity parity tests on the retrieval child
- full `coral-app` test suite
- strict Clippy and rustdoc through top-of-stack `make rust-checks`
- isolated real-source observation/search smoke through the final stack

## Feature flag

Constructs and attaches `SearchObservationHandle` only when the effective `observed_values_search` feature is enabled. Default-off means no source/query publishers, no background writer, and no observation collection; config and process overrides use the shared feature system.

PR #1573, opened 2026-07-07
feat(search): add search maintenance controls

mystic123:
## Feature flag

Catalog search maintenance remains available regardless of `observed_values_search`. The observed-value maintenance paths introduced later in this stack are the paths gated by the default-off feature.

### DIFF
--- a/apps/docs/reference/cli-reference.mdx
+++ b/apps/docs/reference/cli-reference.mdx
@@ -93,7 +93,8 @@ Manage Coral's local search indexes.
 
 ### `coral search-index rebuild`
 
-Rebuild an app-owned local search index.
+Rebuild a local search index. Coral skips the rebuild when the index is already
+up to date unless you pass `--force`.
 
 ```shellscript
 coral search-index rebuild --provider catalog
@@ -111,11 +112,10 @@ coral search-index rebuild --provider catalog --force
 
 Clear Coral's local search data for one workspace.
 
-This command is destructive, so it requires `--yes` and the global
-`--workspace` flag. Use `--workspace NAME` to clear `NAME`. Use bare
-`--workspace` to clear the workspace Coral would otherwise select:
-`CORAL_WORKSPACE` when set, otherwise `default`. The global flag may appear
-after `clear`.
+Clear deletes local search data, so Coral requires both `--yes` and
+`--workspace`. Use `--workspace NAME` to clear that workspace. Use
+`--workspace` with no name to clear the workspace Coral would normally use for
+this command: `CORAL_WORKSPACE` if it is set, or `default` otherwise.
 
 ```shellscript
 coral search-index clear --scope all --workspace --yes
--- a/apps/docs/reference/cli-reference.mdx
+++ b/apps/docs/reference/cli-reference.mdx
@@ -89,7 +89,7 @@ For catalog search, the `catalog_metadata` provider status reports whether searc
 
 ## `coral search-index`
 
-Manage app-owned local search indexes.
+Manage Coral's local search indexes.
 
 ### `coral search-index rebuild`
 
@@ -109,11 +109,13 @@ coral search-index rebuild --provider catalog --force
 
 ### `coral search-index clear`
 
-Clear app-owned local search data for the current workspace.
-The global `--workspace [NAME]` flag is required as destructive confirmation:
-use it bare to confirm the selected workspace, or pass a name to select and
-confirm that workspace. Although `--workspace` is global, it may appear inside
-the `search-index clear` command.
+Clear Coral's local search data for one workspace.
+
+This command is destructive, so it requires `--yes` and the global
+`--workspace` flag. Use `--workspace NAME` to clear `NAME`. Use bare
+`--workspace` to clear the workspace Coral would otherwise select:
+`CORAL_WORKSPACE` when set, otherwise `default`. The global flag may appear
+after `clear`.
 
 ```shellscript
 coral search-index clear --scope all --workspace --yes
--- a/apps/docs/reference/cli-reference.mdx
+++ b/apps/docs/reference/cli-reference.mdx
@@ -97,38 +97,35 @@ Rebuild a local search index. Coral skips the rebuild when the index is already
 up to date unless you pass `--force`.
 
 ```shellscript
-coral search-index rebuild --provider catalog
-coral search-index rebuild --provider catalog --force
+coral search-index rebuild
+coral search-index rebuild --force
 ```
 
 #### Options
 
-| Option               | Type | Default   | Description                                                     |
-| -------------------- | ---- | --------- | --------------------------------------------------------------- |
-| `--provider catalog` | enum | `catalog` | Search index provider to rebuild                                |
-| `--force`            | flag | `false`   | Rebuild even when the stored projection fingerprint is current   |
+| Option    | Type | Default | Description                                                    |
+| --------- | ---- | ------- | -------------------------------------------------------------- |
+| `--force` | flag | `false` | Rebuild even when the stored projection fingerprint is current |
 
 ### `coral search-index clear`
 
 Clear Coral's local search data for one workspace.
 
-Clear deletes local search data, so Coral requires both `--yes` and
-`--workspace`. Use `--workspace NAME` to clear that workspace. Use
-`--workspace` with no name to clear the workspace Coral would normally use for
-this command: `CORAL_WORKSPACE` if it is set, or `default` otherwise.
+Clear deletes local search data, so Coral requires both `--yes` and an explicit
+`--workspace NAME`.
 
 ```shellscript
-coral search-index clear --scope all --workspace --yes
 coral search-index clear --scope all --workspace work --yes
 ```
 
 #### Options
 
-| Option                      | Type        | Default  | Description                                      |
-| --------------------------- | ----------- | -------- | ------------------------------------------------ |
-| `--scope all`               | enum        | required | Search data scope to clear                       |
-| global `--workspace [NAME]` | flag/string | required | Confirm workspace-scoped deletion; with `NAME`, also select that workspace |
-| `--yes`                     | flag        | required | Confirm destructive search-data deletion         |
+| Option                    | Type   | Default  | Description                               |
+| ------------------------- | ------ | -------- | ----------------------------------------- |
+| `--scope all`             | enum   | required | Search data scope to clear                |
+| global `--workspace NAME` | string | required | Workspace whose search data will be cleared |
+| `--yes`                   | flag   | required | Confirm destructive search-data deletion  |
+
 ## `coral source`
 
 Manage configured sources, either the bundled ones, or your custom source specs.
--- a/apps/docs/reference/cli-reference.mdx
+++ b/apps/docs/reference/cli-reference.mdx
@@ -87,6 +87,46 @@ With `--json`, the response separates `results`, `provider_statuses`, and `trunc
 
 For catalog search, the `catalog_metadata` provider status reports whether search found matches, completed with no matches, returned partial coverage, or failed. `truncation` reports when more matches were available than returned.
 
+## `coral search-index`
+
+Manage app-owned local search indexes.
+
+### `coral search-index rebuild`
+
+Rebuild an app-owned local search index.
+
+```shellscript
+coral search-index rebuild --provider catalog
+coral search-index rebuild --provider catalog --force
+```
+
+#### Options
+
+| Option               | Type | Default   | Description                                                     |
+| -------------------- | ---- | --------- | --------------------------------------------------------------- |
+| `--provider catalog` | enum | `catalog` | Search index provider to rebuild                                |
+| `--force`            | flag | `false`   | Rebuild even when the stored projection fingerprint is current   |
+
+### `coral search-index clear`
+
+Clear app-owned local search data for the current workspace.
+The global `--workspace [NAME]` flag is required as destructive confirmation:
+use it bare to confirm the selected workspace, or pass a name to select and
+confirm that workspace. Although `--workspace` is global, it may appear inside
+the `search-index clear` command.
+
+```shellscript
+coral search-index clear --scope all --workspace --yes
+coral search-index clear --scope all --workspace work --yes
+```
+
+#### Options
+
+| Option                      | Type        | Default  | Description                                      |
+| --------------------------- | ----------- | -------- | ------------------------------------------------ |
+| `--scope all`               | enum        | required | Search data scope to clear                       |
+| global `--workspace [NAME]` | flag/string | required | Confirm workspace-scoped deletion; with `NAME`, also select that workspace |
+| `--yes`                     | flag        | required | Confirm destructive search-data deletion         |
 ## `coral source`
 
 Manage configured sources, either the bundled ones, or your custom source specs.
--- a/apps/docs/reference/cli-reference.mdx
+++ b/apps/docs/reference/cli-reference.mdx
@@ -93,7 +93,7 @@ Manage Coral's local search indexes.
 
 ### `coral search-index rebuild`
 
-Rebuild a local search index. Coral skips the rebuild when the index is already
+Rebuild all local search indexes. Coral skips a rebuild when an index is already
 up to date unless you pass `--force`.
 
 ```shellscript
--- a/apps/docs/reference/cli-reference.mdx
+++ b/apps/docs/reference/cli-reference.mdx
@@ -114,6 +114,9 @@ Clear Coral's local search data for one workspace.
 Clear deletes local search data, so Coral requires both `--yes` and an explicit
 `--workspace NAME`.
 
+`all` is currently the only supported scope. It clears all local search data
+for the selected workspace.
+
 ```shellscript
 coral search-index clear --scope all --workspace work --yes
 ```
@@ -122,7 +125,7 @@ coral search-index clear --scope all --workspace work --yes
 
 | Option                    | Type   | Default  | Description                               |
 | ------------------------- | ------ | -------- | ----------------------------------------- |
-| `--scope all`             | enum   | required | Search data scope to clear                |
+| `--scope <SCOPE>`         | enum   | required | Scope to clear. Supported value: `all`    |
 | global `--workspace NAME` | string | required | Workspace whose search data will be cleared |
 | `--yes`                   | flag   | required | Confirm destructive search-data deletion  |
 
--- a/crates/coral-api/proto/coral/v1/search.proto
+++ b/crates/coral-api/proto/coral/v1/search.proto
@@ -29,19 +29,17 @@ message SearchResponse {
   SearchResultTruncation truncation = 3;
 }
 
-// Search index maintenance request.
+// Request to rebuild all app-owned local search indexes.
 message RebuildSearchIndexRequest {
   // Workspace whose local search data should be rebuilt.
   Workspace workspace = 1;
-  // Index provider to rebuild.
-  SearchIndexProvider provider = 2;
   // Force a rebuild even when the persisted fingerprint says the projection is current.
-  bool force = 3;
+  bool force = 2;
 }
 
 // Search index maintenance response.
 message RebuildSearchIndexResponse {
-  // Maintenance results for the providers selected by the request.
+  // Maintenance results for rebuilt providers.
   repeated SearchMaintenanceResult results = 1;
 }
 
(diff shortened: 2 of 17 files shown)


mystic123:
* **#1472** <a href="https://app.graphite.com/github/pr/withcoral/coral/1472?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1573** <a href="https://app.graphite.com/github/pr/withcoral/coral/1573?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1573?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1372** <a href="https://app.graphite.com/github/pr/withcoral/coral/1372?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/docs/reference/cli-reference.mdx
```diff
@@ -81,6 +81,43 @@ answer the question by itself; use the returned `sql_reference` or
 | `--json`  | flag    | `false` | Render the shared machine-readable JSON response          |
 | `--limit` | integer | `10`    | Maximum results to return, from 1 to 50                   |
 
+## `coral search-index`
+
+Manage app-owned local search indexes.
```
AlbertQM: What's an app? (I think whatever your explanation is, would be good to replace the text here and everywhere else the had the same copy)

apps/docs/reference/cli-reference.mdx
```diff
@@ -81,6 +81,47 @@ answer the question by itself; use the returned `sql_reference` or
...
+| Option               | Type | Default   | Description                                                     |
+| -------------------- | ---- | --------- | --------------------------------------------------------------- |
+| `--provider catalog` | enum | `catalog` | Search index provider to rebuild                                |
+| `--force`            | flag | `false`   | Rebuild even when the stored projection fingerprint is current   |
+
+### `coral search-index clear`
+
+Clear app-owned local search data for the current workspace.
+The global `--workspace [NAME]` flag is required as destructive confirmation:
+use it bare to confirm the selected workspace, or pass a name to select and
```
AlbertQM: After reading this, I'm not sure what happens if I don't pass a workspace name. Can you write it in your own words please? Thanks!

apps/docs/reference/cli-reference.mdx
```diff
@@ -81,6 +81,49 @@ answer the question by itself; use the returned `sql_reference` or
 | `--json`  | flag    | `false` | Render the shared machine-readable JSON response          |
 | `--limit` | integer | `10`    | Maximum results to return, from 1 to 50                   |
 
+## `coral search-index`
+
+Manage Coral's local search indexes.
+
+### `coral search-index rebuild`
+
+Rebuild an app-owned local search index.
```
AlbertQM: What's an app..? 🙈 

I saw you update the command above, so should this be something like `Rebuild a search index.`, plus some more information a human would find useful? E.g. the fact that we don't actually rebuild if the index is already up to date

apps/docs/reference/cli-reference.mdx
```diff
@@ -81,6 +81,49 @@ answer the question by itself; use the returned `sql_reference` or
...
+
+### `coral search-index clear`
+
+Clear Coral's local search data for one workspace.
+
+This command is destructive, so it requires `--yes` and the global
+`--workspace` flag. Use `--workspace NAME` to clear `NAME`. Use bare
+`--workspace` to clear the workspace Coral would otherwise select:
+`CORAL_WORKSPACE` when set, otherwise `default`. The global flag may appear
+after `clear`.
```
AlbertQM:
```suggestion
Clear deletes local search data, so Coral requires both `--yes` and `--workspace`. Use `--workspace NAME` to clear that workspace. Use `--workspace` with no name to clear the workspace Coral would normally use for this command: `CORAL_WORKSPACE` if it is set, or `default` otherwise.
```

I'd remove the flag positioning too, as we don't need to explain in our docs where to put --flags

mystic123:
## Review guide

*A short explainer to help review this PR (second of the three stacked PRs: #1470 → #1573 → #1472; the diff is stacked on #1470, so it contains only this PR's changes).*

**What it does:** Adds operator-facing maintenance for the local sqlite search: `coral search-index rebuild` (re-derives the catalog projection, with `--force` for corruption the fingerprint check can't see) and a confirmed destructive `coral search-index clear` that wipes a workspace's search data and compacts the database. Groundwork for #1472, which makes local search state something users may want to purge.

**How it works:** A new transport-neutral `SearchProviderMaintenance` trait (`search/maintenance.rs`) is implemented by `CatalogMetadataProvider`; the not-yet-existing observed-values provider is reported as `Skipped` on rebuild. Rebuild runs in a single SQLite Immediate transaction doing delete-and-reinsert only when `force || projection_changed`; clear deletes rows then runs best-effort compaction (WAL checkpoint + VACUUM) whose failures become status notes, not errors. Two new gRPC RPCs (`RebuildSearchIndex`, `ClearSearchData` in `search.proto`) run the blocking work via `spawn_blocking`. The CLI's clear demands both `--yes` and an explicit `--workspace`, implemented by making the **global** `--workspace` flag optional-valued with a sentinel `default_missing_value`.

**Where to focus review:**
- The `--workspace` rework in `coral-cli/src/lib.rs` is the riskiest change: `num_args = 0..=1` with an in-band magic-string sentinel changes parsing for *every* command — bare `--workspace` elsewhere now fails at runtime instead of parse time, and a workspace literally named the sentinel is unreachable. Check clap value-greediness against adjacent positionals.
- Clear-scope asymmetry: `clear_data` forwards everything to the catalog provider only, so `scope=OBSERVED` gets a confusing error ("catalog search provider supports only all search-data clear scope"), and clear responses never include a `Skipped` observed entry the way rebuild does — a deliberate choice worth settling before #1472.
- Error mapping in `catalog/provider.rs` (`search_sqlite_app_error`): lock contention becomes `Unavailable`, but every other SQLite failure — including I/O errors and corruption — becomes `FailedPrecondition`; `Internal` may be more honest.
- Maintenance results repurpose search-time `ProviderCoverage`/`ResultsFound` semantics (e.g. `eligible_units = returned_count = new_document_count`), which reads oddly for a rebuild and is baked into the public proto.

**Watch-outs:**
- Compaction failure after a clear still exits 0 with a "Cleared N documents" line — the failure is visible only in the compaction status labels — so disk space may not actually be reclaimed on a nominally successful clear.
- Maintenance holds no app-level lock during the SQLite work, so concurrent searches/rebuilds/clears are serialized only by SQLite busy timeouts, surfacing as `UNAVAILABLE` with no server-side retry.
- The proto leaks internal stack naming ("PR 4A supports workspace-scoped clears only") into the public API, and `SearchClearTarget.source_name` passes service validation but always fails downstream as not-implemented.

**Tests:** Good coverage across four layers — sqlite unit tests (forced rebuild repairs a corrupted projection, clear preserves schema metadata, compaction status), gRPC integration tests (rebuild for catalog/all/observed with `Skipped`, clear followed by lazy projection recreation), clap parse-level tests (bare `--workspace` parses everywhere but only `search-index clear` accepts it), and CLI e2e against a mock server. Gaps: no tests for the gRPC error paths (scope=OBSERVED clear, missing/false/empty clear target, lock-contention mapping) and no runtime test that bare `--workspace` on a non-clear command produces the intended error.

crates/coral-cli/src/lib.rs
```diff
@@ -146,10 +151,71 @@ struct SearchArgs {
...
+#[derive(Debug, Clone, Copy, ValueEnum)]
+enum SearchClearScope {
+    All,
+}
+
 #[derive(Debug, Args)]
 struct WorkspaceSelectionArgs {
     /// Workspace to target. Overrides `CORAL_WORKSPACE`.
-    #[arg(long = "workspace", value_name = "NAME", global = true)]
+    #[arg(
```
sauldhernandez, line 206: P1: This overloads the global workspace selector as destructive confirmation, but flag presence and workspace selection can diverge. `--workspace "$WS"` with an empty `$WS`, or an explicitly named workspace `__coral_current_workspace_confirmation__` (valid under current validation), sets `workspace_scope = true`; `selected_workspace_name` then discards that value and clears `CORAL_WORKSPACE`/`default` instead. The optional value is also greedy: `coral --workspace search-index clear --scope all --yes` consumes `search-index` as `NAME` and fails parsing. Please separate destructive confirmation from workspace selection, or require a non-empty explicit workspace, and remove the in-band sentinel. Confidence: high.

crates/coral-app/src/search/sqlite_store.rs
```diff
@@ -133,6 +142,42 @@ impl SqliteSearchStore {
         SqliteCatalogIndex::new().clear_workspace(&mut connection, &self.workspace_name)
     }
 
+    pub(crate) fn compact_after_clear(&self) -> SqliteSearchCompactionResult {
+        let mut notes = Vec::new();
+        let wal_checkpoint_truncate_completed =
+            match self.execute_maintenance_batch("PRAGMA wal_checkpoint(TRUNCATE);") {
```
sauldhernandez: P2: `execute_batch` only proves that the PRAGMA executed; it discards the returned checkpoint row. `PRAGMA wal_checkpoint(TRUNCATE)` reports reader contention as `(busy = 1, log, checkpointed)` while still returning SQLite success. With a held reader I reproduced `1|1|0` and exit status 0, so this sets `wal_checkpoint_truncate_completed = true` while the WAL remains. Please query and validate the result row, report completion only when `busy == 0`, and add a contention test. Confidence: high.

crates/coral-app/src/search/manager.rs
```diff
@@ -28,11 +39,75 @@ impl SearchManager {
...
+        };
+        Ok(RebuildSearchIndexResponse { provider_results })
+    }
+
+    pub(crate) fn clear_data(
+        &self,
+        request: &ClearSearchDataRequest,
+    ) -> Result<ClearSearchDataResponse, SearchManagerError> {
+        self.require_workspace(&request.workspace_name)?;
+        let outcome = self.catalog.clear_data(SearchProviderClearRequest {
```
sauldhernandez, line 90: P2: The public API advertises `SEARCH_DATA_SCOPE_OBSERVED`, and the service accepts it, but every clear is dispatched to the catalog provider, which rejects every scope except `All`. A valid observed-only request therefore always returns `INVALID_ARGUMENT`; `All` also omits the explicit observed-provider `Skipped` result returned by rebuild. Please make the manager own scope-based provider routing/fan-out, or narrow the wire contract until observed clears exist. Confidence: high.

crates/coral-api/proto/coral/v1/search.proto
```diff
@@ -29,6 +29,120 @@ message SearchResponse {
...
+  oneof target {
+    // Clear the selected scope for the whole workspace.
+    bool workspace = 1;
+    // Reserved for source-scoped clear in a later PR.
+    string source_name = 2;
+  }
+}
+
+// Provider-local maintenance result.
+message SearchMaintenanceProviderResult {
```
sauldhernandez: P2: This maintenance result reuses search-only state and coverage types. `SEARCH_PROVIDER_STATE_RESULTS_FOUND` is documented as returned search results, and `SearchProviderCoverage.returned_count` is documented as candidates returned before fusion, but catalog rebuild populates them with document counts and clear reports `EMPTY`. Machine-readable maintenance responses therefore contradict the public schema. Please introduce maintenance-specific state/metrics, or omit search coverage and rely on the maintenance detail messages.

crates/coral-app/src/search/catalog/provider.rs
```diff
@@ -180,6 +196,142 @@ impl CatalogMetadataProvider {
...
+
+fn search_compaction_status(result: SqliteSearchCompactionResult) -> SearchCompactionStatus {
+    SearchCompactionStatus {
+        wal_checkpoint_truncate_completed: result.wal_checkpoint_truncate_completed,
+        vacuum_completed: result.vacuum_completed,
+        note: result.note,
+    }
+}
+
+fn search_sqlite_app_error(error: &SqliteSearchError) -> AppError {
```
sauldhernandez, line 324: P2: This maps every non-lock SQLite failure to `FAILED_PRECONDITION`, including database corruption, disk I/O, permission failures, and disk exhaustion. Clients are told these server/storage failures are caller-fixable preconditions and may handle them as permanent request failures. Please map typed categories: contention to `UNAVAILABLE`, unsupported capability/schema to `FAILED_PRECONDITION`, and I/O/corruption/internal failures to their corresponding server-side status. Confidence: high.

crates/coral-cli/src/lib.rs
```diff
@@ -907,6 +1016,132 @@ async fn run_search(
...
+        SearchClearScope::All => SearchDataScope::All,
+    }
+}
+
+fn print_search_rebuild_response(response: &coral_api::v1::RebuildSearchIndexResponse) {
+    for result in &response.provider_results {
+        match result.detail.as_ref() {
+            Some(search_maintenance_provider_result::Detail::CatalogRebuild(detail)) => {
+                println!(
+                    "Rebuilt {} search index: old documents {}, new documents {}, projection changed {}, rebuild performed {}.",
```
sauldhernandez: P2: This always says `Rebuilt` even when `rebuild_performed` is false. A normal second non-forced invocation therefore prints `Rebuilt ... rebuild performed no`, contradicting the documented behavior that an already-current index is skipped. Please branch on `rebuild_performed` or render the provider note so a no-op does not claim a rebuild. Confidence: high.

apps/docs/reference/cli-reference.mdx
```diff
@@ -81,6 +81,49 @@ answer the question by itself; use the returned `sql_reference` or
...
+```shellscript
+coral search-index rebuild --provider catalog
+coral search-index rebuild --provider catalog --force
+```
+
+#### Options
+
+| Option               | Type | Default   | Description                                                     |
+| -------------------- | ---- | --------- | --------------------------------------------------------------- |
+| `--provider catalog` | enum | `catalog` | Search index provider to rebuild                                |
```
sauldhernandez: Would it make more sense for the default to be all providers?

apps/docs/reference/cli-reference.mdx
```diff
@@ -81,6 +81,49 @@ answer the question by itself; use the returned `sql_reference` or
...
+| `--force`            | flag | `false`   | Rebuild even when the stored projection fingerprint is current   |
+
+### `coral search-index clear`
+
+Clear Coral's local search data for one workspace.
+
+Clear deletes local search data, so Coral requires both `--yes` and
+`--workspace`. Use `--workspace NAME` to clear that workspace. Use
+`--workspace` with no name to clear the workspace Coral would normally use for
```
sauldhernandez: I don't understand the usefulness of allowing `--workspace` with no args. It's fundamentally different to how the `--workspace` arg works anywhere else in the CLI

crates/coral-api/proto/coral/v1/search.proto
```diff
@@ -29,6 +29,120 @@ message SearchResponse {
...
+  // Index provider to rebuild.
+  SearchIndexProvider provider = 2;
+  // Force a rebuild even when the persisted fingerprint says the projection is current.
+  bool force = 3;
+}
+
+// Search index maintenance response.
+message RebuildSearchIndexResponse {
+  // Per-provider maintenance results.
+  repeated SearchMaintenanceProviderResult provider_results = 1;
```
sauldhernandez: The naming here seems to expose too much about the internals. Is an API client aware of what a MaintenanceProvider is?

crates/coral-api/proto/coral/v1/search.proto
```diff
@@ -29,6 +29,120 @@ message SearchResponse {
...
+  SearchDataScope scope = 2;
+  // Removal target. PR 4A supports workspace-scoped clears only.
+  SearchClearTarget target = 3;
+}
+
+// Search data removal response.
+message ClearSearchDataResponse {
+  // Per-provider maintenance results.
+  repeated SearchMaintenanceProviderResult provider_results = 1;
+  // Best-effort SQLite compaction status after row deletion.
```
sauldhernandez: API should not expose whether the backing is sqlite or anything else for that matter

crates/coral-api/proto/coral/v1/search.proto
```diff
@@ -29,6 +29,120 @@ message SearchResponse {
...
+  repeated SearchMaintenanceProviderResult provider_results = 1;
+}
+
+// Search data removal request.
+message ClearSearchDataRequest {
+  // Workspace whose local search data should be removed.
+  Workspace workspace = 1;
+  // Data scope to remove.
+  SearchDataScope scope = 2;
+  // Removal target. PR 4A supports workspace-scoped clears only.
```
sauldhernandez: I don't think we should mention specific PRs here

crates/coral-api/proto/coral/v1/search.proto
```diff
@@ -29,6 +29,120 @@ message SearchResponse {
...
+  bool rebuild_performed = 4;
+}
+
+// Catalog clear detail.
+message CatalogClearResult {
+  // Catalog document rows deleted from the workspace projection.
+  uint32 deleted_document_count = 1;
+}
+
+// Best-effort SQLite compaction status after destructive maintenance.
```
sauldhernandez: Again, we shouldn't expose that here IMO

crates/coral-app/src/search/catalog/provider.rs
```diff
@@ -83,7 +92,14 @@ impl CatalogMetadataProvider {
...
     fn load_catalog(&self, request: &SearchRequest) -> Result<CatalogInfo, AppError> {
-        self.catalog_loader.load_catalog(&request.workspace_name)
+        self.load_catalog_for_workspace(&request.workspace_name)
+    }
+
+    fn load_catalog_for_workspace(
+        &self,
+        workspace_name: &WorkspaceName,
+    ) -> Result<CatalogInfo, AppError> {
+        self.catalog_loader.load_catalog(workspace_name)
```
sauldhernandez: This function does very little, better to inline it.

sauldhernandez (changes requested):

apps/docs/reference/cli-reference.mdx
```diff
@@ -81,6 +81,45 @@ answer the question by itself; use the returned `sql_reference` or
...
+
+```shellscript
+coral search-index clear --scope all --workspace work --yes
+```
+
+#### Options
+
+| Option                    | Type   | Default  | Description                               |
+| ------------------------- | ------ | -------- | ----------------------------------------- |
+| `--scope all`             | enum   | required | Search data scope to clear                |
```
AlbertQM: We never explain or list what other options are available other than `all`, is that intended?

AlbertQM (approved):
Thank you! I think we should extend the docs to explain what `--scope` does, an ideally enumerating the values of the `enum`. If we only have `all` for now, feel free to leave as-is and update in future PRs

crates/coral-app/src/search/sqlite_store.rs
```diff
@@ -181,6 +255,38 @@ impl SqliteSearchError {
                 )
         )
     }
+
+    pub(crate) fn is_storage_exhaustion(&self) -> bool {
```
sauldhernandez, line 259: P2: The error categorization is still incomplete. `is_storage_exhaustion` only recognizes `SqliteSearchError::Sqlite`, so ENOSPC from `create_new_file_private` is converted to `SqliteSearchError::Io` by `ensure_sqlite_file` and falls through to `AppError::Internal`, not `RESOURCE_EXHAUSTED`. `detect_capabilities` also converts every probe failure to `false` via `.is_ok()`, so `SQLITE_FULL`, `SQLITE_NOMEM`, `SQLITE_IOERR`, or corruption during a probe is mislabeled as `UnsupportedCapability` and returned as `FAILED_PRECONDITION`. Please preserve operational error categories through file creation and capability detection, and add tests for those pre-classification paths. Confidence: high.

crates/coral-cli/src/lib.rs
```diff
@@ -907,6 +980,124 @@ async fn run_search(
...
+    workspace: &Workspace,
+    args: SearchIndexArgs,
+) -> Result<(), CliError> {
+    match args.command {
+        SearchIndexCommand::Rebuild(args) => {
+            let response = app
+                .search_client()
+                .rebuild_search_index(Request::new(RebuildSearchIndexRequest {
+                    workspace: Some(workspace.clone()),
+                    provider: SearchIndexProvider::Catalog as i32,
```
sauldhernandez: P2: This does not implement the requested all-provider default; it hardcodes catalog. Narrowing the current wire enum to one provider makes the behavior equivalent today, but when another search index provider lands this command will silently keep rebuilding only catalog unless every caller and this default are revisited. Please either make rebuild an all-provider operation with manager-owned fan-out, or remove the provider field from the public request while catalog is the sole implementation so this default cannot calcify incorrectly. Confidence: high.

sauldhernandez (approved):

mystic123:
### Merge activity

* **Jul 17, 7:09 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1573).
* **Jul 17, 7:10 AM UTC**: [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1573) couldn't merge this PR because it had merge conflicts.
* **Jul 17, 8:40 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1573).
* **Jul 17, 8:40 AM UTC**: @mystic123 merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1573).

Issue #1575, opened 2026-07-07
DigitalOcean OpenAPI v3 import produces no runtime components


simonwhitaker:
## Summary

Importing DigitalOcean's OpenAPI v3 descriptor as a DSL v4 OpenAPI surface succeeds at materialization time but the installed source has no runtime components, so validation warns immediately after `source add`.

This was discovered while validating `sw/openapi-pagination-detection` against public OpenAPI specs. It does not appear related to the pagination changes: the imported semantic IR has no response cardinality for any operation.

## Repro

```yaml
name: digitalocean_openapi_validation
dsl_version: 4
surfaces:
  - id: rest
    type: openapi
    url: https://raw.githubusercontent.com/digitalocean/openapi/main/specification/DigitalOcean-public.v2.yaml
    base_url: https://api.digitalocean.com
```

Run:

```sh
CORAL_CONFIG_DIR=$(mktemp -d) cargo run --quiet -p coral-cli -- source add --file <manifest.yaml>
```

## Observed

`source add` installs the source, then validation warns:

```text
Added source digitalocean_openapi_validation (secrets: none)
Warning: validation failed: code: 'The system is not in a state required for the operation's execution', message: "failed precondition: failed precondition: source 'digitalocean_openapi_validation' has no runtime components"
```

Artifact inspection from the same run:

- raw descriptor size: 110,891 bytes
- semantic IR operations: 663
- operation output cardinality: 663 `none`, 0 `list`/`wrapped_list`/`singleton`
- pagination modes: 663 `none`
- projection catalog entries: 663 tables
- materialization diagnostics: 0

## Expected

For a public OpenAPI v3 descriptor with hundreds of paths, DSL v4 should either:

- generate usable runtime components, or
- emit actionable diagnostics explaining why no response schemas/runtime components can be produced.

Silently importing 663 operations with `cardinality: none`, then failing only at runtime-component validation, is too late and not actionable enough for source authors.

simonwhitaker:
This is a concrete example of #1497; DigitalOcean's OpenAPI spec uses external refs, which Coral doesn't currently resolve.

PR #1576, opened 2026-07-07
feat(spec): infer OpenAPI pagination hints

### LINKED ISSUE #1485: Improve automatic pagination detection
Today, for OpenAPI surfaces, we determine pagination thus:

1. If an operation has `page` and `per_page` parameters, it uses `PaginationMode::Page`
2. Else, it uses `PaginationMode::None`

😳 

Improve those heuristics to cover common pagination strategies.

simonwhitaker:
## Summary

This teaches the DSL v4 OpenAPI importer to infer pagination from common OpenAPI request/response shapes, and extends the HTTP runtime so those inferred hints can actually be executed.

The importer now detects pagination for list-like responses, then chooses the first supported pattern it can prove from the operation shape:

- link-header pagination from a declared `Link` response header
- next-page URL header pagination from headers like `X-Next-Page-Url`
- cursor query pagination from cursor-like query params plus response cursor fields or headers
- X-style `pagination_token` cursor pagination from response `meta.next_token`
- token-style `page` cursor pagination when `page` is a string parameter and the response exposes `next_page` / `nextPage`
- offset pagination from `offset` + `limit`
- page-number pagination from numeric `page`/`pageNumber` + page-size params

Runtime support now includes response-cursor headers for `cursor_query` / `cursor_body` and custom next-page URL headers for `link_header`, with same-origin checks for URL-bearing pagination headers.

## Why

Generated v4 sources currently need too much manual pagination cleanup for ordinary OpenAPI specs. The old inference only recognized the exact `page` + `per_page` pattern, which misses common cursor, offset, header, and next URL pagination contracts. This makes generated sources more useful out of the box without guessing pagination for singleton responses.

## Notable details

- OpenAPI response selection now retains resolved response headers and response cardinality as pagination context.
- Header-backed cursor pagination can use `response_cursor_header` instead of `response_cursor_path`.
- Link-header pagination can use `next_url_header` when APIs return the next URL in a custom header rather than RFC 5988 `Link`.
- Optional string `page` parameters are treated as cursor tokens only when the response has a string next-page cursor; numeric `page` remains page-number pagination.
- `pagination_token` is recognized as a cursor query parameter for APIs that return `next_token` in response metadata.
- Pagination-owned query inputs are still hidden from generated projections via the existing projection sync path.
- Generated source-manifest schema is updated for the new pagination fields.

## Validation

Automated checks:

- `make rust-checks`
- `make schema-check`
- `cargo test -p coral-spec v4::openapi_tests:: -- --nocapture`
- `cargo test -p coral-engine --test engine pagination_ -- --nocapture`

Real OpenAPI import smoke tests via `coral source add --file` with isolated `CORAL_CONFIG_DIR`:

- X API: 165 operations; inferred 38 `cursor_query` paginated operations.
- GitHub REST API: 1194 operations; inferred 193 `link_header` and 68 `page` paginated operations.
- Stripe API: 587 operations; inferred 7 `cursor_query` paginated operations for search endpoints using string `page` + response `next_page`.
- GitLab API: 1729 operations; inferred 1 `cursor_query` and 51 `page` paginated operations.

GitHub note: GitHub's REST docs say paginated responses use `Link`, but this import is constrained by the OpenAPI descriptor. The descriptor currently declares `Link` on about 206 operations, including some singleton-shaped responses that this importer intentionally excludes, so the 193 `link_header` result is close to the explicit descriptor signal. Operations with `page`/`per_page` but no declared `Link` header fall back to page-number pagination.

X note: the descriptor has about 39 list-like operations with `pagination_token` or `next_token` plus response `next_token`. This PR now infers 38 of them; the remaining miss has response metadata behind a nested `$ref`, which needs separate response-schema resolution hardening. Tracked in [#1578](https://github.com/withcoral/coral/issues/1578).

Stripe limitation: Stripe's ordinary list endpoints use `starting_after`/`ending_before` with `has_more`, where the next cursor is derived from the last returned row id. This PR does not add a DSL/runtime primitive for "cursor from last row when has_more", so those endpoints correctly remain uninferred rather than being counted as covered. That follow-up is tracked in #1577.

GitLab limitation: GitLab's OpenAPI descriptor has many more `page`/`per_page` operations than the 51 inferred here. The low count is mainly because many paginated endpoints declare a singular response schema ref rather than an array/list response shape, and this PR intentionally gates inferred pagination on list-like cardinality to avoid paginating singleton resources. Supporting those GitLab cases likely needs a separate, explicit cardinality inference strategy for strongly paginated OpenAPI operations. Tracked in [#1579](https://github.com/withcoral/coral/issues/1579).

While testing DigitalOcean's OpenAPI descriptor, I found an unrelated DSL v4 importer/runtime-component issue where all 663 imported operations had `cardinality: none`; that is tracked separately in #1575.

### DIFF
--- a/crates/coral-engine/src/backends/http/fetch.rs
+++ b/crates/coral-engine/src/backends/http/fetch.rs
@@ -164,7 +164,6 @@ pub(super) async fn fetch_rows(
                 table_headers: &active_request.headers,
                 table_name: target.name(),
                 method: active_request.method,
-                base_url: &base_url,
                 url: &url,
                 query_pairs: &query_pairs,
                 body: body.as_ref(),
--- a/crates/coral-engine/src/backends/http/fetch.rs
+++ b/crates/coral-engine/src/backends/http/fetch.rs
@@ -175,11 +175,12 @@ pub(super) async fn fetch_rows(
                 render_context,
                 allow_404_empty: target.response().allow_404_empty,
                 link_header_require_results: pagination.link_header_require_results,
+                response_cursor_header: target.pagination().response_cursor_header.as_deref(),
             },
         )
         .await?;
 
-        let Some((payload, next_url)) = request else {
+        let Some((payload, pagination_hints)) = request else {
             break;
         };
 
@@ -231,12 +232,13 @@ pub(super) async fn fetch_rows(
         match &pagination.mode {
             ValidatedPaginationMode::None => break,
             ValidatedPaginationMode::CursorQuery | ValidatedPaginationMode::CursorBody => {
-                let next_cursor =
+                let next_cursor = pagination_hints.cursor.or_else(|| {
                     get_path_value(&payload, &target.pagination().response_cursor_path)
                         .and_then(Value::as_str)
                         .map(str::trim)
                         .filter(|s| !s.is_empty())
-                        .map(ToOwned::to_owned);
+                        .map(ToOwned::to_owned)
+                });
                 match next_cursor {
                     Some(cursor) => state.cursor = Some(cursor),
                     None => break,
@@ -265,10 +267,12 @@ pub(super) async fn fetch_rows(
                     })?;
                 state.offset = state.offset.saturating_add(step);
             }
-            ValidatedPaginationMode::LinkHeader | ValidatedPaginationMode::Auto => match next_url {
-                Some(next) => state.next_url = Some(next),
-                None => break,
-            },
+            ValidatedPaginationMode::LinkHeader | ValidatedPaginationMode::Auto => {
+                match pagination_hints.next_url {
+                    Some(next) => state.next_url = Some(next),
+                    None => break,
+                }
+            }
         }
     }
 
--- a/crates/coral-engine/src/backends/http/fetch.rs
+++ b/crates/coral-engine/src/backends/http/fetch.rs
@@ -176,6 +176,7 @@ pub(super) async fn fetch_rows(
                 allow_404_empty: target.response().allow_404_empty,
                 link_header_require_results: pagination.link_header_require_results,
                 response_cursor_header: target.pagination().response_cursor_header.as_deref(),
+                next_url_header: target.pagination().next_url_header.as_deref(),
             },
         )
         .await?;
--- a/crates/coral-spec/src/common.rs
+++ b/crates/coral-spec/src/common.rs
@@ -705,6 +705,8 @@ pub struct PaginationSpec {
     #[serde(default)]
     pub link_header_require_results: bool,
     #[serde(default)]
+    pub next_url_header: Option<String>,
+    #[serde(default)]
     pub max_pages: Option<usize>,
 }
 
@@ -724,6 +726,7 @@ impl Default for PaginationSpec {
             offset_start: 0,
             offset_step: None,
             link_header_require_results: false,
+            next_url_header: None,
             max_pages: None,
         }
     }
--- a/crates/coral-spec/src/common.rs
+++ b/crates/coral-spec/src/common.rs
@@ -788,6 +788,16 @@ impl PaginationSpec {
         table: &str,
         has_page_size: bool,
     ) -> Result<ValidatedPaginationMode> {
+        if self
+            .response_cursor_header
+            .as_deref()
+            .is_some_and(|header| header.trim().is_empty())
+        {
+            return Err(ManifestError::validation(format!(
+                "{schema}.{table} pagination.response_cursor_header must not be empty"
+            )));
+        }
+
         match self.mode {
             PaginationMode::None => Ok(ValidatedPaginationMode::None),
             PaginationMode::Auto => Ok(ValidatedPaginationMode::Auto),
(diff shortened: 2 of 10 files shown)


AlbertQM (approved):

crates/coral-spec/src/v4/surfaces/openapi/operations.rs
```diff
@@ -299,30 +301,255 @@ fn parameter_is_required(parameter_obj: &Map<String, Value>, location: IrInputLo
...
-        .any(|input| input.location == IrInputLocation::Query && input.name == "page");
-    let has_per_page = inputs
+        .find(|(name, header)| {
+            RESPONSE_CURSOR_HEADER_TOKENS.contains(&name_token(name).as_str())
+                && response_header_allows_string(header)
+        })
+        .map(|(name, _)| name.clone())
+}
+
+fn response_next_url_header(context: &OpenApiResponsePaginationContext) -> Option<String> {
```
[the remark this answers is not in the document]
simonwhitaker: Filed #1580 to track this as follow-up heuristic work: https://github.com/withcoral/coral/issues/1580

I agree this is ambiguous, but I do not think we should tune it from one synthetic case in this PR. The issue calls out testing against a larger OpenAPI corpus, including `konfig-sdks/openapi-examples`, and then making the heuristic change that gets us the right answer most often.

crates/coral-spec/src/v4/surfaces/openapi/operations.rs
```diff
@@ -299,30 +301,262 @@ fn parameter_is_required(parameter_obj: &Map<String, Value>, location: IrInputLo
...
+) -> Option<PaginationSpec> {
+    let cursor_input = find_optional_query_input(
+        inputs,
+        &[
+            "after",
+            "continuationtoken",
+            "cursor",
+            "marker",
+            "nextcursor",
+            "nextpage",
```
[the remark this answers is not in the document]
simonwhitaker: Filed #1581 to track this as follow-up heuristic work: https://github.com/withcoral/coral/issues/1581

This is a real ambiguity, but it needs corpus-driven tuning rather than a one-off rule here. The issue calls out testing against a larger OpenAPI set, including `konfig-sdks/openapi-examples`, and then changing the inference toward whichever behavior is right most often.

crates/coral-spec/src/common.rs
```diff
@@ -736,6 +742,16 @@ impl PaginationSpec {
...
+            .is_some_and(|header| header.trim().is_empty())
+        {
+            return Err(ManifestError::validation(format!(
+                "{schema}.{table} pagination.response_cursor_header must not be empty"
+            )));
+        }
+
         match self.mode {
             PaginationMode::None => Ok(ValidatedPaginationMode::None),
             PaginationMode::Auto => Ok(ValidatedPaginationMode::Auto),
```
iaptsiauri, line 803: I think we should remove this. It is same as LinkHeader pagination.

crates/coral-engine/src/backends/http/fetch.rs
```diff
@@ -175,11 +174,13 @@ pub(super) async fn fetch_rows(
                 render_context,
                 allow_404_empty: target.response().allow_404_empty,
                 link_header_require_results: pagination.link_header_require_results,
```
iaptsiauri, line 176: This pagination config is starting to leak across layers. `fetch.rs` passes `link_header_require_results`, `response_cursor_header`, and `next_url_header` into `transport.rs`,
which means transport now needs to understand pagination-specific response semantics.

I think the cleaner design is for `transport.rs` to just execute the HTTP request and return the decoded payload plus response headers. Then the pagination module can own both
sides of pagination:

- request mutation: query params, body fields, page size
- response extraction: `Link`, cursor headers, next-url headers, body cursor paths
- state advancement / stop conditions

That also avoids continuing to add optional pagination fields to `OutgoingHttpRequest` every time we support another mode, e.g. a future `X-Next-Page: 2` style header. Ideally
these response sources should be part of a validated pagination plan/mode, not pulled partly from raw `target.pagination()` and partly from `ValidatedPagination`.

crates/coral-spec/src/v4/surfaces/openapi/operations.rs
```diff
@@ -299,30 +301,263 @@ fn parameter_is_required(parameter_obj: &Map<String, Value>, location: IrInputLo
...
+fn detect_link_header_pagination(
+    inputs: &[IrOperationInput],
+    context: &OpenApiResponsePaginationContext,
+) -> Option<PaginationSpec> {
+    let has_link_header = response_header(context, &["link"]).is_some();
+    let next_url_header = response_next_url_header(context);
+    if !has_link_header && next_url_header.is_none() {
+        return None;
+    }
+    Some(PaginationSpec {
```
iaptsiauri, line 328: In case of `link_header` pagination, `per_page is correctly hidden but page` is leaked as a SQL filter for in the projection

iaptsiauri (commented):
Two bugs but one general design comment we don't have to address the design comment if it is a massive change at the moment

crates/coral-spec/src/common.rs
simonwhitaker: Yeah I agree, the concept of "auto" was never fully implemented, and becomes redundant with DSL v4. Will leave this for now but clean it up once we're completely on DSL v4.

simonwhitaker:
* **#1583** <a href="https://app.graphite.com/github/pr/withcoral/coral/1583?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1576** <a href="https://app.graphite.com/github/pr/withcoral/coral/1576?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1576?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-engine/src/backends/http/fetch.rs
simonwhitaker: Addressed in #1583

PR #1592, opened 2026-07-09
feat: DSL V4 lookup key inference 

iaptsiauri:
**Summary**
- Adds DSL v4 `lookup_keys` parameter metadata, including parsing and validation for generated metadata and user overrides.
- Infers lookup-key exclusions during materialization for pagination, search, and presentation query params, then writes that metadata alongside each REST surface artifact.
- Threads lookup-key metadata through projection generation so REST filter specs can set `lookup_key: true` only when the filter is safe for dependent joins.
- Keeps excluded parameters as normal pushdown filters; exclusion only disables joinability, not SQL exposure.
- Bumps the v4 projection generator to `derive-read-v8` 

Confidence: high.

### DIFF
--- a/crates/coral-app/src/sources/materialization.rs
+++ b/crates/coral-app/src/sources/materialization.rs
@@ -723,7 +723,7 @@ fn write_materialization(
             },
         )));
     }
-    let projections = generate_projection_catalog(manifest, &semantic_irs)
+    let projections = generate_projection_catalog(manifest, &semantic_irs, &BTreeMap::new())
         .map_err(|error| AppError::FailedPrecondition(error.to_string()))?;
     diagnostics.extend(projections.diagnostics.clone());
     for ir in &semantic_irs {
--- a/crates/coral-app/src/sources/materialization.rs
+++ b/crates/coral-app/src/sources/materialization.rs
@@ -6,16 +6,14 @@ use std::path::{Path, PathBuf};
 use std::time::Duration;
 
 use coral_spec::v4::{
-    Diagnostic, DiagnosticSeverity, Fingerprint, FingerprintSurface, GeneratedParameterMetadata,
-    LookupKeysMetadata, MCP_IMPORTER_VERSION, MaterializedSurface, McpToolCatalog,
-    OPENAPI_IMPORTER_VERSION, PROJECTION_GENERATOR_VERSION, ProjectionCatalog,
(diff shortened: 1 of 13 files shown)


crates/coral-app/src/sources/materialization.rs
```diff
@@ -394,16 +402,50 @@ fn read_raw_source_document_artifact(
     })
 }
 
+fn read_generated_lookup_keys(
+    surface_dir: &Path,
+    semantic_ir: &SemanticIr,
+) -> Result<Option<LookupKeysMetadata>, AppError> {
+    let path = surface_dir.join(PARAMETER_METADATA_OVERRIDE_FILENAME);
```
simonwhitaker: This is confusing. At the bare minimum, the constant needs renaming. But also, this file is currently only used for overrides; for pagination, the generated values are in the semantic-ir.yaml file. Does it make sense to store lookup keys in the IR, too? Right now we have:

| purpose | generated? | location |
|---|---|---|
| pagination | yes | {source dir}/surfaces/{surface id}/materialized/v4/semantic-ir.yaml |
| pagination | no | {source dir}/overrides/surfaces/{surface id}/parameter_metadata.yaml |
| lookup keys | yes | {source dir}/surfaces/{surface id}/parameter_metadata.yaml |
| lookup keys | no | {source dir}/overrides/surfaces/{surface id}/parameter_metadata.yaml |

I'm certain agents will get confused by this. Humans too.

simonwhitaker: I _think_ the right way to do this is to add a field to `IrOperationInput` (`exclude_from_lookup_keys`?), and set that at materialisation time instead of writing to a separate `parameter_metadata.yaml`. That file remains used for overrides only.

crates/coral-app/src/sources/materialization.rs
```diff
@@ -639,20 +640,9 @@ fn write_materialization(
                 continue;
             }
         };
+        apply_lookup_key_inference(&mut materialized_surface.semantic_ir);
```
simonwhitaker: I wonder if we can do this inference as we visit each operation during `materialize_surface`, rather than as a second pass across the whole IR. I don't imagine this second pass being massively slow, but for large APIs it would be interesting to see how much latency it adds and whether inlining the check in `materialize_surface` mitigates that.

simonwhitaker (approved):

PR #1595, opened 2026-07-09
feat(reef): stream OAuth source installs

lazybean:
Move source reads and mutations through React Router loaders/actions while keeping Wax card navigation route-free via the CardList as/getCardProps adapter. Add a same-origin OAuth install resource action that relays Coral server-stream events as NDJSON so interactive installs work without browser gRPC-Web.

Constraint: Browser OAuth install must avoid gRPC-Web and preserve normal server action CRUD.\nRejected: Restore browser getSourceClient/getCoralTransport OAuth path | It would reverse the route-owned data path.\nRejected: Send OAuth credential inputs as source secrets | OAuth client inputs must remain provider credential retrievals only.\nConfidence: high\nScope-risk: moderate\nDirective: Keep unary source CRUD in loaders/actions; use manual same-origin streaming only for interactive server-stream installs until the backend offers unary/polling.\nTested: npm run check --prefix apps/reef; npm run typecheck --prefix apps/reef; npm test --prefix apps/reef; npm run build --prefix apps/reef\nNot-tested: Live Sentry device-flow success; provider returned invalid_client outside Reef streaming path.


lazybean:
* **#1595** <a href="https://app.graphite.com/github/pr/withcoral/coral/1595?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1595?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1611** <a href="https://app.graphite.com/github/pr/withcoral/coral/1611?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/reef/app/lib/source-install-form.ts
lazybean, line 1: Is that really a route?. I see no roudefault export, no laoder no action.  
  
should it be in lib? wdyt?

apps/reef/app/__tests__/architecture.test.ts
```diff
@@ -345,11 +348,31 @@ describe('Architectural Tests', () => {
...
+      // Structural check: the OAuth streaming resource route stays outside the
+      // app shell, while rendered app pages are nested inside it and settings
       // keeps its desktop-only conditional spread.
       expect(routeConfig).toMatch(
-        /layout\(\s*'routes\/app-shell\.tsx',\s*\[\s*index\('routes\/index\.tsx'\),\s*route\('sources', 'routes\/sources\.tsx',\s*\[\s*route\(':sourceName', 'routes\/source-detail\.tsx'\),?\s*\]\),\s*route\('traces', 'routes\/traces\.tsx'\),\s*\.\.\.\(\s*isDesktopApp\s*\?\s*\[route\('settings', 'routes\/settings\.tsx'\)\]\s*:\s*\[\]\),?\s*\]\s*\)/,
+        /export default \[\s*(?:\/\/[^\n]*\n\s*)*route\('sources\/:sourceName\/oauth-install', 'routes\/source-oauth-install\.ts'\),\s*layout\(\s*'routes\/app-shell\.tsx',\s*\[\s*index\('routes\/index\.tsx'\),\s*route\('sources', 'routes\/sources\.tsx',\s*\[\s*route\(':sourceName', 'routes\/source-detail\.tsx'\),?\s*\]\),\s*route\('traces', 'routes\/traces\.tsx'\),\s*\.\.\.\(\s*isDesktopApp\s*\?\s*\[route\('settings', 'routes\/settings\.tsx'\)\]\s*:\s*\[\]\),?\s*\]\s*\)/,
+      )
+    })
+
+    it('OAuth source install uses an action-only resource route, not browser gRPC', () => {
```
AlbertQM: Interesting, what is this test trying to prevent? Skipping over React Router patterns?

AlbertQM:
This is going to be so cool!

Could we add some stories please? I was testing this with the actual flow, and noticed a few broken things, e.g. missing gap between text and banner, misaligned icon, etc.
image

We also have stacked dialogs, they might work nicely here 🤔 -- opening a new dialog just with OAuth data rather than adding a banner at the bottom. WDYT?

AlbertQM (approved):
On a second thought.. Feel free to land as is. The SourceCard needs _a lot_ of love anyway. We can do a UI pass after the functionality has been added.

apps/reef/app/__tests__/architecture.test.ts
lazybean: I dont think we should have such test in here actually.  
  
It is trying to avoid using web gRPC (and skipping over React Router pattern yes).  
One reason an agent could be tempted is that the oauth flow is a streaming process, and you cannot mimic that nicely with action. We do it with a resource 'loader' and handling ndjson in some custom client code.

apps/reef/app/views/sources/source-install.tsx
lazybean, line 1: @AlbertQM is right, we need at least some stories (if only to make it easier to improve the UI later).  
  
We should split the components out in components/sources/install/xxx.tsx and add stories for them. At least OAuthProgress and OAuthFields (and what is needed by them)

lazybean:
I agree.  
I'll at least split some components out and storify them.  
Rest will wait, as this is mostly changing glue.

PR #1601, opened 2026-07-09
feat(reef): extract reusable source catalog components

AlbertQM:
## Summary

Moving some of the sources view components into `/components/`, as I plan on reusing the for the onboarding; `/components/` is our shared folder with UI built using `wax`.

No functionality changes in here, just extracting bits of UI.

## Test plan

UI still renders just fine:

image

PR #1602, opened 2026-07-09
feat(reef): add sources onboarding story

AlbertQM:
## Summary

Adding the first screen of the onboarding.

Connecting sources is the most important thing, so that's what we'll show people first.
Only when at least a source is connected you'll able to go to the second screen (coming into another PR).

None of this has been wired, just finalising the screens first, and adding the `/onboarding/` router later.

## Test plan

image

### DIFF
--- a/apps/reef/app/components/onboarding/onboarding-sources-page.tsx
+++ b/apps/reef/app/components/onboarding/onboarding-sources-page.tsx
@@ -1,3 +1,5 @@
+import { Link } from 'react-router'
+
 import { Button, Typography } from '@/wax/components'
 import { Pill } from '@/wax/components/pill'
 
@@ -12,7 +14,6 @@ export interface OnboardingSourcesPageProps {
   entries: SourceCatalogEntry[]
   errorMessage?: string | null
   loadState?: SourceCatalogLoadState
-  onContinue?: () => void
   onRetry?: () => void
   onSearchChange: (search: string) => void
   onSourceSelect: (entry: SourceCatalogEntry) => void
@@ -26,7 +27,6 @@ export function OnboardingSourcesPage({
   entries,
   errorMessage = null,
   loadState = 'idle',
-  onContinue,
   onRetry,
   onSearchChange,
   onSourceSelect,
@@ -89,7 +89,13 @@ export function OnboardingSourcesPage({
               </Typography.BodyLarge>
             </div>
 
-            <Button.Container disabled={!canContinue} onClick={onContinue} variant="secondary">
+            <Button.Container
+              as={Link}
+              disabled={!canContinue}
+              onClick={canContinue ? undefined : (event) => event.preventDefault()}
+              to="?step=query"
+              variant="secondary"
+            >
               <Button.Text>{continueLabel}</Button.Text>
             </Button.Container>
           </div>
--- a/apps/reef/app/components/onboarding/onboarding-sources-page.tsx
+++ b/apps/reef/app/components/onboarding/onboarding-sources-page.tsx
@@ -1,0 +1,114 @@
+import { Button, Typography } from '@/wax/components'
+import { Pill } from '@/wax/components/pill'
+
+import { SourceCatalogSurface } from '@/components/sources'
+import type { SourceCatalogEntry, SourceCatalogLoadState } from '@/components/sources'
+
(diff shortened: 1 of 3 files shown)


apps/reef/app/components/onboarding/onboarding-sources-page.tsx
```diff
@@ -0,0 +1,114 @@
...
+                  rel="noopener noreferrer"
+                  target="_blank"
+                >
+                  write your own spec
+                </a>
+                .
+              </Typography.BodyLarge>
+            </div>
+
+            <Button.Container disabled={!canContinue} onClick={onContinue} variant="secondary">
```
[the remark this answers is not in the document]
AlbertQM: Hmm good point. On one hand, I wouldn't encourage people to onboard on mobile. In a previous version of this, I had the action button on the top right of the header, which could work on mobile. I'll think about this; but wouldn't block on this, and leave as-is for now.

apps/reef/app/components/onboarding/onboarding-sources-page.tsx
```diff
@@ -0,0 +1,114 @@
+import { Button, Typography } from '@/wax/components'
+import { Pill } from '@/wax/components/pill'
+
+import { SourceCatalogSurface } from '@/components/sources'
```
lazybean, line 6: :question:I need to see this.

AlbertQM: It's part of the previous PR on which this is stacked #1601! 😄

apps/reef/app/components/onboarding/onboarding-sources-page.tsx
```diff
@@ -0,0 +1,114 @@
...
+              loadState={loadState}
+              onPick={onSourceSelect}
+              onRetry={onRetry}
+              onSearchChange={onSearchChange}
+              search={search}
+              showTitle={false}
+              variant="compact"
+            />
+          </div>
+        </div>
```
lazybean, line 116: I wonder if this should be an \<Outlet/>  
  
And the actual container (onboarding-sources-page) shoud be 'onboarding-page' a Layout route .  
Isnt the wrapper the same for all step? Just the text/labels changing?

lazybean: Maybe no, it actually is overengineered.  
  
But I still think the wrapper could be reused.

apps/reef/app/components/onboarding/onboarding-sources-page.tsx
```diff
@@ -0,0 +1,114 @@
...
+              <Typography.BodyLarge>
+                Coral ships with{' '}
+                <a
+                  className={styles.inlineLink}
+                  href="https://withcoral.com/docs/reference/bundled-sources"
+                  rel="noopener noreferrer"
+                  target="_blank"
+                >
+                  core sources
+                </a>
```
lazybean, line 69: :question:What is the behaviour in the desktop app?

apps/reef/app/components/onboarding/onboarding-sources-page.tsx
lazybean: That should be rendered as a \<Link/> no? Because we will navigate to next step!

lazybean (approved):
Looks very nice.  
  
A few comments and concern; but nothing blocking yet.  
  
I think the header and explainer could be reused in the other steps, no? Worth extracting.  
  
Still wondering how everything will be wired.

apps/reef/app/components/onboarding/onboarding-sources-page.tsx
AlbertQM: It opens a new tab in a browser!

AlbertQM:
> I think the header and explainer could be reused in the other steps, no? Worth extracting.

Yes! A PR upstream does that (once I add a second page, they'll be extracted) 

> Still wondering how everything will be wired.

Tried a simple version last night #1747

apps/reef/app/components/onboarding/onboarding-sources-page.tsx
AlbertQM: Yes, will update. Thanks!

PR #1606, opened 2026-07-09
feat(xtask/openapi): add standalone openapi tool

### LINKED ISSUE #1497: Support remote refs in OpenAPI specs
When importing OpenAPI specs like Sentry (https://raw.githubusercontent.com/getsentry/sentry/master/api-docs/openapi.json), follow refs to external documents.

### LINKED ISSUE #1575: DigitalOcean OpenAPI v3 import produces no runtime components

simonwhitaker:
## OpenAPI descriptor hydration

OpenAPI documents are often split across path, parameter, response, and schema files. This change adds `xtask openapi-hydrate`, which produces a self-contained, dereferenced JSON document from a local descriptor, `file:` URI, or HTTPS URL.

## Use it

Hydrate a local descriptor:

```sh
cargo run --locked -p xtask -- openapi-hydrate path/to/openapi.yaml > openapi.json
```

HTTPS roots work the same way:

```sh
cargo run --locked -p xtask -- openapi-hydrate https://example.com/openapi.yaml > openapi.json
```

The command writes pretty JSON to standard output. Errors go to standard error and exit non-zero, so it is suitable for scripts.

## Implementation

- `xtask/src/openapi/mod.rs` owns the `openapi-hydrate` arguments and `hydrate` command handler.
- `xtask/src/openapi/lib.rs` contains the resolver implementation.
- `xtask/src/openapi/tests.rs` covers resolver behavior, and `xtask/tests/openapi_hydrate.rs` covers the command surface.

The resolver accepts YAML or JSON, follows only reachable nested external `$ref` values, supports percent-encoded JSON Pointer fragments, and preserves arrays and nested objects while dereferencing. YAML integers beyond serde_json's native integer range remain JSON numbers.

## Resolution boundaries

Only local files and HTTPS are accepted. HTTPS descriptors cannot load `file:` references; HTTPS redirects to another scheme are rejected. Local references must remain under the root descriptor directory, and symlinked descriptors or references are rejected.

Root and external documents are limited to 16 MiB. HTTPS fetches time out after 30 seconds; external documents are fetched in bounded concurrent batches and cached by resolved URI.

## Verification

- `cargo test --locked -p xtask`
- `cargo clippy --locked -p xtask --all-targets --all-features -- -D warnings`
- `make rust-checks`


antonmry (commented):
Does it need to be a separate tool? Could be `xtask` or being a subcommand of coral a valid option? I assume this tool is for authors/users, a separate binary is hard to discover and also to install. If we think coral engineers might be the only initial users, `xtask` could be a good enough option. We can always move it to its own tool in the future if we feel the pain, and we avoid the risk of having a new pattern `crates/tools/`.

antonmry (approved):

PR #1615, opened 2026-07-10
fix(reef): centralize reduced-motion animations

AlbertQM:
## Summary

Says on the tin. We've been hardcoding the same animation in a bunch of places, let's centralise it in WAX and respect users wth `prefers-reduced-motion: true`


martinrajdl (approved):
💅 beauty

PR #1618, opened 2026-07-10
feat(reef): add second onboarding step

AlbertQM:
## Summary

Adding a second step to the onboarding. I wanted to show Coral in action, by running some SQL based on which sources users have connected.
I haven't yet found a neat way of constructing a "ah-ha" query yet, so for now the query will just be against `coral` schema. We can improve this in the future.

## Test plan

image

- [feat(reef): add second onboarding step](https://github.com/withcoral/coral/pull/1618) <-
- [feat(reef): add onboarding flow](https://github.com/withcoral/coral/pull/1747)
- [feat(reef): add agent setup onboarding step](https://github.com/withcoral/coral/pull/1764)

### DIFF
--- a/apps/reef/app/components/onboarding/onboarding-sample-query-page.tsx
+++ b/apps/reef/app/components/onboarding/onboarding-sample-query-page.tsx
@@ -13,6 +13,7 @@ import { Icon } from '@/wax/components/icon'
 
 import { OnboardingPage } from './onboarding-page'
 import * as styles from './onboarding-sample-query-page.css'
+import { getOnboardingStepState } from './onboarding-steps'
 import { pluralise } from '~/utils/pluralise'
 
 export type SampleQueryLoadState = 'error' | 'idle' | 'loading' | 'success'
@@ -36,7 +37,6 @@ export interface OnboardingSampleQueryPageProps {
   onContinue?: () => void
   onRunSampleQuery?: (sql: string) => void
   rows?: OnboardingSampleQueryRow[]
-  stepLabel?: string
 }
 
 export function OnboardingSampleQueryPage({
@@ -48,8 +48,8 @@ export function OnboardingSampleQueryPage({
   onContinue,
   onRunSampleQuery,
   rows = [],
-  stepLabel = 'Step 2/2',
 }: OnboardingSampleQueryPageProps) {
+  const step = getOnboardingStepState('query')
   const sourceCount = connectedSources.length
   const sourceSetKey = connectedSources
     .map((source) => source.name)
@@ -80,7 +80,7 @@ export function OnboardingSampleQueryPage({
         onClick: onContinue,
       }}
       ariaLabel="Confirm connected sources"
-      stepLabel={stepLabel}
+      step={step}
       sideContent={
         <>
           <Typography.BodyLarge>
--- a/apps/reef/app/components/onboarding/onboarding-sample-query-page.tsx
+++ b/apps/reef/app/components/onboarding/onboarding-sample-query-page.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef, type ReactNode } from 'react'
+import type { ReactNode } from 'react'
 
 import { ErrorBanner } from '@/components/error-banner'
 import {
@@ -35,7 +35,7 @@ export interface OnboardingSampleQueryPageProps {
   errorMessage?: string | null
   loadState?: SampleQueryLoadState
   onContinue?: () => void
-  onRunSampleQuery?: (sql: string) => void
+  onRetry?: () => void
   rows?: OnboardingSampleQueryRow[]
 }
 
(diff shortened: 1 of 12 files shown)


apps/reef/app/components/onboarding/onboarding-sample-query-page.tsx
```diff
@@ -0,0 +1,267 @@
...
+          {connectedSourceCount === 1 ? 'source' : 'sources'}.
+        </Typography.Body>
+      </div>
+    ),
+    statusLabel: 'ready',
+    statusTone: 'running',
+  }
+}
+
+function MissingSourceFallback() {
```
martinrajdl, line 205: Should we have many components in one file?

AlbertQM: Do you think we will reuse any of these components individually?
Is there another benefit in splitting these out if the answer to the previous question is no?

apps/reef/app/components/onboarding/onboarding-sources-page.tsx
```diff
@@ -31,90 +28,54 @@ export function OnboardingSourcesPage({
   onSearchChange,
   onSourceSelect,
   search,
-  stepLabel = 'Step 1/3',
+  stepLabel = 'Step 1/2',
```
lazybean: :pick:I wonder how easy it would be to be clever and figure out total step and current step?  
There must be somewhere the list and order of step? Maybe just not in a easy proccessable way.

apps/reef/app/components/onboarding/onboarding-sources-page.tsx
```diff
@@ -31,90 +28,54 @@ export function OnboardingSourcesPage({
...
-            />
-          </div>
-        </div>
-      </div>
-    </section>
+    <OnboardingPage
+      action={{
+        disabled: !canContinue,
+        label: continueLabel,
+        to: '?step=query',
```
lazybean: :pick:Same nitty comment as above.

apps/reef/app/components/onboarding/onboarding-sources-page.tsx
```diff
@@ -31,90 +28,54 @@ export function OnboardingSourcesPage({
...
+      }}
+      stepLabel={stepLabel}
+      sideContent={
+        <>
+          <Typography.BodyLarge>
+            Sources turn APIs, services, and local datasets into tables that Coral can query.
+          </Typography.BodyLarge>
+          <Typography.BodyLarge>
+            Coral ships with{' '}
+            <OnboardingLink href="https://withcoral.com/docs/reference/bundled-sources">
```
lazybean, line 55: :question:I know I already asked. But want to be sure. How does this behave in the desktop/electron app? Open in default browser new tab? Is that the best?

apps/reef/app/components/onboarding/onboarding-sample-query-page.tsx
```diff
@@ -0,0 +1,267 @@
...
+      !sourceSetKey ||
+      loadState !== 'idle' ||
+      lastAutoRunSourceSet.current === sourceSetKey
+    ) {
+      return
+    }
+
+    lastAutoRunSourceSet.current = sourceSetKey
+    onRunSampleQuery(ONBOARDING_SAMPLE_QUERY)
+  }, [loadState, onRunSampleQuery, sourceSetKey])
```
lazybean: :wrench: This could and I think should be run in the page loader.  
Maybe you can defer (ie return the promise + Suspense + loading/progrees fallback+ use) if you want to show it is running.  
  
We have only one route for the whole onboarding right? (step is a search pararm)

lazybean: aka : No data fetching directly in the component!  
  
aka: I dislike `useEffect`

lazybean (commented):
Thank you!  
I got more comments!

apps/reef/app/components/onboarding/onboarding-sources-page.tsx
AlbertQM: > Open in default browser new tab? 

Yup!

> Is that the best?

What else did you have in mind? 👀 That's what other tools do, all redirecting to the good ol' browser

apps/reef/app/components/onboarding/onboarding-sources-page.tsx
AlbertQM: Sure thing, updated both this and the `to`! 😄

apps/reef/app/components/onboarding/onboarding-sample-query-page.tsx
AlbertQM: Updated! The other part of this lives in the next stacked PR. The query now runs in the loader, BUT since we might need to access the keychain (which takes longer than the SSR timeout), the promise dance is handled on the client.

apps/reef/app/components/onboarding/onboarding-sources-page.tsx
AlbertQM: There's some extra cleanup in the PR stacked on top of this too 😄

apps/reef/app/components/onboarding/onboarding-sample-query-page.tsx
lazybean: That keychain issue should only happen in dev no? Not sure it worth complicated the client for that.

AlbertQM: Fair point, will update 👀  if it somehow happens, we still fail gracefully now and show a retry button
(it's in #1747)

lazybean (approved):
Thank you

PR #1620, opened 2026-07-10
feat(app): register gRPC health service

sauldhernandez:
## Intent

Expose the standard grpc.health.v1 endpoint as a process-liveness signal for remote Coral deployments.

## What it enables

Load balancers and operators can probe the aggregate service without depending on workspace, catalog, or credential readiness; named services remain unregistered and streaming watch is explicitly unsupported.

## Usage examples

Call grpc.health.v1.Health/Check with an empty service name and expect SERVING while the Coral process is accepting gRPC requests.

## Validation

cargo test --locked -p coral-app --test grpc grpc_health_is_process_liveness_only; cargo check -p coral-app --all-features; make rust-checks; make perf-check; git diff --check

PR #1623, opened 2026-07-10
feat(app): add server bind configuration

sauldhernandez:
## Intent

Establish app-owned `[server]` configuration and distinguish short-lived embedded gRPC lifecycles from a long-running standalone server without changing Coral's unauthenticated defaults.

## What it enables

- `EphemeralGrpc` binds to `127.0.0.1:0` for short-lived in-process servers.
- `StandaloneGrpc` uses `[server].bind_addr`, including loopback or non-loopback addresses, without requiring server authentication.
- Embedded UI remains loopback-only.

## Usage examples

Configure the standalone bind address in `config.toml`:

```toml
[server]
bind_addr = "127.0.0.1:14555"
```

The descendant CLI layer starts this configured runtime with `coral server`.

### DIFF
--- a/crates/coral-app/src/bootstrap/mod.rs
+++ b/crates/coral-app/src/bootstrap/mod.rs
@@ -7,6 +7,7 @@ mod env;
 mod error;
 mod health;
 mod server;
+mod server_config;
 
 use crate::state::AppStateLayout;
 use crate::telemetry::{
--- a/crates/coral-app/src/bootstrap/server.rs
+++ b/crates/coral-app/src/bootstrap/server.rs
@@ -114,7 +114,7 @@ impl ServerConfig {
     pub(crate) fn new() -> Self {
         Self {
             config_dir: None,
-            mode: ServerMode::NativeGrpc,
+            mode: ServerMode::EphemeralGrpc,
             engine_extensions_providers: Vec::new(),
             user_principal_provider: Arc::new(SingleUserPrincipalProvider),
             feedback_publisher: Arc::new(HostedFeedbackPublisher::new()),
@@ -160,11 +160,11 @@ impl ServerConfig {
 /// transport or asset-serving knob.
 #[derive(Clone)]
 pub enum ServerMode {
-    /// Native gRPC for CLI, MCP, and local client callers.
-    NativeGrpc,
-    /// Native gRPC bound to an explicit address for the long-running server.
-    RemoteGrpc {
-        /// Address to bind. Non-loopback addresses require authentication.
+    /// Ephemeral native gRPC for CLI, MCP, and local client callers.
+    EphemeralGrpc,
+    /// Native gRPC bound to an explicit address for a standalone server.
+    StandaloneGrpc {
+        /// Address to bind.
         bind: SocketAddr,
     },
     /// Loopback gRPC-Web server that also serves embedded UI assets.
@@ -179,21 +179,10 @@ pub enum ServerMode {
 impl ServerMode {
     fn bind_addr(&self) -> SocketAddr {
         match self {
-            Self::NativeGrpc => SocketAddr::from((Ipv4Addr::LOCALHOST, 0)),
-            Self::RemoteGrpc { bind } => *bind,
+            Self::EphemeralGrpc => SocketAddr::from((Ipv4Addr::LOCALHOST, 0)),
+            Self::StandaloneGrpc { bind } => *bind,
             Self::EmbeddedUi { port, .. } => SocketAddr::from((Ipv4Addr::LOCALHOST, *port)),
         }
-    }
-
-    fn validate_bind(&self) -> Result<(), AppError> {
-        if let Self::RemoteGrpc { bind } = self
-            && !bind.ip().is_loopback()
-        {
-            return Err(AppError::FailedPrecondition(
-                "non-loopback gRPC bind requires configured server authentication".to_string(),
-            ));
-        }
-        Ok(())
     }
 }
 
@@ -205,7 +194,7 @@ pub struct ServerBuilder {
 
 impl ServerBuilder {
     #[must_use]
-    /// Creates a builder for the default native gRPC local server.
+    /// Creates a builder for the default ephemeral native gRPC local server.
     pub fn new() -> Self {
         Self {
             config: ServerConfig::new(),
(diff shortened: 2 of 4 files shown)


sauldhernandez:
* **#1628** <a href="https://app.graphite.com/github/pr/withcoral/coral/1628?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1624](https://github.com/withcoral/coral/pull/1624) <a href="https://app.graphite.com/github/pr/withcoral/coral/1624?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1631](https://github.com/withcoral/coral/pull/1631) <a href="https://app.graphite.com/github/pr/withcoral/coral/1631?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1623** <a href="https://app.graphite.com/github/pr/withcoral/coral/1623?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1623?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

simonwhitaker (changes requested):
## Review — feat(app): add server bind configuration

**Stacked PR** (base: `sauldhernandez/oss-serve-grpc-health`). Adds three things: a `[server]` config section loader (`ServerSettings` in new `bootstrap/server_config.rs`), a rename of `ServerMode::NativeGrpc` → `EphemeralGrpc`, and a new `ServerMode::StandaloneGrpc { bind }` + `ServerBuilder::standalone_grpc(addr)` seam for a future `coral serve` command. Nothing outside tests consumes the new pieces yet (explicitly acknowledged via the `expect(dead_code)` attribute).

### 🔴 Main concern: the PR body's security claim doesn't match the code

The body says:

> this layer **fails closed before state initialization for every non-loopback bind** until server authentication exists

There is no such check anywhere in this diff. In fact the opposite is tested and codified: `standalone_grpc_binds_the_requested_non_loopback_address` asserts that an **unauthenticated** server successfully binds `0.0.0.0` and comes up. As merged, `ServerBuilder::standalone_grpc` is a public API that will happily expose the full gRPC surface on all interfaces with no auth, and `ServerSettings::load` accepts any `bind_addr` without validation.

Either the guard was intended in this layer and is missing, or it's deferred to the descendant `coral serve` PR and the body is stale. If it's deferred, I'd still argue the check belongs *here*: the seam itself should reject non-loopback addresses (e.g. a `bind.ip().is_loopback()` check in `standalone_grpc` or at `start_server`), so no future caller can accidentally bypass it. Relying on every descendant caller to remember the check is exactly the fail-open pattern the body says it avoids. At minimum, the non-loopback test should be flipped to assert rejection, and the body corrected.

### ✅ What's good

- **`ServerSettings::load` is nicely scoped** — parses only `[server]`, tolerates unknown sections, defaults to `127.0.0.1:0` when the file or section is absent. The "loads without claiming other sections" test is a good guard for the config-ownership model.
- **Rename `NativeGrpc` → `EphemeralGrpc`** is accurate (it binds port 0) and safe: I found no callers of `native_grpc()`/`NativeGrpc` outside `bootstrap/server.rs`, and external consumers go through `ServerBuilder::new()`.
- **Test coverage of the config layer is solid**: default, explicit value, and the `localhost:14555` rejection case (documents that `SocketAddr` requires an IP literal — a useful behavioral pin, since users will naturally try hostnames).
- The `preserves_unrelated_sections_when_rendering_with_existing_config` extension correctly verifies `[server]` survives config rewrites by the source catalog.

### 🟡 Minor points

- **Hostname rejection UX**: `bind_addr = 'localhost:14555'` failing is asserted but the error the user sees is a raw serde/TOML parse error. When `coral serve` lands, consider mapping this to a hint ("use an IP address, e.g. 127.0.0.1:14555"). Fine to defer.
- **Incidental fixture reformatting** in `state/config.rs`: the `[features]` block was de-indented (tabs removed) while inserting `[server]`. Harmless — other sections keep the odd tab indentation so formatting-preservation coverage remains — but it's an unrelated hunk worth a commit-message mention or reverting.
- **`try_exists()` + read** has a small TOCTOU window; matching `ErrorKind::NotFound` on the read would be marginally tighter. Cosmetic.
- The `0.0.0.0` bind test can trigger macOS firewall prompts on dev machines when running the suite locally — another small reason to make that case a rejection test instead.
- **Docs**: if `[server]` is user-facing config, there's no documentation update in this PR; presumably deferred to the `coral serve` PR — worth confirming.

### Verdict

The mechanics are clean and well-tested, but I'd hold approval until the fail-closed discrepancy is resolved — either add the loopback guard in this layer (my preference, with the non-loopback test inverted) or explicitly restate in the PR body that non-loopback binds are allowed here and gated in a descendant, with a pointer to where that gate lands.

---
*Review by Claude Code on behalf of @simonwhitaker*

simonwhitaker (changes requested):
## Re-review — follow-up on the fail-closed question

Thanks for the update. The body/code inconsistency I flagged is resolved: the PR body no longer claims non-loopback binds fail closed, and the code, tests, and body now agree. Appreciate that.

But the *direction* of the resolution raises a bigger question than the one it answers, so I'm keeping changes requested until we've talked it through.

### What I'm seeing

The first commit in this branch (`e4d6cfd22`) contains exactly the guard my review asked for: a `validate_bind()` check rejecting non-loopback binds before state initialization, with a fail-closed test. The new second commit (`d3b665e0f`, "allow unauthenticated standalone gRPC binds") deliberately removes it and inverts the test to assert an unauthenticated `0.0.0.0` bind succeeds.

Meanwhile #1631 (`feat(app): enforce static bearer authentication`) — the descendant where the auth gate would have landed — was closed a few hours after my review, with no explanation. And #1628's body now states "Coral does not gate non-loopback binds on server authentication" and documents `bind_addr = "0.0.0.0:14555"` as a supported configuration.

So as the stack now stands, `coral server` will expose the full gRPC surface on all interfaces with no authentication, and the safeguard that briefly existed was removed rather than kept.

### Questions before I can approve

1. **Why was #1631 closed?** If static bearer auth is deferred rather than abandoned, is there an issue or PR tracking where it lands? My original review was fine with the gate living in a descendant — but there needs to *be* a descendant.
2. **Is unauthenticated non-loopback serving an intentional product decision?** If so, that deserves an explicit call-out (in the PR body and ideally the `coral server` docs/output — e.g. a warning when binding non-loopback), not just the removal of the guard. Operators typing `0.0.0.0` into `bind_addr` should understand they're exposing an unauthenticated API to their network.
3. If it's *not* a settled decision, I'd still advocate restoring `validate_bind()` from `e4d6cfd22` — the seam rejecting non-loopback binds until auth exists is cheap, and it's much easier to relax a guard later than to retrofit one after callers depend on the open behavior.

The minor points from my previous review (fixture reformatting in `state/config.rs`, `try_exists` TOCTOU, hostname-rejection UX, docs) all still stand but none are blocking.

---
*Review by Claude Code on behalf of @simonwhitaker*

simonwhitaker (approved):

sauldhernandez:
### Merge activity

* **Jul 21, 1:46 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1623).
* **Jul 21, 1:48 PM UTC**: [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1623) rebased this pull request as part of a merge.
* **Jul 21, 2:09 PM UTC**: @sauldhernandez merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1623).

PR #1634, opened 2026-07-10
feat(app): scaffold OAuth authorization server

sauldhernandez:
## Stack overview

This stack lets a long-running Coral server authenticate users through an external OpenID Connect provider and protect its gRPC and MCP HTTP APIs.

## Where this PR fits

The previous PR adds the temporary state required by an OAuth authorization flow. This PR assembles the configuration, token issuer, state store, HTTP listener, and server metadata into one server shell. The next PR adds safe handling for the external URLs that later login steps will need to fetch.

## Intent

Create the fail-closed configuration and HTTP lifecycle for Coral's OAuth authorization server before adding the actual login routes.

## What it enables

This PR introduces `AuthSettings`, which parses the `[auth]` section from a config snapshot supplied by the caller. The OAuth server receives those validated settings instead of reopening `config.toml`, so the file is read only once by the composition layer.

`AuthSettings` has exactly one public constructor, `AuthSettings::from_toml`. The type that serde deserializes into is private to the parsing module, so no caller can obtain a value that has not been validated — holding one is proof that validation ran, and nothing downstream revalidates.

Validation covers:

- **`[auth.session]`** — at most one signing-key source (`signing_key_env` or `signing_key_file`; configuring neither falls back to the `CORAL_SESSION_SIGNING_KEY` environment variable), a non-empty env var name, and an access-token TTL that is greater than zero and no more than a year. The upper bound keeps session tokens from outliving any reasonable rotation window, and closes the extreme where a near-`u64::MAX` TTL parses cleanly and then overflows the expiry computation at every token issuance.
- **`[auth.authorization_server]`** — a required `issuer` that is a valid https (or loopback http) URL with no credentials, query, or fragment, mounted at the root.
- **`[auth.provider]`** — a required `issuer`, `client_id`, and `redirect_uri`; exactly one of `client_secret` or `client_secret_env`; and a `redirect_uri` whose origin matches the authorization server's issuer, since the upstream IdP sends the browser back to a callback route under that origin.
- **`[auth]`** — a bind address that must be loopback unless `allow_insecure_remote_http_bind` is set, because the endpoints are served as cleartext.

The server can then start an HTTP listener and publish `/.well-known/oauth-authorization-server`. It also owns the session-token issuer and in-memory OAuth state store that later routes will use.

Failures surface as `AuthServerError`, so a caller can tell a listener failure from an invalid configuration by matching rather than by inspecting message text.

[...]

PR #1632, opened 2026-07-10
feat(mcp-http): enforce OAuth bearer sessions

sauldhernandez:
## Stack overview

This stack lets a long-running Coral server authenticate users through an external OpenID Connect provider and protect its gRPC and MCP HTTP APIs.

## Where this PR fits

The previous PR adds a loopback-only MCP HTTP server with authentication disabled. This PR adds the authenticated transport that later serving code can use. The next PR only exposes the local mode through Coral's server configuration, so this authenticated path is a library capability at this point in the stack.

## Intent

Protect remote MCP HTTP sessions with OAuth bearer tokens without allowing an authenticated request to fall back to Coral's shared unauthenticated client.

## What it enables

The authenticated server:

- publishes OAuth protected-resource metadata;
- requires and validates a bearer token before handling MCP work;
- creates each MCP handler with an `AppClient` that forwards that bearer token to gRPC;
- requires the exact same bearer token on every later request for that MCP session;
- keeps the live handler, token fingerprint, and last-used time in one session record;
- closes a session after one hour of inactivity when it is next accessed; and
- closes both authenticated and authentication-disabled sessions through the same shutdown path.

Only a SHA-256 fingerprint of the bearer is kept with the session; the raw bearer is not stored there.

Live MCP handlers remain in the process that created them. A deployment with more than one replica must route every request for an `Mcp-Session-Id` back to the same replica. This PR does not add shared session persistence or cross-replica handler restoration.

Browser calls remain unsupported, so any request carrying an `Origin` header is rejected.

## Usage examples

```rust
use coral_mcp::{
    McpOptions,
    http::{
        AuthenticatedMcpHttpConfig, AuthenticatedMcpHttpRuntime,
        start_authenticated,
    },
};

let config = AuthenticatedMcpHttpConfig::new(
    "0.0.0.0:14556".parse()?,
    "https://coral.example.com/mcp",
    "https://login.example.com",
    "coral:mcp",
)?;

let runtime = AuthenticatedMcpHttpRuntime::new(
    validate_bearer,
    app_client_for_bearer,
    McpOptions::default(),
    check_grpc_readiness,
);

let server = start_authenticated(config, runtime).await?;
```

`app_client_for_bearer` must return an `AppClient` that forwards the supplied bearer to gRPC. If the server runs behind a load balancer, configure session stickiness using `Mcp-Session-Id`.

## Validation

- `cargo test --locked -p coral-mcp --all-features`
[...]

PR #1631, opened 2026-07-10
feat(app): enforce static bearer authentication

sauldhernandez:
## Intent
Protect long-running Coral gRPC services with an operator-configured static bearer while preserving unauthenticated local modes.

## What it enables
Allows authenticated non-loopback serving behind a TLS-terminating reverse proxy, with a reusable AuthValidator seam for later session and HTTP gates.

## Usage examples
Configure [server.auth] and provide CORAL_SERVER_AUTH_TOKEN; clients send the token with the Bearer authorization scheme.

## Validation
- make rust-checks
- make perf-check (201 ms mean; 750 ms threshold)
- make docs-check
- focused static-auth, config-secrecy, fail-before-state, and authenticated gRPC tests
- strict coral-app and coral-cli clippy

PR #1629, opened 2026-07-10
feat(app): add in-memory OAuth state store

sauldhernandez:
## Stack overview

This stack lets a long-running Coral server authenticate users through an external OpenID Connect provider and protect its gRPC and MCP HTTP APIs.

## Where this PR fits

The previous PR defines the session token Coral will issue after login. This PR adds the temporary, single-use records needed while an OAuth login moves through redirects and code exchange. The next PR puts those pieces behind an OAuth server shell.

## Intent

Track in-progress OAuth logins and short-lived authorization codes without adding a database migration or storing their lookup secrets in plain text.

## What it enables

The in-memory store keeps two kinds of record:

- an authorization session, indexed by the state sent to the external OpenID Connect provider; and
- an authorization code, returned to the client after a successful login.

Lookup secrets are stored as SHA-256 hashes. Records expire after five minutes and are consumed at most once. An authorization code is removed only when its client ID, redirect URI, and PKCE challenge all match the redemption request.

The store allows at most 4,096 combined sessions and codes. Expired entries are removed when new records are stored. Restarting Coral intentionally invalidates every login that was still in progress.

The internal `StateStore` trait separates the OAuth handlers from storage operations, but this PR includes only the process-local implementation. It does not add Redis, database persistence, or multi-replica coordination.

## Usage examples

The later OAuth handlers use the store in this order:

1. `/oauth/authorize` stores the client request together with generated OpenID Connect state.
2. The provider callback consumes that state exactly once.
3. The callback stores a short-lived authorization code for the authenticated user.
4. `/oauth/token` consumes the code only after the client ID, redirect URI, and PKCE proof match.

## Validation

- `cargo test --locked -p coral-app --all-features auth::state_store`
- `cargo clippy --locked -p coral-app --all-targets --all-features -- -D warnings`
- `make rust-checks`
- `git diff --check`

PR #1635, opened 2026-07-10
feat(app): retain validated OIDC provider configuration

sauldhernandez:
## Stack overview

This stack lets a long-running Coral server authenticate users through an external OpenID Connect provider and protect its gRPC and MCP HTTP APIs.

## Where this PR fits

The previous PR adds safe URL handling for configured endpoints and untrusted metadata. This PR applies that foundation to OpenID Connect provider configuration and keeps the validated values that later discovery, redirect, token exchange, and claim checks will need.

## Intent

Turn the `[auth.provider]` section into complete, runtime-ready OpenID Connect provider settings instead of validating the section and then discarding its useful values.

## What it enables

Coral now retains and validates:

- the provider type, issuer, client ID, and redirect URI;
- exactly one inline or environment-backed client secret;
- requested scopes, which default to `openid`, `email`, and `profile`;
- the claims used for the Coral user ID and display name;
- additional authorization parameters; and
- claims that must have specific values in the provider's ID token.

Environment-backed secrets are read once while runtime dependencies are resolved. Resolving them produces `ResolvedAuthSettings`, whose existence is proof the secret was fetched, so reading it later cannot fail. Resolved secrets are kept in zeroizing storage and replaced with `<redacted>` in debug output. Configuration errors also avoid including secret values.

Provider endpoints use the configured-endpoint URL policy from the previous PR. Reserved OAuth parameters cannot be overridden through `auth_params`, and the requested scopes must include `openid`.

The `issuer` must be written exactly as the provider publishes it, because later discovery compares it byte for byte. OpenID Connect Discovery 1.0 §3 and RFC 8414 §2 both require a published issuer to be a normalized URL, so spellings the URL parser would rewrite — a mixed-case host, an explicit `:443`, dot segments — are ones no compliant provider can publish. Accepting them would not make them work; it would only move the failure from startup to every login, so they are rejected at startup, with the canonical form named in the error. A trailing slash is kept as written, since some providers publish one and some do not.

The `redirect_uri` is retained verbatim for the same reason: OAuth 2.0 compares it against the registration by exact string match, and Coral sends the configured spelling. The URL parser silently rewrites a few spellings for HTTP and HTTPS URLs — `.../auth\oidc/callback` parses as if it were written with a slash — so the value that satisfies validation is not always the value that goes on the wire. Both endpoints therefore reject the parser's syntax violations (a backslash, a non-URL code point, a truncated percent escape) rather than passing them on to the provider.

This PR retains configuration only. It does not perform provider discovery, redirect a user, exchange a code, or validate an ID token.

## Usage examples

```toml
[auth.provider]
type = "oidc"
issuer = "https://login.example.com/tenant"
client_id = "coral"
client_secret_env = "CORAL_OIDC_CLIENT_SECRET"
redirect_uri = "https://coral.example.com/auth/oidc/callback"
scopes = ["openid", "email", "groups"]
principal_claim = "sub"
display_name_claim = "email"

[auth.provider.auth_params]
prompt = "select_account"

[auth.provider.required_claims]
tenant = "example"
```

An inline `client_secret` can be used instead of `client_secret_env`, but configuring both is rejected.

### DIFF
--- a/crates/coral-app/src/auth/authorization_server.rs
+++ b/crates/coral-app/src/auth/authorization_server.rs
@@ -92,12 +92,13 @@ impl CoralAuthorizationServer {
                 get(authorization_server_metadata),
             )
             .with_state(state);
-        let listener = TcpListener::bind(bind_addr)
-            .await
-            .map_err(|source| AuthServerError::Bind {
-                address: bind_addr,
-                source,
-            })?;
+        let listener =
+            TcpListener::bind(bind_addr)
+                .await
+                .map_err(|source| AuthServerError::Bind {
+                    address: bind_addr,
+                    source,
+                })?;
         let endpoint_uri = format!(
             "http://{}",
             listener.local_addr().map_err(AuthServerError::LocalAddr)?
@@ -275,7 +276,11 @@ mod tests {
             panic!("expected unsafe bind to be rejected");
         };
 
-        assert!(error.to_string().contains("allow_insecure_remote_http_bind"));
+        assert!(
+            error
+                .to_string()
+                .contains("allow_insecure_remote_http_bind")
+        );
     }
 
     #[test]
--- a/crates/coral-app/src/auth/authorization_server.rs
+++ b/crates/coral-app/src/auth/authorization_server.rs
@@ -13,7 +13,7 @@ use tokio::net::TcpListener;
 use tokio::sync::oneshot;
 use tokio::task::JoinHandle;
 
-use super::config::{AuthSettings, signing_key_env_error};
+use super::config::{AuthSettings, ResolvedAuthSettings, signing_key_env_error};
 use super::error::AuthServerError;
 use super::session::SessionTokenIssuer;
 use super::state_store::{InMemoryStateStore, StateStore};
@@ -22,7 +22,7 @@ const SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(5);
 
 /// Prepared Coral authorization server with validated settings and runtime dependencies.
 pub struct CoralAuthorizationServer {
-    settings: Arc<AuthSettings>,
+    settings: Arc<ResolvedAuthSettings>,
     session_tokens: SessionTokenIssuer,
     state_store: Arc<dyn StateStore>,
 }
@@ -31,8 +31,9 @@ impl CoralAuthorizationServer {
     /// Builds the Coral authorization server from validated auth settings.
     ///
     /// [`AuthSettings`] can only be produced by [`AuthSettings::from_toml`],
-    /// so the settings arrive already validated and are not rechecked here;
-    /// only runtime readiness (a resolved provider secret) is confirmed.
+    /// so the settings arrive already validated and are not rechecked here.
+    /// This resolves what parsing cannot: the provider secret and the session
+    /// signing key.
     ///
     /// `config_path` is used only to resolve a relative session signing-key
     /// path. Config parsing and the `config.toml` filesystem read remain the
@@ -55,17 +56,19 @@ impl CoralAuthorizationServer {
     }
 
     pub(crate) fn from_resolved_settings(
-        settings: AuthSettings,
+        settings: ResolvedAuthSettings,
         session_tokens: SessionTokenIssuer,
     ) -> Result<Self, AuthServerError> {
-        settings.validate_runtime_ready()?;
         if !settings.matches_session_token_issuer(&session_tokens) {
             return Err(AuthServerError::SessionIssuerMismatch);
         }
         Ok(Self::from_validated_parts(settings, session_tokens))
     }
 
-    fn from_validated_parts(settings: AuthSettings, session_tokens: SessionTokenIssuer) -> Self {
+    fn from_validated_parts(
+        settings: ResolvedAuthSettings,
+        session_tokens: SessionTokenIssuer,
+    ) -> Self {
         Self {
             settings: Arc::new(settings),
             session_tokens,
@@ -173,7 +176,7 @@ impl Drop for RunningCoralAuthorizationServer {
 
 #[derive(Clone)]
 struct AuthorizationServerHttpState {
-    settings: Arc<AuthSettings>,
+    settings: Arc<ResolvedAuthSettings>,
     #[expect(
         dead_code,
         reason = "used by the OAuth token endpoint in a descendant PR"
@@ -288,9 +291,11 @@ mod tests {
         let dir = authorization_server("");
         let config_path = dir.path().join("config.toml");
         let raw = fs::read_to_string(&config_path).expect("config snapshot");
-        let settings = AuthSettings::from_toml(&raw)
+        let (settings, _issuer) = AuthSettings::from_toml(&raw)
             .expect("valid config")
-            .expect("auth settings");
+            .expect("auth settings")
+            .resolve_runtime_dependencies(&config_path, &|_| Ok(None))
+            .expect("resolved runtime dependencies");
         let signing_key = fs::read(dir.path().join("session.key")).expect("session key");
         let session_tokens = super::SessionTokenIssuer::new(
             Some("http://localhost:9999"),
--- a/crates/coral-app/src/auth/authorization_server.rs
+++ b/crates/coral-app/src/auth/authorization_server.rs
@@ -270,6 +270,9 @@ mod tests {
             .expect("prepared from snapshot");
     }
 
+    /// `AuthSettings` has no public constructor other than `from_toml`, so the
+    /// unsafe-bind guard cannot be bypassed by building a value directly; the
+    /// parsing boundary is the only place it needs to hold.
     #[test]
     fn rejects_unsafe_bind_at_the_only_construction_site() {
         let raw = format!(
--- a/crates/coral-app/src/auth/authorization_server.rs
+++ b/crates/coral-app/src/auth/authorization_server.rs
@@ -31,7 +31,8 @@ impl CoralAuthorizationServer {
     /// Builds the Coral authorization server from validated auth settings.
     ///
     /// [`AuthSettings`] can only be produced by [`AuthSettings::from_toml`],
-    /// so the settings arrive already validated and are not rechecked here.
+    /// so the settings arrive already validated and are not rechecked here;
+    /// only runtime readiness (a resolved provider secret) is confirmed.
     ///
     /// `config_path` is used only to resolve a relative session signing-key
     /// path. Config parsing and the `config.toml` filesystem read remain the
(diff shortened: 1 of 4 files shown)


sauldhernandez:
* **#1635** <a href="https://app.graphite.com/github/pr/withcoral/coral/1635?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1635?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-app/src/auth/config.rs
```diff
@@ -419,6 +735,203 @@ mod tests {
...
+    #[test]
+    fn retains_every_issuer_spelling_a_provider_can_publish() {
+        for issuer in [
+            "https://accounts.example.test",
+            "https://accounts.example.test/",
+            "https://accounts.example.test/tenant",
+            "https://accounts.example.test/tenant/",
+            "https://ACCOUNTS.example.test/tenant",
+            "https://accounts.example.test:443/tenant",
+            "https://accounts.example.test/tenant/../other",
```
simonwhitaker: These three assertions contradict the PR description, which says:

> Spellings no provider could publish — a mixed-case host, an explicit `:443`, or dot segments — are rejected at startup rather than failing every login.

The test says the opposite: all three are accepted and retained byte-for-byte. The code agrees with the test — `provider_issuer_endpoint` only checks `url::SyntaxViolation`s, `ConfiguredEndpointUrl::parse` checks scheme/host/credentials/fragment only (dot-segment rejection lives exclusively on `PublicMetadataUrl`), and `self.issuer` is overwritten with the raw trimmed string, discarding the parsed URL.

Worth deciding which one is right, because these spellings produce exactly the failure mode the PR says it prevents. Take `https://a.test/tenant/../other`: the discovery descendant fetches `{issuer}/.well-known/openid-configuration`, and any URL-resolving join collapses `..` to `/other/...`. The document there publishes `issuer: "https://a.test/other"`, which can never byte-match the configured string — so *every* login fails at runtime, in a later PR. Same argument for the mixed-case host and `:443`: RFC 8414 / OIDC Discovery issuers are normalized, so those spellings can only ever mismatch.

The doc comment above ("every valid spelling a provider publishes has to survive validation unchanged") is actually the argument for rejecting these three, since no provider publishes them. `has_dot_path_segment` is already available next door in `outbound_url_policy`, and the host-case / default-port checks are a couple of lines against the `ConfiguredEndpointUrl` you already have in hand at the call site.

If the intent really is to accept them, the description and this doc comment both need updating — but I'd lean toward implementing the rejection.

crates/coral-app/src/auth/config.rs
```diff
@@ -253,41 +333,183 @@ impl OidcProviderSettings {
...
+        }
+
+        self.issuer = provider_required("issuer", &self.issuer)?;
+        let issuer = provider_issuer_endpoint(&self.issuer)?;
+        if issuer.as_url().query().is_some() {
+            return Err(invalid_provider("issuer must not include a query"));
+        }
+        self.client_id = provider_required("client_id", &self.client_id)?;
+        self.redirect_uri = provider_required("redirect_uri", &self.redirect_uri)?;
+        let redirect_uri = provider_endpoint("redirect_uri", &self.redirect_uri)?;
```
simonwhitaker: `redirect_uri` is stored raw here and the parsed URL is used only for the origin check, then dropped — so like `issuer`, it's byte-for-byte significant later, both against the IdP's registration and against Coral's own callback route. But it goes through `provider_endpoint`, not `provider_issuer_endpoint`, so it gets none of the syntax-violation checking.

```toml
redirect_uri = 'http://localhost:9080/auth\oidc/callback'
```

The `url` crate normalizes `\` to `/` for special schemes, so the origin check passes, and the value is stored and later sent to the IdP with the backslash intact. Same for `%1`-style escapes and non-URL code points. That's a startup-detectable misconfiguration that instead surfaces as a failed login — the thing this PR is otherwise careful to avoid.

Routing `redirect_uri` through `provider_issuer_endpoint` (renamed to something like `provider_literal_endpoint`) would close this, and would leave `provider_endpoint` with no remaining callers to fold away.

crates/coral-app/src/auth/config.rs
```diff
@@ -317,40 +539,85 @@ fn required(label: &str, value: &str) -> Result<String, AuthServerError> {
...
-    if root_only && !matches!(url.path(), "" | "/") {
-        return Err(config_error(format!("{label} must mount at the root")));
+}
+
+fn provider_endpoint(field: &str, value: &str) -> Result<ConfiguredEndpointUrl, AuthServerError> {
+    ConfiguredEndpointUrl::parse(value)
+        .map_err(|error| invalid_provider(format!("{field} is invalid: {error}")))
+}
+
+fn provider_issuer_endpoint(value: &str) -> Result<ConfiguredEndpointUrl, AuthServerError> {
```
simonwhitaker: Minor: this parses the URL twice — once for the violation callback, once inside `ConfiguredEndpointUrl::parse`. Startup-only so the cost is irrelevant, just noting it as a simplification candidate if `ConfiguredEndpointUrl` ever grows a `from_parsed`.

crates/coral-app/src/auth/config.rs
```diff
@@ -253,41 +333,183 @@ impl OidcProviderSettings {
...
 
-struct SecretMarker;
+/// Provider settings whose client secret has been read from its source.
+///
+/// [`OidcProviderSettings::resolve_secret`] is the only way to build one, so
+/// holding one is proof the secret is present. The configured fields are
+/// reached through [`Deref`]; only the secret needs an accessor, because it is
+/// the one value the config may merely have pointed at.
+#[derive(Clone, Debug)]
+pub(super) struct ResolvedOidcProvider {
```
simonwhitaker, line 484: The secret ends up stored twice: here, and as `settings.client_secret`. Harmless in practice (same `Arc` in the inline case), but it slightly undercuts the doc comment's invariant, since `settings.client_secret` is still an `Option` reachable from inside the module. Clearing it during resolution, or having `ResolvedOidcProvider` own a settings type that doesn't carry the raw secret fields at all, would make "holding one is proof the secret is present" airtight rather than conventional.

crates/coral-app/src/auth/config.rs
```diff
@@ -253,41 +333,183 @@ impl OidcProviderSettings {
...
+    #[cfg_attr(
+        not(test),
+        expect(dead_code, reason = "used by the OIDC federation descendant")
+    )]
+    pub(super) fn client_secret(&self) -> &str {
+        self.client_secret.as_str()
+    }
+}
+
+impl Deref for ResolvedOidcProvider {
```
simonwhitaker: Deref polymorphism on a non-pointer type is the one spot where this PR departs from the module's style — `ResolvedAuthSettings` directly above uses explicit accessors (`bind_addr()`, `authorization_server()`). It does read well at the call sites, so this is a judgement call rather than a defect; mainly flagging so it's a deliberate choice and not drift.

crates/coral-app/src/auth/config.rs
```diff
@@ -236,14 +289,41 @@ impl AuthorizationServerSettings {
...
+    pub(super) issuer: String,
+    pub(super) client_id: String,
+    client_secret: Option<ProviderSecret>,
+    client_secret_env: Option<String>,
+    pub(super) redirect_uri: String,
+    pub(super) scopes: Vec<String>,
+    pub(super) principal_claim: String,
+    pub(super) display_name_claim: String,
+    pub(super) auth_params: BTreeMap<String, String>,
+    pub(super) required_claims: BTreeMap<String, Value>,
```
simonwhitaker: `serde_json::Value` as the deserialize target for TOML input has two rough edges for the descendant PR: a TOML datetime (`expires = 2030-01-01`) fails with a raw serde type error rather than a Coral-shaped one, and TOML/JSON number coercion will need thought when these get compared against ID-token claims (a TOML integer becomes a JSON integer, but the same claim may arrive as a float). Nothing to fix here — worth a note for whoever implements the comparison.

crates/coral-app/src/auth/config.rs
```diff
@@ -253,41 +333,183 @@ impl OidcProviderSettings {
...
+                return Err(invalid_provider(
+                    "client_secret_env must be a nonempty string",
+                ));
+            }
+            if env_name.bytes().any(|byte| matches!(byte, b'=' | b'\0')) {
+                return Err(invalid_provider("client_secret_env is invalid"));
+            }
+        }
+
+        if self.scopes.is_empty() {
```
simonwhitaker, line 387: An explicit `scopes = []` silently becomes `["openid", "email", "profile"]` rather than erroring. Defensible — an empty list can't satisfy the `openid` requirement below anyway — but empty-means-default and absent-means-default aren't distinguishable in the config, and unlike the other guards in this function there's no comment saying that's intentional.

crates/coral-app/src/auth/config.rs
```diff
@@ -419,6 +735,203 @@ mod tests {
...
+        }
+    }
+
+    #[test]
+    fn rejects_invalid_provider_fields() {
+        let cases = vec![
+            (
+                valid("").replace(
+                    "issuer = 'https://accounts.example.test'",
+                    "type = 'oauth'\nissuer = 'https://accounts.example.test'",
```
simonwhitaker: Only the reject path for `type` is exercised. Worth one case asserting that an explicit `type = 'oidc'` is accepted — right now a typo that made the comparison always fail would still pass this table.

crates/coral-app/src/auth/authorization_server.rs
```diff
@@ -264,9 +270,6 @@ mod tests {
             .expect("prepared from snapshot");
     }
 
-    /// `AuthSettings` has no public constructor other than `from_toml`, so the
-    /// unsafe-bind guard cannot be bypassed by building a value directly; the
-    /// parsing boundary is the only place it needs to hold.
     #[test]
     fn rejects_unsafe_bind_at_the_only_construction_site() {
```
simonwhitaker, line 277: The doc comment explaining *why* this single test is sufficient was deleted here, but the test itself is unchanged:

> `AuthSettings` has no public constructor other than `from_toml`, so the unsafe-bind guard cannot be bypassed by building a value directly; the parsing boundary is the only place it needs to hold.

That reasoning still holds after this PR — `AuthSettings::from_toml` is still the only constructor. Was the removal intentional? Otherwise it's worth restoring, since without it the test name is the only hint that one construction site is the whole story.

simonwhitaker (approved):

sauldhernandez:
### Merge activity

* **Jul 29, 12:35 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1635).
* **Jul 29, 12:46 PM UTC**: [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1635) rebased this pull request as part of a merge.
* **Jul 29, 12:53 PM UTC**: @sauldhernandez merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1635).

PR #1636, opened 2026-07-10
feat(app): resolve MCP HTTP serve configuration

sauldhernandez:
## Stack overview

This stack lets a long-running Coral server authenticate users through an external OpenID Connect provider and protect its gRPC and MCP HTTP APIs.

## Where this PR fits

The previous PR provides the MCP HTTP transport APIs. This PR lets `coral-app` resolve the supported local listener settings without starting that listener or coupling MCP HTTP startup to gRPC startup. The next PR makes the CLI use the resolved settings.

## Intent

Give the process that owns server startup a small, validated configuration object for the optional local MCP HTTP listener.

## What it enables

`ServerBuilder` can now read `[server.mcp_http]` and return an `McpHttpServeConfig` before gRPC starts. MCP HTTP remains disabled by default.

At this point in the stack, configuration supports only the authentication-disabled local mode:

- the MCP bind address must be loopback;
- OAuth metadata fields under `[server.mcp_http]` are rejected rather than ignored;
- the removed `[server.auth]` configuration is rejected;
- the accompanying gRPC server must listen on a loopback or wildcard address so the MCP listener can reach it locally; and
- embedded gRPC-Web mode cannot be paired with this listener.

The running gRPC server also exposes its bound socket address so the CLI can connect the MCP listener to the actual port selected by the operating system.

This PR resolves settings only. It does not start MCP HTTP and does not expose the authenticated transport through configuration.

## Usage examples

Enable the local listener in `config.toml`:

```toml
[server.mcp_http]
enabled = true
bind = "127.0.0.1:14556"
```

A caller can resolve those settings independently of server startup:

```rust
let mcp_http = builder.resolve_mcp_http_serve_config()?;
```

Leaving out `[server.mcp_http]`, or setting `enabled = false`, returns no MCP HTTP configuration.

PR #1628, opened 2026-07-10
feat(cli): add long-running server command

sauldhernandez:
## Intent

Expose Coral's native gRPC runtime as a long-running CLI process for deployments and clients that need a stable server lifecycle. `coral server` reads its bind address exclusively from `[server].bind_addr`; it intentionally provides no command-line host or port overrides.

## What it enables

- Start a standalone native gRPC server that runs until `Ctrl-C` or, on Unix, `SIGTERM`, then shuts down cleanly.
- Print the selected endpoint and a `CORAL_ENDPOINT` hint for connecting clients.
- Listen on loopback or non-loopback addresses selected in configuration. Coral does not gate non-loopback binds on server authentication.
- Fall back to an ephemeral port on `127.0.0.1` when no `[server]` section is configured.

## Usage examples

Configure a fixed endpoint:

```toml
[server]
bind_addr = "127.0.0.1:14555"
```

Start the server:

```shell
coral server
```

Coral reports the selected endpoint:

```text
Coral gRPC server listening on http://127.0.0.1:14555
Connect clients with CORAL_ENDPOINT=http://127.0.0.1:14555
Press Ctrl-C to stop the server.
```

To listen beyond loopback, set a non-loopback address directly in `config.toml`, for example:

```toml
[server]
bind_addr = "0.0.0.0:14555"
```

## Validation

- `cargo test --locked -p coral-app --lib configured_standalone_grpc`
- `cargo test --locked -p coral-app --lib auth`
- `make rust-checks`
- `make docs-check`
- `make perf-check` (226.5 ms mean; 750 ms threshold on the fully restacked branch)
- `git diff --check`

PR #1626, opened 2026-07-10
feat(mcp-http): add loopback Streamable HTTP server

sauldhernandez:
## Stack overview

This stack lets a long-running Coral server authenticate users through an external OpenID Connect provider and protect its gRPC and MCP HTTP APIs.

## Where this PR fits

This PR and the next one add Streamable HTTP as a second transport for Coral's existing MCP tools. This PR establishes the local, authentication-disabled server; the next PR adds the bearer-authenticated version on top of the same lifecycle.

## Intent

Allow local programs to connect to Coral's MCP server over HTTP instead of requiring the stdio transport.

## What it enables

`coral-mcp` can now start an MCP Streamable HTTP listener backed by an existing `AppClient`. This first serving mode is deliberately restricted to loopback addresses because all sessions share one unauthenticated local client.

The server also:

- exposes `/livez` and `/readyz` health endpoints;
- limits MCP request bodies to 1 MiB;
- rejects requests carrying an `Origin` header because browser clients are not supported;
- rejects non-initialization requests that do not have a session before creating an MCP handler;
- restricts accepted `Host` values to the bound address; and
- closes active sessions and drains in-flight requests during shutdown.

The same configured MCP tools are available over stdio and HTTP; choosing HTTP does not reduce the tool set.

## Usage examples

```rust
use coral_mcp::{
    McpOptions,
    http::{McpHttpConfig, start_auth_disabled},
};

let config = McpHttpConfig::new("127.0.0.1:0".parse()?)?;
let server = start_auth_disabled(config, app, McpOptions::default()).await?;

println!("MCP is available at http://{}/mcp", server.local_addr());

server.shutdown().await?;
```

A non-loopback address such as `0.0.0.0:14556` is rejected by this authentication-disabled entry point.

## Validation

- `cargo test --locked -p coral-mcp --all-features`
- `cargo clippy --locked -p coral-mcp --all-targets --all-features -- -D warnings`
- `make rust-checks`
- `git diff --check`

PR #1625, opened 2026-07-10
feat(app): add ES256 session token core

sauldhernandez:
## Stack overview

This stack lets a long-running Coral server authenticate users through an external OpenID Connect provider and protect its gRPC and MCP HTTP APIs.

## Where this PR fits

The earlier PRs establish local gRPC and MCP HTTP serving. This PR starts the authentication portion of the stack by defining the tokens Coral will issue after a user logs in. The next PR adds the short-lived state needed while that login is in progress.

## Intent

Give Coral a session-token format that can eventually be verified by a detached MCP HTTP server without sharing Coral's private signing key.

## What it enables

Coral can issue and validate session tokens signed with ES256 using a P-256 private key.

Each token includes the Coral issuer, audience, provider, user subject, issue time, not-before time, and expiry. Validation pins the signing algorithm and key ID, requires every expected claim, checks the configured issuer and audience, and rejects tokens whose lifetime exceeds the configured limit.

The issuer derives a public JWK from the private key and gives it an RFC 7638 thumbprint as its key ID. That public key can be published later so another process can validate Coral tokens. This PR intentionally does not add a JWKS HTTP endpoint.

The token code is not connected to login routes, gRPC authentication, or MCP authentication in this PR.

## Usage examples

Load a raw DER-encoded PKCS#8 P-256 private key from a file relative to `config.toml`:

```toml
[auth.session]
issuer = "https://coral.example.com"
audience = "https://coral.example.com/mcp"
access_token_ttl_seconds = 900
signing_key_file = "session-signing-key.pk8"
```

Alternatively, point to an environment variable containing the same key encoded with standard base64:

```toml
[auth.session]
signing_key_env = "CORAL_SESSION_SIGNING_KEY"
```

Configuring both key sources is rejected. When neither key source is configured, Coral reads the key from the `CORAL_SESSION_SIGNING_KEY` environment variable.

## Validation

- `cargo test -p coral-app --all-features auth::session::tests`
- `make rust-checks`

PR #1637, opened 2026-07-10
feat(app): add OIDC discovery and code exchange

sauldhernandez:
## Stack context

**This part of the stack:** PR #1635 keeps the validated provider settings, this PR uses those settings to communicate with the provider, and PR #1639 verifies the identity token returned by the provider.

**This PR:** Adds Coral's internal OpenID Connect client for provider discovery, browser authorization requests, and authorization-code exchange.

## Intent

Coral cannot safely begin an external login from an issuer URL alone. It must discover the provider's exact endpoints, protect the browser round trip, and exchange the returned code without leaking credentials or trusting an unverified identity.

## What it enables

- Downloads the provider's standard discovery document and requires its issuer to exactly match the configured issuer.
- Validates the discovered authorization, token, and signing-key endpoints before using them.
- Creates fresh state, nonce, and PKCE values for every authorization request.
- Builds the upstream authorization URL and exchanges the returned code for an ID token.
- Refuses HTTP redirects, ignores system proxies, limits request time and response size, and returns errors that do not contain provider secrets.
- Passes the still-unverified ID token and signing-key information to the validation layer added by PR #1639.

This PR adds internal provider communication only; it does not expose a new login route by itself.

## Usage examples

When a later authorization route starts a login, it asks this client for:

- the provider URL to redirect the browser to;
- the state and nonce needed to verify the callback; and
- the PKCE verifier needed to exchange the returned code.

When the provider redirects back, the callback passes its code and the stored verifier to this client. The resulting ID token is then verified by the next PR before Coral accepts the user.

### DIFF
--- a/crates/coral-app/src/auth/provider_client.rs
+++ b/crates/coral-app/src/auth/provider_client.rs
@@ -1,6 +1,6 @@
 #![cfg_attr(not(test), expect(dead_code, reason = "wired by OAuth descendants"))]
 use super::config::ResolvedOidcProvider;
-use crate::outbound_url_policy::{ConfiguredEndpointUrl, read_bounded_body};
+use crate::outbound_url_policy::{ConfiguredEndpointUrl, DiscoveredEndpointUrl, read_bounded_body};
 use base64::Engine as _;
 use base64::engine::general_purpose::URL_SAFE_NO_PAD;
 use reqwest::header::ACCEPT;
@@ -26,14 +26,14 @@ pub(super) struct OidcAuthorizationRequest {
 }
 pub(super) struct OidcCodeExchange {
     id_token: Zeroizing<String>,
-    jwks_uri: ConfiguredEndpointUrl,
+    jwks_uri: DiscoveredEndpointUrl,
     signing_algorithms: Vec<String>,
 }
 impl OidcCodeExchange {
     pub(super) fn id_token(&self) -> &str {
         self.id_token.as_str()
     }
-    pub(super) fn jwks_uri(&self) -> &ConfiguredEndpointUrl {
+    pub(super) fn jwks_uri(&self) -> &DiscoveredEndpointUrl {
         &self.jwks_uri
     }
     pub(super) fn signing_algorithms(&self) -> &[String] {
@@ -144,6 +144,8 @@ impl OidcProviderClient {
         &self,
         provider: &ResolvedOidcProvider,
     ) -> Result<ValidatedDiscovery, OidcProviderClientError> {
+        let issuer = ConfiguredEndpointUrl::parse(&provider.issuer)
+            .map_err(|_error| OidcProviderClientError::Discovery)?;
         let url = discovery_url(&provider.issuer)?;
         let response = self
             .http
@@ -174,17 +176,22 @@ impl OidcProviderClient {
             authorization_endpoint: discovered_endpoint(
                 "authorization_endpoint",
                 &document.authorization_endpoint,
+                &issuer,
             )?,
-            token_endpoint: discovered_endpoint("token_endpoint", &document.token_endpoint)?,
-            jwks_uri: discovered_endpoint("jwks_uri", &document.jwks_uri)?,
+            token_endpoint: discovered_endpoint(
+                "token_endpoint",
+                &document.token_endpoint,
+                &issuer,
+            )?,
+            jwks_uri: discovered_endpoint("jwks_uri", &document.jwks_uri, &issuer)?,
             signing_algorithms: document.id_token_signing_alg_values_supported,
         })
     }
 }
 struct ValidatedDiscovery {
-    authorization_endpoint: ConfiguredEndpointUrl,
-    token_endpoint: ConfiguredEndpointUrl,
-    jwks_uri: ConfiguredEndpointUrl,
+    authorization_endpoint: DiscoveredEndpointUrl,
+    token_endpoint: DiscoveredEndpointUrl,
+    jwks_uri: DiscoveredEndpointUrl,
     signing_algorithms: Vec<String>,
 }
 #[derive(Deserialize)]
@@ -209,11 +216,12 @@ fn discovery_url(issuer: &str) -> Result<ConfiguredEndpointUrl, OidcProviderClie
 fn discovered_endpoint(
     label: &'static str,
     value: &str,
-) -> Result<ConfiguredEndpointUrl, OidcProviderClientError> {
+    issuer: &ConfiguredEndpointUrl,
+) -> Result<DiscoveredEndpointUrl, OidcProviderClientError> {
     if value.trim() != value {
         return Err(OidcProviderClientError::InvalidEndpoint(label));
     }
-    let url = ConfiguredEndpointUrl::parse(value)
+    let url = DiscoveredEndpointUrl::parse(value, issuer)
         .map_err(|_error| OidcProviderClientError::InvalidEndpoint(label))?;
     let reserved = url.as_url().query_pairs().any(|(key, _value)| {
         let key = key.to_ascii_lowercase();
@@ -328,6 +336,22 @@ mod tests {
             &format!("{}/token?tenant=one", server.uri()),
             &format!("{}/jwks", server.uri()),
         )
+    }
+
+    #[test]
+    fn remote_issuer_cannot_discover_loopback_http_endpoints() {
+        let issuer =
+            ConfiguredEndpointUrl::parse("https://accounts.example.test/tenant").expect("issuer");
+        for (label, endpoint) in [
+            ("authorization_endpoint", "http://127.0.0.1:9000/authorize"),
+            ("token_endpoint", "http://127.0.0.1:9000/token"),
+            ("jwks_uri", "http://127.0.0.1:9000/jwks"),
+        ] {
+            assert_eq!(
+                discovered_endpoint(label, endpoint, &issuer).expect_err("loopback downgrade"),
+                OidcProviderClientError::InvalidEndpoint(label)
+            );
+        }
     }
 
     #[tokio::test]
--- a/crates/coral-app/src/auth/provider_client.rs
+++ b/crates/coral-app/src/auth/provider_client.rs
@@ -1,5 +1,5 @@
 #![cfg_attr(not(test), expect(dead_code, reason = "wired by OAuth descendants"))]
-use super::config::OidcProviderSettings;
+use super::config::ResolvedOidcProvider;
 use crate::outbound_url_policy::{ConfiguredEndpointUrl, read_bounded_body};
 use base64::Engine as _;
 use base64::engine::general_purpose::URL_SAFE_NO_PAD;
@@ -69,7 +69,7 @@ impl OidcProviderClient {
 
     pub(super) async fn authorization_request(
         &self,
-        provider: &OidcProviderSettings,
+        provider: &ResolvedOidcProvider,
     ) -> Result<OidcAuthorizationRequest, OidcProviderClientError> {
         let discovery = self.discover(provider).await?;
         let state = random_url_token()?;
@@ -101,7 +101,7 @@ impl OidcProviderClient {
     }
     pub(super) async fn exchange_code(
         &self,
-        provider: &OidcProviderSettings,
+        provider: &ResolvedOidcProvider,
         code: &str,
         code_verifier: &str,
     ) -> Result<OidcCodeExchange, OidcProviderClientError> {
@@ -142,7 +142,7 @@ impl OidcProviderClient {
     }
     async fn discover(
         &self,
-        provider: &OidcProviderSettings,
+        provider: &ResolvedOidcProvider,
     ) -> Result<ValidatedDiscovery, OidcProviderClientError> {
         let url = discovery_url(&provider.issuer)?;
         let response = self
@@ -262,12 +262,16 @@ mod tests {
     use wiremock::matchers::{method, path};
     use wiremock::{Mock, MockServer, ResponseTemplate};
 
+    use base64::engine::general_purpose::STANDARD as BASE64_STANDARD;
+    use ring::signature::{ECDSA_P256_SHA256_FIXED_SIGNING, EcdsaKeyPair};
+    use std::path::Path;
+
     use super::*;
     use crate::auth::config::AuthSettings;
 
     const CLIENT_SECRET: &str = "client-secret-must-not-leak";
 
-    fn provider(issuer: &str) -> OidcProviderSettings {
+    fn provider(issuer: &str) -> ResolvedOidcProvider {
         let settings = AuthSettings::from_toml(&format!(
             "[auth]
              [auth.session]
@@ -284,6 +288,18 @@ mod tests {
         ))
         .expect("valid auth config")
         .expect("auth settings");
+        // A resolved provider is only reachable through runtime resolution, so
+        // the helper supplies the session signing key the same way a running
+        // server would.
+        let signing_key = BASE64_STANDARD.encode(
+            EcdsaKeyPair::generate_pkcs8(&ECDSA_P256_SHA256_FIXED_SIGNING, &SystemRandom::new())
+                .expect("P-256 signing key"),
+        );
+        let (settings, _issuer) = settings
+            .resolve_runtime_dependencies(Path::new("config.toml"), &|name| {
+                Ok((name == "CORAL_SESSION_SIGNING_KEY").then(|| signing_key.clone()))
+            })
+            .expect("resolved runtime dependencies");
         settings.provider().clone()
     }
 
--- a/crates/coral-app/src/auth/provider_client.rs
+++ b/crates/coral-app/src/auth/provider_client.rs
@@ -1,0 +1,586 @@
+#![cfg_attr(not(test), expect(dead_code, reason = "wired by OAuth descendants"))]
+use super::config::OidcProviderSettings;
+use crate::outbound_url_policy::{ConfiguredEndpointUrl, read_bounded_body};
+use base64::Engine as _;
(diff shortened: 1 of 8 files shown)


sauldhernandez:
* **#1637** <a href="https://app.graphite.com/github/pr/withcoral/coral/1637?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1637?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-app/src/auth/provider_client.rs
```diff
@@ -0,0 +1,626 @@
...
+) -> Result<DiscoveredEndpointUrl, OidcProviderClientError> {
+    if value.trim() != value {
+        return Err(OidcProviderClientError::InvalidEndpoint(label));
+    }
+    let url = DiscoveredEndpointUrl::parse(value, issuer)
+        .map_err(|_error| OidcProviderClientError::InvalidEndpoint(label))?;
+    let reserved = url.as_url().query_pairs().any(|(key, _value)| {
+        let key = key.to_ascii_lowercase();
+        match label {
+            "authorization_endpoint" => matches!(
```
simonwhitaker: **`response_mode` is missing from this list.**

`config.rs`'s `RESERVED_PROVIDER_AUTH_PARAMS` includes `response_mode`, with a comment spelling out why: `fragment` and `form_post` both stop the authorization code from ever reaching the GET callback route, so the login silently never completes and leaves nothing in the logs.

A discovery document advertising `authorization_endpoint: "https://idp/authorize?response_mode=form_post"` slips past this check and produces exactly the failure the config path goes out of its way to prevent — and from a *less* trusted source than the config file.

These two lists encode the same policy and shouldn't be able to drift. Could the authorization arm reuse `RESERVED_PROVIDER_AUTH_PARAMS` directly?

crates/coral-app/src/auth/provider_client.rs
```diff
@@ -0,0 +1,626 @@
...
+    id_token: Zeroizing<String>,
+}
+fn discovery_url(issuer: &str) -> Result<ConfiguredEndpointUrl, OidcProviderClientError> {
+    ConfiguredEndpointUrl::parse(&format!(
+        "{}/.well-known/openid-configuration",
+        issuer.trim_end_matches('/')
+    ))
+    .map_err(|_error| OidcProviderClientError::Discovery)
+}
+fn discovered_endpoint(
```
simonwhitaker, line 400: **Stringly-typed dispatch loses exhaustiveness.**

The `match label { ... _ => false }` below means a typo, or a third endpoint added later, silently disables the reserved-param check rather than failing to compile. Given this check is a security control, I'd rather the compiler enforced coverage.

An enum (`Endpoint::Authorization` / `Endpoint::Token`) with a `fn label(&self) -> &'static str` for the error variant gets that back at no runtime cost, and the `_ => false` arm disappears.

crates/coral-app/src/auth/provider_client.rs
```diff
@@ -0,0 +1,626 @@
...
+            code_verifier,
+        })
+    }
+    pub(super) async fn exchange_code(
+        &self,
+        provider: &ResolvedOidcProvider,
+        code: &str,
+        code_verifier: &str,
+    ) -> Result<OidcCodeExchange, OidcProviderClientError> {
+        let discovery = self.discover(provider).await?;
```
simonwhitaker, line 207: **Discovery is refetched on every call.**

`authorization_request` (line 74) and `exchange_code` (here) each run a full `discover()`. That's two discovery GETs plus the token POST per login, where one discovery would do.

It also means the `authorization_endpoint` used to build the redirect URL and the `token_endpoint` used to redeem the code can come from two different documents. Both are validated so I don't think it's exploitable, but a TTL-cached document would cut outbound requests by roughly two thirds and close the window.

crates/coral-app/src/auth/provider_client.rs
```diff
@@ -0,0 +1,626 @@
...
+    }
+    pub(super) fn jwks_uri(&self) -> &DiscoveredEndpointUrl {
+        &self.jwks_uri
+    }
+    pub(super) fn signing_algorithms(&self) -> &[String] {
+        &self.signing_algorithms
+    }
+}
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Error)]
+pub(super) enum OidcProviderClientError {
```
simonwhitaker, line 114: **Opaque errors are right; unobservable failures aren't.**

Every call site does `map_err(|_error| ...)` and nothing is logged anywhere in this module. Keeping the *returned* error free of provider data is correct and well-tested — but an operator debugging a broken login gets `"OIDC provider discovery failed"` with no status code, no cause, and no trace to correlate.

A `tracing::warn!` at each failure point carrying the HTTP status and the redacted URL would make this diagnosable without weakening anything: `DiscoveredEndpointUrl`'s `Debug` impl already redacts the path via `RedactedUrl`.

crates/coral-app/src/auth/provider_client.rs
```diff
@@ -0,0 +1,626 @@
...
+        provider: &ResolvedOidcProvider,
+        code: &str,
+        code_verifier: &str,
+    ) -> Result<OidcCodeExchange, OidcProviderClientError> {
+        let discovery = self.discover(provider).await?;
+        let response = self
+            .http
+            .post(discovery.token_endpoint.as_url().clone())
+            .header(ACCEPT, "application/json")
+            .form(&[
```
simonwhitaker, line 212: **Token-endpoint auth method isn't negotiated.**

This always uses `client_secret_post`, and `token_endpoint_auth_methods_supported` from the discovery document is ignored. A provider that only accepts `client_secret_basic` will fail here with an opaque `TokenExchange` — which, combined with the lack of logging, is close to undiagnosable.

At minimum, reading that field during `discover()` and failing with a specific error when `client_secret_post` isn't offered turns a mystery runtime failure into a clear one.

crates/coral-app/src/auth/provider_client.rs
```diff
@@ -0,0 +1,626 @@
...
+            .map_err(|_error| OidcProviderClientError::Discovery)?;
+        if !response.status().is_success() {
+            return Err(OidcProviderClientError::Discovery);
+        }
+        let body = read_bounded_body(response, RESPONSE_MAX_BYTES)
+            .await
+            .map_err(|_error| OidcProviderClientError::Discovery)?;
+        let document: DiscoveryDocument =
+            serde_json::from_slice(&body).map_err(|_error| OidcProviderClientError::Discovery)?;
+        if document.issuer != provider.issuer {
```
simonwhitaker, line 266: **Byte-exact match against the raw config string.**

Exact comparison is what the spec asks for, so no objection to the check itself. But `provider.issuer` is the operator's trimmed input — `provider_issuer_endpoint` parses to validate and then discards the normalized form — so a config of `https://ACCOUNTS.example.test` against a document returning `https://accounts.example.test` fails at the first login attempt rather than at startup.

Normalizing `issuer` at config time (or comparing normalized origins here) moves that class of typo into config validation, where the error message can actually say what's wrong.

crates/coral-app/src/auth/provider_client.rs
```diff
@@ -0,0 +1,626 @@
...
+const RESPONSE_MAX_BYTES: usize = 64 * 1024;
+#[derive(Clone)]
+pub(super) struct OidcProviderClient {
+    http: reqwest::Client,
+}
+pub(super) struct OidcAuthorizationRequest {
+    pub(super) url: Url,
+    pub(super) state: String,
+    pub(super) nonce: String,
+    pub(super) code_verifier: String,
```
simonwhitaker: **PKCE verifier isn't zeroized.**

`OidcCodeExchange::id_token` gets `Zeroizing<String>`, but `code_verifier` — a PKCE secret with much the same handling requirements — plus `state` and `nonce` are plain `String`s here. Given how carefully the rest of the module handles secret material, this reads as an oversight rather than a decision. `Zeroizing<String>` for at least `code_verifier` would make the treatment consistent.

crates/coral-app/src/outbound_url_policy.rs
```diff
@@ -230,15 +281,15 @@ pub(crate) async fn read_bounded_body(
...
+    let mut body = Zeroizing::new(Vec::with_capacity(limit));
     while let Some(chunk) = response
         .chunk()
         .await
         .map_err(OutboundUrlPolicyError::BodyRead)?
     {
         append_bounded_chunk(&mut body, &chunk, limit)?;
     }
-    Ok(body)
+    Ok(std::mem::take(&mut *body))
```
simonwhitaker, line 321: **The `Zeroizing` wrapper is a no-op on the success path.**

`std::mem::take(&mut *body)` moves the buffer out, so what `Zeroizing` drops is an empty `Vec` — the secret bytes leave in the returned `Vec`, unprotected. The wrapper only does real work on the error paths (`BodyTooLarge`, and `BodyRead` via `?`).

The actual success-path guarantee is the `with_capacity(limit)`, which prevents a realloc from stranding a copy of partially-read secret data — and that part is genuinely valuable. But the code reads as if it's doing more than it is, and a future reader may well "clean up" the `mem::take` into something that copies. Worth a comment stating explicitly that ownership transfer is deliberate and zeroization is the error path only.

Two smaller notes:
- `with_capacity(limit)` now allocates the full bound regardless of body size. Irrelevant at 64 KiB with one caller, but it turns `limit` into an allocation knob for any future one.
- In `provider_client.rs`, `exchange_code` wraps the result in `Zeroizing` but `discover` binds it to a plain `Vec`. Fine — discovery documents aren't secret — but the asymmetry deserves a word so it doesn't look accidental.

crates/coral-app/src/outbound_url_policy.rs
```diff
@@ -751,9 +836,11 @@ mod tests {
         let response = reqwest::get(format!("{}/metadata", server.uri()))
             .await
             .expect("response");
-        assert_eq!(
-            read_bounded_body(response, 5).await.expect("body"),
-            b"hello"
+        let body = read_bounded_body(response, 5).await.expect("body");
+        assert_eq!(body, b"hello");
+        assert!(
+            body.capacity() >= 5,
```
simonwhitaker: This assertion holds for *any* `Vec` containing 5 bytes, so it doesn't test the reserve-upfront behaviour the message claims. Asserting `capacity() == limit` with a limit well above the body length (say, read a 5-byte body with `limit = 4096`) would actually pin the invariant.

crates/coral-app/src/auth/provider_client.rs
```diff
@@ -0,0 +1,626 @@
...
+        }
+    });
+    (!reserved)
+        .then_some(url)
+        .ok_or(OidcProviderClientError::InvalidEndpoint(label))
+}
+
+fn random_url_token() -> Result<String, OidcProviderClientError> {
+    let mut bytes = Zeroizing::new([0_u8; 32]);
+    SystemRandom::new()
```
simonwhitaker: Nit: a fresh `SystemRandom::new()` on each call, three calls per authorization request. Worth hoisting to a field on `OidcProviderClient` or a single local in `authorization_request`.

crates/coral-app/src/auth/provider_client.rs
```diff
@@ -0,0 +1,626 @@
...
+use base64::engine::general_purpose::URL_SAFE_NO_PAD;
+use reqwest::header::ACCEPT;
+use ring::rand::{SecureRandom as _, SystemRandom};
+use serde::Deserialize;
+use sha2::{Digest as _, Sha256};
+use std::time::Duration;
+use thiserror::Error;
+use url::Url;
+use zeroize::Zeroizing;
+const CONNECT_TIMEOUT: Duration = Duration::from_secs(3);
```
simonwhitaker, line 38: Style nits for the module as a whole:

- No module-level `//!` doc and no `///` on any item, unlike `outbound_url_policy.rs`, `error.rs`, and `config.rs` right next door. This module encodes some subtle trust decisions that would benefit from the same treatment.
- No blank lines between top-level items here (lines 14–68) or between the `impl` methods (lines 101–143), which reads as accidental rather than intentional.
- Import grouping differs from neighbouring files, which use `std` → external → `crate`. Same in the test module.

crates/coral-app/src/auth/provider_client.rs
```diff
@@ -0,0 +1,626 @@
...
+        for value in [&request.state, &request.nonce, &request.code_verifier] {
+            assert_eq!(value.len(), 43);
+            assert!(
+                value
+                    .bytes()
+                    .all(|byte| byte.is_ascii_alphanumeric() || matches!(byte, b'-' | b'_'))
+            );
+        }
+        assert_ne!(request.state, request.nonce);
+        assert_ne!(request.nonce, request.code_verifier);
```
simonwhitaker: Nit: `state` vs `nonce` and `nonce` vs `code_verifier` are compared, but not `state` vs `code_verifier`. Almost certainly fine given they come from independent draws, but the third assertion completes the set for free.

crates/coral-app/src/auth/provider_client.rs
```diff
@@ -0,0 +1,626 @@
...
+            assert_eq!(query.get(key).map(String::as_str), Some(expected));
+        }
+        assert_eq!(query.get("state"), Some(&request.state));
+        assert_eq!(query.get("nonce"), Some(&request.nonce));
+        let expected = URL_SAFE_NO_PAD.encode(Sha256::digest(request.code_verifier.as_bytes()));
+        assert_eq!(query.get("code_challenge"), Some(&expected));
+    }
+
+    #[tokio::test]
+    async fn discovery_requires_exact_issuer_and_validates_every_endpoint() {
```
simonwhitaker: Nit: this covers three distinct behaviours — the endpoint rejection table, private-HTTPS acceptance, and the RS256 requirement. Splitting into three tests would make a failure self-describing rather than requiring a line number to interpret.

simonwhitaker (approved):

crates/coral-app/src/auth/provider_client.rs
sauldhernandez: Good catch, and it is addressed higher in the stack rather than here: #2002 caches the discovery document behind a configurable TTL (five minutes by default), collapses concurrent misses into a single refresh, and caches only successes so a provider blip never outlives itself.

It sits above #1655 deliberately. Once CIMD makes `client_id` something any caller can supply, `GET /oauth/authorize` turns each inbound request into an outbound one to the IdP — so the cache is a prerequisite for that PR, not just an optimization for this one. Keeping it there also keeps this PR to the client itself.

The two-documents window you spotted closes with it.

crates/coral-app/src/auth/provider_client.rs
sauldhernandez: Agreed, and the gap is wider than this module — there is no `tracing` anywhere in `auth/` or in `outbound_url_policy.rs`. That makes it a decision about the serving stack rather than a line fix here, so I am doing it as its own PR rather than logging in one module and nowhere else.

Recording the shape you propose so it does not get relitigated: `warn!` at each failure point carrying the HTTP status and the redacted URL, leaning on the existing `RedactedUrl` `Debug` impl, with the returned error left exactly as opaque as it is now.

crates/coral-app/src/auth/provider_client.rs
sauldhernandez: This one landed downstack between you reading it and now — `34e07a53e` on the config PR added `validate_canonical_issuer`, so your exact example is a startup error rather than a first-login failure:

```
issuer must be copied exactly as the provider publishes it; `https://ACCOUNTS.example.test` is not a canonical URL (its canonical form is `https://accounts.example.test/`)
```

It rejects any spelling the parser would rewrite, and tolerates only the trailing slash, since providers publish issuers both ways. The byte-exact comparison here is then comparing against a value already known to be canonical.

(`provider_issuer_endpoint` is `provider_literal_endpoint` now, same reason.)

crates/coral-app/src/auth/provider_client.rs
sauldhernandez: Done for `code_verifier`, which is now `Zeroizing<String>`. Deliberately not for `state` and `nonce` — both travel in the clear in the authorization URL and are anti-replay values rather than secrets, so wrapping them would suggest a protection that the flow does not actually give them.

crates/coral-app/src/auth/provider_client.rs
sauldhernandez: Done, with one deliberate deviation worth flagging.

`discover` now reads `token_endpoint_auth_methods_supported` and fails with a named `UnsupportedTokenEndpointAuth` before the user is redirected anywhere, rather than opaquely after they come back.

The deviation: OIDC Discovery 1.0 §3 says an *omitted* list defaults to `client_secret_basic`. Applying that literally would reject every provider that accepts `client_secret_post` without advertising it, which is common enough that it would be a compatibility regression on an optional field. So an omitted list is treated as unknown and left for the token request to settle; a list that is *present* is believed, including an empty one. The reasoning is on `offers_token_endpoint_auth_method` so it does not read as an oversight.

Happy to tighten it to the spec default if you would rather have the strict reading.

sauldhernandez:
### Merge activity

* **Jul 29, 12:35 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1637).
* **Jul 29, 12:55 PM UTC**: [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1637) rebased this pull request as part of a merge.
* **Jul 29, 1:03 PM UTC**: @sauldhernandez merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1637).

PR #1639, opened 2026-07-10
feat(app): validate OIDC ID tokens

sauldhernandez:
## Stack context

**Whole stack:** This stack lets a long-running Coral server authenticate users through an external OpenID Connect provider and protect its gRPC and MCP HTTP APIs.

**This part of the stack:** PR #1637 obtains an ID token from the configured provider, this PR proves that the token is genuine and intended for Coral, and PR #1640 begins the browser-facing authorization flow that relies on that verification.

**This PR:** Adds strict validation of provider ID tokens and returns only the user identity fields Coral needs.

## Intent

An ID token received from a provider is untrusted input. Coral must verify its signature, intended recipient, lifetime, and browser nonce before treating any claim in it as a user identity.

## What it enables

- Downloads the provider's public signing-key document with a strict size limit and refuses redirects or malformed responses.
- Accepts only supported public-key signature algorithms; shared-secret token signatures are rejected.
- Requires exactly one signing key matching the token's key ID and declared algorithm.
- Checks the signature, issuer, audience, authorized party, expiration, issue time, optional not-before time, and expected nonce.
- Applies the configured required claims and maps the configured identity and display-name claims.
- Returns only the verified principal and optional display name. Token contents and provider response details are not exposed through errors.

This PR adds identity verification inside the provider client; it does not add a public HTTP route.

## Usage examples

After an authorization code is exchanged, the provider client passes this validator:

- the returned ID token;
- the nonce saved before the browser redirect;
- the provider's advertised signing algorithms; and
- the provider's public signing keys.

A valid token produces a verified Coral identity. A wrong signature, issuer, audience, nonce, lifetime, key, or required claim produces the same generic validation failure.

PR #1638, opened 2026-07-10
feat(cli): compose MCP HTTP with gRPC

sauldhernandez:
## Stack overview

This stack lets a long-running Coral server authenticate users through an external OpenID Connect provider and protect its gRPC and MCP HTTP APIs.

## Where this PR fits

The previous PR teaches `coral-app` how to resolve local MCP HTTP settings. This PR completes the local serving path by making the CLI start and stop the gRPC and MCP HTTP listeners together. The next PR begins the separate user-authentication foundation by adding Coral session tokens.

## Intent

Let `coral server` optionally expose the same in-process Coral server through loopback MCP Streamable HTTP.

## What it enables

When `[server.mcp_http]` is enabled, `coral server` now:

1. starts the configured gRPC server;
2. derives a safe loopback address for that server;
3. connects an `AppClient` to gRPC;
4. starts the loopback MCP HTTP listener; and
5. prints both endpoints.

The CLI owns this orchestration so gRPC and MCP HTTP can be separated later without putting transport-specific startup logic in `coral-app`.

Startup and shutdown are coordinated. If MCP HTTP fails to start after gRPC has started, the CLI shuts gRPC down. On exit, it stops MCP HTTP first and then gRPC, while preserving errors from either listener.

This PR starts only the authentication-disabled, loopback MCP HTTP mode. It does not make the authenticated transport from the earlier PR available through `coral server`.

## Usage examples

```toml
[server]
bind_addr = "127.0.0.1:14555"

[server.mcp_http]
enabled = true
bind = "127.0.0.1:14556"
```

Then run:

```shell
coral server
```

The CLI prints the gRPC endpoint and `http://127.0.0.1:14556/mcp`. Press `Ctrl-C`, or send `SIGTERM` on Unix, to stop both listeners.

PR #1624, opened 2026-07-10
feat(client): support authenticated remote connections

sauldhernandez:
## Intent

Allow Coral clients to attach validated outbound gRPC metadata, including bearer authorization, through the typed `AppClient` connection path without changing unauthenticated local connections.

## What it enables

- `AppClient::connect_with_bearer` sends bearer metadata only to public `https://` endpoints.
- `AppClient::connect_with_metadata` supports other validated outbound metadata while preserving repeated values and sensitive authorization handling.
- Public plaintext `http://` endpoints fail before dialing when authorization metadata is present; `AppClient::connect` remains available for unauthenticated local servers.

## Usage examples

Use `AppClient::connect` for an unauthenticated local server. For an authenticated remote server, construct a `BearerToken` and call `AppClient::connect_with_bearer` with an `https://` endpoint.

PR #1622, opened 2026-07-10
refactor(mcp): export tool core

sauldhernandez:
## Intent

Make Coral’s MCP handlers reusable by transports other than stdio without hard-coding a transport-specific tool set.

## What it enables

Adds a public `CoralMcpServerFactory` that creates a fresh handler for each MCP session. The handler exposes the tools allowed by `McpOptions`, so stdio and alternate transports use the same capability rules.

Transports still own networking, lifecycle, and authentication. An authenticated transport must construct the factory with an `AppClient` scoped to the validated session authorization and must not fall back to a shared unauthenticated client. The existing stdio server continues to work as before.

## Usage examples

```rust
let factory = CoralMcpServerFactory::new(app, options);
let handler = factory.create();
```

Call `create()` once per MCP session. Enable task and feedback tools through `McpOptions`; the transport does not restrict them independently.

## Validation

- `make rust-checks`
- `cargo test --locked -p coral-mcp --all-features factory_shares_client_and_configured_tools_across_sessions`
- `git diff --check`

PR #1640, opened 2026-07-10
feat(app): handle OAuth authorization requests

sauldhernandez:
## Stack context

**This part of the stack:** PR #1639 establishes trusted provider identities, this PR validates the client's initial authorization request and redirects the browser upstream, and PR #1641 handles the provider's callback.

**This PR:** Adds the `GET /oauth/authorize` endpoint and saves the information needed to verify the rest of the login.

## Intent

Before sending a browser to an identity provider, Coral must know that the client and callback are allowed, that the request targets this Coral server, and that the client supplied a valid PKCE proof.

## What it enables

- Bounds the query size, parameter count, names, and values, and rejects duplicate parameters.
- Requires an exact approved client and redirect-URI match before redirecting errors back to the client.
- Supports only the authorization-code flow with PKCE S256.
- Requires a `resource` that names this Coral server.
- Rejects any requested `scope`. This server advertises no `scopes_supported`, so a compliant client omits the parameter.
- Creates the upstream provider request and stores a single-use record containing the client, callback, PKCE challenge, resource, provider, provider nonce, and provider PKCE verifier.
- Redirects the browser upstream only after that record has been saved.

This layer intentionally has no production source of approved clients yet. A later PR connects the endpoint to Client ID Metadata Documents.

## Usage examples

In the completed stack, a web app's backend-for-frontend (BFF) starts login with a request shaped like:

```text
GET /oauth/authorize
  ?response_type=code
  &client_id=https://client.example/oauth-client.json
  &redirect_uri=https://client.example/oauth/callback
  &resource=https://coral.example/mcp
  &code_challenge=<S256 challenge>
  &code_challenge_method=S256
  &state=<client state>
```

At this point in the stack, tests provide the approved client entry. Production client lookup is added later.

### DIFF
--- a/crates/coral-app/src/auth/authorization_server/authorize.rs
+++ b/crates/coral-app/src/auth/authorization_server/authorize.rs
@@ -9,6 +9,7 @@ use super::super::PROVIDER_ID;
 use super::super::provider_client::OidcAuthorizationRequest;
 use super::super::state_store::OAuthAuthorizationSessionRecord;
 use super::{AuthorizationServerHttpState, canonical_authorization_resource};
+use crate::outbound_url_policy::{BrowserRedirect, EndpointUrl};
 
 const MAX_QUERY_BYTES: usize = 8 * 1024;
 const MAX_PARAMETERS: usize = 32;
@@ -149,6 +150,13 @@ fn parse_query(raw: &str) -> Result<AuthorizeQuery, ()> {
     Ok(query)
 }
 
+/// Resolves the registered redirect URI a request names, under
+/// [`BrowserRedirect`].
+///
+/// The registry is remote input once client metadata documents populate it, so
+/// the policy is applied here rather than trusted to have been applied at
+/// registration: a URI the policy rejects resolves to `None`, and the request
+/// fails with a direct error instead of becoming a redirect.
 fn registered_redirect(
     registered_clients: &BTreeMap<String, Vec<String>>,
     client_id: &str,
@@ -158,7 +166,8 @@ fn registered_redirect(
         .get(client_id)?
         .iter()
         .find(|uri| uri.as_str() == redirect_uri)
-        .and_then(|uri| Url::parse(uri).ok())
+        .and_then(|uri| EndpointUrl::<BrowserRedirect>::parse(uri).ok())
+        .map(EndpointUrl::into_url)
 }
 
 fn valid_s256_challenge(value: &str) -> bool {
@@ -405,6 +414,25 @@ mod tests {
         assert!(location(&response).is_none());
         let body = to_bytes(response.into_body(), 4096).await.expect("body");
         assert!(!String::from_utf8_lossy(&body).contains("raw-secret"));
+    }
+
+    #[tokio::test]
+    async fn registered_redirect_outside_browser_policy_never_becomes_a_redirect() {
+        let store = Arc::new(InMemoryStateStore::new());
+        for uri in [
+            "javascript:alert(1)",
+            "data:text/html,<script>alert(1)</script>",
+            "http://client.example.test/callback",
+        ] {
+            let mut auth_state = state("https://provider.invalid", store.clone());
+            auth_state.registered_clients =
+                Arc::new(BTreeMap::from([(CLIENT_ID.into(), vec![uri.into()])]));
+            let mut request_pairs = pairs();
+            replace(&mut request_pairs, "redirect_uri", Some(uri));
+            let response = request(auth_state, &request_pairs).await;
+            assert_eq!(response.status(), StatusCode::BAD_REQUEST, "accepted {uri}");
+            assert!(location(&response).is_none(), "redirected to {uri}");
+        }
     }
 
     #[tokio::test]
--- a/crates/coral-app/src/auth/authorization_server/authorize.rs
+++ b/crates/coral-app/src/auth/authorization_server/authorize.rs
@@ -5,7 +5,6 @@ use axum::http::{StatusCode, header};
 use axum::response::{IntoResponse, Response};
 use url::{Url, form_urlencoded};
 
-use super::super::PROVIDER_ID;
 use super::super::provider_client::OidcAuthorizationRequest;
 use super::super::state_store::OAuthAuthorizationSessionRecord;
 use super::{AuthorizationServerHttpState, canonical_authorization_resource};
@@ -85,7 +84,6 @@ pub(super) async fn oauth_authorize(
         code_verifier: oidc_code_verifier,
     } = request;
     let session = OAuthAuthorizationSessionRecord {
-        provider_id: PROVIDER_ID.to_string(),
         client_id: client_id.to_string(),
         redirect_uri: redirect_uri.to_string(),
         client_state: query.state,
@@ -500,7 +498,6 @@ mod tests {
             .await
             .expect("store")
             .expect("stored before redirect");
-        assert_eq!(session.provider_id, PROVIDER_ID);
         assert_eq!(session.client_id, CLIENT_ID);
         assert_eq!(session.redirect_uri, REDIRECT_URI);
         assert_eq!(session.client_state.as_deref(), Some(CLIENT_STATE));
@@ -534,7 +531,6 @@ mod tests {
         provider.reset().await;
         mount_discovery(&provider, 200).await;
         let filler = OAuthAuthorizationSessionRecord {
-            provider_id: PROVIDER_ID.into(),
             client_id: CLIENT_ID.into(),
             redirect_uri: REDIRECT_URI.into(),
             client_state: None,
--- a/crates/coral-app/src/auth/authorization_server/authorize.rs
+++ b/crates/coral-app/src/auth/authorization_server/authorize.rs
@@ -1,0 +1,563 @@
+use std::collections::{BTreeMap, BTreeSet};
+
+use axum::extract::{RawQuery, State};
+use axum::http::{StatusCode, header};
+use axum::response::{IntoResponse, Response};
+use url::{Url, form_urlencoded};
+
+use super::super::PROVIDER_ID;
+use super::super::provider_client::OidcAuthorizationRequest;
+use super::super::state_store::OAuthAuthorizationSessionRecord;
+use super::{AuthorizationServerHttpState, canonical_authorization_resource};
+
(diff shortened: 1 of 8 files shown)


sauldhernandez:
* **#1640** <a href="https://app.graphite.com/github/pr/withcoral/coral/1640?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1640?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-app/src/auth/authorization_server/authorize.rs
```diff
@@ -0,0 +1,563 @@
...
+    };
+    let Ok(resource) = canonical_authorization_resource(resource) else {
+        return trusted.error("invalid_target", "resource is invalid");
+    };
+    if !state.authorization_resources.contains(&resource) {
+        return trusted.error("invalid_target", "resource does not match this server");
+    }
+    let request = match state
+        .provider_client
+        .authorization_request(state.settings.provider())
```
simonwhitaker, line 74: **Every authorize request triggers an uncached upstream discovery fetch.**

`OidcProviderClient::authorization_request` calls `self.discover(provider)` unconditionally, and `discover` has no cache — it issues a fresh `GET {issuer}/.well-known/openid-configuration` with a 3s connect / 10s request timeout.

`/oauth/authorize` is unauthenticated, and `client_id` + `redirect_uri` are public values. Anyone who can reach this endpoint can force one outbound HTTPS round-trip to the IdP per request. That gives unbounded request amplification against the provider, tail latency on the happy path, and a Coral-side availability coupling to provider responsiveness. The router has no `tower` layers either — no concurrency limit, timeout, or rate limit.

Caching discovery (honouring `Cache-Control`, or a fixed TTL behind a single-flight guard) looks like the highest-value follow-up in this stack. If it's deliberately deferred to a descendant PR, a `TODO` naming that PR would be worth adding here.

crates/coral-app/src/auth/authorization_server/authorize.rs
```diff
@@ -0,0 +1,563 @@
...
+fn registered_redirect(
+    registered_clients: &BTreeMap<String, Vec<String>>,
+    client_id: &str,
+    redirect_uri: &str,
+) -> Option<Url> {
+    registered_clients
+        .get(client_id)?
+        .iter()
+        .find(|uri| uri.as_str() == redirect_uri)
+        .and_then(|uri| Url::parse(uri).ok())
```
simonwhitaker: **Registered redirect URIs bypass the crate's outbound URL trust policy.**

This is plain `Url::parse` — no scheme, credential, or transport check. `javascript:`, `data:`, and plain `http://` to arbitrary hosts all parse successfully and would flow straight into `Location:`.

Not exploitable today: browsers don't follow `javascript:` from a `Location` header, and the registry is test-only at this point in the stack. But this PR fixes the shape that the CIMD PR will feed untrusted data into, and at that point the registry's provenance changes entirely.

The crate already has `PublicMetadataUrl` / `ConfiguredEndpointUrl` for precisely this. Applying a policy here — public HTTPS, or explicit-loopback HTTP for native clients — rather than trusting the registry to have done it is a few lines of defence in depth, and cheaper to add now than to retrofit once something else is populating the map.

crates/coral-app/src/auth/authorization_server/authorize.rs
```diff
@@ -0,0 +1,563 @@
...
+            return trusted.error("invalid_request", "PKCE S256 code_challenge is required");
+        }
+    };
+    if query.scope.is_some() {
+        return trusted.error("invalid_scope", "scope is not supported");
+    }
+    let Some(resource) = query.resource.as_deref() else {
+        return trusted.error("invalid_target", "resource is required");
+    };
+    let Ok(resource) = canonical_authorization_resource(resource) else {
```
simonwhitaker, line 66: **`ConfiguredEndpointUrl` is being built from attacker-controlled input, against its own documented contract.**

`outbound_url_policy.rs` is explicit about this:

> This profile permits plain-HTTP loopback, and HTTPS to any host including a private one, with no check on what the host resolves to. Both are sound only for a value an operator authored — a config file or environment variable.
>
> **Do not build one from a value a remote party controls.**

`canonical_authorization_resource` runs `ConfiguredEndpointUrl::parse` on the client-supplied `resource` parameter.

This is safe as written, because the result is only compared against `authorization_resources` and never fetched — so the profile's permissiveness about private hosts and loopback HTTP never turns into an outbound request. But that argument lives nowhere in the code, and the shared helper means the config path and the request path are now coupled: a future change that makes the canonical form feed anything other than a set lookup would silently inherit an SSRF-shaped profile.

Either add a comment at this call site recording why compare-only makes it sound, or split canonicalization from the transport policy so the two paths don't share a type whose docs forbid this use.

simonwhitaker (approved):
Approving. The security-critical ordering is right: validating `client_id`/`redirect_uri` before anything else, and returning a direct JSON 400 rather than a redirect until that check passes, is exactly the property that keeps this endpoint from becoming an open redirect. Duplicate rejection on *decoded* parameter names closes a parser-differential class of bug that most implementations miss, and applying it to unknown parameters too is the correct conservative call. Errors carry only `&'static str`, so nothing client- or provider-supplied can reach the redirect query.

Test quality is high — `empty_resource_allowlist_fails_closed`, the store-before-redirect ordering assertion, and the `!provider_url.contains(secret)` checks all target the properties that matter rather than line coverage.

Three things to follow up on, left as inline comments. None block the merge, but the first two should be tracked before the stack reaches production:

1. Every authorize request triggers an uncached upstream discovery fetch, on an unauthenticated endpoint.
2. Registered redirect URIs bypass the crate's outbound URL trust policy — latent today, live once CIMD feeds the registry.
3. `ConfiguredEndpointUrl` is built from attacker-controlled input, against its own documented contract. Safe as written, but the reasoning isn't recorded anywhere.

crates/coral-app/src/auth/authorization_server/authorize.rs
sauldhernandez: Deliberately deferred — the discovery fetch is covered by two descendants of this PR:

- **#1654** splits this handler into `oauth_authorize_get` / `oauth_authorize_post` and moves `authorization_request` into the POST, behind the consent step. The unauthenticated GET stops touching the provider entirely.
- **#2002** adds the cache you describe, taking the second of your two options: a fixed TTL (5 min default, 1 h maximum, `0` disables) with concurrent misses collapsed behind a single-flight guard. Failures are deliberately not cached, so a provider blip can't be pinned for a whole TTL.

Router-level limits are a broader concern than this endpoint, so I'm following that up separately rather than in this thread.

Resolving as covered downstream.

crates/coral-app/src/auth/authorization_server/authorize.rs
sauldhernandez: Addressed in e977e01b8: `registered_redirect` now resolves the URI under `EndpointUrl<BrowserRedirect>` instead of raw `Url::parse`, so `javascript:`, `data:`, and arbitrary-host plain HTTP resolve to `None` and the request fails with a direct error rather than becoming a redirect. Covered by a regression test.

crates/coral-app/src/auth/authorization_server/authorize.rs
sauldhernandez: Took the split option in 80f2b7b94: canonicalization now runs under a dedicated `ResourceIdentifier` policy rather than `Configured`. It shares `Configured`'s transport rule but withholds `FetchablePolicy`, so a remote-controlled `resource` can be compared and recorded but never fetched — the config path and the request path no longer share a type whose docs forbid remote input.

crates/coral-app/src/outbound_url_policy.rs
```diff
@@ -27,112 +43,215 @@ use crate::bootstrap::is_loopback_ip;
...
-    /// Parses an HTTPS endpoint or an explicit loopback HTTP endpoint.
+/// A profile that both fetches and permits a private or loopback host is sound
+/// only because of where the URL came from. A policy that omits this marker
+/// describes a URL Coral only ever compares, stores, or hands to a browser, and
+/// [`EndpointUrl::get`] and [`EndpointUrl::post`] do not exist for it.
+pub(crate) trait FetchablePolicy: UrlPolicy {}
+
+impl<P: UrlPolicy> EndpointUrl<P> {
+    /// Returns the validated URL.
+    pub(crate) fn as_url(&self) -> &Url {
```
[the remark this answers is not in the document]
sauldhernandez: Accurate as stated: `as_url`/`into_url` hand back the raw `Url` for every policy, so `FetchablePolicy` gates only the `get`/`post` helpers, not raw extraction.

That's by design, though — the finding is really pointing at a doc over-claim rather than a missing boundary. `EndpointUrl<P>` is a *proof that a URL passed policy `P`'s checks*; it isn't a capability token, and it can't be. A `Url` you can see is a `Url` you can fetch (clone it, or rebuild it from the string), so "this value can never be fetched" is not a guarantee the type is in a position to make.

The actual SSRF protections don't live in usage-gating:

- `PublicMetadata` is the only profile Coral fetches from attacker-controlled input, and it's guarded at validation time and again at fetch time by `PublicMetadataResolver`, which rejects any hostname resolving to a non-public address (DNS rebinding included).
- `ResourceIdentifier` and `BrowserRedirect` are safe because no call site fetches them — a property of the code, not the type.

Narrowing `as_url`/`into_url` to `FetchablePolicy` would only relocate the same escape hatch (the extracted `Url` is still fetchable) while forcing awkward accessors onto the compare-only and redirect profiles, so we're not doing that. What we have fixed is the two doc passages that described the non-fetchable policies as ones "no request can be built from" / that "cannot" become a request — that phrasing invited reading the marker as a confinement boundary it was never meant to be. Resolving on that basis.

sauldhernandez:
### Merge activity

* **Jul 31, 8:06 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1640).
* **Jul 31, 8:06 AM UTC**: @sauldhernandez merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1640).

PR #1621, opened 2026-07-10
feat(app): add outbound URL policy profiles

sauldhernandez:
## Stack overview

This stack lets a long-running Coral server authenticate users through an external OpenID Connect provider and protect its gRPC and MCP HTTP APIs.

## Where this PR fits

The previous PR creates the OAuth server shell. This PR adds the URL checks needed before that server can safely contact external providers or fetch metadata supplied by a client. The next PR applies the configured-endpoint policy while retaining usable OpenID Connect provider settings.

## Intent

Use different security rules for URLs chosen by an operator and URLs supplied by an untrusted OAuth client.

## What it enables

`ConfiguredEndpointUrl` accepts HTTPS endpoints and explicit loopback HTTP endpoints. It is intended for values such as an operator-configured OpenID Connect issuer or callback. Its loopback rule is the same `bootstrap::is_loopback_ip` helper used by the server bind guards and the auth URL validator, so all of them treat the same spellings as loopback — including IPv4-mapped forms such as `[::ffff:127.0.0.1]`. The type documents that it is only sound for operator-authored values; endpoints a remote party controls, such as those inside a discovery document, need a stricter profile.

`PublicMetadataUrl` is stricter because a remote client controls the value. It requires a public HTTPS document URL, rejects credentials, fragments, localhost, and non-public IP addresses, and rejects path traversal in every spelling the URL parser accepts — including scheme forms without `//` and inputs carrying embedded tab or newline characters, which the parser would otherwise silently canonicalize.

The IP blocklists implement the IANA special-purpose registries as a denylist, with each branch annotated with the range and registry entry it encodes. The IPv6 table includes the post-2024 entries: SRv6 SIDs (`5f00::/16`), the newer documentation range (`3fff::/20`), and the dummy prefix (`100:0:0:1::/64`).

The matching HTTP client:

- ignores environment proxy settings;
- refuses redirects;
- applies a five-second connection and request timeout;
- rejects a hostname if any DNS answer is non-public; and
- lets callers cap the number of response bytes read into memory, whether or not the response declares a length.

Debug output shows the scheme, host, and port but redacts paths and query strings.

These helpers are added for later OAuth, OpenID Connect, and Client ID Metadata Document work. This PR does not make any outbound requests itself.

## Usage examples

Operator-configured endpoints use the less restrictive wrapper:

```rust
let issuer =
    ConfiguredEndpointUrl::parse("https://login.example.com/tenant")?;
```

A client-supplied metadata URL uses the public-only policy and bounded client:

[...]

PR #1641, opened 2026-07-10
feat(app): complete upstream OIDC callbacks

sauldhernandez:
## Stack context

**This part of the stack:** PR #1640 starts the browser login and saves its state, this PR completes the provider callback, and PR #1642 tests the callback's failure handling before token issuance is added.

**This PR:** Adds the provider callback route and converts a verified external login into a single-use Coral authorization code.

## Intent

A provider callback is not proof of identity by itself. Coral must match it to the login it started, exchange the provider code, verify the returned identity, and preserve the original client's security bindings.

## What it enables

- Adds `GET /auth/oidc/callback`. Coral authenticates against exactly one configured provider, so the path is fixed rather than provider-scoped.
- Pins `auth.provider.redirect_uri` to that path during configuration validation, so a redirect URI the config accepts is always one this server answers rather than a startup-clean setting that strands every login on a 404.
- Requires and consumes the stored callback state before making provider requests, preventing reuse.
- Exchanges the provider code using the stored PKCE verifier and validates the ID token using the stored nonce.
- Creates a random, one-time Coral authorization code bound to the user, client ID, exact redirect URI, client PKCE challenge, and resource.
- Redirects only to the callback saved in the original request and preserves the client's state.
- Returns fixed OAuth errors with no-cache and no-referrer headers instead of exposing provider or token details.

The authorization code is created here; PR #1643 adds the endpoint that redeems it for a Coral session token.

## Usage examples

The configured provider redirects the browser to:

```text
/auth/oidc/callback?state=<saved state>&code=<provider code>
```

After successful exchange and identity validation, Coral redirects the browser to the previously approved client callback:

```text
https://client.example/oauth/callback?code=<one-time Coral code>&state=<client state>
```

If the user denies access or validation fails, the same approved callback receives a sanitized OAuth error and no usable code.

### DIFF
--- a/crates/coral-app/src/auth/authorization_server/callback.rs
+++ b/crates/coral-app/src/auth/authorization_server/callback.rs
@@ -1,21 +1,15 @@
-use std::collections::BTreeSet;
-
 use axum::extract::{RawQuery, State};
-use axum::http::{StatusCode, header};
-use axum::response::{IntoResponse, Response};
+use axum::response::Response;
 use base64::Engine as _;
 use base64::engine::general_purpose::URL_SAFE_NO_PAD;
 use ring::rand::{SecureRandom as _, SystemRandom};
-use url::{Url, form_urlencoded};
+use url::Url;
 use zeroize::Zeroizing;
 
 use super::super::state_store::{OAuthAuthorizationCodeRecord, OAuthAuthorizationSessionRecord};
 use super::AuthorizationServerHttpState;
-
-const MAX_QUERY_BYTES: usize = 8 * 1024;
-const MAX_PARAMETERS: usize = 32;
-const MAX_PARAMETER_NAME_BYTES: usize = 64;
-const MAX_PARAMETER_VALUE_BYTES: usize = 2 * 1024;
+use super::query;
+use super::response::{TrustedRedirect, direct_error};
 
 pub(super) async fn oidc_callback(
     State(state): State<AuthorizationServerHttpState>,
@@ -37,7 +31,7 @@ pub(super) async fn oidc_callback(
             "OIDC callback state is invalid or expired",
         );
     };
-    let Some(trusted) = TrustedRedirect::new(&session) else {
+    let Some(trusted) = trusted_redirect(&session) else {
         return direct_error("server_error", "authorization failed");
     };
     let provider_code = match classify_result(&query) {
@@ -53,21 +47,30 @@ pub(super) async fn oidc_callback(
         }
     };
     let provider = state.settings.provider();
-    let Ok(exchange) = state
+    let exchange = match state
         .provider_client
         .exchange_code(provider, &provider_code, &session.oidc_code_verifier)
         .await
-    else {
-        return trusted.error("server_error", "authorization failed");
+    {
+        Ok(exchange) => exchange,
+        Err(error) => {
+            tracing::warn!(%error, "OIDC callback could not exchange the provider code");
+            return trusted.error("server_error", "authorization failed");
+        }
     };
-    let Ok(identity) = state
+    let identity = match state
         .provider_client
         .validate_code_exchange(provider, exchange, &session.oidc_nonce)
         .await
-    else {
-        return trusted.error("server_error", "authorization failed");
+    {
+        Ok(identity) => identity,
+        Err(error) => {
+            tracing::warn!(%error, "OIDC callback could not validate the provider identity");
+            return trusted.error("server_error", "authorization failed");
+        }
     };
     let Ok(authorization_code) = random_code() else {
+        tracing::warn!("OIDC callback could not generate an authorization code");
         return trusted.error("server_error", "authorization failed");
     };
     let authorization = OAuthAuthorizationCodeRecord {
@@ -77,12 +80,12 @@ pub(super) async fn oidc_callback(
         code_challenge: session.code_challenge,
         resource: session.resource,
     };
-    if state
+    if let Err(error) = state
         .state_store
         .store_authorization_code(&authorization_code, authorization)
         .await
-        .is_err()
     {
+        tracing::warn!(%error, "OIDC callback could not store the authorization code");
         return trusted.error("server_error", "authorization failed");
     }
     trusted.success(&authorization_code)
@@ -97,31 +100,17 @@ struct CallbackQuery {
 }
 
 fn parse_query(raw: &str) -> Result<CallbackQuery, ()> {
-    if raw.len() > MAX_QUERY_BYTES {
-        return Err(());
-    }
     let mut query = CallbackQuery::default();
-    let mut seen = BTreeSet::new();
-    for (index, (name, value)) in form_urlencoded::parse(raw.as_bytes()).enumerate() {
-        if index >= MAX_PARAMETERS
-            || name.len() > MAX_PARAMETER_NAME_BYTES
-            || value.len() > MAX_PARAMETER_VALUE_BYTES
-        {
-            return Err(());
-        }
-        let name = name.into_owned();
-        if !seen.insert(name.clone()) {
-            return Err(());
-        }
-        let target = match name.as_str() {
+    query::scan(raw, |name, value| {
+        let target = match name {
             "state" => &mut query.state,
             "code" => &mut query.code,
             "error" => &mut query.error,
             "error_description" => &mut query.error_description,
-            _ => continue,
+            _ => return,
         };
         *target = Some(value.into_owned());
-    }
+    })?;
     Ok(query)
 }
 
@@ -150,73 +139,20 @@ fn random_code() -> Result<String, ()> {
     Ok(URL_SAFE_NO_PAD.encode(bytes.as_slice()))
 }
 
-struct TrustedRedirect {
-    url: Url,
-    client_state: Option<String>,
-}
-
-impl TrustedRedirect {
-    fn new(session: &OAuthAuthorizationSessionRecord) -> Option<Self> {
-        Some(Self {
-            url: Url::parse(&session.redirect_uri).ok()?,
-            client_state: session.client_state.clone(),
-        })
-    }
-
-    fn success(&self, code: &str) -> Response {
-        self.redirect("code", code, None)
-    }
-
-    fn error(&self, error: &'static str, description: &'static str) -> Response {
-        self.redirect("error", error, Some(description))
-    }
-
-    fn redirect(&self, key: &str, value: &str, description: Option<&str>) -> Response {
-        let mut url = self.url.clone();
-        let mut query = url.query_pairs_mut();
-        query.append_pair(key, value);
-        if let Some(description) = description {
-            query.append_pair("error_description", description);
-        }
-        if let Some(state) = &self.client_state {
-            query.append_pair("state", state);
-        }
-        drop(query);
-        redirect(url.as_str())
-    }
-}
-
-fn direct_error(error: &'static str, description: &'static str) -> Response {
-    let body = serde_json::json!({
-        "error": error,
-        "error_description": description,
-    })
-    .to_string();
-    (
-        StatusCode::BAD_REQUEST,
-        security_headers(),
-        [(header::CONTENT_TYPE, "application/json")],
-        body,
-    )
-        .into_response()
-}
-
-fn redirect(location: &str) -> Response {
-    (
-        StatusCode::FOUND,
-        security_headers(),
-        [(header::LOCATION, location)],
-        "",
-    )
-        .into_response()
-}
-
-fn security_headers() -> [(header::HeaderName, &'static str); 3] {
-    [
-        (header::CACHE_CONTROL, "no-store"),
-        (header::PRAGMA, "no-cache"),
-        (header::REFERRER_POLICY, "no-referrer"),
-    ]
+/// Rebuilds the client callback this login was started for.
+///
+/// The `None` arm is unreachable. `redirect_uri` is only ever stored after
+/// `authorize`'s `registered_redirect` matched it against a registered URI and
+/// parsed that URI under
+/// [`BrowserRedirect`](crate::outbound_url_policy::BrowserRedirect), so a value
+/// that reaches here has already parsed once. It is checked rather than
+/// unwrapped because the store sits between the two handlers, and a panic in a
+/// callback would be a worse answer than a failed login. Re-applying
+/// `BrowserRedirect` here would not add a check — it would only move where the
+/// same policy was applied.
+fn trusted_redirect(session: &OAuthAuthorizationSessionRecord) -> Option<TrustedRedirect> {
+    let url = Url::parse(&session.redirect_uri).ok()?;
+    Some(TrustedRedirect::new(url, session.client_state.clone()))
 }
 
 #[cfg(test)]
@@ -226,10 +162,12 @@ mod tests {
     use std::sync::Arc;
     use std::time::Duration;
 
+    use axum::http::{StatusCode, header};
     use base64::engine::general_purpose::STANDARD;
     use ring::rand::SystemRandom;
     use ring::signature::{ECDSA_P256_SHA256_FIXED_SIGNING, EcdsaKeyPair};
     use serde_json::{Value, json};
+    use url::form_urlencoded;
     use wiremock::matchers::{method, path};
     use wiremock::{Mock, MockServer, ResponseTemplate};
 
@@ -392,15 +330,7 @@ mod tests {
 
     #[tokio::test]
     async fn malformed_queries_are_bounded_and_do_not_consume_state() {
-        for raw in [
-            "x".repeat(MAX_QUERY_BYTES + 1),
-            format!("{}=x", "n".repeat(MAX_PARAMETER_NAME_BYTES + 1)),
-            format!("x={}", "v".repeat(MAX_PARAMETER_VALUE_BYTES + 1)),
-            (0..=MAX_PARAMETERS)
-                .map(|index| format!("x{index}=1"))
-                .collect::<Vec<_>>()
-                .join("&"),
-        ] {
+        for raw in query::rejected_queries() {
             let Err(()) = parse_query(&raw) else {
                 panic!("malformed callback query must be rejected");
             };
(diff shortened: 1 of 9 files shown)


sauldhernandez:
* **#1641** <a href="https://app.graphite.com/github/pr/withcoral/coral/1641?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1641?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-app/src/auth/authorization_server/callback.rs
```diff
@@ -0,0 +1,547 @@
...
+            return trusted.error("access_denied", "authorization was denied");
+        }
+        Ok(CallbackResult::ProviderError) => {
+            return trusted.error("server_error", "authorization failed");
+        }
+        Err(()) => {
+            return trusted.error("invalid_request", "OIDC callback response is invalid");
+        }
+    };
+    let provider = state.settings.provider();
```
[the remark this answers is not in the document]
sauldhernandez: Both contracts quoted here were errors in the PR description, not in the implementation — the description has since been corrected.

- **Callback path.** The route is `/auth/oidc/callback`, never `/auth/oidc/{provider}/callback`. Coral authenticates against exactly one configured provider, so there is no provider segment to scope by. Router registration and configuration validation now read a single constant, `auth::OIDC_CALLBACK_PATH`, and validation pins `auth.provider.redirect_uri` to that path — which is also the fix for the 404 flagged in the sibling finding.
- **Scope.** Authorization codes are not scope-bound and never were. `/oauth/authorize` rejects a `scope` parameter with `invalid_scope`, and `OAuthAuthorizationCodeRecord` has no scope field. The description no longer claims otherwise.

You also caught one binding that had been missed: the description listed the code as bound to "the provider", which `OAuthAuthorizationCodeRecord` does not carry either. That is now removed, so the documented bindings are exactly the record's fields — `user_id`, `client_id`, `redirect_uri`, `code_challenge`, and `resource`.

No code change was needed for this finding; the mismatch was entirely in the description.

crates/coral-app/src/auth/authorization_server/callback.rs
```diff
@@ -0,0 +1,540 @@
...
+use base64::Engine as _;
+use base64::engine::general_purpose::URL_SAFE_NO_PAD;
+use ring::rand::{SecureRandom as _, SystemRandom};
+use url::{Url, form_urlencoded};
+use zeroize::Zeroizing;
+
+use super::super::state_store::{OAuthAuthorizationCodeRecord, OAuthAuthorizationSessionRecord};
+use super::AuthorizationServerHttpState;
+
+const MAX_QUERY_BYTES: usize = 8 * 1024;
```
simonwhitaker: **~100 lines are now duplicated with `authorize.rs`, and they have already drifted.** These four constants, `parse_query`'s structure, `TrustedRedirect`, `direct_error`, and `redirect` are all near-copies of the siblings in `authorize.rs`.

The drift is the tell: `callback.rs::direct_error` sets `referrer-policy: no-referrer` via the new `security_headers()` helper, but `authorize.rs::direct_error` does not. That divergence is harmless, which is exactly why it's worth acting on now — the next one might not be.

Suggest hoisting a shared `query.rs` + `response.rs` into `authorization_server/` and letting `authorize.rs` pick up the missing header. Cheaper now than after the token endpoint in #1643 copies the pair a third time.

crates/coral-app/src/auth/authorization_server/callback.rs
```diff
@@ -0,0 +1,540 @@
...
+        }
+        Ok(CallbackResult::ProviderError) => {
+            return trusted.error("server_error", "authorization failed");
+        }
+        Err(()) => {
+            return trusted.error("invalid_request", "OIDC callback response is invalid");
+        }
+    };
+    let provider = state.settings.provider();
+    let Ok(exchange) = state
```
simonwhitaker: **Every failure path here discards a typed error and logs nothing.** This `let Ok(…) else`, the `validate_code_exchange` one below it, and the `store_authorization_code(…).is_err()` check all drop information that was built to be recorded:

- `OidcProviderClientError` distinguishes `Discovery` / `IssuerMismatch` / `InvalidEndpoint` / `UnsupportedTokenEndpointAuth` / `TokenExchange`, and `provider_client`'s module doc explicitly says errors "name the stage that failed and nothing else" — i.e. they are safe to log.
- `StateStoreError` distinguishes `CapacityExceeded` from `ExpiryOverflow`.

As written, an operator debugging a broken IdP sees `error=server_error` in the browser and nothing at all server-side. There are zero `tracing::` calls anywhere under `crates/coral-app/src/auth/`, so this is a stack-wide gap rather than a regression introduced here — but the callback is where it costs most, since network, crypto, and store failures all converge on this one handler.

Suggest `tracing::warn!(%error, "…")` on the three `server_error` paths.

crates/coral-app/src/auth/authorization_server/callback.rs
```diff
@@ -0,0 +1,540 @@
...
+    Ok(query)
+}
+
+enum CallbackResult {
+    Code(String),
+    AccessDenied,
+    ProviderError,
+}
+
+fn classify_result(query: &CallbackQuery) -> Result<CallbackResult, ()> {
```
simonwhitaker, line 123: Minor: the `error_description.is_some()` discriminant means `code=…` accompanied by an `error_description` (and no `error`) is classified `Err(())` → `invalid_request`, *after* the state has already been consumed — so the login fails unrecoverably and the user must restart.

Defensible strictness, and no spec-compliant provider does this. But it's a nonstandard-IdP failure mode with no diagnostic attached. Consider ignoring `error_description` when `code` is present and `error` is absent.

crates/coral-app/src/auth/authorization_server/callback.rs
```diff
@@ -0,0 +1,540 @@
...
+        query.error_description.is_some(),
+    ) {
+        (Some(code), None, false) if !code.is_empty() => Ok(CallbackResult::Code(code.into())),
+        (None, Some("access_denied"), _) => Ok(CallbackResult::AccessDenied),
+        (None, Some(error), _) if !error.is_empty() => Ok(CallbackResult::ProviderError),
+        _ => Err(()),
+    }
+}
+
+fn random_code() -> Result<String, ()> {
```
simonwhitaker, line 136: Nit: this zeroizes the 32 raw bytes but returns a plain `String`, and the base64 form is the actual secret. Since the code then travels in a URL and lands in browser history, and the store keys it by SHA-256 rather than retaining the plaintext, this is cosmetic — but `Zeroizing<String>` would match the stated intent of the `Zeroizing` on the byte array.

crates/coral-app/src/auth/authorization_server/callback.rs
```diff
@@ -0,0 +1,540 @@
...
+    Ok(URL_SAFE_NO_PAD.encode(bytes.as_slice()))
+}
+
+struct TrustedRedirect {
+    url: Url,
+    client_state: Option<String>,
+}
+
+impl TrustedRedirect {
+    fn new(session: &OAuthAuthorizationSessionRecord) -> Option<Self> {
```
simonwhitaker: This `None` branch is unreachable, and it would help to say so. `session.redirect_uri` is only ever stored after `authorize.rs::registered_redirect` string-matched it against a registered URI that had already parsed under `EndpointUrl::<BrowserRedirect>` — so `Url::parse` cannot fail here.

Fine as defense-in-depth, but as written it reads like the callback relies on bare `Url::parse` where authorize relies on a policy type. A one-line comment — that the value arrives already policy-checked and this is belt-and-braces — would stop the next reader from either writing a test for the dead branch or "fixing" it by re-applying `BrowserRedirect`.

crates/coral-app/src/auth/authorization_server/callback.rs
```diff
@@ -0,0 +1,540 @@
...
+
+    fn redirect(&self, key: &str, value: &str, description: Option<&str>) -> Response {
+        let mut url = self.url.clone();
+        let mut query = url.query_pairs_mut();
+        query.append_pair(key, value);
+        if let Some(description) = description {
+            query.append_pair("error_description", description);
+        }
+        if let Some(state) = &self.client_state {
+            query.append_pair("state", state);
```
simonwhitaker: **A client `redirect_uri` carrying a reserved query param yields a duplicate here.** `config.rs` in this same PR reasons carefully about exactly this collision for the *provider* redirect URI ("the provider appends its own `state`, and a configured one collides with it") — but there is no equivalent guard on the *client* side.

`registered_redirect` validates under `BrowserRedirect`, whose `check` only calls `check_https_or_explicit_loopback` — it permits any query. So a client registering `?state=x` (or `?code=x`) gets two `state`/`code` params via `append_pair`, and a client reading the first occurrence gets the attacker-or-operator-supplied value instead of ours. Not an auth bypass, but a confusion vector.

Unreachable today, since `registered_clients` is always an empty `BTreeMap` — it becomes reachable in the CIMD/registration PR, which is where rejecting reserved params belongs. Note the success test's `count == 1` assertion passes only because the fixture uses `tenant`; a `?state=` fixture would fail it. `authorize.rs::TrustedRedirect::error` has the same gap.

crates/coral-app/src/auth/config.rs
```diff
@@ -368,6 +369,19 @@ impl OidcProviderSettings {
                 served_origin.ascii_serialization()
             )));
         }
+        // The origin alone does not make the URI reachable: the router serves
+        // one callback path, so any other path is accepted here and then 404s
+        // once the provider redirects the browser to it. A query is rejected
+        // for the same reason — the provider appends its own `state`, and a
+        // configured one collides with it and fails the callback.
+        if redirect_uri.as_url().path() != OIDC_CALLBACK_PATH {
```
simonwhitaker, line 379: **This pin breaks a path-prefixed reverse proxy.** `CoralAuthorizationServer::start`'s doc says the server is intended to run "on loopback or behind a TLS-terminating reverse proxy." If that proxy mounts Coral under a path prefix, then `auth.authorization_server.issuer = https://example.com/coral` passes the origin check just above, but `https://example.com/coral/auth/oidc/callback` fails this exact-path check at startup — and the error message then tells the operator to use `/auth/oidc/callback`, which their proxy won't route.

Comparing against the issuer's path joined with `OIDC_CALLBACK_PATH` would handle both the prefixed and unprefixed deployments. Since nothing documents `auth.provider.redirect_uri` yet, this isn't a break for existing users — worth settling before it is, either with the join or with an explicit "no path prefix" constraint in the doc comment.

crates/coral-app/src/auth/mod.rs
```diff
@@ -21,6 +21,15 @@ pub use authorization_server::{CoralAuthorizationServer, RunningCoralAuthorizati
...
 pub use error::AuthServerError;
 
+/// Path the authorization server answers the upstream provider's callback on.
+///
+/// The router registers exactly this path, and configuration validation pins
+/// `auth.provider.redirect_uri` to it. Both sides read the constant so an
+/// accepted redirect URI is always one this server actually serves: without the
+/// pin, a redirect URI on the right origin but a different path passes startup
+/// and then strands every login on a 404 the provider reports, not Coral.
+pub(crate) const OIDC_CALLBACK_PATH: &str = "/auth/oidc/callback";
```
simonwhitaker: Nit: only `auth::config` and `auth::authorization_server` read this, so `pub(in crate::auth)` would match the tighter visibility used elsewhere in the module (`ResolvedOidcProvider` is `pub(super)`, `id_token::tests` is `pub(in crate::auth)`).

simonwhitaker (approved):
Approving. The security shape is right: the login session is consumed once under the store mutex before any provider I/O, no provider-supplied data reaches a response, error descriptions are static, and every path sets `no-store`/`no-cache`/`no-referrer`. Unknown state correctly yields a direct 400 rather than a redirect to an unvalidated URI.

Verified locally on `9b17557`: `cargo test -p coral-app --lib auth::` → 87 passed, 0 failed (includes the 3 new callback tests); `cargo clippy -p coral-app --lib --all-targets` → clean.

The tests are the strongest part of this PR. The concurrent double-callback via `tokio::join!` is a real race test and is deterministic because `take_authorization_session` removes under the mutex. `%73tate=duplicate` exercises duplicate-key detection *after* percent-decoding — the case a naive `contains("state=")` check would miss. And asserting that `VERIFIER`/`PROVIDER_SECRET`/`PROVIDER_CODE` never appear in the `Location` header pins down the leak-prevention claim directly rather than by inspection.

Inline comments below. None block merge. Two are worth doing before the next handler in the stack lands (the shared-helper extraction and the missing diagnostics), and two are decisions rather than defects (the reverse-proxy path constraint, and reserved query params in client redirect URIs — deferring the latter to the CIMD/registration PR is a fine answer).

crates/coral-app/src/auth/authorization_server/callback.rs
sauldhernandez: Keeping the strictness — recording the reasoning here rather than loosening it.

The discriminant is over a response that contradicts itself: it carries `code` (this succeeded) alongside `error_description` (here is why something failed). There is no reading of that response which is safe to act on. Treating it as success means exchanging a provider code that arrived together with the provider's own account of something going wrong, and deciding whether an external login is trustworthy is the entire job of this handler. `invalid_request` is the honest answer to a response that does not say what it means.

On the unrecoverability: that isn't a cost of this particular classification. `take_authorization_session` consumes the state before any provider I/O, deliberately, so a callback cannot be replayed — the concurrent-callback test pins exactly that ordering. Every failure downstream of that point needs a fresh login, `access_denied` and `server_error` included. This path isn't special in what it costs the user; it's only rarer.

And agreed that no spec-compliant provider produces it, which is what keeps the real-world cost at zero.

crates/coral-app/src/auth/authorization_server/callback.rs
sauldhernandez: Agreed it's cosmetic — and I'd rather not, because `Zeroizing<String>` here would advertise a protection it cannot deliver.

By the time that binding drops, the base64 string has been appended to a URL by `TrustedRedirect::success` and handed to the browser in a `Location` header. That copy — and the browser-history entry behind it — is the one that matters, and neither is reachable from this function. Zeroizing the single copy we can reach, while the copies that actually carry risk outlive it, is the sort of gesture that gets read as a guarantee by the next person to touch the file.

The `Zeroizing` on the byte array is doing real work by contrast: those 32 bytes are raw RNG output, nothing else ever sees them, and the base64 encoding is the only thing derived from them. So the asymmetry is deliberate — the array is protected because it can be, and the string is not because it can't be. I'd rather keep it visible than smooth it over.

(You're right that the store isn't a third copy: `store_authorization_code` keys by `secret_hash`, so only the SHA-256 is retained.)

crates/coral-app/src/auth/config.rs
sauldhernandez: The pin is consistent here rather than the constraint — the prefix is ruled out a step earlier, at the issuer.

`auth.authorization_server.issuer` is validated with `root_only`, so `https://example.com/coral` is rejected at startup with *"must mount at the root"*, and that check runs (`RawAuthSettings::validate`) before the provider's `redirect_uri` is looked at. So the prefixed issuer in the scenario never loads; the exact-path pin isn't what the operator hits first.

That's not incidental — it's the server's shape. Metadata advertises every endpoint relative to the issuer (`{issuer}/oauth/authorize`, `{issuer}/oauth/token`), the router mounts them at the origin root, and the callback is pinned to a root path; a root issuer is what keeps the three agreeing. Joining `issuer.path()` onto the callback path would only mean something if the issuer could carry a path, and it can't without relaxing `root_only` — which then pulls in RFC 8414 well-known insertion and the router mount points. That's "support prefix-mounted deployment," a feature rather than a one-line change, and not this PR.

So an origin-only proxy (`https://auth.example.com/` → Coral) works as intended; a prefix-mounting one is unsupported, and it's the issuer rule that says so, not the redirect_uri pin. The gap you're pointing at is real but it's a documentation one — `start()` said "behind a TLS-terminating reverse proxy" without noting the proxy has to forward at the origin root. I've tightened that doc (`f3d5ec4`). Thanks for the push on it.

crates/coral-app/src/auth/authorization_server/response.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// Both strings are `&'static str` so an error can only ever say what this
+    /// crate wrote, never what a provider or a store reported.
+    pub(super) fn error(&self, error: &'static str, description: &'static str) -> Response {
+        self.redirect("error", error, Some(description))
+    }
+
+    fn redirect(&self, key: &str, value: &str, description: Option<&str>) -> Response {
+        let mut url = self.url.clone();
+        let mut query = url.query_pairs_mut();
+        query.append_pair(key, value);
```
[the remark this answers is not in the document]
sauldhernandez: Deferred deliberately — and documented at the point in question. `TrustedRedirect::redirect` appends rather than replaces, so a registered callback URI that itself carried `code`/`state`/`error`/`error_description` would emit a duplicate of that key.

It is unreachable in this PR: `registered_clients` is an empty `BTreeMap`, so no such URI can be registered yet. The `TrustedRedirect` doc records this as a "Known gap," and the fix belongs at registration — rejecting reserved query keys where a redirect URI is first accepted — which lands with the client-metadata-document (CIMD) work later in this stack. Stripping or rewriting response keys at send time here would be the wrong layer, so there is no change in the redirect builder.

Resolving as tracked-and-deferred.

crates/coral-app/src/auth/authorization_server/callback.rs
sauldhernandez: Agreed, and deferring to the CIMD/registration PR as you called out — that's where a reserved-key reject belongs, at the point the redirect URI is first accepted.

This is now recorded in the code too: the `TrustedRedirect` doc carries it as a "Known gap," noting the append-not-replace behavior, that it's unreachable while `registered_clients` is empty, and that the check belongs at registration. Resolving here on that basis.

sauldhernandez:
### Merge activity

* **Jul 31, 8:06 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1641).
* **Jul 31, 8:08 AM UTC**: [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1641) rebased this pull request as part of a merge.
* **Jul 31, 8:14 AM UTC**: @sauldhernandez merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1641).

PR #1642, opened 2026-07-10
test(app): harden OAuth callback failures

sauldhernandez:
## Stack context

**Whole stack:** This stack lets a long-running Coral server authenticate users through an external OpenID Connect provider and protect its gRPC and MCP HTTP APIs.

**This part of the stack:** PR #1641 implements the provider callback, this PR exercises its most sensitive failure cases, and PR #1643 adds authorization-code redemption and session-token issuance.

**This PR:** Adds adversarial callback tests without changing production behavior.

## Intent

The callback handles provider codes, ID tokens, signing keys, stored PKCE values, and newly issued Coral codes. Its failure paths need direct tests proving that none of those values escape and that a failed login cannot accidentally produce a usable code.

## What it enables

The new tests prove that:

- a failed provider token exchange;
- a failed public-key fetch;
- an ID-token nonce mismatch; and
- a failure while storing the new Coral authorization code

all produce the same sanitized `server_error`, include no client authorization code, retain the required no-cache and no-referrer headers, and do not expose provider response bodies or secrets.

Production callback logic is unchanged by this PR.

## Usage examples

If a provider's token or signing-key endpoint returns an internal error body, the client sees only:

```text
error=server_error
```

If the stored nonce is wrong or Coral cannot save the one-time authorization code, the client receives the same fixed error and no `code` parameter.

PR #1643, opened 2026-07-10
feat(app): issue OAuth session tokens

sauldhernandez:
## Stack context

**This part of the stack:** PR #1642 locks down callback failures, this PR redeems the resulting Coral authorization code for a session token, and PR #1645 requires that token on Coral's served APIs.

**This PR:** Adds `POST /oauth/token` and completes Coral's authorization-code flow.

## Intent

After a successful browser login, the client still needs to prove that it started the request and exchange the short-lived Coral code for a token that Coral's APIs can validate.

## What it enables

- Accepts bounded `application/x-www-form-urlencoded` token requests and rejects duplicate or malformed fields.
- Supports only `grant_type=authorization_code`.
- Rejects client secrets and HTTP client authentication because these clients use PKCE instead.
- Requires a `resource` indicator, and rejects a request whose resource is missing, malformed, or different from the one the code was issued for.
- Recomputes the PKCE challenge and atomically consumes a code bound to the exact client ID, redirect URI, challenge, and resource.
- Prevents replay and concurrent double redemption.
- Issues a signed Bearer session token audienced to that resource with the configured session TTL, and returns its remaining lifetime.
- Marks every response, including the token response, `Cache-Control: no-store`, and uses fixed OAuth error responses that expose no codes, verifiers, or token details.

## Usage examples

After the browser callback returns a Coral code, the OAuth client exchanges it with:

```shell
curl -X POST https://auth.example/oauth/token \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  --data-urlencode 'grant_type=authorization_code' \
  --data-urlencode 'code=<Coral authorization code>' \
  --data-urlencode 'client_id=https://client.example/oauth-client.json' \
  --data-urlencode 'redirect_uri=https://client.example/oauth/callback' \
  --data-urlencode 'code_verifier=<original PKCE verifier>' \
  --data-urlencode 'resource=https://api.example/mcp'
```

A successful response contains `access_token`, `token_type: "Bearer"`, and `expires_in`. Coral grants no scopes: a `scope` parameter is rejected with `invalid_scope`, and the token names the resource as its audience instead.

### DIFF
--- a/crates/coral-app/src/auth/authorization_server/token.rs
+++ b/crates/coral-app/src/auth/authorization_server/token.rs
@@ -80,7 +80,6 @@ pub(super) async fn oauth_token(
         Err(_error) => return token_error(TokenError::ServerError),
     };
     let issued = match state.session_tokens.issue_access_token(
-        &authorization.provider_id,
         &authorization.user_id,
         &authorization.client_id,
         &authorization.resource,
@@ -298,7 +297,6 @@ mod tests {
     use url::form_urlencoded;
 
     use super::*;
-    use crate::auth::PROVIDER_ID;
     use crate::auth::config::AuthSettings;
     use crate::auth::provider_client::OidcProviderClient;
     use crate::auth::session::SessionTokenIssuer;
@@ -334,7 +332,7 @@ type = "oidc"
 issuer = "https://accounts.example.test"
 client_id = "upstream-client"
 client_secret = "test-secret"
-redirect_uri = "http://127.0.0.1:14554/auth/oidc/alpha/callback"
+redirect_uri = "{AUTH_ISSUER}/auth/oidc/callback"
 "#
         );
         let settings = AuthSettings::from_toml(&raw)
@@ -366,7 +364,6 @@ redirect_uri = "http://127.0.0.1:14554/auth/oidc/alpha/callback"
             .store_authorization_code(
                 code,
                 OAuthAuthorizationCodeRecord {
-                    provider_id: PROVIDER_ID.into(),
                     user_id: "raw/provider/subject".into(),
                     client_id: CLIENT.into(),
                     redirect_uri: REDIRECT.into(),
@@ -672,7 +669,6 @@ redirect_uri = "http://127.0.0.1:14554/auth/oidc/alpha/callback"
             .expect("valid access token");
         assert_eq!(validated.audience, RESOURCE);
         assert_eq!(validated.client_id, CLIENT);
-        assert_eq!(validated.provider, PROVIDER_ID);
         assert_eq!(validated.subject, "raw/provider/subject");
         invalid_grant(state.clone(), "success-code").await;
 
--- a/crates/coral-app/src/auth/authorization_server/token.rs
+++ b/crates/coral-app/src/auth/authorization_server/token.rs
@@ -107,16 +107,23 @@ pub(super) async fn oauth_token(
 fn client_auth_rejection(headers: &axum::http::HeaderMap) -> Option<Response> {
     let mut values = headers.get_all(header::AUTHORIZATION).iter();
     let value = values.next()?;
-    let challenge = value
-        .to_str()
-        .ok()
-        .and_then(|value| value.split_once(' '))
-        .filter(|(scheme, _credentials)| {
-            values.next().is_none() && axum::http::HeaderName::from_bytes(scheme.as_bytes()).is_ok()
-        })
-        .and_then(|(scheme, _credentials)| {
-            axum::http::HeaderValue::try_from(format!("{scheme} realm=\"Coral OAuth\"")).ok()
-        });
+    // RFC 6749 section 5.2 requires the 401 challenge to name the scheme the
+    // client authenticated with. A duplicated or malformed header names no
+    // single scheme, so it stays a plain `invalid_client` without a challenge.
+    let challenge = if values.next().is_some() {
+        None
+    } else {
+        value
+            .to_str()
+            .ok()
+            .and_then(|value| value.split_once(' '))
+            .filter(|(scheme, _credentials)| {
+                axum::http::HeaderName::from_bytes(scheme.as_bytes()).is_ok()
+            })
+            .and_then(|(scheme, _credentials)| {
+                axum::http::HeaderValue::try_from(format!("{scheme} realm=\"Coral OAuth\"")).ok()
+            })
+    };
     let mut response = token_error(TokenError::InvalidClient);
     if let Some(challenge) = challenge {
         *response.status_mut() = StatusCode::UNAUTHORIZED;
--- a/crates/coral-app/src/auth/authorization_server/token.rs
+++ b/crates/coral-app/src/auth/authorization_server/token.rs
@@ -684,19 +684,12 @@ redirect_uri = "{AUTH_ISSUER}/auth/oidc/callback"
         assert_eq!(validated.audience, RESOURCE);
         assert_eq!(validated.client_id, CLIENT);
         assert_eq!(validated.subject, "raw/provider/subject");
-        invalid_grant(state.clone(), "success-code").await;
-
-        seed(store.as_ref(), "concurrent-code", VERIFIER).await;
-        let first = redeem(state.clone(), "concurrent-code", VERIFIER);
-        let second = redeem(state, "concurrent-code", VERIFIER);
-        let (first, second) = tokio::join!(first, second);
-        assert_eq!(
-            [first.0, second.0]
-                .into_iter()
-                .filter(|status| *status == StatusCode::OK)
-                .count(),
-            1
-        );
-        assert!(first.1.contains("invalid_grant") || second.1.contains("invalid_grant"));
+        // Single use is the store's guarantee rather than this handler's:
+        // `take_authorization_code_for_request` does its get-check-remove under
+        // one lock, so two redemptions of a code cannot both observe it. This
+        // handler only turns the resulting `None` into `invalid_grant`, which
+        // is what the replay below asserts. The atomicity itself is covered by
+        // `state_store`'s `authorization_codes_are_consumed_only_after_all_bindings_match`.
+        invalid_grant(state, "success-code").await;
     }
 }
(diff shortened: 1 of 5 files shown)


sauldhernandez:
* **#1643** <a href="https://app.graphite.com/github/pr/withcoral/coral/1643?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1643?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-app/src/auth/authorization_server/token.rs
```diff
@@ -0,0 +1,695 @@
...
+use base64::engine::general_purpose::URL_SAFE_NO_PAD;
+use sha2::{Digest as _, Sha256};
+use zeroize::Zeroizing;
+
+use super::AuthorizationServerHttpState;
+
+const MAX_BODY_BYTES: usize = 8 * 1024;
+const MAX_PARAMETERS: usize = 32;
+const MAX_PARAMETER_NAME_BYTES: usize = 64;
+const MAX_PARAMETER_VALUE_BYTES: usize = 2 * 1024;
```
simonwhitaker: These four limits duplicate `query.rs`, whose module doc says exactly why they live in one place: *"a handler that gets its own copy of the limits is a handler whose limits can drift from the rest."* `MAX_BODY_BYTES` is `MAX_QUERY_BYTES`, and the other three are identical to their `query.rs` counterparts.

The fork itself looks defensible — `query::scan` yields `Cow<str>` from `form_urlencoded`, which can't be wrapped in `Zeroizing`, and this is the only handler holding a verifier — but nothing in the file says so. Two suggestions:

- Re-export the limits from `query.rs` as `pub(super) const` and use them here, so the two parsers can't drift apart even though their decoding differs.
- Add a module doc (`//!`) explaining why the token endpoint decodes its own form rather than calling `query::scan`, in the style of the `query.rs` / `response.rs` headers.

Same pattern in the tests: `authorize.rs` and `callback.rs` both extend their limit cases with `query::rejected_queries()` — which exists so a limit added in one place is exercised by every handler — while `rejects_content_type_body_and_malformed_or_unbounded_forms` hand-writes the equivalents (`"n".repeat(65)`, `"x".repeat(2049)`, 33 params). Those are the literals that will silently stop matching if a limit ever moves.

crates/coral-app/src/auth/authorization_server/token.rs
```diff
@@ -0,0 +1,695 @@
...
+}
+
+fn json_response(status: StatusCode, body: &serde_json::Value) -> Response {
+    (
+        status,
+        [
+            (header::CONTENT_TYPE, "application/json"),
+            (header::CACHE_CONTROL, "no-store"),
+            (header::PRAGMA, "no-cache"),
+            (header::REFERRER_POLICY, "no-referrer"),
```
simonwhitaker: `response.rs::security_headers()` already returns exactly these three, and its doc names the bug it was extracted to end: *"a handler cannot ship with a weaker header set than its siblings — the state this module was extracted to end, where one handler set `referrer-policy` on its direct errors and the other did not."* Re-listing them here reopens that gap.

Suggest making `security_headers()` `pub(super)` and composing it with the `CONTENT_TYPE` tuple, so a header added there reaches the token endpoint too.

crates/coral-app/src/auth/authorization_server/token.rs
```diff
@@ -0,0 +1,695 @@
...
+        &authorization.user_id,
+        &authorization.client_id,
+        &authorization.resource,
+    ) {
+        Ok(issued) => issued,
+        Err(_error) => return token_error(TokenError::ServerError),
+    };
+    let now = match SystemTime::now().duration_since(UNIX_EPOCH) {
+        Ok(duration) => duration.as_secs(),
+        Err(_error) => return token_error(TokenError::ServerError),
```
simonwhitaker: Three distinct server-side failures — state-store error, signing failure, clock error — are each discarded into a bare 500 with no log line. `callback.rs` warns on every comparable failure (`tracing::warn!(%error, "OIDC callback could not store the authorization code")` and three siblings), so this is a departure from how the rest of the authorization server reports its own faults.

It matters most at line 80: the store failure is the one case where the code may already be gone, so the user restarts the entire browser flow while the operator gets nothing to correlate against. From the client side a store outage is indistinguishable from an ordinary `invalid_grant` loop.

A `tracing::warn!(%error, ...)` at each of the three, with the response body left exactly as-is, would close this without changing what a client can observe.

crates/coral-app/src/auth/authorization_server/token.rs
```diff
@@ -0,0 +1,695 @@
...
+            .expect("valid access token");
+        assert_eq!(validated.audience, RESOURCE);
+        assert_eq!(validated.client_id, CLIENT);
+        assert_eq!(validated.subject, "raw/provider/subject");
+        invalid_grant(state.clone(), "success-code").await;
+
+        seed(store.as_ref(), "concurrent-code", VERIFIER).await;
+        let first = redeem(state.clone(), "concurrent-code", VERIFIER);
+        let second = redeem(state, "concurrent-code", VERIFIER);
+        let (first, second) = tokio::join!(first, second);
```
simonwhitaker: This doesn't exercise a real race. `#[tokio::test]` is the current-thread runtime, and `tokio::join!` polls its futures in order — nothing between the in-memory body read and the uncontended `Mutex::lock().await` inside `take_authorization_code_for_request` ever returns `Pending`, so `first` runs to completion before `second` is polled for the first time. What's being asserted is a replay, which `invalid_grant(state.clone(), "success-code")` already covers three lines up.

Either `#[tokio::test(flavor = "multi_thread", worker_threads = 2)]` (still not a guaranteed interleaving, but at least a genuine one is possible), or drop the concurrency framing and note that the atomicity guarantee lives in the store's take-under-lock rather than in the handler.

simonwhitaker (approved):
Approving — the security posture here is solid: uniform `invalid_grant` across every binding mismatch (and tested as indistinguishable), atomic single-use redemption, no secrets in any response body, `no-store` everywhere. No correctness bug found in the happy or unhappy paths.

Three non-blocking notes inline: two shared helpers get forked rather than reused (the very drift their module docs exist to prevent), three distinct server-side failures collapse to a silent 500, and the double-redemption test doesn't actually run concurrently.

crates/coral-app/src/auth/authorization_server/token.rs
sauldhernandez: Shared in `a3ab77247`, narrowed slightly from the suggestion.

The three parameter limits moved into `query.rs` as `pub(super)` and are imported here, so `parse_form` and `scan` are bounded by one set of constants. The limit test now asserts against a new `query::rejected_parameter_lists()`; `rejected_queries()` became that list plus the query-size case, so `authorize.rs` and `callback.rs` are unchanged.

`MAX_BODY_BYTES` deliberately stays local, with a doc comment saying why: a request body and a URL are bounded for different reasons and only happen to coincide at 8 KiB today, so folding it into `MAX_QUERY_BYTES` would be the wrong kind of sharing.

To check the sharing isn't cosmetic I temporarily set `MAX_PARAMETERS = 3` and re-ran: three token tests failed, confirming this endpoint's own parser really does read the shared constant — while the limit test stayed green, because its cases derive from the same value.

One thing I didn't do: the `//!` module header explaining why this endpoint decodes its own form instead of calling `query::scan`. `query.rs` now notes *that* it does, but not your `Cow<str>`/`Zeroizing` reason, which is the actual answer. Happy to add it if you want it where a reader will hit it.

crates/coral-app/src/auth/authorization_server/token.rs
sauldhernandez: Done in `a3ab77247`. `security_headers()` is `pub(super)` and this `json_response` composes it with its own `content-type`, matching `direct_error`'s shape, so the three headers live in one place again.

While verifying that, I found the invariant has one more exception — documented rather than changed: the discovery metadata handler has its own `json_response` in `authorization_server.rs` that sets only `content-type`. I think that one is correct, since RFC 8414 metadata is public and meant to be cached, so `no-store` there would work against it. But it did make "every response" too strong a claim, so `security_headers()`'s doc now says it reaches every response carrying a code, a token, or an error, and names the metadata document as the deliberate exception. Say the word if you'd rather that exception also be called out in the module header.

crates/coral-app/src/auth/authorization_server/token.rs
sauldhernandez: Added in `23771460d` — `tracing::warn!(%error, ...)` at each, with the response bodies left exactly as they were, so nothing a client can observe moved.

Four sites rather than three: alongside the store, signing, and clock failures there's the `expires_in` guard just below, which also returned a bare 500. It has no `error` to interpolate, but reaching it means the issuer minted an already-expired token — a clock or config fault worth the same visibility.

Agreed that the store case is the one that matters most; that's the one where the code is already gone and the user restarts the entire browser flow.

crates/coral-app/src/auth/authorization_server/token.rs
sauldhernandez: Taken in `f9c3c1a0e`, via the second option — because the first one wouldn't actually have worked.

`tokio::join!` drives both futures from a single task, so it is concurrent but never parallel: it only reaches the second future if the first returns `Pending`. `flavor = "multi_thread"` doesn't change that — no worker count makes two futures joined on one task interleave. A genuine race would need a `tokio::spawn` per redemption, and even then the interleaving would be possible rather than guaranteed, so it still couldn't fail deterministically.

Your diagnosis of why nothing pends was exactly right, and it holds at both awaits: the request body is a `Full<Bytes>` that is already in memory, and the uncontended `tokio::sync::Mutex` inside `take_authorization_code_for_request` acquires through the semaphore fast path.

So the block is gone, with a comment in its place recording where the guarantee really lives: single use is the store's get-check-remove under one lock, not anything this handler does. The handler only turns the resulting `None` into `invalid_grant`, which the replay assertion above already covers, and `state_store`'s `authorization_codes_are_consumed_only_after_all_bindings_match` covers the atomicity directly.

sauldhernandez:
### Merge activity

* **Aug 3, 2:59 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1643).
* **Aug 3, 3:00 PM UTC**: @sauldhernandez merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1643).

PR #1645, opened 2026-07-10
feat(app): accept session tokens on served surfaces

sauldhernandez:
## Stack context

**This part of the stack:** PR #1643 issues Coral session tokens, this PR requires and validates them on the served APIs, and PR #1647 starts and stops the OAuth authorization server alongside those listeners.

**This PR:** Requires a session token on the served surfaces, and resolves the one configuration both of them read. Starting the authorization server itself is #1647's job, so an instance configured from this slice alone validates tokens but has nothing issuing them yet. The reference documentation for this configuration surface lives in #2042, at the top of the stack, so this slice is code and code comments only.

## Intent

Issuing a token is useless unless Coral's APIs require and validate it consistently. Both served surfaces must agree on the issuer, the signing key, the accepted audience, and the user identity a token resolves to.

## What it enables

- Strictly parses a single Bearer credential, validates its signature, issuer, audience, type, key id, and expiry, and maps its subject to a stable opaque Coral user ID.
- Distinguishes public from private surfaces. A public surface has a `public_url` that identifies it as an OAuth protected resource, and that URL is the audience of the tokens minted for it — MCP HTTP is the only one today. The native gRPC API is private: it is reached through the surfaces in front of it, so it has no identifier of its own and accepts a token minted for any configured public surface, while a public surface accepts only its own. An authenticated instance therefore requires at least one public surface, or nothing could issue a token at all.
- Validates an MCP token before creating a session, then forwards it over the loopback gRPC connection so the private API validates it again.
- Canonicalizes every OAuth resource URL through one shared canonicalizer, so the audience Coral mints always matches the resource it advertises.
- Keeps composition at the call site. The gRPC bootstrap resolves configuration and starts a gRPC server; `coral serve` turns those settings into services and owns their lifecycle, so the builder never constructs a transport it does not run.
- Separates the two questions health probes ask. `grpc.health.v1.Health`'s empty service name reports process liveness from a constant, so a degraded engine never tells a liveness prober to restart the process; engine readiness moves to a `coral.readiness` service that answers from the catalog, with the answer cached briefly and shared between concurrent callers. Readiness reports on the engine's reachability, not on whether every question about it succeeds: a rejection the catalog would answer the same way on every replica — a missing default workspace, two sources claiming one runtime schema — leaves the instance ready rather than emptying the fleet. A probe bounds how long it waits, never the resolution behind it, so a slow catalog converges to ready on a later poll instead of restarting from nothing on every one.
- Keeps ephemeral servers, embedded UI operation, and loopback auth-disabled MCP HTTP working as before, and leaves health checks unauthenticated so probes need no credential.

## Usage examples

Require tokens on both surfaces:

```toml
[auth]
http_bind_addr = "127.0.0.1:14557"

[auth.authorization_server]
issuer = "https://auth.example"

[auth.session]
signing_key_file = "session.key"
access_token_ttl_seconds = 3600

[auth.provider]
issuer = "https://accounts.example"
client_id = "coral"
client_secret_env = "CORAL_OIDC_CLIENT_SECRET"
redirect_uri = "https://auth.example/auth/oidc/callback"

[server.mcp_http]
enabled = true
bind = "0.0.0.0:14556"
public_url = "https://coral.example/mcp"
```

The gRPC listener needs no identifier of its own; it is gated as soon as `[auth]` is configured:

```toml
[server]
bind_addr = "0.0.0.0:14555"
```

Clients send the issued token as `Authorization: Bearer <token>` to gRPC or MCP HTTP.

## Things to know

Any valid session token grants full access to the instance and every source configured in it — Coral does not scope access per user, per workspace, or per source today. `access_token_ttl_seconds` defaults to 30 days (`2592000`), so set it explicitly. Shorter is not automatically better: an MCP session is bound to the token that opened it, so the TTL bounds session lifetime, and a request carrying a freshly issued token under the same `Mcp-Session-Id` is answered `403` rather than the `404` that tells an MCP client to re-initialize. Coral issues no refresh tokens either, so every rotation is a full authorization round trip. Pick a value short enough to bound what a leaked token grants and long enough that clients are not reconnecting constantly.

Coral terminates no TLS, so every listener speaks cleartext and any bind outside loopback exposes that traffic: without `[auth]`, full access to the instance; with it, the bearer tokens clients send. Binding remotely is a deployment decision Coral does not gate — terminate TLS in front of the listener and keep untrusted clients off it. `coral server` warns on startup for a non-loopback bind on either served surface, gRPC or MCP HTTP.

Authentication is wired by `coral serve`, so `[auth]` is `coral server` configuration rather than something the builder reads. The builder always holds a principal provider — the local one unless a caller installs its own — so nothing downstream has to handle a server that might have no way to name its caller. Composing the session provider from the resolved `[auth]` settings is `coral serve`'s job, and that it does so is a property of that composition rather than of the server it hands a provider to: `session_authenticated_companion_forwards_bearer` drives the real serve path with `[auth]` configured and asserts the gRPC surface answers `Unauthenticated` without a token. A caller that drives `ServerBuilder` directly and installs no provider therefore gets the local principal, which is what the in-process ephemeral and embedded-UI modes rely on — they read the same config directory but are loopback listeners for the local user, who holds no session token to present.


sauldhernandez:
* **#1645** <a href="https://app.graphite.com/github/pr/withcoral/coral/1645?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1645?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-cli/src/lib.rs
```diff
@@ -793,6 +792,30 @@ fn server_endpoint_is_loopback(endpoint: &str) -> bool {
...
+/// this line on stderr is the only notice they get — which is why it fires
+/// whether or not authentication is configured. Coral terminates no TLS: without
+/// `[auth]` anyone reachable gets in, and with it the bearer tokens cross the wire
+/// in cleartext. [`grpc_exposure_warning`] words each case.
+fn server_requires_security_warning(endpoint: &str) -> bool {
+    !server_endpoint_is_loopback(endpoint)
+}
+
+/// The exposure warning for a non-loopback gRPC listener.
+fn grpc_exposure_warning(endpoint: &str, authentication_enabled: bool) -> String {
```
simonwhitaker, line 825: This warning is now the *only* mitigation for a remote bind — the PR removes `auth.allow_insecure_remote_http_bind` and its guard, so a non-loopback `auth.http_bind_addr` is no longer gated. Reframing that as an operator decision is defensible, but the notice doesn't cover what the gate used to.

Right now this fires for the gRPC bind only. An operator who binds `auth.http_bind_addr` or `server.mcp_http.bind` to `0.0.0.0` gets no gate *and* no warning — on the two listeners that carry the OAuth authorization code and the bearer tokens respectively. Given the docs' emphasis that Coral terminates no TLS, those are arguably the more dangerous of the three.

Suggest extending the startup check to all three binds. `server_requires_security_warning` is currently just `!server_endpoint_is_loopback(...)` with a long doc comment and no behaviour of its own; giving it the other two addresses would earn the indirection.

crates/coral-app/src/bootstrap/server.rs
```diff
@@ -271,19 +271,18 @@ impl ServerBuilder {
...
+    /// authorization server, and running the transports they belong to, is the
+    /// caller's job: this builder starts a gRPC server and nothing else.
     ///
     /// # Errors
     ///
-    /// Returns [`AppError`] if the config directory cannot be determined or
-    /// the MCP HTTP configuration cannot use the selected gRPC mode.
-    pub fn resolve_mcp_http_serve_config(&self) -> Result<Option<McpHttpServeConfig>, AppError> {
+    /// Returns [`AppError`] if the configuration cannot be loaded or validated.
+    pub fn serve_settings(&self) -> Result<ServeSettings, AppError> {
```
simonwhitaker, line 291: This is the highest-consequence fail-open in the change: with `[auth]` in `config.toml`, `ServerBuilder::start()` called directly serves an unauthenticated gRPC listener falling back to `LocalPrincipalProvider`. The config says "authenticate" and the server says nothing.

The PR body documents this as intended (`[auth]` is `coral server` configuration), and I confirmed the shipped path is safe — `coral server` routes through `serve::start` → `compose_session_policies`. But `start_ui_server` and `bootstrap()` both call `.start()` directly, so the invariant is "nobody has wired it wrong yet" rather than something the type system or the code enforces.

`resolved_mode` already loads the config via `LoadedServerConfig` a few lines up, so the check is nearly free: if `[auth]` parses and `self.config.principal_provider` is still the default, error. That turns a silent unauthenticated listener into a startup failure, which is the right direction for a guard whose only current protection is call-site discipline.

crates/coral-app/src/bootstrap/health.rs
```diff
@@ -1,13 +1,144 @@
...
+    pub(super) fn fixed(ready: bool) -> Self {
+        Self::new(Arc::new(move || Box::pin(std::future::ready(ready))))
+    }
+
+    /// Reads the cached answer while it is fresh, else probes once and caches.
+    ///
+    /// Exactly one probe runs at a time. Callers that queue behind it re-read the
+    /// cache first, so a burst of health traffic resolves the catalog once rather
+    /// than once per request.
+    async fn is_ready(&self) -> bool {
```
simonwhitaker, line 134: The `probing` mutex is held across `(self.probe)().await` with no deadline. If `catalog_info` wedges — which is precisely the condition this service exists to report — the health RPC hangs instead of returning `NotServing`, and every caller that queues behind it hangs too.

The MCP `/readyz` path is covered by `READINESS_TIMEOUT`, but a direct `grpc_health_probe -service=coral.readiness` is not. Most orchestrators treat a hung probe as a failure, so this degrades acceptably today — but that's a property of the prober's config, not of this code.

Wrapping the probe in `tokio::time::timeout` and treating expiry as `false` would make the answer deterministic regardless of who's asking.

(The single-flight design itself is right, and the reasoning in the field comment about unauthenticated traffic fanning out into engine work is well argued.)

crates/coral-app/src/bootstrap/health.rs
```diff
@@ -1,13 +1,144 @@
...
+        Self {
+            probe,
+            cached: Arc::new(Mutex::new(None)),
+            probing: Arc::new(AsyncMutex::new(())),
+        }
+    }
+
+    /// Reports readiness by resolving the default workspace's catalog, the same
+    /// work the auth-disabled readiness probe drives through `ListCatalog`.
+    pub(super) fn from_query_manager(queries: QueryManager) -> Self {
```
simonwhitaker, line 96: Two things about tying readiness to a full catalog resolution of the default workspace:

1. **Blast radius.** If a single misconfigured source can fail `catalog_info`, one bad source marks the whole instance unready and pulls it out of rotation. Is `catalog_info` resilient to a per-source failure? If not, it's worth deciding whether readiness should mean "the engine responds" rather than "every source resolves" — they fail very differently in production.

2. **The `false` path is untested against the real probe.** The unit tests all use `EngineReadiness::fixed(false)`, and the integration test in `serve/tests.rs` covers transport failure (gRPC shut down), not engine failure. Nothing exercises `catalog_info` actually returning `Err` through this closure.

crates/coral-app/src/auth/authorization_server.rs
```diff
@@ -325,21 +321,6 @@ mod tests {
     /// `AuthSettings` has no public constructor other than `from_toml`, so the
     /// unsafe-bind guard cannot be bypassed by building a value directly; the
     /// parsing boundary is the only place it needs to hold.
```
simonwhitaker: Leftover from the deleted `rejects_unsafe_bind_at_the_only_construction_site` test — the `#[test]` went but the doc comment stayed. It now sits above `rejects_session_tokens_resolved_from_different_settings`, which it has nothing to do with, and describes a guard that no longer exists anywhere in the codebase.

Delete lines 321-323.

apps/docs/reference/configuration.mdx
```diff
@@ -96,6 +96,96 @@ bind = "127.0.0.1:14556"
...
+  Any valid session token grants full access to the instance and every source
+  configured in it. Coral does not scope access per user, per workspace, or per
+  source today, so issuing a token is equivalent to granting complete access.
+</Warning>
+
+The signing key file must contain a PKCS#8 DER-encoded P-256 private key and is
+resolved relative to Coral's config directory. `access_token_ttl_seconds`
+defaults to 30 days (`2592000`), which is a long life for a bearer token that
+grants everything — set it explicitly to the shortest value your clients can
+re-authenticate within. A public surface accepts only tokens whose audience is its
```
simonwhitaker: This advises setting the TTL "to the shortest value your clients can re-authenticate within", but the page doesn't mention what that costs on the MCP side.

An authenticated MCP session is bound to a bearer fingerprint at creation, and its forwarded gRPC client carries that original token for the session's lifetime. So token lifetime bounds *session* lifetime: when the token expires the session dies, and when the client rotates to a fresh token `authorize_bound_session` returns `Mismatch` → **403**, not 401. Plenty of MCP clients won't re-initialize on a 403.

The binding predates this PR, but this is the PR that puts it on the served path alongside advice to shorten TTLs, so the two should be documented together. Something like: token lifetime bounds MCP session lifetime, and rotation requires a new session.

crates/coral-mcp/Cargo.toml
```diff
@@ -30,11 +30,11 @@ url.workspace = true
 uuid.workspace = true
 
 coral-api.workspace = true
+coral-app.workspace = true
```
simonwhitaker, line 33: Promoting `coral-app` from a dev-dependency to a real one pulls the entire app crate — engine, catalog, storage — into `coral-mcp` for `CanonicalOauthUrl` and `OauthUrlError`. It also inverts the natural layering, since `coral-mcp` is a transport sitting *in front of* `coral-app`.

The "one canonicalizer" property is absolutely worth having (the whole point of `oauth_resource.rs`), but it doesn't require this edge. A small `coral-oauth-url` crate, or moving the type into an existing shared crate, gets the same guarantee without the dependency inversion or the compile-time cost. Same applies to `READINESS_SERVICE_NAME`, which is a `&'static str`.

crates/coral-app/src/identity.rs
```diff
@@ -98,6 +98,39 @@ impl Principal {
...
+    /// prefixes for the same user or orphaning every existing attribution row.
+    ///
+    /// The digest is stable and collision-free, but it is not opaque against a
+    /// guesser: it is unkeyed, and subjects are low-entropy (emails, numeric
+    /// provider ids), so anyone holding this value and a candidate list can confirm a
+    /// match offline. Nothing deployment-specific enters the preimage either, so
+    /// the same subject yields the same id everywhere — two databases join on it
+    /// directly. Keying the derivation is the fix, and it has to happen before real
+    /// deployments store `federated-*` rows, since afterwards there is no backfill.
+    pub(crate) fn for_federated(subject: &str) -> Self {
```
simonwhitaker, line 136: The doc comment on this is admirably candid — unkeyed SHA-256 over a low-entropy subject, nothing deployment-specific in the preimage, so IDs are confirmable offline by anyone holding a candidate list, and two unrelated databases join on them directly.

But it also states a hard deadline with no owner: *"it has to happen before real deployments store `federated-*` rows, since afterwards there is no backfill."* Nothing in the code or the PR enforces that ordering, and the derivation being one-way with the upstream subject persisted nowhere means there's genuinely no recovery — the comment is right that changing it later means either dual-prefix acceptance forever or orphaning every attribution row.

A comment that says "this must be fixed before X happens" needs either a linked tracking issue or the fix done now. Keying it while there's zero stored data is close to free; doing it after the first real deployment is impossible.

simonwhitaker (approved):
Approving. The architecture here is right: the public/private audience split (private gRPC accepts every public surface's audience, each public surface accepts only its own) is the correct model, `CanonicalOauthUrl` collapsing three copies of the canonicalizer into one removes a whole class of silent auth breakage, and the end-to-end test minting against the *advertised* resource rather than the configured string is exactly the drift that would otherwise slip through.

I checked a few things that looked risky and they're fine:

- An empty accepted-audience list is rejected inside `validate_access_token`, so `SessionPrincipalProvider` can't be constructed into a fail-open state.
- `loopback_grpc_endpoint_uri` still rejects `::ffff:127.0.0.1`, so the IPv4-mapped case covered by the deleted `mcp_http_resolution_accepts_wildcard_and_rejects_public_grpc_binds` test is still fail-closed — just at startup rather than config resolution.
- The canonicalized `authorization_server` string does equal the `issuer` the AS metadata publishes (both trim the root path), so there's no RFC 8414 mismatch for discovery clients.

Inline comments below. Nothing blocking, but a few are cheap and restore fail-closed behaviour the PR otherwise gives up:

- **Worth doing before merge:** the orphaned doc comment (trivial delete), and the `ServerBuilder::start()` fail-open guard.
- **Worth a follow-up:** readiness probe timeout, warning coverage for the other two binds, the TTL/session-binding doc note.
- **Worth a tracked issue:** the `Principal::for_federated` keying, on the code's own reasoning that there's no backfill afterwards.

crates/coral-app/src/auth/authorization_server.rs
sauldhernandez: Deleted in 4e6c0f384. Removed 321-324 rather than the 321-323 you named: taking only the three `///` lines leaves the blank lines on either side adjacent, and rustfmt collapses consecutive blanks, so that patch would have failed `cargo fmt --check`.

You called it a leftover, but it is worse than dead scaffolding — the guard it describes was removed deliberately in b3b5c363a as part of this PR, so the comment was asserting a security property at a boundary this PR had just decided not to enforce.

crates/coral-app/src/bootstrap/server.rs
sauldhernandez: Fixed in 7a9ea715d, with one deliberate change from what you wrote — worth flagging, because the literal version breaks the CLI.

Your wording was "if `[auth]` parses and `self.config.principal_provider` is still the default, error", unscoped, naming `bootstrap()` and `start_ui_server` as the callers to catch. But those two read the *same* config directory as `coral server` — all three build the layout from `AppEnvironment::discover().app_state_layout(config_dir)` with `config_dir == None`. So on a host configured for `coral server`, `coral query ...` would go `bootstrap()` → `ServerBuilder::new()` (EphemeralGrpc, `127.0.0.1:0`) → guard fires → every CLI command fails to start. Neither of those is a fail-open: both are in-process loopback listeners for the local user, who holds no session token to present.

The guard is therefore scoped to standalone mode, matching how the adjacent `reject_removed_auth` is already scoped in the same match. `local_server_modes_resolve_with_configured_auth` and `ephemeral_grpc_starts_with_configured_auth` pin that.

Two other notes. It gates *both* standalone arms, not just the `Explicit(StandaloneGrpc)` one your line numbers pointed at — `coral server` actually reaches startup through `ConfiguredStandaloneGrpc` (`bootstrap.rs:87`), so gating only the explicit arm would have missed the production path. And `Arc<dyn PrincipalProvider>` has no equality, so "still the default" is not directly testable; the field became `Option<Arc<dyn PrincipalProvider>>` with the fallback applied at startup.

On severity, for the record: I could not find a reachable path at this commit. `standalone_grpc()` has no production callers, every `.start()` caller is either composed or loopback-only, and both crates are `publish = false`. This is hardening of an API invariant rather than a live hole — which does not argue against fixing it, and the asymmetry you are really pointing at is that MCP HTTP already had this guard in `MissingSessionProvider` while gRPC did not.

crates/coral-app/src/bootstrap/health.rs
sauldhernandez: Fixed in 3d34de65f, though not by wrapping the probe — that turns out to make things worse, and it took an adversarial pass to catch it.

A `tokio::time::timeout` around `(self.probe)()` cancels the resolution to answer on time. Nothing on that path caches partial progress and `CoralQuery::prepare` builds a fresh runtime per call, so an engine whose catalog honestly resolves slower than the deadline would have its work thrown away and restarted from nothing on every poll, and could **never** report ready. That is strictly worse than today, where a caller patient enough to finish one probe at least warms the cache for everyone after it. Picking a value below coral-mcp's `READINESS_TIMEOUT` does not avoid this; it just moves the cliff from 2s to 1.5s.

So the deadline now bounds the *answer* and not the resolution. The probe runs in a spawned task that survives both the deadline and a caller disconnecting, and records what it finds — a slow engine converges to ready on a later poll. `a_slow_engine_reports_ready_on_a_later_poll` is the regression test; it fails against the cancel-the-probe version.

That task is also what serializes probes now, so the `probing` mutex is gone. Better in the wedged case: a permanently stuck engine holds one resolution rather than accumulating one per poll.

One limit worth stating plainly, since the doc comment now says it: this bounds the asynchronous part only. `load_query_sources` takes the state lock with a blocking `flock` on the runtime thread, and no timeout can preempt that — it still ties up a worker until it returns. The RPC does now answer, which is the part that matters here, but the blocking lock is a real thing and out of scope for this PR.

crates/coral-app/src/bootstrap/health.rs
sauldhernandez: Two halves, and they landed differently.

**Blast radius** — the premise does not hold. `catalog_info` is already resilient to a per-source failure: load failures collect into `failed_source_names` and continue (`query/manager.rs:584-604`), registration failures return `SourceFailurePolicy::Ignore` (`:157-167`), and `resolve_catalog` returns `Ok` with a partial catalog. `resolve_catalog_reports_load_failure_and_keeps_healthy_metadata` and `resolve_catalog_reports_registration_failure_and_keeps_healthy_metadata` already pin it. One bad source cannot pull the instance out of rotation.

**But the question was the right one to ask**, because the residual `Err` paths are worse than per-source ones — they are instance-wide, and therefore *correlated across replicas*. A missing default workspace (`WorkspaceNotFound`) or two sources claiming one runtime schema (`InvalidInput`) made every replica answer 503 simultaneously. Meanwhile the auth-disabled `/readyz` asks the identical question through `ListCatalog` and runs it through `catalog_rejection_is_reachable`, which deliberately treats `NotFound`/`InvalidArgument`/`FailedPrecondition` as engine-reachable → 204. So turning authentication on silently changed what readiness *means*, and that asymmetry is new in this PR.

Fixed in 3d34de65f: `readiness_from_catalog` mirrors coral-mcp's exclusion list rather than collapsing everything through `.is_ok()`. I mirrored the negative list verbatim instead of writing a narrower allow-list of the three codes — a narrower one would have re-created a smaller version of the divergence being fixed.

**Test gap** — confirmed and closed, with two corrections. The unit tests did not all use `fixed(false)` (`health.rs:242` and `:264` already used real closures), and the production closure's `Ok` path was already covered end to end by `grpc_health_reports_liveness_and_readiness_without_a_principal`. The genuinely uncovered surface was the `Err` branch, which `instance_wide_catalog_rejections_still_report_ready` and `infrastructure_faults_report_unready` now drive with real `QueryManagerError` values.

crates/coral-cli/src/lib.rs
sauldhernandez: Fixed for MCP HTTP in bc320082f; the `auth.http_bind_addr` half is deliberately left for #1647.

On the shape: rather than adding a second alias beside the first, `server_endpoint_is_loopback` and `server_requires_security_warning` collapsed into `server_endpoint_address` + a `server_requires_security_warning(SocketAddr)` with two real callers, which is the bar you set. It also reconciles a gap neither of us mentioned — the old rule used bare `ip().is_loopback()` while coral-app's `is_loopback_ip` also treats IPv4-mapped IPv6 as loopback, so a `::ffff:127.0.0.1` bind that bootstrap accepts as local was being reported as exposed. Both rules now agree, with the mapped case tested in each direction.

Two corrections to the finding. `server.mcp_http.bind` was not ungated: `server_config.rs:203-207` still hard-rejects a non-loopback MCP bind when auth is *disabled*, so the more dangerous case was already fail-closed and unchanged from base. Only the authenticated MCP bind accepts a remote address, and that is new capability in this PR rather than a removed guard — which is why the warning needs no unauthenticated wording.

And `auth.http_bind_addr` is not reachable at this commit at all: `into_authorization_server` has no non-test caller, so `coral server` binds nothing for it. Warning about it here would be dead code. It belongs in #1647 alongside the listener that actually starts.

The strongest form of your argument turned out to be one you did not make: this PR's own documented authenticated example binds *both* `[server].bind_addr` and `server.mcp_http.bind` to `0.0.0.0`, and the docs then said the warning covers the gRPC bind. That is now fixed too.

apps/docs/reference/configuration.mdx
sauldhernandez: Fixed in b983c886c. I verified every mechanism before writing it: the fingerprint is `Sha256::digest(token)` stored on the record at session creation, `authorize` is a bare equality compare with no rebind path anywhere, and `Missing => 404` / `Mismatch => 403` is the only mapping.

More in scope than you allowed, incidentally — you called it a courtesy pairing because the binding predates this PR, but the TTL advice paragraph is *added by* this PR, so it was a gap in text this PR introduces. The old "set it explicitly to the shortest value your clients can re-authenticate within" clause was trimmed to "set it explicitly", since leaving it would contradict the new paragraph's balanced guidance two sentences later. The PR body carried the same advice and has been updated to match.

Two refinements to how it is worded. I avoided "the session dies" — the record survives the 403 with its handle and admission permit intact, and only an explicit DELETE or shutdown removes it. And expiry is not required to hit this: any client that rotates proactively trips the 403 while the old token is still perfectly valid, so the docs describe rotation rather than expiry. The page also now notes there are no refresh tokens, which is what makes a short TTL cost more than "clients re-authenticate".

One adjacent thing I did **not** fix, flagging it rather than burying it: `AuthenticatedSessions::records` shrinks only in `close_session` and `close_all`, and `AUTHENTICATED_SESSION_IDLE_TIMEOUT` is passed as rmcp's `keep_alive`, which stops the worker without removing coral's record or releasing its `OwnedSemaphorePermit`. So every session abandoned at a 403 permanently consumes one of `MAX_AUTHENTICATED_SESSIONS = 4096` until restart — and short TTLs make abandonment the normal path. Pre-existing coral-mcp behavior, but the advice on this page accelerates it. Worth its own issue.

crates/coral-mcp/Cargo.toml
sauldhernandez: I dug into this one and I do not think it holds — leaving it open rather than resolving, in case you disagree.

**No new compile cost.** `coral-app` was already a normal transitive dependency of `coral-mcp` at base, via `coral-client` (`coral-client/Cargo.toml:29`, present before this PR). `cargo tree -p coral-mcp -i coral-app` shows the path. The engine, catalog and storage were already being compiled before `coral-mcp` on every build; making the edge direct adds no crate, no feature-unification change, and no codegen unit. The 4 added lines in `Cargo.lock` are `coral-client` gaining `tonic-health`, not this.

**Not an inversion.** There is no `coral-app` → `coral-mcp` edge in any section, so there was nothing to invert — and by your own framing (a transport sitting in front of `coral-app`) depending on `coral-app` is the correct direction, not the reverse one.

**Wrong crate for the constant.** `coral-mcp` never references `READINESS_SERVICE_NAME`. Its only new consumer is `coral-client/src/client.rs:18`, riding a `coral-client` → `coral-app` edge that predates this PR.

**The small crate is not small.** `CanonicalOauthUrl::parse` delegates to `crate::outbound_url_policy::{EndpointUrl, ResourceIdentifier, OutboundUrlPolicyError}`, all `pub(crate)` inside a 1232-line typed URL-policy module that itself pulls `crate::bootstrap::is_loopback_ip` and `zeroize`. Extracting it means moving 300-500 LoC across four-plus files into a new workspace member — or reimplementing the transport rules standalone, which recreates exactly the second copy this PR removes.

One thing in your favor that I think is the better version of the finding: at base `coral-app` was declared as a `coral-mcp` dev-dependency and was entirely unused — the only occurrence in the crate is a trace-filter string in `src/tests.rs`. So the change is better read as replacing a dead dev-dep with a live real one than as promoting a dep.

crates/coral-app/src/identity.rs
sauldhernandez: Agreed on substance, and we are deferring the fix to #1647 rather than doing it here. Leaving this open so it stays visible.

The reasoning, including one correction to the premise:

**"There is no backfill" is not true**, and it is load-bearing for the escalation — the comment says it, you amplified it to "genuinely no recovery" and "impossible after the first real deployment", and I repeated it back to Saul before checking. The subject is re-presented at *every* login, so a v1 id is recomputable: a lazy `UPDATE ... WHERE created_by_principal_id = <v1>` at next login rekeys each user with no dual-prefix acceptance and no orphaned rows. The deadline is soft. Worth noting the overstatement was introduced by 1b7e4ff58, which was itself correcting an earlier one; that comment needs fixing whenever the derivation is.

**The derivation is unreachable in production at this commit.** `into_authorization_server` has no caller, and `serve.rs` builds verifier-only principal providers, so no token endpoint is served and no `federated-*` row can exist until #1647 lands. That makes "fix before real deployments store rows" a #1647 gate rather than a #1645 blocker — which is the honest answer to your "deadline with no owner", and why it is being tracked there rather than here.

**One trap for whoever does it.** The obvious keying material is the session signing key, and that is the wrong choice: `SessionTokenVerifier` already models a multi-key `JwkSet` precisely so that key can be rotated, and keying principal ids off it would make a routine JWT key rotation destroy every attribution row irrecoverably. It needs a salt derived once and never rotated. Roughly 45-60 lines across five files, no migration, no golden hashes in the tests.

Also, for the record on the sibling comment: the preimage does not contain the provider, only a domain tag, a length, and the subject — the doc comment explains why the provider is deliberately absent.

crates/coral-app/src/identity.rs
```diff
@@ -95,9 +104,43 @@ impl Principal {
...
+        let mut identity = Vec::with_capacity(provider.len() + subject.len() + 32);
+        identity.extend_from_slice(b"coral-federated-user-v1\0");
+        identity.extend_from_slice(&(provider.len() as u64).to_be_bytes());
+        identity.extend_from_slice(provider.as_bytes());
+        identity.extend_from_slice(&(subject.len() as u64).to_be_bytes());
+        identity.extend_from_slice(subject.as_bytes());
+        Self {
+            id: PrincipalId(format!(
+                "federated-{}",
+                crate::hash::sha256_hex(&identity)
```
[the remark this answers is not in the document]
sauldhernandez: Tracked for #1647 rather than fixed here — leaving this open.

Two corrections to the finding. The preimage is not "`provider` and `subject`": it is a domain tag, a length, and the subject, with the provider deliberately absent (the doc comment explains why, given the single-provider assumption). And the derivation is unreachable in production at this commit — `into_authorization_server` has no caller, so no token endpoint is served and no `federated-*` row can be written until #1647 lands.

The underlying point stands: an unkeyed SHA-256 over a low-entropy subject is confirmable offline against a candidate list, and two unrelated deployments join on the id directly. When it is fixed, the per-deployment key must **not** be the session signing key — that one is expected to rotate (`SessionTokenVerifier` already models a multi-key `JwkSet`), and keying principal ids off it would destroy attribution on the first rotation. It needs a salt derived once and never rotated.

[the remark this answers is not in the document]
sauldhernandez: We decided not to key this after all. The plan we landed on: Coral will eventually need a users table anyway (you can't do authorization management without being able to list users), and once that exists, each user gets a random row ID and the provider's `sub` stored alongside it. That gives us a genuinely opaque ID for free, without a per-deployment secret to manage — it's basically the "persisted random ID mapping" you suggested.

Keying the hash now would just be an interim step that gets thrown away, and both changes cost the same rekey-on-login migration, so we'd pay it twice.

Until then the hash stays as it is. That's acceptable because it only ever lands in one DB column, never in any API response or log, and knowing an ID grants nothing.

`06b851a73` updates the comment to explain all this.

crates/coral-app/src/identity.rs
sauldhernandez: You were right that "fix it before X" with no owner wasn't good enough, and thinking that through changed the plan: we're not going to key this at all.

The longer-term need is a users table — authorization management means listing users and showing them to an admin, which a one-way hash can't do no matter how well it's keyed. Once that table exists, the ID becomes a random row key with the `sub` stored next to it: more opaque than a keyed hash, and no secret to distribute or protect from rotation.

So instead of a deadline there's now a direction, written into the doc comment in `06b851a73`. The current hash stays until then — it's confined to one DB column and grants nothing, which felt acceptable for a placeholder.

sauldhernandez:
### Merge activity

* **Aug 3, 2:59 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1645).
* **Aug 3, 3:01 PM UTC**: [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1645) rebased this pull request as part of a merge.
* **Aug 3, 3:10 PM UTC**: @sauldhernandez merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1645).

PR #1648, opened 2026-07-11
test(cli): cover session-authenticated gRPC and MCP

sauldhernandez:
> **Stack:** This stack lets a long-running Coral server authenticate users through an external OpenID Connect provider and protect its gRPC and MCP HTTP APIs.
>
> **This part of the stack:** This side branch checks the assembled authenticated server rather than adding runtime behavior. This PR covers the basic allow-and-deny path; the next PR adds invalid-token and restart coverage.

## Intent

Prove that the same signed Coral session token protects both direct gRPC calls and MCP over HTTP when the CLI starts the combined server.

## What it enables

The integration test starts the real composed server with ES256 session-token settings. It verifies that a gRPC catalog request without credentials is rejected, then verifies that one valid token authorizes both the gRPC catalog request and an MCP `list_catalog` call.

This catches wiring regressions between CLI startup, gRPC authentication, and MCP bearer forwarding. It mints the session token directly, so it does not test client metadata, the browser authorization flow, or token issuance.

## Usage examples

Run the focused test with:

```shell
cargo test --locked -p coral-cli --all-features serve::tests::session_authenticated_companion_gates_grpc_and_mcp
```

PR #1653, opened 2026-07-11
feat(app): store OAuth authorization approvals

sauldhernandez:
> **Stack:** This stack lets a long-running Coral server authenticate users through an external OpenID Connect provider and protect its gRPC and MCP HTTP APIs.
>
> **This part of the stack:** This section adds a user-confirmation step before Coral sends the browser to the identity provider. The previous PR validates client details, this PR stores the pending request, and the next PR adds the confirmation page.

## Intent

Give the confirmation step a server-side record it can trust. This PR adds short-lived, single-use approval tickets so Coral does not need to accept the client ID, redirect URL, PKCE challenge, or target resource back from the browser.

## What it enables

Coral can store the client name and ID, exact redirect URL, client state, PKCE challenge, and target resource under a random ticket. The ticket is drawn from the system random number generator and filled in place, so its bytes never exist outside the value that wipes them when it goes out of scope. Only a SHA-256 hash of the ticket is kept as the lookup key. A ticket expires after five minutes and can be claimed only once, even when two requests arrive at the same time.

The records live in memory and disappear when Coral restarts. They use the same 4,096-entry limit the rest of the OAuth state uses, counted separately for each record type, so a flood of approval requests cannot crowd out authorization sessions or codes.

This PR also splits the one `StateStore` trait into `ApprovalStore`, `SessionStore`, and `CodeStore`. Each request handler now depends only on the records it touches, and a test double implements only the kind it actually fakes.

This PR adds the storage building block only; it does not add an HTTP endpoint or change the authorization flow yet.

## Usage examples

There is no new command or public API in this PR. The next PR uses this flow: validate the request, mint a ticket, store the record under that ticket, put only the ticket in the confirmation form, and remove the stored record when the user chooses Continue or Cancel.

### DIFF
--- a/crates/coral-app/src/auth/authorization_server/callback.rs
+++ b/crates/coral-app/src/auth/authorization_server/callback.rs
@@ -179,7 +179,10 @@ mod tests {
     };
     use crate::auth::provider_client::OidcProviderClient;
     use crate::auth::session::SessionTokenIssuer;
-    use crate::auth::state_store::{InMemoryStateStore, StateStore, StateStoreError};
+    use crate::auth::state_store::{
+        InMemoryStateStore, OAuthAuthorizationApprovalRecord, OAuthAuthorizationApprovalTicket,
+        StateStore, StateStoreError,
+    };
 
     const OIDC_STATE: &str = "oidc-state";
     const CLIENT_STATE: &str = "client&error=injected";
@@ -507,6 +510,21 @@ mod tests {
 
     #[async_trait::async_trait]
     impl StateStore for FailCodeStore {
+        async fn store_authorization_approval(
+            &self,
+            ticket: &OAuthAuthorizationApprovalTicket,
+            approval: OAuthAuthorizationApprovalRecord,
+        ) -> Result<(), StateStoreError> {
+            self.0.store_authorization_approval(ticket, approval).await
+        }
+
+        async fn take_authorization_approval(
+            &self,
+            ticket: &OAuthAuthorizationApprovalTicket,
+        ) -> Result<Option<OAuthAuthorizationApprovalRecord>, StateStoreError> {
+            self.0.take_authorization_approval(ticket).await
+        }
+
         async fn store_authorization_session(
             &self,
             state: &str,
--- a/crates/coral-app/src/auth/authorization_server/callback.rs
+++ b/crates/coral-app/src/auth/authorization_server/callback.rs
@@ -22,7 +22,7 @@ pub(super) async fn oidc_callback(
         return direct_error("invalid_request", "OIDC callback state is required");
     };
     let Ok(Some(session)) = state
-        .state_store
+        .session_store
         .take_authorization_session(oidc_state)
         .await
     else {
@@ -81,7 +81,7 @@ pub(super) async fn oidc_callback(
         resource: session.resource,
     };
     if let Err(error) = state
-        .state_store
+        .code_store
         .store_authorization_code(&authorization_code, authorization)
         .await
     {
@@ -179,10 +179,7 @@ mod tests {
     };
     use crate::auth::provider_client::OidcProviderClient;
     use crate::auth::session::SessionTokenIssuer;
-    use crate::auth::state_store::{
-        InMemoryStateStore, OAuthAuthorizationApprovalRecord, OAuthAuthorizationApprovalTicket,
-        StateStore, StateStoreError,
-    };
+    use crate::auth::state_store::{CodeStore, InMemoryStateStore, SessionStore, StateStoreError};
 
     const OIDC_STATE: &str = "oidc-state";
     const CLIENT_STATE: &str = "client&error=injected";
@@ -220,7 +217,7 @@ mod tests {
         settings
     }
 
-    fn state(issuer: &str, state_store: Arc<dyn StateStore>) -> AuthorizationServerHttpState {
+    fn state(issuer: &str, store: Arc<InMemoryStateStore>) -> AuthorizationServerHttpState {
         let signing_key =
             EcdsaKeyPair::generate_pkcs8(&ECDSA_P256_SHA256_FIXED_SIGNING, &SystemRandom::new())
                 .expect("P-256 signing key");
@@ -233,7 +230,8 @@ mod tests {
         AuthorizationServerHttpState {
             settings: Arc::new(settings(issuer)),
             session_tokens,
-            state_store,
+            session_store: store.clone(),
+            code_store: store,
             provider_client: OidcProviderClient::new().expect("client"),
             registered_clients: Arc::new(BTreeMap::new()),
             authorization_resources: Arc::new(BTreeSet::from([RESOURCE.into()])),
@@ -268,14 +266,14 @@ mod tests {
         callback_raw(state, query(pairs)).await
     }
 
-    async fn seed(store: &dyn StateStore, key: &str) {
+    async fn seed(store: &dyn SessionStore, key: &str) {
         store
             .store_authorization_session(key, session())
             .await
             .expect("store");
     }
 
-    async fn take(store: &dyn StateStore, key: &str) -> Option<OAuthAuthorizationSessionRecord> {
+    async fn take(store: &dyn SessionStore, key: &str) -> Option<OAuthAuthorizationSessionRecord> {
         store.take_authorization_session(key).await.expect("store")
     }
 
@@ -506,38 +504,11 @@ mod tests {
         assert_provider_failure(200, 200, "wrong-nonce").await;
     }
 
-    struct FailCodeStore(InMemoryStateStore);
+    /// Code store whose writes always fail, standing in for a full store.
+    struct FailCodeStore;
 
     #[async_trait::async_trait]
-    impl StateStore for FailCodeStore {
-        async fn store_authorization_approval(
-            &self,
-            ticket: &OAuthAuthorizationApprovalTicket,
-            approval: OAuthAuthorizationApprovalRecord,
-        ) -> Result<(), StateStoreError> {
-            self.0.store_authorization_approval(ticket, approval).await
-        }
-
-        async fn take_authorization_approval(
-            &self,
-            ticket: &OAuthAuthorizationApprovalTicket,
-        ) -> Result<Option<OAuthAuthorizationApprovalRecord>, StateStoreError> {
-            self.0.take_authorization_approval(ticket).await
-        }
-
-        async fn store_authorization_session(
-            &self,
-            state: &str,
-            session: OAuthAuthorizationSessionRecord,
-        ) -> Result<(), StateStoreError> {
-            self.0.store_authorization_session(state, session).await
-        }
-        async fn take_authorization_session(
-            &self,
-            state: &str,
-        ) -> Result<Option<OAuthAuthorizationSessionRecord>, StateStoreError> {
-            self.0.take_authorization_session(state).await
-        }
+    impl CodeStore for FailCodeStore {
         async fn store_authorization_code(
             &self,
             _code: &str,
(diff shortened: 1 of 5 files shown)


sauldhernandez:
* **#1653** <a href="https://app.graphite.com/github/pr/withcoral/coral/1653?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1653?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-app/src/auth/state_store.rs
```diff
@@ -105,6 +156,44 @@ impl InMemoryStateStore {
...
+        ticket: &OAuthAuthorizationApprovalTicket,
+        approval: OAuthAuthorizationApprovalRecord,
+    ) -> Result<(), StateStoreError> {
+        let key = opaque_hash(ticket.as_bytes());
+        let mut inner = self.inner.lock().await;
+        let now = Instant::now();
+        let expires_at = self.expires_at(now)?;
+        inner.purge_expired(now);
+        ensure_capacity(
+            inner.approvals.len(),
```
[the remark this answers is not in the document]
sauldhernandez: Not taking this one — the per-record-type bound is deliberate and predates this PR.

There is no shared-budget contract in the code to violate: the constant is `MAX_OAUTH_STATE_ENTRIES_PER_KIND`, the field is `max_entries_per_kind`, and the existing test is `expiration_is_lazy_and_capacity_is_bounded_per_record_type`. This PR adds a third map that follows the same rule the sessions and codes maps already followed. The "documented shared limit" was a wording slip in the PR description, which I have corrected.

Merging the maps into one budget would weaken the flow rather than harden it. Per-kind limits act as bulkheads. An attacker hammering `/authorize` creates the cheapest possible record; under a shared budget that flood would exhaust the pool and block authorization *codes* for users who are already mid-login, breaking their redemption. Keeping the maps separate confines the flood to new approvals.

The memory ceiling worth tightening is per-record size rather than map count — `client_state` is caller-controlled and bounded only by the HTTP request-line limit. That is tracked separately and is out of scope here.

crates/coral-app/src/auth/state_store.rs
```diff
@@ -6,13 +6,45 @@ use std::time::Duration;
...
 type SecretHash = [u8; 32];
 
+pub(crate) struct OAuthAuthorizationApprovalTicket([u8; 32]);
+
+impl OAuthAuthorizationApprovalTicket {
+    #[cfg_attr(
+        not(test),
+        expect(dead_code, reason = "wired by the stacked authorization approval flow")
+    )]
+    pub(super) fn from_bytes(bytes: [u8; 32]) -> Self {
```
simonwhitaker: Two things this constructor leaves on the table, both fixable with the same change:

1. `[u8; 32]` is `Copy`, so `from_bytes(bytes)` copies — the caller's original array is untouched by the `Drop` impl, which only clears the struct's copy. The test below demonstrates it: `raw_ticket` is still live and unzeroed after the ticket exists.
2. Nothing in the type stops a caller minting `from_bytes([0; 32])`. Entropy is load-bearing for this secret but lives entirely outside the type.

A `generate()` constructor in this module that fills the array from a CSPRNG in place and never hands out a second copy fixes both, and drops the `expect(dead_code)` here as a bonus. Related: `as_bytes` is `pub(super)`, so the raw secret is visible to all of `auth` when only `opaque_hash` in this file needs it — if the next PR needs a URL-safe encoding for the form field, putting `Display`/`FromStr` here rather than exporting raw bytes keeps the secret from circulating.

crates/coral-app/src/auth/state_store.rs
```diff
@@ -6,13 +6,45 @@ use std::time::Duration;
...
+
+#[derive(Clone, PartialEq, Eq)]
+pub(crate) struct OAuthAuthorizationApprovalRecord {
+    pub(crate) client_id: String,
+    pub(crate) client_name: String,
+    pub(crate) redirect_uri: String,
+    pub(crate) client_state: Option<String>,
+    pub(crate) code_challenge: String,
+    pub(crate) resource: String,
+}
```
simonwhitaker, line 68: The PR description says Coral stores "the validated provider, client name and ID, exact redirect URL, client state, PKCE challenge, and scope". The record has neither provider nor scope: there's a single configured provider (`state.settings.provider()`), and `authorize.rs` rejects any `scope` parameter outright with `invalid_scope` ("scope is not supported"). The field is `resource`.

Worth correcting the description so reviewers of the next PR aren't hunting for a scope field that was never meant to exist.

crates/coral-app/src/auth/state_store.rs
```diff
@@ -231,15 +322,35 @@ fn secret_hash(secret: &str) -> Result<SecretHash, StateStoreError> {
     if secret.is_empty() {
         return Err(StateStoreError::EmptySecret);
     }
-    Ok(Sha256::digest(secret.as_bytes()).into())
+    Ok(opaque_hash(secret.as_bytes()))
+}
+
+fn opaque_hash(secret: &[u8]) -> SecretHash {
```
simonwhitaker: Nit: `opaque_hash` vs `secret_hash` reads as a difference in secrecy, but the actual difference is bytes-vs-checked-string — both hash a secret, one just skips the emptiness guard because a fixed-size array can't be empty. `hash_bytes` would read truer.

crates/coral-app/src/auth/state_store.rs
```diff
@@ -262,6 +373,151 @@ mod tests {
         }
     }
 
+    #[tokio::test]
+    async fn authorization_approvals_are_hashed_and_atomically_single_use() {
```
simonwhitaker, line 454: The technique here is a real improvement over `tokio::join!` — it *forces* both futures onto the mutex queue rather than hoping for the interleaving. Two notes:

- The two spawn blocks are byte-for-byte identical across ~30 lines. Worth extracting the poll-until-parked-then-await body into a helper.
- That makes the inconsistency more visible: `authorization_sessions_are_hashed_and_single_use` still uses the weaker `join!` form for the same property. Either share the helper across all three record kinds or leave all three at the simpler form — having exactly one kind rigorously covered is the odd state.

Minor: `assert!(!inner.approvals.contains_key(&raw_ticket))` is asserting `sha256(x) != x`, which can't fail. Consistent with the session test's equivalent, so not carrying weight either way.

crates/coral-app/src/auth/authorization_server/callback.rs
```diff
@@ -507,6 +510,21 @@ mod tests {
 
     #[async_trait::async_trait]
     impl StateStore for FailCodeStore {
+        async fn store_authorization_approval(
```
simonwhitaker: `StateStore` is 6 methods now, and this double gains 12 lines of pure delegation that will never be exercised — `FailCodeStore` only cares about codes. With more record kinds coming in this stack, worth considering splitting into per-concern traits (`ApprovalStore` / `SessionStore` / `CodeStore`) so a double only implements what it actually fakes.

simonwhitaker (approved):
Approving — scope is tight, the new record kind reuses the existing TTL/capacity/purge machinery exactly, and no correctness bugs turned up.

Verified locally on the PR head: `cargo clippy -p coral-app --all-targets` clean, all 7 `state_store` tests pass, and the new deterministic-interleaving test was stable across 40 consecutive runs.

Comments below are all non-blocking. The one I'd actually push on before the next PR lands is the ticket constructor — it's cheapest to fix while nothing constructs a ticket yet.

Three forward-looking notes for the next PR in the stack, not for this one:

- **Fail-closed capacity with no eviction.** At 4,096 outstanding approvals, new *legitimate* requests get `CapacityExceeded` for up to 5 minutes. Approvals will be created by unauthenticated `/authorize` traffic, so this becomes the cheapest denial surface in the stack. Same policy as the existing kinds so it isn't a regression here — but worth storing the approval as late in validation as possible, and deciding whether evicting the oldest beats rejecting the newest.
- **`client_name` is remote input**, fetched from a client-supplied metadata URL. Length-bounded upstream at `MAX_CLIENT_NAME_BYTES`, so memory is fine, but it needs HTML-escaping when the confirmation page renders it.
- **Ticket delivery.** The ticket is the only thing binding the confirmation POST to the validated request — there's no browser-session binding — so it doubles as the CSRF token. It needs to travel in the form body, never in a URL (Referer / history / logs).

crates/coral-app/src/auth/state_store.rs
sauldhernandez: Done — `generate(&dyn SecureRandom)` now fills the ticket in place, `from_bytes` is `#[cfg(test)]`, and `as_bytes` is private on this PR. I took `&dyn SecureRandom` rather than an internal `SystemRandom::new()` so it matches ring's own signatures and the next PR can hold one RNG on the HTTP state.

Two corrections worth recording. The `expect(dead_code)` does not disappear — it moves from `from_bytes` to `generate`, which nothing calls until #1654. And on #1654 both `from_bytes` and `as_bytes` had to come back to `pub(super)`: the confirmation page encodes the ticket into the form and decodes it back out of the submission, so a non-test byte constructor is genuinely needed there. The minting path is hardened everywhere — `confirmation::new_ticket` calls `generate` now instead of filling a buffer and copying — but your `Display`/`FromStr` consolidation is still open on that PR.

Related, and left deliberately alone: `OAuthAuthorizationApprovalBrowserBinding` on #1654 has exactly the shape you flagged here. Changing it during the rebase would have been smuggling a fix into another PR.

crates/coral-app/src/auth/state_store.rs
sauldhernandez: Correct on both counts — there is no provider field, and `authorize.rs` rejects `scope` outright. Description fixed: it says target resource now. Since this was the second inaccuracy in it, I re-read the whole thing rather than patching the one clause.

crates/coral-app/src/auth/state_store.rs
sauldhernandez: Renamed to `hash_bytes`. Agreed the old name implied a difference in secrecy that is not there.

crates/coral-app/src/auth/state_store.rs
sauldhernandez: Extracted `park_takes_on_store_lock` and used it for all three record kinds, so the duplication is gone and the coverage is consistent.

One thing your note understated: codes had no concurrent single-use test at all. `authorization_codes_are_consumed_only_after_all_bindings_match` covers the binding checks, not the race — so that one is new rather than a rewrite. Sessions moved off the `join!` form.

I mutation-checked the new codes test before trusting it: swapping `codes.remove(&key)` for a non-consuming `get(&key).cloned()` makes it fail, so it is not vacuous. Left the `sha256(x) != x` assert alone, as you suggested.

crates/coral-app/src/auth/authorization_server/callback.rs
sauldhernandez: Split into `ApprovalStore` / `SessionStore` / `CodeStore`. `FailCodeStore` is now a unit struct implementing only `CodeStore`, with the take arm as `unreachable!` — the callback issues authorization codes, it never redeems them. callback.rs ends up smaller than it was before this PR touched it.

You were right that the cost grows with the stack. It took two conflict stops across the nine branches upstack, and #1655's `AuthorizationServerHttpState::new` absorbed the fan-out from one concrete store into the three trait-object fields, which reads better than what this PR does on its own.

sauldhernandez:
### Merge activity

* **Aug 3, 2:59 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1653).
* **Aug 3, 3:30 PM UTC**: [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1653) rebased this pull request as part of a merge.
* **Aug 3, 3:37 PM UTC**: @sauldhernandez merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1653).

PR #1654, opened 2026-07-11
feat(app): confirm OAuth authorization redirects

sauldhernandez:
> **Stack:** This stack lets a long-running Coral server authenticate users through an external OpenID Connect provider and protect its gRPC and MCP HTTP APIs.
>
> **This part of the stack:** This section adds a user-confirmation step before Coral sends the browser to the identity provider. The previous PR stores each pending request, this PR adds the confirmation page and form, and the next PR connects the flow to client-hosted metadata.

## Intent

Require an explicit Continue or Cancel decision before Coral redirects a user to the external identity provider.

## What it enables

`GET /oauth/authorize` now validates the request, stores the trusted values under a short-lived ticket, and shows a locked-down page with the client name and the redirect host and port. `POST /oauth/authorize` accepts only that ticket and the user's decision.

Continue consumes the ticket and starts the provider login. Cancel consumes it and returns `access_denied` to the previously validated redirect URL. Replayed, expired, malformed, or restart-lost tickets fail directly instead of redirecting anywhere. The form is size-limited, the page escapes client text, and the response disables caching, framing, scripts, and referrer sharing.

This PR supplies the confirmation machinery. The next PR replaces the temporary client lookup used here with the production client-metadata lookup.

## Usage examples

A valid authorization request first returns an “Approve access?” page. Choosing Continue redirects the browser to the configured identity provider once. Choosing Cancel redirects to the validated client callback with `error=access_denied`; submitting the same form again returns a direct error.

### DIFF
--- a/crates/coral-app/src/auth/authorization_server.rs
+++ b/crates/coral-app/src/auth/authorization_server.rs
@@ -11,7 +11,7 @@ use super::config::{AuthSettings, ResolvedAuthSettings, signing_key_env_error};
 use super::error::AuthServerError;
 use super::provider_client::OidcProviderClient;
 use super::session::SessionTokenIssuer;
-use super::state_store::{CodeStore, InMemoryStateStore, SessionStore};
+use super::state_store::{ApprovalStore, CodeStore, InMemoryStateStore, SessionStore};
 use crate::oauth_resource::{CanonicalOauthUrl, OauthUrlError};
 use axum::Router;
 use axum::extract::State;
@@ -36,7 +36,7 @@ pub struct CoralAuthorizationServer {
     settings: Arc<ResolvedAuthSettings>,
     session_tokens: SessionTokenIssuer,
     state_store: Arc<InMemoryStateStore>,
-    registered_clients: Arc<BTreeMap<String, Vec<String>>>,
+    registered_clients: Arc<BTreeMap<String, (String, Vec<String>)>>,
     authorization_resources: BTreeSet<String>,
 }
 
@@ -126,6 +126,7 @@ impl CoralAuthorizationServer {
         let state = AuthorizationServerHttpState {
             settings: self.settings,
             session_tokens: self.session_tokens,
+            approval_store: self.state_store.clone(),
             session_store: self.state_store.clone(),
             code_store: self.state_store,
             provider_client: OidcProviderClient::new()
@@ -138,7 +139,10 @@ impl CoralAuthorizationServer {
                 "/.well-known/oauth-authorization-server",
                 get(authorization_server_metadata),
             )
-            .route("/oauth/authorize", get(authorize::oauth_authorize))
+            .route(
+                "/oauth/authorize",
+                get(authorize::oauth_authorize_get).post(authorize::oauth_authorize_post),
+            )
             .route("/oauth/token", post(token::oauth_token))
             .route(OIDC_CALLBACK_PATH, get(callback::oidc_callback))
             .with_state(state);
@@ -231,10 +235,11 @@ impl Drop for RunningCoralAuthorizationServer {
 struct AuthorizationServerHttpState {
     settings: Arc<ResolvedAuthSettings>,
     session_tokens: SessionTokenIssuer,
+    approval_store: Arc<dyn ApprovalStore>,
     session_store: Arc<dyn SessionStore>,
     code_store: Arc<dyn CodeStore>,
     provider_client: OidcProviderClient,
-    registered_clients: Arc<BTreeMap<String, Vec<String>>>,
+    registered_clients: Arc<BTreeMap<String, (String, Vec<String>)>>,
     authorization_resources: Arc<BTreeSet<String>>,
 }
 
--- a/crates/coral-app/src/auth/authorization_server/authorize.rs
+++ b/crates/coral-app/src/auth/authorization_server/authorize.rs
@@ -1,18 +1,26 @@
 use std::collections::BTreeMap;
 
-use axum::extract::{RawQuery, State};
+use axum::extract::{RawQuery, Request, State};
+use axum::http::HeaderMap;
 use axum::response::Response;
 use url::Url;
 
 use super::super::provider_client::OidcAuthorizationRequest;
-use super::super::state_store::OAuthAuthorizationSessionRecord;
+use super::super::state_store::{
+    OAuthAuthorizationApprovalRecord, OAuthAuthorizationSessionRecord,
+};
 use super::response::{TrustedRedirect, direct_error, redirect};
 use super::{AuthorizationServerHttpState, canonical_authorization_resource, query};
 use crate::outbound_url_policy::{BrowserRedirect, EndpointUrl};
 
-pub(super) async fn oauth_authorize(
+mod confirmation;
+
+use confirmation::ApprovalDecision;
+
+pub(super) async fn oauth_authorize_get(
     State(state): State<AuthorizationServerHttpState>,
     RawQuery(raw_query): RawQuery,
+    headers: HeaderMap,
 ) -> Response {
     let Ok(query) = parse_query(raw_query.as_deref().unwrap_or_default()) else {
         return direct_error("invalid_request", "authorization request is malformed");
@@ -22,11 +30,12 @@ pub(super) async fn oauth_authorize(
     else {
         return direct_error("invalid_request", "client_id and redirect_uri are required");
     };
-    let Some(callback) = registered_redirect(&state.registered_clients, client_id, redirect_uri)
+    let Some((client_name, callback)) =
+        registered_client(&state.registered_clients, client_id, redirect_uri)
     else {
         return direct_error("invalid_request", "client_id or redirect_uri is invalid");
     };
-    let trusted = TrustedRedirect::new(callback, query.state.clone());
+    let trusted = TrustedRedirect::new(callback.clone(), query.state.clone());
 
     match query.response_type.as_deref() {
         Some("code") => {}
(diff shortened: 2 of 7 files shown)


sauldhernandez:
* **#1654** <a href="https://app.graphite.com/github/pr/withcoral/coral/1654?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1654?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-app/src/auth/authorization_server.rs
```diff
@@ -36,7 +36,7 @@ pub struct CoralAuthorizationServer {
     settings: Arc<ResolvedAuthSettings>,
     session_tokens: SessionTokenIssuer,
     state_store: Arc<dyn StateStore>,
-    registered_clients: Arc<BTreeMap<String, Vec<String>>>,
+    registered_clients: Arc<BTreeMap<String, (String, Vec<String>)>>,
```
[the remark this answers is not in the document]
sauldhernandez: Already fixed — this was true at the reviewed commit (6ce54ef0f) but is stale at HEAD. `6d3545e7b` rebuilds the fixture in the `(client_name, redirect_uris)` shape at authorize.rs:762-765. Verified with a cache-busted rebuild: `cargo clippy -p coral-app --all-targets --all-features --locked -- -D warnings` exits 0 with zero diagnostics, and `cargo fmt --all -- --check` is clean.

One accurate residual, noted for the record rather than actioned here: the tuple widening landed in this PR's first commit (`984758a78`) while the fixture stayed `Vec`-only until the last, so five of the six commits fail `--all-targets`. Only the tip is green. Not a defect in the merged result; worth squashing if per-commit bisectability matters.

crates/coral-app/src/auth/authorization_server.rs
```diff
@@ -137,7 +137,10 @@ impl CoralAuthorizationServer {
                 "/.well-known/oauth-authorization-server",
                 get(authorization_server_metadata),
             )
-            .route("/oauth/authorize", get(authorize::oauth_authorize))
+            .route(
+                "/oauth/authorize",
+                get(authorize::oauth_authorize_get).post(authorize::oauth_authorize_post),
```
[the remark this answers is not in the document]
sauldhernandez: Correct observation, but it does not belong to this PR.

The dead route predates this branch. The base (`sauldhernandez/oss-serve-oauth-approval-store`) already had the empty map at authorization_server.rs:89, the `/oauth/authorize` route at :140, and the same gate with the identical error string at authorize.rs:25. `git log -S "registered_clients: Arc::new(BTreeMap::new())"` blames `1107905bd` (#1640), which is already on `main` — `origin/main` ships the same unreachable route today. This PR's entire footprint on that file is nine lines: widening the map value to `(String, Vec<String>)` and adding `.post(...)` to the existing route.

The mechanism sentence is also inaccurate for the half that is actually new here: `oauth_authorize_post` (authorize.rs:110-166) never consults `registered_clients` at all — it gates on `take_authorization_approval` against the state store.

Wiring lands in the next PR, as the description says. #1655 removes `registered_clients` from `AuthorizationServerHttpState` entirely and replaces it with a `ClientMetadataResolver` backed by a production `HttpClientMetadataResolver`, turning the gate into `resolve_client(&state, client_id).await`. Doing it here would be work #1655 immediately deletes.

Fair point retained: this PR ships a large amount of code whose production reachability depends on #1655 landing, and the map-value widening is churn that #1655 discards.

crates/coral-app/src/auth/authorization_server/authorize.rs
```diff
@@ -1,18 +1,26 @@
 use std::collections::BTreeMap;
```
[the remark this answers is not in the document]
sauldhernandez: The line count is exact and this PR did cause the crossing (519 → 1,038), but the "1,000-line maintainability limit" is not a rule this repository has. I searched clippy.toml, the workspace `Cargo.toml` lint tables, rustfmt config, all nine `AGENTS.md` files, `CONTRIBUTING.md`, the `Makefile`, and `.github/workflows/` — there is no file-size or line-count rule anywhere, and clippy has no file-length lint. 52 files under `crates/` already exceed 1,000 lines (largest: `sources/manager.rs` at 4,040), so the threshold is a review heuristic rather than a standard this file is breaking.

On the substance of the suggestion: the split is 245 production lines and 793 test lines, and `<module>/tests.rs` is an established in-repo pattern — 8 occurrences, including `coral-mcp/src/http.rs` (1,058 production lines) → `http/tests.rs`, the nearest structural analogue. So moving the test module to `authorize/tests.rs` would be conventional rather than deviant, and the cost is near zero: module nesting is identical either way, so the existing `use super::*;` and `use super::super::{...};` need no edits.

Leaving it as-is for now — reasonable suggestion, no rule compels it, and it is orthogonal to what this PR is for. Resolving rather than actioning; happy to revisit if the file keeps growing.

crates/coral-app/src/auth/authorization_server/authorize.rs
```diff
@@ -309,7 +489,230 @@ mod tests {
...
+            "x".repeat(257),
+        ];
+        for body in malformed {
+            let request = Request::builder()
+                .uri(format!(
+                    "/oauth/authorize?ticket={ticket}&decision=continue"
+                ))
+                .header(header::CONTENT_TYPE, "application/x-www-form-urlencoded")
+                .body(Body::from(body))
+                .expect("request");
```
simonwhitaker: This test never reaches the body parser it's testing.

`parse_submission` checks `has_exact_origin` first, and these requests carry only `Content-Type` — no `Origin`, no `Cookie`. So every case returns `Err(())` at the origin check, and all six malformed bodies pass for a reason that has nothing to do with what the test is named after. The trailing wrong-content-type case below has the same problem.

That leaves the whole submission parser with no coverage: the `index >= 2` field cap, the duplicate-name guards, the `decision` whitelist, the `TICKET_LENGTH` filter, and `MAX_FORM_BYTES`. The `?ticket=…&decision=continue` on the URI is likewise never proven to be ignored, so the "no query bypass" half of the name isn't tested either. Only the happy path is exercised, via `submission_request`.

Adding `.header(header::ORIGIN, AUTH_ISSUER)` and `.header(header::COOKIE, browser_cookie())` to both builders fixes it. Worth also asserting one *well-formed* submission parses `Ok` inside this same test — otherwise a future refactor that inverts an early guard can quietly make it vacuous again in exactly this way.

crates/coral-app/src/auth/authorization_server/authorize/confirmation.rs
```diff
@@ -0,0 +1,267 @@
...
+}
+
+fn browser_binding_from_headers(
+    headers: &HeaderMap,
+    secure: bool,
+) -> Option<OAuthAuthorizationApprovalBrowserBinding> {
+    let expected_name = approval_cookie_name(secure);
+    let mut encoded = None;
+    for header_value in headers.get_all(header::COOKIE) {
+        let header_value = header_value.to_str().ok()?;
```
simonwhitaker: The `?` here returns `None` from the whole function, not just from this header value.

`HeaderValue::to_str` fails on any byte outside visible ASCII, and browsers do send raw non-ASCII cookie octets when `document.cookie` was set that way. So a single unrelated cookie on the auth-server host — set by anything else served from that origin — makes every `POST /oauth/authorize` a 400.

The failure mode is the unpleasant part: the GET path quietly mints a fresh binding and re-sets the cookie each time, so the user loops through the consent page indefinitely with no signal about why the decision never takes. `else { continue }` fails closed just as safely without the wedge.

crates/coral-app/src/auth/authorization_server/authorize.rs
```diff
@@ -61,6 +70,71 @@ pub(super) async fn oauth_authorize(
...
+        &approval.client_id,
+        &callback,
+        &browser_binding,
+        secure_cookie,
+    ) else {
+        return trusted.error("server_error", "authorization failed");
+    };
+    if state
+        .state_store
+        .store_authorization_approval(&ticket, &browser_binding, approval)
```
simonwhitaker, line 101: Worth flagging for the client-metadata PR on top of this one: this write now happens before any user interaction and before the provider is contacted, where previously a store entry required a successful discovery round trip first. A reload of the consent page also mints a new ticket while reusing the binding, so entries accumulate per page view rather than per authorization.

With 4096 shared entries, anyone holding a registered `client_id` + `redirect_uri` pair — both public by construction — can fill the approval store cheaply and turn real authorizations into `server_error` for the TTL.

Not exploitable today, since `registered_clients` is `BTreeMap::new()`. But the next PR in the stack is exactly what populates it, so this is the moment the cost changes. A per-binding cap on outstanding approvals would bound it without needing rate-limiting infrastructure.

simonwhitaker (approved):
The security design here holds up well under scrutiny. A few things I specifically checked rather than assumed:

- The `Origin` comparison in `has_exact_origin` is exactly right, not accidentally right: `validate_issuer` is `root_only` and trims the trailing slash, and `Url` normalization drops default ports, so the stored issuer string is precisely the serialized-origin form a browser sends.
- Binding a wrong cookie misses the approval map *without* consuming the entry, so a mismatched binding can't be used to burn someone else's pending approval.
- Rejecting duplicate cookie names closes cookie-shadowing, which is the usual hole in ticket+cookie schemes.

Approving. Three inline comments below — the first is the one I'd land before merge, since a security-relevant parser is currently shipping with tests that appear to cover it but don't. The other two are worth doing but don't block.

Two smaller things not worth inline comments:

- `confirmation::response` hand-rolls `cache-control`/`pragma`/`referrer-policy` instead of calling `security_headers()`. The `response.rs` module doc says the module exists so "a handler cannot ship with a weaker header set than its siblings — the state this module was extracted to end." That's the same drift, reintroduced. Relatedly, `direct_error` gains `nosniff` here but `redirect` doesn't; if it belongs on one it probably belongs in `security_headers()`.
- `Continue` is the first submit button, so it's the implicit Enter-key default on the consent page. Putting `Cancel` first in DOM order makes the safe choice the default one.

I didn't run the suite — my checkout is on `main` and this targets a stacked base.

crates/coral-app/src/auth/authorization_server/authorize.rs
sauldhernandez: Fixed in #1655 (`081fd481b`) — though not as a per-binding cap, because that would not have bounded it.

A caller that sends no cookie is handed a fresh binding on every request by `browser_binding_for_page`, so every entry lands under a distinct binding, the per-binding count never reaches its limit, and the approval budget fills anyway. A per-binding cap would have bounded only callers that keep their cookie — honest browsers reloading the page — not the attack.

The cap is per `client_id` instead: 64 approvals per client beneath the existing budget (`state_store.rs:31`, enforced at `:260`). The record already carries `client_id`, so it needed no new plumbing. The count walks the map rather than maintaining a tally — the store path already runs an O(n) `purge_expired` on every write, so it costs nothing asymptotically, and a count taken from the entries cannot drift out of step with them.

That shifts the cost the way you were describing: a flood now stays inside the client that caused it, so one `client_id` can no longer deny every other. Reaching the 4096 approval budget takes 64 distinct client metadata documents rather than one — though that is worth less than it sounds, since a single host with a wildcard route serves 64 of them trivially, so containment is the part carrying the weight. `provider_and_store_failures_are_generic_trusted_errors` is evidence the bound bites: it filled all 4096 slots under a single `CLIENT_ID`, which is precisely the attack, and now names a distinct client per entry.

One scope note, in case "shared" meant one pool across record kinds: the 4096 is per kind — #1653 gave approvals their own budget (`MAX_OAUTH_STATE_ENTRIES_PER_KIND` at `state_store.rs:13`, counted against `inner.approvals` alone). So filling approvals stops new authorizations but has never touched in-flight sessions or code redemption.

The part that made this worth bounding — the write moving ahead of any user interaction, and a reload minting a new ticket per page view — is unchanged.

sauldhernandez:
### Merge activity

* **Aug 3, 2:59 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1654).
* **Aug 3, 3:39 PM UTC**: [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1654) rebased this pull request as part of a merge.
* **Aug 3, 3:46 PM UTC**: @sauldhernandez merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1654).

PR #1715, opened 2026-07-12
fix(app): cancel identity OAuth on shutdown

sauldhernandez:
## Intent

Cancel active USER and WORKSPACE identity OAuth operations when the local server shuts down, without allowing shutdown to report `Cancelled` after durable credential persistence has begun.

## What it enables

- `RunningServer` owns one root cancellation token and gives each identity service a child token.
- Explicit shutdown and `Drop` signal identity OAuth cancellation before the Tonic server is stopped.
- Pre-transaction cancellation returns a static redacted `Cancelled` status and drops the operation before persistence.
- Once a serializable transaction starts, shutdown awaits the same pinned manager future through its real terminal result, preventing a false cancelled response over committed encrypted credentials.
- USER and WORKSPACE identity OAuth both preserve the same behavior across deterministic restart.

## Usage examples

- If `RunningServer::shutdown().await` begins while an omitted-setup identity OAuth token response is still pending, the client receives `Code::Cancelled` and no identity or identity document is persisted.
- Restarting Coral against the same database after that shutdown returns typed `IDENTITY_NOT_FOUND` for both attempted identities.
- Fixed-token identity creation and source-import OAuth behavior are unchanged.

## Validation

- Focused shutdown-phase, manager-race, and dual-owner gRPC tests passed.
- Strict all-target `coral-app` Clippy passed with `-D warnings`.
- `make postgres-tests` passed all three Docker-backed gates.
- `make perf-check` passed at 209.6 ms mean against the 750 ms threshold.
- Exact final `make rust-checks` passed 1,748/1,748 tests with 3 skipped plus clean fmt, workspace Clippy, and warning-denied rustdoc.
- Final diff: 541 additions and 51 deletions, 592 changed lines.

## Review

The exact-head five-lane review enumerated 17 safe credential flows. It confirmed the prior commit/cancellation P1 is closed, accepted and closed one P2 transition-test gap, and found no additional P1/P2 risk at 0.97 supervisor confidence.

A separate mandatory direct child, B5d1c, will abort owned authorization-code callback workers when their parent operation is dropped. That child remains required before B5d2 and before any affected PR can become ready.

## Stack

Parent: #1714. This PR is intentionally draft and must remain draft.

PR #1460, opened 2026-07-01
test(app): cover credential encryption primitives

sauldhernandez:
## Intent

Keep credential encryption unit coverage separate from the core encryption primitives so the stack stays under the PR size cap.

## What it enables

This verifies AES-GCM round trip behavior, workspace/source AAD binding, tamper rejection, unknown key rejection, DEK rewrap behavior, and local key file reuse.

## Usage examples

Run focused encryption coverage:

```shell
cargo test --locked -p coral-app credentials::encryption -- --nocapture
```

## Validation

Review-swarm run `.context/review-swarm/20260701T075247Z-sauldhernandez-rdbms-hardening-docs-ci-branch`; final pass recorded 0 active findings and 0 supervisor questions.

- `cargo test --locked -p coral-app credentials::encryption -- --nocapture`
- `make rust-checks`

PR #1716, opened 2026-07-12
fix(app): cancel OAuth callback workers

sauldhernandez:
## Intent

OAuth loopback callback connections were handled by detached tasks. If the owning OAuth future was dropped by client cancellation, source-import cancellation, timeout, or server shutdown, an already accepted idle connection could remain alive and later emit a stale HTTP 200 after the session no longer had a receiver.

Own every accepted callback worker in the callback receiver so dropping that receiver also aborts its outstanding connections.

## What it enables

- Bounds accepted callback-worker lifetime to the OAuth operation that owns the listener.
- Prevents an idle preconnection from returning callback authority or a stale success response after cancellation.
- Preserves the existing valid, ignored, error, timeout, and worker-failure behavior.
- Applies the same cleanup to identity OAuth and the shared source-import OAuth callback path without adding new public cancellation plumbing.

## Usage examples

There is no new command or API. Existing USER and WORKSPACE OAuth creation streams behave the same while active. If the stream is disconnected or the server shuts down before a callback completes, accepted callback sockets are closed with the operation instead of surviving as detached workers.

The shared source-import authorization-code flow receives the same abort-on-drop cleanup when its owning future is cancelled.

## Validation

- Focused callback, source-import, and server-shutdown regression selection: 6/6 passed.
- Strict all-target coral-app Clippy passed.
- `make rust-checks`: 1,749/1,749 tests passed, 3 skipped; formatting, workspace Clippy, and rustdoc clean.
- The new deterministic regression proves an accepted idle preconnection receives zero response bytes after the callback receiver is dropped.

## Review

- Exact final five-lane review swarm: zero findings and zero questions.
- Eleven credential flows enumerated; all safe, with no unknown transport guards or verdicts.
- Independent supervisor missed-risk sweep found no remaining P1/P2 risk at 0.97 confidence.

## Stack

- Direct draft child of #1715 at exact parent `7017a6abb58981f515439fa4ba9019d0b1781e09`.
- Submitted through an isolated, exact-parent Graphite checkout; pre-submit was one `Create` and post-submit is one `No-op`.
- PR #1460 remains the immutable mission ancestor at `23074065dc7220ee327edbc5cdf13791a6c6b710`.

### DIFF
--- a/crates/coral-app/src/credentials/oauth.rs
+++ b/crates/coral-app/src/credentials/oauth.rs
@@ -21,7 +21,7 @@ use serde_json::{Value, json};
 use sha2::{Digest as _, Sha256};
 use tokio::io::{AsyncReadExt as _, AsyncWriteExt as _};
 use tokio::net::TcpListener;
-use tokio::sync::mpsc;
+use tokio::task::JoinSet;
 use url::{Host, Url, form_urlencoded};
 use uuid::Uuid;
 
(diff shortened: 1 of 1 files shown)


sauldhernandez:
* **#1716** <a href="https://app.graphite.com/github/pr/withcoral/coral/1716?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1716?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `sauldhernandez/dsl-v4-identity-v3-b5d1b-oauth-shutdown-cancellation`


<h2></h2>

PR #1731, opened 2026-07-13
feat(desktop): release-shaped universal mac packaging

martinrajdl:
Part 1/4 of splitting #1519 into reviewable pieces. This PR establishes the local and CI packaging path; it does not ship a desktop release.

- `npm run package:mac` builds a **universal** (arm64 + x86_64) macOS app: `CORAL_DESKTOP_UNIVERSAL=1` makes `stage-coral.mjs` compile the Coral sidecar for both targets and combine them with `lipo`.
- electron-builder produces **DMG + ZIP** with version-less `coral-desktop-mac-*` names, so a future `releases/latest/download` link can remain stable across releases.
- The `github` publish provider generates `latest-mac.yml` update metadata and embeds `app-update.yml`; nothing consumes them yet (the updater lands in part 4).
- Unsigned packaging and artifact verification live in the reusable `.github/workflows/desktop-package.yml`. Validate calls it only for distribution-sensitive PRs: desktop files, packaging/validation/release workflow changes, Cargo workspace or toolchain configuration, and crate manifests or build scripts. Ordinary Rust, Reef, and UI changes continue to use their existing checks.
- The reusable workflow also runs nightly on `main` and by manual dispatch as an unsigned packaging preflight. Every run verifies the DMG, ZIP, blockmaps, and `latest-mac.yml`; PR and manual runs retain the roughly 700 MB artifact for seven days, while nightly runs verify it without uploading it.
- PRs may restore the trusted desktop Rust cache, but only nightly and manual runs can update it. The stale `actions/setup-node` pins are also refreshed so the workflow security audit recognizes the current immutable revision.
- Contributor guidance now distinguishes distribution-sensitive PRs from ordinary Rust/Reef/UI validation and records nightly integrated-package coverage.

The validation artifact is unsigned and is never a release input. Signing/notarization and desktop release publishing remain in stacked follow-ups #1732 and #1735; that release path will rebuild the desktop package from a clean checkout.

Based on the paths changed by the 50 most recently created PRs that later merged (created July 7–17, 2026), the narrowed filter is expected to reduce this expensive PR job from roughly 84% of PRs to roughly 6%, while retaining packaging coverage on sensitive changes and nightly integration coverage.

Stack: this → #1732 signing → #1735 release publish → #1736 auto-update. Release remains ordered before the updater deliberately: the first release verifies signing/notarization/upload end to end, and the release after #1736 verifies the update flow against it.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -40,6 +40,7 @@ jobs:
       ui-tests: ${{ steps.filter.outputs.ui-tests }}
       reef-checks: ${{ steps.filter.outputs.reef-checks }}
       desktop-checks: ${{ steps.filter.outputs.desktop-checks }}
+      desktop-package-macos: ${{ steps.filter.outputs.desktop-package-macos }}
       proto-lint: ${{ steps.filter.outputs.proto-lint }}
       source-lint: ${{ steps.filter.outputs.source-lint }}
       docs-freshness: ${{ steps.filter.outputs.docs-freshness }}
@@ -96,6 +97,15 @@ jobs:
             desktop-checks:
               - '.github/workflows/validate.yml'
               - 'apps/desktop/**'
+            desktop-package-macos:
+              - '.github/workflows/validate.yml'
+              - '.github/workflows/release.yml'
+              - 'Cargo.toml'
+              - 'Cargo.lock'
+              - 'crates/**'
+              - 'apps/desktop/**'
+              - 'apps/reef/**'
+              - 'apps/ui/**'
             proto-lint:
               - '.github/workflows/validate.yml'
               - 'Makefile'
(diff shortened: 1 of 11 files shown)


apps/desktop/electron-builder.config.ts
```diff
@@ -3,7 +3,13 @@ import type { Configuration } from 'electron-builder'
 const config: Configuration = {
   appId: 'com.withcoral.desktop',
   productName: 'Coral',
-  artifactName: 'coral-desktop-${os}-${arch}.${ext}',
+  publish: [
```
jsummerfield, line 7: Could we make the non-release publishing policy explicit for all package scripts? Adding this provider globally means the existing `package` and `package:dmg` commands inherit electron-builder's CI/tag publish default because they do not pass `--publish never`. `package:mac` is protected, but a token-bearing tagged invocation of either older command can publish before Part 3. Adding `--publish never` to every non-release packaging command would preserve the PR's “nothing ships until Part 3” contract.

.github/workflows/validate.yml
```diff
@@ -96,6 +97,15 @@ jobs:
             desktop-checks:
               - '.github/workflows/validate.yml'
               - 'apps/desktop/**'
+            desktop-package-macos:
+              - '.github/workflows/validate.yml'
+              - '.github/workflows/release.yml'
+              - 'Cargo.toml'
```
jsummerfield, line 105: Could this filter include `rust-toolchain.toml`? This job is the package gate for Rust inputs, but a toolchain-only change currently skips it. That leaves compiler upgrades unvalidated against the universal build.

apps/desktop/electron-builder.config.ts
```diff
@@ -37,9 +43,10 @@ const config: Configuration = {
     },
   ],
   mac: {
+    artifactName: 'coral-desktop-mac-${arch}.${ext}',
```
jsummerfield: Could we retain the existing global `artifactName` and only add the mac targets/provider changes? Moving the key under `mac` removes stable names from generic Windows/Linux packaging, which is unrelated to this mac-only PR.

jsummerfield (commented):

martinrajdl:
good calls on all 3, fixes in the two comments 

1. --publish never on all non-release scripts → done on package and package:dmg.
2. rust-toolchain.toml in the filter → done (added both rust-toolchain.toml and rust-toolchain).
3. Retain global artifactName → done (dropped the mac-only override).

martinrajdl:
The zizmor validation seems unrelated but needs fixing on main it seems

AlbertQM:
Took over this PR; added a separate job for the app packaging. We don't want this to run on every PR, just on release.

To test:
- Let this PR finish building the package (40+ min)
- Download those assets

```
gh run download 29574506791 \
    --repo withcoral/coral \
    --name coral-desktop-macos-unsigned \
    --dir /tmp/coral-desktop-pr1731
```

image

Open the .zip, Coral runs as expected.

image
image

lazybean:
> We don't want this to run on every PR, just on release.
>   
>   

Thank you, maybe some tiny updates needed:

- PR body says the reusable workflow “runs nightly on `main` and by manual dispatch”.
- `AGENTS.md` says Validate calls desktop packaging for distribution-sensitive PRs and mentions manual dispatch preflights.
- Actual workflow only has `workflow_call` and `workflow_dispatch`.
- `validate.yml` currently calls it for matching PRs, including this PR.

lazybean (approved):

lazybean:
> > We don't want this to run on every PR, just on release.
>
> Thank you, maybe some tiny updates needed:
>
> - PR body says the reusable workflow “runs nightly on `main` and by manual dispatch”.
> - `AGENTS.md` says Validate calls desktop packaging for distribution-sensitive PRs and mentions manual dispatch preflights.
> - Actual workflow only has `workflow_call` and `workflow_dispatch`.
> - `validate.yml` currently calls it for matching PRs, including this PR.

​Actually, I think the goal was to run on every PR, as a verification.  
THen on releaese we run the sign build

lazybean (changes requested):
That should definitively run on every PR

lazybean:
it is actually very long to run

lazybean (approved):

PR #1732, opened 2026-07-13
feat(desktop): mac signing and notarization config

martinrajdl:
Part 2/4 of splitting #1519 (stacked on #1731).

- Hardened runtime + JIT/unsigned-memory/library-validation entitlements (required for Electron).
- `CORAL_DESKTOP_RELEASE=1` gates strict behavior: release builds `forceCodeSigning` and notarize; local/CI packaging stays unsigned.

**Inert until something sets `CORAL_DESKTOP_RELEASE=1`** — that happens in the release workflow (part 3). Signing/notarization itself can only be verified on a real release run since the secrets live in the `release` environment.


apps/desktop/resources/entitlements.mac.inherit.plist
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+  <dict>
+    <key>com.apple.security.cs.allow-jit</key>
```
jsummerfield, line 10: Could you add comments why so we can remember

apps/desktop/electron-builder.config.ts
```diff
@@ -1,8 +1,11 @@
 import type { Configuration } from 'electron-builder'
 
+const releaseBuild = process.env.CORAL_DESKTOP_RELEASE === '1'
+
 const config: Configuration = {
   appId: 'com.withcoral.desktop',
   productName: 'Coral',
+  forceCodeSigning: releaseBuild,
```
jsummerfield: Could we make unsigned non-release builds explicit? `forceCodeSigning: false` only makes a missing identity non-fatal; app-builder-lib still auto-discovers a Developer ID identity from the user's keychain. A developer with a certificate therefore gets a signed local package despite the PR contract. `mac.identity: releaseBuild ? undefined : null` (or disabling identity discovery in every non-release script) would make the behavior deterministic.

apps/desktop/electron-builder.config.ts
```diff
@@ -45,7 +48,11 @@ const config: Configuration = {
   mac: {
     artifactName: 'coral-desktop-mac-${arch}.${ext}',
     category: 'public.app-category.developer-tools',
+    entitlements: 'resources/entitlements.mac.plist',
+    entitlementsInherit: 'resources/entitlements.mac.inherit.plist',
+    hardenedRuntime: true,
     icon: 'resources/icons/icon.icns',
+    notarize: releaseBuild,
```
jsummerfield: Could release mode preflight notarization credentials instead of relying on this boolean? In locked app-builder-lib 26.15.3, `notarize: true` still warns and returns successfully when it cannot construct credential options. The later release workflow's stapler validation catches this, but this config alone is not the strict mode described here. An explicit credential check before packaging would fail at the cause.

jsummerfield (commented):

martinrajdl:
1. "Add comments why" on the entitlements → done.

2. mac.identity deterministic unsigned → done 

3. notarize-credential-preflight - addressed in #1735's xtask preflight

lazybean (approved):

PR #1733, opened 2026-07-13
feat(desktop): auto-update from GitHub Releases

martinrajdl:
Part 4/4 of splitting #1519 (stacked on #1734). Closes the update loop: the first release after this merges ships an app that keeps itself current.

- `electron-updater` (GitHub provider): checks 5s after startup **and every 4h thereafter** (long-running sessions pick up releases without a restart), auto-downloads, notifies, installs on quit.
- Manual **Coral → Check for Updates...** menu item with result dialogs.
- Guarded by `app.isPackaged` — dev builds are untouched.
- `release-please` now also syncs the desktop `package-lock.json` version so the packaged app version exactly matches the release tag the updater compares against (the lockfile picks up the previously-missed 0.5.2 sync).

End-to-end verification once merged: install the app from the previous release, cut the next one, watch it self-update.

Lockfile note: the diff carries `peer` flag churn — it is byte-identical to a fresh `npm install --package-lock-only`, i.e. npm's own normalization, not hand edits.

### DIFF
--- a/apps/desktop/README.md
+++ b/apps/desktop/README.md
@@ -28,7 +28,7 @@ app lifecycle, window and menu, and MCP client configuration.
 - Coral MCP configuration for Codex and Claude Code — from the Settings page and
   the app menu — via `add-mcp`
 - signed, notarized macOS DMG and ZIP packaging via `electron-builder`
-- GitHub Releases update metadata for packaged desktop builds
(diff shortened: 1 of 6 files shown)


martinrajdl:
Auto-closed when its base branch was briefly deleted during a stack reorder. Superseded by #1736.

PR #1734, opened 2026-07-13
feat(desktop): publish mac app in the release workflow

martinrajdl:
Part 3/4 of splitting #1519 (stacked on #1732). **This is the go-live switch for the download** — the first release after this merges ships the desktop DMG/ZIP.

- `build-desktop-macos` release job: builds, signs, and notarizes the universal mac app. Decodes the App Store Connect `.p8` key to a file first because `@electron/notarize` expects `APPLE_API_KEY` to be a **path**, not base64 contents (would have failed the first real release run).
- `APPLE_TEAM_ID` from a release environment variable, or extracted from `APPLE_CODESIGN_IDENTITY`; fails loudly otherwise.
- CI verifies `codesign`/`stapler`, `latest-mac.yml` presence, and exact artifact counts before upload; a failed desktop build fails the whole release, so update metadata can never silently go missing.
- Uploads `coral-desktop-mac-universal.dmg` + `.zip` (+ blockmaps, `latest-mac.yml`) to the same GitHub Release as the CLI, included in `checksums.sha256`. Stable download link: `https://github.com/withcoral/coral/releases/latest/download/coral-desktop-mac-universal.dmg`

Known and accepted: a release cut between this PR and #1733 ships an app **without the auto-updater** — installers from that window must re-download manually once. The link is not yet published anywhere, so exposure is effectively zero.

Prerequisite before merging: `APPLE_TEAM_ID` set as a release environment variable, or present in `APPLE_CODESIGN_IDENTITY`.

PR #1735, opened 2026-07-13
feat(desktop): publish mac app in the release workflow

martinrajdl:
Part 3/4 of splitting #1519 (stacked on #1732; replaces #1734, which Graphite auto-closed during a restack). **This is the go-live switch for the download** — the first release after this merges ships the desktop DMG/ZIP.

- `build-desktop-macos` release job: builds, signs, and notarizes the universal mac app. Decodes the App Store Connect `.p8` key to a file first because `@electron/notarize` expects `APPLE_API_KEY` to be a **path**, not base64 contents (would have failed the first real release run).
- `APPLE_TEAM_ID` from a release environment variable, or extracted from `APPLE_CODESIGN_IDENTITY`; fails loudly otherwise.
- CI verifies `codesign`/`stapler`, `latest-mac.yml` presence, and exact artifact counts before upload; a failed desktop build fails the whole release, so update metadata can never silently go missing.
- Uploads `coral-desktop-mac-universal.dmg` + `.zip` (+ blockmaps, `latest-mac.yml`) to the same GitHub Release as the CLI, included in `checksums.sha256`. Stable download link: `https://github.com/withcoral/coral/releases/latest/download/coral-desktop-mac-universal.dmg`

Known and accepted: a release cut between this PR and the auto-update PR ships an app **without the auto-updater** — installers from that window must re-download manually once. The link is not yet published anywhere, so exposure is effectively zero.

Prerequisite before merging: `APPLE_TEAM_ID` set as a release environment variable, or present in `APPLE_CODESIGN_IDENTITY`.


.github/workflows/release.yml
```diff
@@ -318,9 +318,100 @@ jobs:
...
+      - name: Set up Node
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
+        with:
+          node-version-file: apps/desktop/.node-version
+          package-manager-cache: false
+
+      - name: Install desktop dependencies
+        run: npm ci --prefix apps/desktop
+
+      - name: Build, sign, and notarize desktop app
```
jsummerfield: Could this orchestration move behind an `xtask` command, leaving the workflow to provide environment values and invoke it? This step now owns team-ID resolution, secret decoding, signing/notarization, and postcondition checks, while `AGENTS.md` explicitly assigns release signing/notarization automation to `xtask/src/release.rs`. Centralizing it there would also make the preflight and cleanup testable.

.github/workflows/release.yml
```diff
@@ -318,9 +318,100 @@ jobs:
...
+            exit 1
+          fi
+          codesign --verify --deep --strict --verbose=2 "$app_bundle"
+          xcrun stapler validate "$app_bundle"
+          test -s apps/desktop/dist/latest-mac.yml
+
+          shopt -s nullglob
+          dmg_artifacts=(apps/desktop/dist/coral-desktop-*.dmg)
+          zip_artifacts=(apps/desktop/dist/coral-desktop-*.zip)
+          if [ "${#dmg_artifacts[@]}" -ne 1 ] || [ "${#zip_artifacts[@]}" -ne 1 ]; then
```
jsummerfield: Could this verifier require every update asset referenced by `latest-mac.yml`, especially the ZIP blockmap? It currently counts only the DMG and ZIP. `upload-artifact` succeeds when those files and the YAML exist, and the later `nullglob` silently drops a missing blockmap, so the “metadata can never silently go missing” guarantee is not enforced. Sharing this verifier with Validate would also prevent the two checks from drifting.

.github/workflows/release.yml
```diff
@@ -332,9 +423,15 @@ jobs:
           merge-multiple: true
       - name: Generate checksums
         run: |
+          set -euo pipefail
+          shopt -s nullglob
           cd artifacts
-          sha256sum coral-*.tar.gz > checksums.sha256
-          sha256sum coral-*.zip >> checksums.sha256
+          artifacts=(coral-*.tar.gz coral-*.zip coral-*.dmg)
```
jsummerfield: Could the checksum inventory include `coral-*.blockmap` and `latest-mac.yml`? Both are published release assets and the PR explicitly says they are included in checksums, but this array currently selects only archives and DMGs.

jsummerfield (commented):

martinrajdl:
1. Move orchestration into an xtask command → done (release-desktop-macos-package).

2. Verifier requires every latest-mac.yml-referenced asset + the ZIP blockmap, shared with Validate → done via verify-dist.mjs 

3. Checksums include coral-*.blockmap and latest-mac.yml → done.

lazybean (approved):

PR #1736, opened 2026-07-13
feat(desktop): auto-update from GitHub Releases

martinrajdl:
Part 4/4 of splitting #1519 (stacked on #1735; replaces #1733, which was auto-closed when its base branch was deleted). Closes the update loop: the first release after this merges ships an app that keeps itself current.

- `electron-updater` (GitHub provider): checks 5s after startup **and every 4h thereafter** (long-running sessions pick up releases without a restart), auto-downloads, notifies, installs on quit.
- Manual **Coral → Check for Updates...** menu item with result dialogs.
- Guarded by `app.isPackaged` — dev builds are untouched.
- `release-please` now also syncs the desktop `package-lock.json` version so the packaged app version exactly matches the release tag the updater compares against (the lockfile picks up the previously-missed 0.5.2 sync).

End-to-end verification once merged: install the app from the previous release, cut the next one, watch it self-update.

Lockfile note: the diff carries `peer` flag churn — it is byte-identical to a fresh `npm install --package-lock-only`, i.e. npm's own normalization, not hand edits.


apps/desktop/src/main/auto-update.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  await dialog.showMessageBox({
+    type: 'info',
+    message: `Coral ${result.updateInfo.version} is downloading`,
+    detail: 'You will be notified when the update is ready. It will install after Coral quits.',
+  })
+}
+
+export function installAutoUpdater(): void {
```
jsummerfield, line 47: Could we add automated coverage for this updater state before landing it? This introduces one-time installation, startup/periodic timers, interactive/background branches, and downloaded-update state, while desktop CI still only type-checks. Focused tests with a fake updater and fake timers would catch overlap, repeated-update, and error-path behavior.

apps/desktop/src/main/auto-update.ts
```diff
@@ -0,0 +1,108 @@
...
+  if (!app.isPackaged || installed) return
```
jsummerfield: Could updater activation also require `process.platform === 'darwin'`? This stack publishes only `latest-mac.yml`. Existing generic packaging can still produce packaged Windows/Linux apps, which would show the menu and poll for `latest.yml` or `latest-linux*.yml` that do not exist. The same platform guard should drive the menu.

apps/desktop/src/main/auto-update.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+export async function checkForDesktopUpdates({ interactive }: { interactive: boolean }): Promise<void> {
+  if (!app.isPackaged) {
+    if (interactive) await showManualResult(null)
+    return
+  }
+
+  try {
+    const result = await autoUpdater.checkForUpdatesAndNotify()
```
jsummerfield: `checkForUpdatesAndNotify()` does not dedupe notifications here. In locked electron-updater 6.8.9, only `checkForUpdates()` dedupes the check promise; every caller attaches another notification continuation to the shared download promise. Two overlapping calls produced two notifications in a direct harness, and later cached checks can notify again. Could we call `checkForUpdates()` and notify from one `update-downloaded` listener while tracking the staged version?

release-please-config.json
```diff
@@ -11,7 +11,9 @@
       "package-name": "coral",
       "extra-files": [
         "Cargo.toml",
-        { "type": "json", "path": "apps/desktop/package.json", "jsonpath": "$.version" }
+        { "type": "json", "path": "apps/desktop/package.json", "jsonpath": "$.version" },
```
jsummerfield, line 14: These paths correctly keep the normal release-please flow in sync, but the release workflow also accepts any reachable manual `vX.Y.Z` tag and never compares it with this version. That can publish `latest-mac.yml` with a stale desktop version. Could the release preflight assert that `apps/desktop/package.json` equals `${TAG_NAME#v}` before building?

jsummerfield (commented):

martinrajdl:
1. Automated updater tests → done (vitest suite).

2. darwin guard on activation + menu → done (and tightened further to require a signed release build).

3. checkForUpdates() + one update-downloaded listener tracking the staged version → done 

4. tag/version - addressed in #1735's release preflight.

lazybean (approved):

PR #1740, opened 2026-07-13
docs(search): document catalog search

mystic123:
## Summary

- add a human-first **Search the catalog** guide focused on catalog discovery, with a real output excerpt and a handoff to the existing query tutorial
- link catalog search from the docs homepage and place it after **Use Coral over MCP** in Guides navigation
- explain catalog search over MCP without requiring users to name the `search` tool in their prompts
- document machine-readable search fields and provider statuses in the CLI reference
- list catalog search under **Now** and search over observed values under **Next** on the roadmap

## Why

Catalog search shipped through the CLI and MCP in #1462, but the public documentation did not make it discoverable or explain its boundaries. This documents the released capability as **catalog search**: it finds tables, table functions, columns, filters, arguments, and result columns, but does not retrieve source rows or present the future observed-value and source-native search providers as currently available.

## Validation

- `make docs-check`
- `npx --yes mint validate`
- `npx --yes mint broken-links --check-anchors --check-redirects`
- `git diff --check main...HEAD`

### DIFF
--- a/apps/docs/guides/search-the-catalog.mdx
+++ b/apps/docs/guides/search-the-catalog.mdx
@@ -7,7 +7,7 @@ Use catalog search when you know what data you need but not which table or table
 
 Catalog search helps you find where data can be queried. It does not retrieve rows from GitHub, Slack, or another connected source.
 
-## Find the right table or table function
+## Use catalog search from the CLI
 
 ```shellscript
 coral search github issues assigned to me
--- a/apps/docs/guides/search-the-catalog.mdx
+++ b/apps/docs/guides/search-the-catalog.mdx
@@ -3,9 +3,11 @@ title: "Search the catalog"
 description: "Find relevant tables, table functions, columns, and filters before you write SQL."
 ---
 
-Use catalog search when you know what data you need but not which table or table function exposes it. Enter a natural language query and Coral ranks matching tables, table functions, columns, and filters from the sources installed in your current workspace.
+Catalog search helps you find the right table or table function when you know what you want to query but not where to start. Enter a natural language description, and Coral ranks matching tables, table functions, columns, and filters in your current workspace.
 
-Catalog search helps you find where data can be queried. It does not retrieve rows from GitHub, Slack, or another connected source.
+<Note>
+Catalog search looks only at Coral's catalog metadata. It does not query connected sources or return source rows.
+</Note>
 
 ## Use catalog search from the CLI
 
@@ -26,7 +28,7 @@ Results
 
 The `SQL` line gives you the table reference. For a table function, `SQL reference` identifies the function and `Call` shows its required arguments.
 
-Catalog search only identifies these queryable surfaces; it does not execute them or retrieve rows. To fetch rows, use the returned `SQL` reference or `Call` example with `coral sql`. See [Query your data](/getting-started/quickstart#3-query-your-data) for a complete example.
+To fetch rows, use the result's `SQL` reference or `Call` example in a query with `coral sql`. See [Query your data](/getting-started/quickstart#3-query-your-data) for a complete example.
 
 By default, Coral returns up to 10 results. Use `--limit` to request between 1 and 50. See the complete [`coral search` reference](/reference/cli-reference#coral-search).
 
--- a/apps/docs/guides/search-the-catalog.mdx
+++ b/apps/docs/guides/search-the-catalog.mdx
@@ -1,0 +1,46 @@
+---
+title: "Search the catalog"
+description: "Find relevant tables, table functions, columns, and filters before you write SQL."
+---
+
+Use catalog search when you know what data you need but not which table or table function exposes it. Enter a natural language query and Coral ranks matching tables, table functions, columns, and filters from the sources installed in your current workspace.
+
+Catalog search helps you find where data can be queried. It does not retrieve rows from GitHub, Slack, or another connected source.
+
+## Find the right table or table function
+
+```shellscript
+coral search github issues assigned to me
+```
+
+Among the ranked results, you might see:
+
+```text
+Results
+1. [catalog_metadata] table github.issues
+   SQL: github.issues
+```
+
+The `SQL` line gives you the table reference. When a result is a table function, Coral also shows a `Call` example with its arguments.
+
+Catalog search only identifies these queryable surfaces; it does not execute them or retrieve rows. To fetch rows, use the returned `SQL` reference or `Call` example with `coral sql`. See [Query your data](/getting-started/quickstart#3-query-your-data) for a complete example.
+
+## Use catalog search over MCP
+
+Coral exposes catalog search to connected agents through its read-only `search` MCP tool. An agent can use it to find the relevant table or table function before querying your data.
+
+> "Use Coral to show the open GitHub issues assigned to me."
+
+Ask for the result you want; you do not need to mention the `search` tool. The agent can decide when catalog search is needed.
+
+[Set up Coral over MCP](/guides/use-coral-over-mcp).
+
+## Search options
+
+By default, Coral returns up to 10 results. Use `--limit` to request between 1 and 50:
+
+```shellscript
+coral search --limit 5 slack messages text
+```
+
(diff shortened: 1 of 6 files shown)


mystic123:
* **#1740** <a href="https://app.graphite.com/github/pr/withcoral/coral/1740?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1740?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/docs/index.mdx
```diff
@@ -120,6 +122,9 @@ For the full internals, crates, gRPC transport, DataFusion integration, see the
 ## Quick links
 
 <CardGroup cols={2}>
+  <Card title="Search the catalog" href="/guides/search-the-catalog">
+    Find the right table, table function, column, or filter before you write SQL
+  </Card>
```
AlbertQM: The search is mostly for agents, no? I wouldn't add this to the top of the quick links.

apps/docs/index.mdx
```diff
@@ -9,6 +9,8 @@ Agents are able to make fewer, more precise tool calls with Coral than they do w
 
 Coral supports a number of [popular data sources](/reference/bundled-sources) bundled in, plus [community source specs](/reference/community-sources) you can import with `coral source add --file`. You can also extend it to accommodate others by writing your own [source specs](/reference/source-spec-reference). You can run SQL from the [CLI](/reference/cli-reference) or through [MCP](/guides/use-coral-over-mcp). And everything is local; your data, credentials and usage history never leave your machine.
 
+When you know what data you need but not which table or table function exposes it, [search Coral's catalog](/guides/search-the-catalog) with a natural language query before you write SQL.
+
```
AlbertQM: Let's keep this clean please, no need to add this sentence at the top of the first page of our docs

apps/docs/guides/search-the-catalog.mdx
```diff
@@ -0,0 +1,46 @@
+---
+title: "Search the catalog"
```
AlbertQM, line 2: We don't mention "Catalog" anywhere else in the docs, except from some really hidden places in the reference. Are we settling on this name?

WDYT about making this title a bit less abstract:
- Explore your data efficiently
- Find the data you need
- Find data in your catalog

But please feel free to come up with something yourself, you've built this after all :P

apps/docs/guides/search-the-catalog.mdx
```diff
@@ -0,0 +1,46 @@
...
+description: "Find relevant tables, table functions, columns, and filters before you write SQL."
+---
+
+Use catalog search when you know what data you need but not which table or table function exposes it. Enter a natural language query and Coral ranks matching tables, table functions, columns, and filters from the sources installed in your current workspace.
+
+Catalog search helps you find where data can be queried. It does not retrieve rows from GitHub, Slack, or another connected source.
```
AlbertQM: I re-wrote this manually as a suggestion, but please feel free to rewrite _manually_ again! I'm sure you have some cool context in your head that I'm missing completely, and that AI can't express in a nice-to-read way

```suggestion
Catalog search helps you find the data you need when you don't know which table or table function exposes it. You can ask Coral a natural language question and get matching tables, table functions, columns, and filters from the sources installed in your current workspace.

<Note>
Catalog search does not retrieve actual data from connected data sources. It only gives hints on where that data might live.
</Note>
```

apps/docs/guides/search-the-catalog.mdx
```diff
@@ -0,0 +1,46 @@
...
+Results
+1. [catalog_metadata] table github.issues
+   SQL: github.issues
+```
+
+The `SQL` line gives you the table reference. When a result is a table function, Coral also shows a `Call` example with its arguments.
+
+Catalog search only identifies these queryable surfaces; it does not execute them or retrieve rows. To fetch rows, use the returned `SQL` reference or `Call` example with `coral sql`. See [Query your data](/getting-started/quickstart#3-query-your-data) for a complete example.
+
+## Use catalog search over MCP
```
AlbertQM, line 35: Is this is "Use catalog search over MCP", should the other be "Use catalog search over CLI"?

apps/docs/guides/search-the-catalog.mdx
```diff
@@ -0,0 +1,46 @@
...
+coral search github issues assigned to me
+```
+
+Among the ranked results, you might see:
+
+```text
+Results
+1. [catalog_metadata] table github.issues
+   SQL: github.issues
+```
```
AlbertQM, line 27: Love an example! WDYT about extending this a bit, and maybe adding a table function too?

apps/docs/guides/search-the-catalog.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+## Search options
+
+By default, Coral returns up to 10 results. Use `--limit` to request between 1 and 50:
+
+```shellscript
+coral search --limit 5 slack messages text
+```
+
+Use `--json` when a script or another program needs the structured response. See [`coral search` in the CLI reference](/reference/cli-reference#coral-search) for the complete response fields and provider-status details.
```
AlbertQM: I think we can condense this whole section into just the first sentence, and then link to the full reference like you've already done! 😄 
And would probably move this under the first section (where we use `coral cli`), so users are aware that's an option there


```suggestion
By default, Coral returns up to 10 results. Use `--limit` to request between 1 and 50. See the complete [`coral search` reference](/reference/cli-reference#coral-search).
```

apps/docs/guides/use-coral-over-mcp.mdx
```diff
@@ -209,9 +209,9 @@ When your agent has many tools available, name Coral explicitly:
 
 > "Use Coral to compare recent deploy errors in Datadog with related Slack messages."
 
-> "Use Coral. First inspect the catalog, then query the right tables."
+> "Use Coral. Search the catalog for the relevant table or table function, then query the best match."
 
-Good agents will inspect Coral's catalog before writing SQL, then answer from query results. If an answer needs data from multiple connected sources, ask the agent to combine the data in Coral when possible.
+When the right table or table function is not obvious, ask the agent to search the catalog before querying. You do not need to know the MCP tool names; Coral exposes both catalog search and read-only SQL through MCP. Catalog search identifies queryable surfaces but does not retrieve source rows. For a CLI walkthrough, see [Search the catalog](/guides/search-the-catalog). If an answer needs data from multiple connected sources, ask the agent to combine the data in Coral when possible.
```
AlbertQM: How do you feel about the redundancy? Ie could we keep this leaner, and only talk about that the search won't return rows only inside the guide? and what about things like "You do not need to know the MCP tool names"?

```suggestion
When the right table or table function is not obvious, ask the agent to [search the catalog](/guides/search-the-catalog) before querying. Coral exposes both catalog search and read-only SQL through MCP. Catalog search identifies queryable surfaces but does not retrieve source rows. If an answer needs data from multiple connected sources, ask the agent to combine the data in Coral when possible.
```

AlbertQM (commented):
Thanks for adding these and updating the roadmap! The functionality makes sense, looking forward for my agent to retrieve my data even more efficiently and accurately.

Left some hand-written comments as guideline; feel free to adopt as-is, or tweak/overhaul completely by hand like a caveman 😂
Haven't refined our dos guardrails enough for agents to do their thing in a non-sloppy way.

apps/docs/guides/search-the-catalog.mdx
mystic123: It's Catalog search rn, Observed Values search soon and then Fan out in the close future, although I think only the first name is settled.  
I prefer to keep "catalog" explicit, because it's only metadata we are searching for now, not the "data". And even with observed values search later, we can't really say it's "finding the data you need", because we can only search in the sample we observed. I want to specifically avoid false negatives, becase the absence of data we return doesn't imply the true absence.

AlbertQM: Good point, thanks!

AlbertQM (approved):
Thank you!

PR #1505, opened 2026-07-03
feat(app): add request principal boundary

sauldhernandez:
## Intent

Establish the app-owned request-principal boundary needed for multi-user identity without changing client transport behavior or choosing a hosted authentication policy.

## What it enables

- A validated `UserPrincipal` model and provider abstraction.
- Fail-closed `Unauthenticated` error mapping at the app boundary.
- A request context layer that makes the selected principal available to domain services.
- A compatible local single-user default while later stack units add centralized server authentication.

## Usage examples

Existing local-server callers need no changes. The default provider selects the reserved local principal, and the transport boundary installs it into `RequestContext` before dispatching a request.

PR #1745, opened 2026-07-13
feat(app): add request principal boundary

sauldhernandez:
## Intent

Establish the app-owned request-principal boundary needed for multi-user identity while preserving existing local single-user behavior. This is a standalone, `main`-based replacement for #1505 so later identity work can depend on it without inheriting the earlier stack history.

## What it enables

- A validated `UserPrincipal` model and provider abstraction.
- A configurable `UserPrincipalProvider` composition seam on `ServerBuilder`, with the reserved local principal as the default.
- One authentication and request-context layer across every registered native gRPC and gRPC-Web route.
- Fail-closed gRPC status mapping for unauthenticated, unavailable, and internal provider failures.
- Request-scoped principal access for app-owned domain services while static UI assets remain outside the authentication gate.

## Usage examples

Existing local-server callers require no changes. The default provider selects the reserved local principal and installs it into `RequestContext` before dispatching each registered gRPC request.

Product runtimes can install a metadata-aware provider without rebuilding individual services:

```rust
let server = ServerBuilder::new()
    .with_user_principal_provider(Arc::new(MyPrincipalProvider))
    .start()
    .await?;
```

## Validation

- `make rust-checks` — passed: formatting, workspace-wide Clippy with warnings denied, 1,598 tests passed (2 skipped), and Rust documentation built.
- Focused native gRPC coverage proves that `ServerBuilder` installs and enforces a custom rejecting provider.
- Focused gRPC-Web coverage proves registered routes are authenticated while HTML, asset, SPA, and unknown-route fallbacks remain ungated.
- 529 additions and 38 deletions across 8 files (567 changed lines), within the repository's 600-line PR hard cap.

### DIFF
--- a/crates/coral-app/src/request_context.rs
+++ b/crates/coral-app/src/request_context.rs
@@ -1,8 +1,5 @@
 //! Request-scoped app context selected at the transport boundary.
 
-use tonic::{Request, Status};
-
-use crate::bootstrap::{AppError, app_status};
 use crate::identity::UserPrincipal;
 
 /// Request-scoped data that domain services may need after authentication.
@@ -20,29 +17,11 @@ impl RequestContext {
         Self { principal }
     }
 
-    // Installed server-wide by the immediate P1a child and consumed at service
-    // boundaries by P1b.
     #[cfg_attr(
         not(test),
         expect(
             dead_code,
-            reason = "installed server-wide by P1a and consumed at service boundaries by P1b"
-        )
-    )]
-    pub(crate) fn from_request<T>(request: &Request<T>) -> Result<Self, Status> {
-        request.extensions().get::<Self>().cloned().ok_or_else(|| {
-            app_status(AppError::Unauthenticated(
-                "missing request principal".to_string(),
-            ))
-        })
-    }
-
-    // Consumed by the QueryContext introduced in the immediate P1b child.
-    #[cfg_attr(
-        not(test),
-        expect(
-            dead_code,
-            reason = "consumed by the QueryContext introduced in the P1b child"
+            reason = "request principals stay encapsulated for downstream authorization and attribution"
         )
     )]
     pub(crate) fn principal(&self) -> &UserPrincipal {
@@ -60,13 +39,5 @@ mod tests {
         let context = RequestContext::new(principal.clone());
 
         assert_eq!(context.principal(), &principal);
-    }
-
-    #[test]
-    fn rejects_requests_without_a_principal() {
-        let status = RequestContext::from_request(&Request::new(()))
-            .expect_err("missing context should fail closed");
-
-        assert_eq!(status.code(), tonic::Code::Unauthenticated);
     }
 }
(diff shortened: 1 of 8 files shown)


sauldhernandez:
* **#1745** <a href="https://app.graphite.com/github/pr/withcoral/coral/1745?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1745?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-app/src/request_context.rs
```diff
@@ -0,0 +1,72 @@
...
+    // Installed server-wide by the immediate P1a child and consumed at service
+    // boundaries by P1b.
+    #[cfg_attr(
+        not(test),
+        expect(
+            dead_code,
+            reason = "installed server-wide by P1a and consumed at service boundaries by P1b"
+        )
+    )]
+    pub(crate) fn from_request<T>(request: &Request<T>) -> Result<Self, Status> {
```
jsummerfield: Could we keep `RequestContext` as a data-only value and move this extraction and failure mapping into `transport.rs`? As written, the context model owns `tonic::Request` lookup and creates a `tonic::Status`, which couples every downstream context to gRPC. A private transport helper such as `require_request_context(&Request<T>)` would preserve the fail-closed behavior while keeping the context reusable.

crates/coral-app/src/transport.rs
```diff
@@ -49,31 +64,105 @@ impl Extractor for MetadataExtractor<'_> {
     }
 }
 
-/// Wraps a generated tonic service and stores inbound request context on the request.
+/// Compatibility wrapper used until the next stack unit installs one layer
```
jsummerfield: Could we apply `GrpcRequestContextLayer` once around the complete route tree instead of landing this compatibility wrapper? The wrapper duplicates the `Service` and `NamedService` surface, and its constructor hard-codes `SingleUserPrincipalProvider`, so the publicly exported provider abstraction cannot actually be installed by a product runtime. Having `ServerBuilder` own `Arc<dyn UserPrincipalProvider>` with the local provider as its default would make authentication coverage and substitution explicit in one place.

crates/coral-app/src/request_context.rs
```diff
@@ -0,0 +1,72 @@
...
+        Self { principal }
+    }
+
+    // Installed server-wide by the immediate P1a child and consumed at service
+    // boundaries by P1b.
+    #[cfg_attr(
+        not(test),
+        expect(
+            dead_code,
+            reason = "installed server-wide by P1a and consumed at service boundaries by P1b"
```
jsummerfield: Could we replace the `P1a` and `P1b` references with a durable behavior-based reason, or omit the unused method until its consumer lands? Those labels only exist in the stack history; after merge they do not explain the invariant, and this `expect(dead_code)` can mask that the seam remains unused longer than intended.

jsummerfield (commented):
Thanks for keeping the principal model narrow and preserving local behavior. I found three composition and cohesion issues worth tightening so this boundary remains durable as later consumers are added.

jsummerfield (approved):

PR #1747, opened 2026-07-13
feat(reef): add onboarding flow

AlbertQM:
## Summary

Adding a new route that shows the onboarding. Nothing links to it yet; in a follow-up PR, I'll implement a way to store the "has gone through onboarding" per user.

## Test plan

`window.location.assign('/onboarding')` to go to the route

https://github.com/user-attachments/assets/34901c88-375f-4909-a9bb-e1d4351f6282

Try adding a source

https://github.com/user-attachments/assets/545819d9-36e4-4c04-93c4-88f5cb6a686b

https://github.com/user-attachments/assets/2c0db2dd-d2dc-40d0-8835-654cfe350ce8

- [feat(reef): add second onboarding step](https://github.com/withcoral/coral/pull/1618)
- [feat(reef): add onboarding flow](https://github.com/withcoral/coral/pull/1747) <-
- [feat(reef): add agent setup onboarding step](https://github.com/withcoral/coral/pull/1764)


martinrajdl:
Instead of the third step, would it make sense to do to - connect your agent - (similar to from the temp settings page) where you can connect codex app, claude app, and prompt agent - open app with prefilled questions which uses coral to query 
I can send a demo of this if helpful

AlbertQM:
> Instead of the third step, would it make sense to do to - connect your agent - (similar to from the temp settings page) where you can connect codex app, claude app, and prompt agent - open app with prefilled questions which uses coral to query I can send a demo of this if helpful

Yes, the plan is to add a third step with useful things, like connecting MCPs, installing skills, and some other things we will think of. However, each of those requires some thinking (e.g. I wouldn't reuse the settings, because a) it doesn't look like what it was built for (the proof being last week's demo where people misunderstood what it did), and b) it is not clear to me how this would work on the browser.

So yes, good intuition. There will be follow up PRs each adding some of this functionality incrementally and methodically. I thought this was a good point to draw the line and wire up the UI components built so far 😄

apps/reef/app/routes/onboarding.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+export async function loader({ request }: Route.LoaderArgs) {
+  return loadSourcesRouteData(request)
+}
+
+export async function action({ request }: Route.ActionArgs): Promise<SourcesActionData> {
+  return runSourcesAction(request)
+}
+
+export default function OnboardingRoute({ actionData, loaderData }: Route.ComponentProps) {
```
lazybean, line 33: We should aim at keeping the route file as little as possible (agent did not get the memo yet)  
This means the default export (the view) should be very tiny, usually passing actionData and loaderData to a relative component from views folder

lazybean (changes requested):
This is cool and just works!   
...almost.  
  
Adding a source with the oAuth flow will exit the onboarding (redirecting to /sources).  
Sending back to you for fixing.

apps/reef/app/routes/onboarding.tsx
```diff
@@ -0,0 +1,179 @@
...
+  const runSampleQuery = useCallback(async (sql: string) => {
+    const sequence = ++runSequence.current
+    setLoadState('loading')
+    setQueryError(null)
+
+    try {
+      const queryClient = await getQueryClient()
+      const response = await queryClient.executeSql(
+        create(ExecuteSqlRequestSchema, { sql, workspace: WORKSPACE }),
+      )
```
lazybean: :warning: :no_entry:Is that the client querying coral directly? Please don't.  
  
Talk to the bff.  
This likely mean you nead a resource route (action or loader) to execute query

apps/reef/app/routes/onboarding.tsx
AlbertQM: > (agent did not get the memo yet)

Will add an architecture.test for it, thanks!

lazybean (approved):
I am onboard with that PR!

PR #1762, opened 2026-07-14
fix(reef): format source display names

AlbertQM:
## Test plan

image
image

### DIFF
--- a/apps/reef/app/components/sources/source-name.ts
+++ b/apps/reef/app/components/sources/source-name.ts
@@ -1,0 +1,3 @@
+export function formatSourceName(name: string): string {
+  return name.replaceAll('_', ' ')
+}
--- a/apps/reef/app/wax/components/card/card.css.ts
+++ b/apps/reef/app/wax/components/card/card.css.ts
@@ -43,6 +43,7 @@ export const header = style({
 export const title = style({
   flexShrink: 1,
   minWidth: 0,
+  textTransform: 'capitalize',
 })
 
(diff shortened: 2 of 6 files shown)


apps/reef/app/wax/components/card/card.css.ts
```diff
@@ -43,6 +43,7 @@ export const header = style({
 export const title = style({
   flexShrink: 1,
   minWidth: 0,
+  textTransform: 'capitalize',
```
AlbertQM, line 46: We might need to revisit this in the future, but default capitalize looks better than lowercase in the card title 👀

apps/reef/app/components/sources/source-name.ts
```diff
@@ -0,0 +1,3 @@
+export function formatSourceName(name: string): string {
+  return name.replaceAll('_', ' ')
```
lazybean, line 2: That is not going to work for all sources, but can only be better than what we have right now.  
  
Ideally the source itself would give us a human readable name. no?

lazybean (approved):
Thank you that is better.  
But we may need a  display name coming from the source author directly, no?

apps/reef/app/components/sources/source-name.ts
AlbertQM: >Ideally the source itself would give us a human readable name. no?

Yup, that's be the best! I will open a PR and see if source codeowners ✅; thanks for pushing for this!

PR #1764, opened 2026-07-14
feat(reef): add agent setup onboarding step

AlbertQM:
## Summary

Adding an initial draft of the 3rd and last onboarding page. It's quite simple for now; one addition would be 1-click buttons to connect user agents to Coral MCP; that will differ between Desktop (which can) and web (which can't), so will add as a separate PR;

## Test plan

image

image


lazybean:
I think you can leave the content empty. We should reuse (and update for web if needed) what Martin is doing.

apps/reef/app/views/onboarding/onboarding.tsx
```diff
@@ -29,12 +32,27 @@ export function OnboardingView({
   const navigate = useNavigate()
   const { step } = loaderData
 
+  if (step.step === 'next-steps') {
```
lazybean: :pick:As the number of steps grow, we could benefit of 'switch' and ts making sure we handle all cases.

apps/reef/app/views/onboarding/onboarding.tsx
```diff
@@ -73,6 +91,50 @@ export function OnboardingView({
...
+      .catch((error: unknown) => {
+        if (!cancelled) {
+          setMcpLaunchConfig({ message: desktopErrorMessage(error), status: 'error' })
+        }
+      })
+
+    return () => {
+      cancelled = true
+    }
+  }, [])
```
lazybean: That think c/should be in the loader no?

apps/reef/app/components/onboarding/onboarding-next-steps-page.tsx
```diff
@@ -0,0 +1,229 @@
...
+  '',
+  'Coral is an open-source data access layer for AI agents: one place to connect sources such as GitHub, Slack, Linear, Datadog, and local files, then query them with SQL.',
+  '',
+  'I have completed Coral onboarding and connected at least one source. Start by identifying how I am running Coral. Talk it through with me, inspect my environment when possible, and recommend a setup path that uses my existing installation:',
+  '',
+  'Coral entrypoints:',
+  '- Desktop app: a native app with a bundled Coral runtime for a regular desktop environment.',
+  '- CLI (`coral`): for a terminal, headless machine, or server environment.',
+  'Both use local Coral state and can expose the same stdio MCP server to a coding agent.',
+  '',
```
lazybean: I think we know that. We knew if coral is running in the sidecar (deskpot) or as cli or somewhere.   
We should get that from loader and pass that info.

lazybean (approved):
Lets ship it and see! thank you!

apps/reef/app/views/onboarding/onboarding.tsx
AlbertQM: Will do, thanks!

apps/reef/app/components/onboarding/onboarding-next-steps-page.tsx
AlbertQM: Much better, will update. Thanks!

apps/reef/app/views/onboarding/onboarding.tsx
AlbertQM: :blob-yes:

PR #1772, opened 2026-07-15
fix(app): preserve compatible DSL v4 artifacts

### LINKED ISSUE #1744: Make fingerprint mismatch non-fatal
Today, if we upgrade Coral such that installed DSL v4 fingerprints are invalidated, then Coral fails to run until all sources have been re-added.

We should not do this! Instead, we should ensure that updates to Coral remain backwards compatible with existing persisted data.

simonwhitaker:
## Summary

- Treat DSL v4 fingerprints, producer versions, identity metadata, and raw-document hashes as advisory provenance that emits tracing diagnostics.
- Load readable, structurally compatible artifacts while retaining strict rejection of ambiguous runtime structure.
- Recover at surface and source granularity for compatibility failures; unreadable installed manifests and backend-wide credential failures remain fail-closed.

## Contributor behavior

- Update repository guidance so provenance drift is diagnostic rather than a runtime gate.
- Preserve the user-owned materialization lifecycle: loading never refreshes or rewrites persisted artifacts.

## Validation

- make rust-checks
  - formatting and Clippy clean
  - 1,701 tests passed; 2 skipped
  - rustdoc warnings denied
- cargo test -p coral-spec --all-features
- cargo test -p coral-app --all-features
- focused diagnostic-state and fatal-degradation regression tests
- make perf-check built the release binary successfully, but the benchmark could not run because hyperfine is not installed locally.


crates/coral-app/src/sources/materialization.rs
```diff
@@ -236,207 +383,447 @@ pub(crate) fn load_v4_materialization(
...
+    manifest_yaml: &str,
+    manifest: &V4SourceManifest,
+    path: &Path,
+    diagnostics: &mut Vec<Diagnostic>,
+) -> Option<Fingerprint> {
+    let fingerprint = match read_yaml::<Fingerprint>(path) {
+        Ok(fingerprint) => fingerprint,
+        Err(error) => {
+            diagnostics.push(materialization_warning(
+                "V4_FINGERPRINT_UNAVAILABLE",
```
simonwhitaker, line 578: Note to reviewers; I know these strings are a bit of a mess. I'll replace these with enums, and rationalise them, in a separate PR.

crates/coral-app/src/query/manager.rs
```diff
@@ -378,21 +418,24 @@ impl QueryManager {
...
+                        &source.name,
+                    );
+                    loaded_sources.push(loaded_source);
                 }
-                Err(error) => {
-                    tracing::warn!(
-                        source = %source.name,
-                        detail = %error,
-                        "skipping source during query-source load"
+                Err(error) if is_source_scoped_load_failure(&error) => {
```
iaptsiauri: Any per-source failure not listed in this function call will fail every query in the workspace

crates/coral-app/src/query/manager.rs
```diff
@@ -56,6 +57,16 @@ enum CredentialResolutionMode {
     StoredOnly,
 }
 
+fn is_source_scoped_load_failure(error: &AppError) -> bool {
+    matches!(
+        error,
+        AppError::MissingSourceInputs { .. }
+            | AppError::MissingOrIncompatibleV4Materialization { .. }
+            | AppError::InvalidV4ProjectionOverride { .. }
+            | AppError::Credentials(CredentialsError::Io(_) | CredentialsError::Parse(_))
```
iaptsiauri: I think here we reclassification Credentials from backend failure to source error. This might be a valid use case as well. Just pointing out

crates/coral-spec/src/v4/artifacts.rs
```diff
@@ -125,6 +115,43 @@ pub fn validate_materialized_source(
...
+        }
+        if manifest.surface(&materialized_surface.surface_id).is_none() {
+            return Err(ManifestError::validation(format!(
+                "DSL v4 materialized surface '{}' is not declared",
+                materialized_surface.surface_id
+            )));
+        }
+    }
+    let mut projection_names = BTreeSet::new();
+    for projection in &materialized.projections.projections {
```
iaptsiauri, line 144: Two same-name projections with different namespace strings pass validation and produce two identically named TableSpecs in one schema

crates/coral-app/src/query/manager.rs
```diff
@@ -378,21 +418,24 @@ impl QueryManager {
...
+                    );
+                    loaded_sources.push(loaded_source);
                 }
-                Err(error) => {
-                    tracing::warn!(
-                        source = %source.name,
-                        detail = %error,
-                        "skipping source during query-source load"
+                Err(error) if is_source_scoped_load_failure(&error) => {
+                    self.diagnostic_reporter.report_source_load_failure(
```
iaptsiauri, line 420: `MissingOrIncompatibleV4Materialization` and `InvalidV4ProjectionOverride` were fail-closed with user-facing guidance on main; they're now skip-with-tracing-warn.

crates/coral-app/src/sources/materialization.rs
```diff
@@ -236,207 +383,447 @@ pub(crate) fn load_v4_materialization(
...
+    )
+    .map_err(|error| {
+        Box::new(materialization_warning(
+            "V4_PARAMETER_METADATA_OVERRIDE_FAILED",
+            error.to_string(),
+            Some(surface.id.clone()),
+        ))
+    })?;
+    if let Some(fingerprint) = fingerprint {
+        match std::fs::read(&raw_source_document_path) {
```
iaptsiauri: `load_query_sources` runs on every `execute_sql/list_tables/describe_table` call, and each one re-reads and sha256-hashes every surface's full raw document  purely to maybe emit a warning

crates/coral-app/src/sources/materialization.rs
```diff
@@ -39,6 +40,166 @@ pub(crate) const FINGERPRINT_FILENAME: &str = "fingerprint.yaml";
...
+    pub(crate) fn report_runtime_surface_diagnostics(
+        &self,
+        workspace_name: &WorkspaceName,
+        source_name: &SourceName,
+        diagnostics: &[Diagnostic],
+    ) {
+        self.report_source_diagnostics(workspace_name, source_name, "runtime", diagnostics.iter());
+    }
+
+    pub(crate) fn clear_source(&self, workspace_name: &WorkspaceName, source_name: &SourceName) {
```
iaptsiauri, line 124: `clear_source` is only called on single-source removal. `delete_workspace` never clears the reporter.

crates/coral-spec/src/v4/artifacts.rs
simonwhitaker: Ugh. We should probably kill projection.namespace. I'll do that as part of removing support for multiple surfaces per source. For now, I'll have the dupe detection logic use surface ID instead of namespace.

simonwhitaker: Fixed in 56c911d50. Duplicate runtime identity now uses `(surface_id, projection.name)` while `projection.namespace` remains advisory. The affected surface is rejected before duplicate table specs reach the engine, with a defensive runtime check as well. Regression coverage includes identical names with differing namespace strings.

crates/coral-app/src/sources/materialization.rs
simonwhitaker: Fixed in c7f75a47b. Raw-document fingerprint validation is cached per workspace/source/surface and expected SHA in the shared diagnostic reporter, so repeated query and catalog loads no longer reread and SHA-256-hash unchanged documents. Fingerprint changes revalidate, read failures retry, and source removal clears the cache.

crates/coral-app/src/sources/materialization.rs
simonwhitaker: Fixed in 8dd0075fb. WorkspaceManager now receives the shared diagnostic reporter and clears both diagnostic deduplication state and raw-document validation cache after workspace deletion commits. Failed deletions preserve state, and cleanup is scoped so other workspaces are unaffected. The full Rust gate passes: 1,746 tests passed, 2 skipped; Clippy, formatting, and rustdoc are clean.

crates/coral-app/src/query/manager.rs
simonwhitaker: Ah yeah... and any non-source-scoped errors get raised outside this loop anyway. Duh. Removing this.

iaptsiauri (approved):

jsummerfield (approved):

PR #1774, opened 2026-07-15
feat(reef): cleanup source keyboard interaction

AlbertQM:
## Summary

Cleaning up the existing `cmd + f` interaction (we have a component for those type of shortcuts)

## Test plan

Can tab into the cards, cmd +f focuses the search input

### DIFF
--- a/apps/reef/app/views/sources/sources-index.tsx
+++ b/apps/reef/app/views/sources/sources-index.tsx
@@ -1,5 +1,7 @@
-import { useEffect, useRef, useState } from 'react'
+import { useCallback, useRef, useState } from 'react'
 import { useRevalidator } from 'react-router'
+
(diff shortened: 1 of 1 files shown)


lazybean:
One can _tab_ through those source cards already.   
What are the reason for adding arrows? It seems to add some complexity for something browser already do well.

lazybean:
> One can _tab_ through those source cards already.  
> What are the reason for adding arrows? It seems to add some complexity for something browser already do well.

​Or we can use the roving tabindex pattern.  
Tab gets you to the list, wont get you in. Once the card list is focused, you can navigate with arrows.  
But if we do that I'd argue we should bake support in the card-list directly. WDYT?

AlbertQM:
Chatted offline, will simplify this PR to just a cleanup 😄

lazybean (approved):
:rocket:

PR #1780, opened 2026-07-16
feat(reef): polish  source setup dialogs

AlbertQM:
## Summary

Says on the tin. Round of polish with no functionality changes.

## Test plan

Before

https://github.com/user-attachments/assets/29e1243a-900c-468d-96a3-8cc705454531

After

https://github.com/user-attachments/assets/c23f486b-b67b-4126-9165-b33099c0cfce

### DIFF
--- a/apps/reef/app/components/sources/install/oauth-fields.tsx
+++ b/apps/reef/app/components/sources/install/oauth-fields.tsx
@@ -35,8 +35,9 @@ export function OAuthFields({
     <div className={styles.fields}>
       {fields.map(({ key, label, secret, defaultValue }) => (
         <div key={key} className={styles.field}>
-          <Typography.Body className={styles.label}>{label}</Typography.Body>
+          <Typography.BodyStrong variant="primary">{label}</Typography.BodyStrong>
           <TextInput
+            ariaLabel={label}
             name={`oauth:${inputKey}:${key}`}
             type={secret ? 'password' : 'text'}
             value={values[key] ?? ''}
--- a/apps/reef/app/components/sources/install/oauth-fields.tsx
+++ b/apps/reef/app/components/sources/install/oauth-fields.tsx
@@ -1,3 +1,5 @@
+import { useId } from 'react'
+
 import { TextInput } from '@/wax/components/inputs/text'
 import { Typography } from '@/wax/components/typography'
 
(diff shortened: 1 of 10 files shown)


lazybean:
I find this (the 3rd tab title being truncated) a bit ugly.  
Can something be made ot make it nicer? (Either at the wax tab level (wrapping on new line? meh not sure; or using different title , it is always connecting right? )  
device code / OAuth app / token

apps/reef/app/components/sources/install/oauth-fields.tsx
```diff
@@ -35,8 +35,9 @@ export function OAuthFields({
     <div className={styles.fields}>
       {fields.map(({ key, label, secret, defaultValue }) => (
         <div key={key} className={styles.field}>
-          <Typography.Body className={styles.label}>{label}</Typography.Body>
+          <Typography.BodyStrong variant="primary">{label}</Typography.BodyStrong>
           <TextInput
+            ariaLabel={label}
```
lazybean: :thinking_face: What about rendering the typography as 'label' with 'labelFor={`oauth:${inputKey}:${key}`}'  
  
So that you can click on the label to focus the input.

apps/reef/app/views/sources/source-install.test.tsx
```diff
@@ -38,9 +38,118 @@ const entry: CatalogEntry = {
...
+      required: true,
+    },
+  ],
+}
+
 afterEach(() => vi.unstubAllGlobals())
 
 describe('SourceInstallDialog', () => {
+  it('uses Wax tabs to switch between source setup methods', async () => {
+    const router = createMemoryRouter(
```
lazybean: I think 'createRouteStub' is reccomended for test and storybook

apps/reef/app/components/sources/install/oauth-fields.tsx
AlbertQM: Nice, done. Thanks!

apps/reef/app/components/sources/install/oauth-fields.tsx
```diff
@@ -14,12 +14,14 @@ export function OAuthFields({
   disabled,
   fields,
   inputKey,
+  measurementOnly = false,
```
lazybean: :question: what is that for?

lazybean (approved):

apps/reef/app/components/sources/install/oauth-fields.tsx
AlbertQM: To avoid making the hidden fields (used for measurement) to be interactive -- changed instead to prefix each input with an ID, so that clicking the visible label will definitely trigger the visible input. Thanks!

PR #1761, opened 2026-07-14
feat(search): honor source-authored do-not-index policy

mystic123:
## Summary

- add source-authored `do_not_index` column policy for v3 manifests and v4 projections, defaulting to `false`
- carry the policy through runtime schemas and Arrow field metadata
- exclude opted-out fields before value inspection or persistence, while retaining Coral's mandatory sensitive-name and sensitive-value suppression
- document the field, default, SQL behavior, and a concrete YAML example in the public source-spec reference

This is intentionally a child of the observed-values queue substrate: PR #1472 remains policy-agnostic, and this PR owns the source-author policy contract and its reference documentation.

## Testing

- `cargo test -p coral-spec -p coral-engine -p coral-app do_not_index`
- `make docs-check`
- covered by the top-of-stack `make rust-checks`

## Feature flag

Source-authored `do_not_index` metadata remains valid regardless of the runtime flag. Its observed-value collection effect is dormant while `observed_values_search` is disabled.

PR #1756, opened 2026-07-14
docs(search): unify catalog and observed-value search guidance

mystic123:
## Summary

- replace the standalone catalog-search guide with one **Search with Coral** guide
- keep always-available catalog search first, then explain the default-off observed-value search feature
- document coverage limits, local-only storage, retention and storage bounds, source-authored exclusions, and maintenance operations
- align navigation, internal links, architecture, roadmap, MCP guidance, and the CLI reference
- redirect `/guides/search-the-catalog` to `/guides/search-with-coral`

## What reviewers should know

This is a docs-only PR: all nine changed files are under `apps/docs`.

Catalog search remains the default workflow. Observed-value collection, search, and maintenance remain experimental and disabled by default. The guide states explicitly that stored values stay in the workspace's local search database and are not uploaded to a remote search service.

## Review guide

1. `guides/search-with-coral.mdx` — combined user workflow and storage/privacy contract
2. `reference/cli-reference.mdx` — search results plus rebuild, drain, and clear commands
3. `docs.json` — one navigation replacement and one permanent redirect for the previous catalog guide
4. Remaining files — terminology and cross-link alignment

## Stack boundary

Runtime feature gating and search surfaces are implemented in #1593. Maintenance behavior and CLI commands are implemented in #1594. The source-authored `do_not_index` policy is defined in #1761. This PR owns only the public documentation.

## Validation

- `git diff --check`
- `env RUSTC_WRAPPER= make docs-check`
- `npx --yes mint validate` using Node 24
- `npx --yes mint broken-links --check-redirects` using Node 24
- verified the permanent `/guides/search-the-catalog` redirect and canonical `/guides/search-with-coral` path

PR #1612, opened 2026-07-10
feat(search): bound observed values storage

mystic123:
## Summary

Observed values are best-effort, local search data. This PR adds the storage guardrails that must exist before #1593 makes those values queryable.

It prevents observed-value projection from growing a workspace's search SQLite database past its live-data budget. Cleanup is bounded, and a projection that crosses the limit is rolled back in full before Coral decides whether to reclaim space, retry, leave the job queued, or drop the best-effort work.

**Stack position:** #1591 adds observed-value projection -> **this PR adds storage and retention governance** -> #1593 adds policy-aware retrieval.

## Behavior at a glance

| Control | Behavior |
| --- | --- |
| Write guard | Treats the workspace search database as full at 256 MiB, with 32 MiB reserved for WAL headroom (224 MiB usable live-page budget) |
| Accounting | Counts live database pages (`page_count - freelist_count`), not the reusable high-water size of the WAL file |
| Retention | Purges observed rows older than 365 days before evicting live data |
| Eviction | Removes the oldest observed rows first, within one shared 256-row budget per drain |
| Projection safety | Uses a savepoint so canonical and FTS writes either both commit or both roll back |
| FTS cleanup | Enables FTS5 secure-delete and performs small, bounded merge steps for legacy tombstones |

## What happens under storage pressure

1. Purge stale observed rows.
2. If the database is still full, evict the oldest observed rows.
3. Run one bounded FTS merge step when tombstones still occupy live pages.
4. Retry the same queued projection; reclamation retries do not consume its `max_jobs` allowance.
5. Stop when the shared time budget or eviction-row budget is exhausted, or when reclamation makes no progress. A fresh job stays queued for a later drain when more bounded work is possible; new best-effort observations are rejected while storage remains full.

## Reviewer focus

The important contracts are:

- a boundary-crossing projection never leaves partial canonical or FTS writes
- retention, eviction, projection retry, and FTS cleanup share one cooperative drain budget
- no-progress reclamation stops immediately instead of spinning
- eviction preserves an identical value owned by another installed source
- upgraded v2 databases can reclaim legacy FTS tombstones during drain or enqueue
- disabled observed-value paths perform no storage work

Suggested review order:

1. `observed/governance.rs` - accounting, retention, eviction, and bounded FTS merging
2. `observed/sqlite_projection.rs` - rollback, reclamation, retry, and queue accounting
3. `observed/sqlite_store.rs` - orchestration and shared budgets
[...]

PR #1594, opened 2026-07-09
feat(search): expose observed values maintenance

mystic123:
## What this adds

This is the operator-facing maintenance layer for local search data:

- `coral search-index drain` processes queued observed values with a time budget.
- `coral search-index rebuild` refreshes catalog metadata and/or the observed-value FTS projection.
- `coral search-index clear` removes catalog, observed-value, or all local search data at workspace or source scope.
- The same operations are available through app-owned gRPC APIs with structured per-provider results.

## Safety and correctness

- Rebuild drains queued work first and reports that work in the rebuild result.
- Observed FTS rebuilds are bounded and restartable: rows and payload bytes are capped per transaction, each write revalidates current state, and a fixed read snapshot does not reserve SQLite's sole writer.
- Rebuild repairs missing, stale, duplicate, malformed, legacy-rowid, and collision-cycle FTS rows while preserving failed-source and foreign-workspace data.
- WAL and storage guards stop safely instead of allowing rebuild batches to grow local storage without bound.
- Workspace/source `all` clear is one SQLite transaction across catalog documents, catalog FTS, persisted ownership, observed canonical/FTS rows, queue jobs, and generation changes.
- Source clears use persisted runtime-schema ownership, so they do not need to reload source manifests. The schema migration invalidates disposable catalog rows that predate durable ownership while preserving observed data.

## Result semantics

- Drain results include processed, stale, failed, and storage-dropped jobs; canonical/FTS writes; retention purges; eviction; remaining depth; budget exhaustion; and storage pressure.
- Rebuild results include the nested pre-rebuild drain plus canonical rows scanned and FTS rows rebuilt.
- Disabled observed-value search returns `Skipped` for drain/rebuild without opening storage. Explicit clear remains available so existing local data can still be removed.
- Source names may contain internal whitespace, but surrounding whitespace, separators, and dot segments are rejected.

## Review guide

1. API and CLI shape: `search.proto`, `search/service.rs`, and `coral-cli/src/lib.rs`.
2. Atomic clear and persisted catalog ownership: `search/sqlite_store.rs`, `catalog/sqlite_index.rs`, and migration `0004_catalog_source_ownership.sql`.
3. Bounded observed rebuild and projection recovery: `observed/sqlite_projection.rs` and its store tests.

## Validation

- `make rust-checks`: strict workspace Clippy, 1,964/1,964 Nextest tests, and rustdoc with warnings denied.
- Observed-value unit suite: 124/124.
- Catalog search suite: 43/43.
[...]

PR #1781, opened 2026-07-16
fix(app): unify Universal Search with runtime catalog

Bradley-Butcher:
## Intent

Make Universal Search use the workspace's queryable catalog so installed UDFs are discoverable alongside source-defined table functions.

Coral feedback report: `f008dd7f-792d-43c2-ab52-4bf3590aa74f`.

## What Changed

- Route Universal Search and catalog index maintenance through `CatalogDiscovery` and the runtime `CatalogInfo` assembly used by catalog listing. Remove the Search-only manifest snapshot loader.
- `QueryManager::resolve_catalog()` returns an app-internal `CatalogResolution`: the catalog, the names of sources skipped during app-side load or runtime registration (captured through the existing `SourceDecorator::source_failed` hook), and runtime schema ownership for the search projection. Catalog listing projects only the catalog; `coral-engine`'s public API is unchanged.
- Search degrades instead of hiding failures. Skipped sources produce healthy catalog results plus a `Partial` provider status with nonzero `failed_units`, `stale_index = true`, and a bounded note naming the skipped sources without raw error text. A whole-resolution error becomes a catalog provider `Error` while observed-value results survive.
- Catalog resolution is preloaded asynchronously; workspace lifecycle revision validation, SQLite catalog projection refresh/search, observed retrieval, and candidate fusion remain in the blocking phase. A resolution made stale by a concurrent source/UDF mutation is discarded and retried, bounded, then fails closed with a retryable status.
- Workspace creation and deletion hold lifecycle markers through directory staging, and source/function mutations take workspace-scoped guards, so an in-flight search can neither republish a stale projection after a mutation nor recreate storage for a deleted workspace.
- Catalog rebuild uses the same async preload followed by blocking projection work and composes with observed-value rebuild; clear remains local.
- Docs (search guide, architecture, CLI reference) and the MCP search tool description state the precise contract: runtime preparation may read stored credentials, initialize source providers, and inspect file metadata in local or object storage, but it does not execute the user's data query or return source rows.

## Ownership / Boundaries

`coral-app` remains the composition boundary. `CatalogDiscovery` owns workspace catalog access; `QueryManager` assembles the runtime catalog, JIT-resolves UDF metadata, and reports per-source failures; Universal Search only projects, indexes, and ranks the resolved catalog.

`search.sqlite3` remains a rebuildable search projection, not an authoritative catalog.

- Installed UDFs and source-defined table functions use the same table-function search path and ranking inputs.
- Search inherits the same source, credential, runtime-preparation, and failure semantics as catalog listing, and now reports degraded coverage instead of silently dropping skipped sources.
- A search racing a workspace mutation retries against the post-mutation catalog; a search racing workspace deletion fails with `NotFound` and leaves no recreated workspace storage.
- No proto, CLI, MCP, persisted-state, or public API shape changes.

## Not in This PR

- Ranking-weight changes or ranking-quality thresholds.
- Provenance labels distinguishing installed UDFs from source-defined table functions.
- Per-workspace lifecycle revisions. The revision is global, so a mutation in one workspace can force another workspace's in-flight search to retry; correctness is unaffected.
- A separate offline or degraded catalog browser.

## Validation

- `cargo test -p coral-app --test grpc search_tests` — 28 passed
- `make rust-checks` — 1,987 tests passed; 2 skipped
- `make perf-check` — 217 ms mean against the 750 ms threshold


crates/coral-app/src/search/catalog/provider.rs
```diff
@@ -47,23 +46,20 @@ struct CatalogProjectionState {
...
 
-    pub(crate) fn search(
+    pub(crate) async fn search(
         &self,
         request: &SearchRequest,
-        _attribution: &QueryAttribution,
+        attribution: &QueryAttribution,
     ) -> ProviderSearchOutcome {
-        let catalog = match self.load_catalog(request) {
+        let catalog = match self.load_catalog(request, attribution).await {
```
mystic123: [P1] Revalidate the workspace after catalog resolution

This await can resolve workspace revision N while a source add/delete or UDF install commits revision N+1 and clears the search projection; the stale request can then refresh SQLite with N after the mutation completed. Capture a `WorkspaceLifecycleLock` snapshot/revision before resolution, reacquire it after the await, retry when it changed, and hold it through projection refresh/search. Cover deletion separately as well: `open_workspace` creates the directory, so a paused search can recreate state after workspace deletion unless deletion-in-progress/tombstone coordination prevents it.

crates/coral-app/src/search/catalog/provider.rs
```diff
@@ -82,8 +78,17 @@ impl CatalogMetadataProvider {
...
+    async fn load_catalog(
+        &self,
+        request: &SearchRequest,
+        attribution: &QueryAttribution,
+    ) -> Result<CatalogInfo, QueryManagerError> {
+        // Search intentionally shares list_catalog's runtime resolution semantics, including
+        // credential-backed sources and JIT UDF inference. A local manifest projection would
+        // create a second workspace catalog with different contents and failure behavior.
+        self.catalog
+            .catalog_info(&request.workspace_name, None, attribution)
```
mystic123: [P1] Preserve skipped-source failures in search status

Runtime preparation records per-source failures and can continue with healthy sources, but this seam exposes only `CatalogInfo`. Search can therefore return ResultsFound/Empty with `failed_units = 0`, hiding degraded coverage and violating the stack's Partial/stale-index contract. Return an internal catalog-plus-failures outcome; let catalog listing project only the catalog, while search reports Partial/nonzero `failed_units` with a sanitized note and preserves healthy results. Please cover healthy+broken sources, recovery, and catalog/observed fusion.

crates/coral-app/src/search/manager.rs
```diff
@@ -36,6 +35,6 @@ impl SearchManager {
         self.workspaces
             .require_workspace(&request.workspace_name)
             .await?;
-        Ok(self.engine.search(request, attribution))
+        Ok(self.engine.search(request, attribution).await)
```
mystic123: [P1] Preserve the stack's blocking search phase

On #1593/#1594, catalog projection/search, observed retrieval, and fusion are intentionally performed in `spawn_blocking`. Awaiting the whole engine here creates a semantic conflict, and #1594's synchronous rebuild still calls the now-removed `catalog_loader` even though `provider.rs` auto-merges. Resolve the runtime catalog asynchronously first, then run lifecycle validation plus all SQLite/observed work in the blocking phase. Make catalog rebuild an async preload followed by blocking projection work; clear can remain local.

crates/coral-app/src/search/catalog/provider.rs
```diff
@@ -82,8 +78,17 @@ impl CatalogMetadataProvider {
...
     }
 
-    fn load_catalog(&self, request: &SearchRequest) -> Result<CatalogInfo, AppError> {
-        self.catalog_loader.load_catalog(&request.workspace_name)
+    async fn load_catalog(
+        &self,
+        request: &SearchRequest,
+        attribution: &QueryAttribution,
+    ) -> Result<CatalogInfo, QueryManagerError> {
+        // Search intentionally shares list_catalog's runtime resolution semantics, including
```
mystic123: [P1] Update the local-only contract in #1756

This is an intentional externally visible behavior change, not merely an implementation swap. #1756 currently describes catalog search/rebuild as local-only, always available, and not accessing connected sources, while runtime preparation may resolve credentials, start MCP providers, infer schemas, or access providers. Update the guide, architecture, CLI reference, and MCP descriptions in the same stack. A precise contract is that search constructs the workspace query runtime, but does not execute the user's data query or return source rows.

crates/coral-app/tests/grpc/search_tests.rs
```diff
@@ -4,15 +4,242 @@
...
+    harness: &GrpcHarness,
+    workspace: &Workspace,
+    schema_name: &str,
+    item_name: &str,
+) -> Option<CatalogItem> {
+    harness
+        .search_client()
+        .search(Request::new(SearchRequest {
+            workspace: Some(workspace.clone()),
+            query: format!("{schema_name}.{item_name}"),
```
mystic123, line 168: [P2] Keep a descriptive-ranking regression

This exact qualified-name query with `limit: 50` proves item parity, but it can pass if descriptive metadata stops being indexed or the UDF falls outside the default top ten. Keep this parity helper and add a separate natural-language/default-limit assertion (for example, “GitHub review queue table function”) with a meaningful rank bound.

mystic123 (changes requested):
Requesting changes after reviewing this head against live `main` and the full observed-values stack (#1470, #1573, #1472, #1761, #1480, #1591, #1612, #1593, #1594, #1756).

The direction—using the queryable runtime catalog as the source of truth—is sound, and the focused PR test passes, but the current implementation has production correctness problems and is not yet safely composable with the stack:

- The async catalog-resolution path is not protected by one workspace lifecycle revision. A source/UDF mutation can complete during the await, clear the projection, and then have an older search republish stale catalog contents. Workspace deletion can also race with `open_workspace` and recreate deleted search state.
- Runtime preparation can skip broken sources, but `catalog_info` returns only `CatalogInfo`. Search therefore reports complete/empty coverage with `failed_units = 0` instead of the stack's partial/degraded status while preserving healthy results.
- PR-specific textual conflicts begin at #1573. At stack top there are exactly five PR-specific conflict paths:
  - `crates/coral-app/src/bootstrap/server.rs`
  - `crates/coral-app/src/search/catalog/local_snapshot/tests.rs`
  - `crates/coral-app/src/search/engine.rs`
  - `crates/coral-app/src/search/manager.rs`
  - `crates/coral-app/tests/grpc/search_tests.rs`
  The server/test conflicts are mostly mechanical; `engine.rs` and `manager.rs` require a semantic async/blocking split. In addition, `provider.rs` auto-merges but #1594's maintenance implementation still calls the removed `catalog_loader`, so the combined tree does not compile without redesign.
- #1756 still promises local-only/always-available catalog search and rebuild with no connected-source access. This PR intentionally adopts runtime preparation, which may resolve credentials, start MCP providers, infer schemas, or access providers.
- The new UDF test proves exact-name parity at limit 50, but no longer protects natural-language/default-limit discovery and ranking.

The integration seam I recommend is: resolve the runtime catalog asynchronously; then enter the blocking phase for lifecycle revalidation, SQLite catalog projection/search, observed retrieval, and candidate fusion. Catalog rebuild should use the same async preload followed by blocking projection work; clear can remain local. Please add concurrency, degraded-source, maintenance, cross-provider, deletion, and natural-language ranking coverage before restacking.

For separation: the live-main + stack-top simulation also has three pre-existing CLI conflicts (`crates/coral-cli/src/lib.rs`, `crates/coral-cli/tests/cli_e2e.rs`, and `crates/coral-cli/tests/harness/mod.rs`); those are baseline stack drift, not caused by this PR.

crates/coral-app/src/search/catalog/provider.rs
Bradley-Butcher: Implemented per your recipe: search captures the lifecycle revision before resolution, revalidates after the await, holds the snapshot through projection refresh/search, and retries (bounded) when it changed. Deletion is fenced with lifecycle markers held through directory staging rather than a tombstone — `search_paused_in_catalog_resolution_does_not_recreate_deleted_workspace_storage` covers the resurrection case.

crates/coral-app/src/search/catalog/provider.rs
Bradley-Butcher: Adopted as an app-internal `CatalogResolution { catalog, failed_source_names, runtime_schema_owners }` returned by `QueryManager::resolve_catalog()`; catalog listing projects only the catalog. Search reports `Partial` with nonzero `failed_units`, `stale_index = true`, and a bounded note naming skipped sources without raw errors. Healthy+broken, recovery-after-repair, and registration-failure capture (via the `SourceDecorator::source_failed` hook) are covered in tests.

crates/coral-app/src/search/manager.rs
Bradley-Butcher: Engine is synchronous again. Catalog resolution is an async preload; lifecycle validation, SQLite projection refresh/search, observed retrieval, and fusion all run in `spawn_blocking`. Rebuild uses the same preload-then-blocking shape and composes with observed rebuild; clear stays local. The restack onto #1594/#1756 is done and green.

crates/coral-app/src/search/catalog/provider.rs
Bradley-Butcher: Docs updated in the guide, architecture page, CLI reference, and MCP description. One correction to the wording: preparation doesn't start MCP processes or make provider HTTP/MCP calls (tools/list discovery happens at materialization; per-call spawns at scan time). It can read stored credentials and inspect file/object-store schema metadata, and the docs now state that contract precisely.

crates/coral-app/tests/grpc/search_tests.rs
Bradley-Butcher: Kept the parity helper and added `natural_language_review_queue_query_ranks_installed_udf_in_top_three` — natural-language query, default limit, rank bound of three.

mystic123 (approved):

PR #1782, opened 2026-07-16
feat(reef): unify source detail presentation

AlbertQM:
## Summary

Polishing the cards of sources that have already been added; they should both look very much alike.

## Test plan

Before

image

image

After

image

image


lazybean (approved):
Looks good!

PR #1868, opened 2026-07-20
feat(reef): add Wax textarea input

lazybean:
## Summary

- add a reusable Wax `TextArea` that remains a native textarea for form, editing, selection, and accessibility behavior
- let Wax ScrollArea render native intrinsic scrollports so TextArea overflow state and scrollbar behavior come from the shared primitive
- show directional top/bottom fades only where content remains hidden
- preserve focus, hover, disabled, resize, controlled, and uncontrolled behavior
- add default, controlled, disabled, and overflowing Storybook coverage

## Stack

- stacked on the SourceField label-association parent in [#1866](https://github.com/withcoral/coral/pull/1866)
- provides the ScrollArea viewport seam consumed by Wax Radio in [#1869](https://github.com/withcoral/coral/pull/1869)
- consumed by the source-creation flow in [#1800](https://github.com/withcoral/coral/pull/1800)

## Validation

- `npm run check --prefix apps/reef`
- `npm run typecheck --prefix apps/reef`
- `npm test --prefix apps/reef` (86 files, 364 tests)
- `npm run build --prefix apps/reef`
- combined TextArea/Radio/ScrollArea stories (3 files, 10 tests)
- source-creator integration tests (2 files, 13 tests)
- Playwright Storybook checks for native editing, focusability, scroll metrics, resize style, stable border, and directional overflow fades

PR #1800, opened 2026-07-17
feat(reef): source creation wizard for DSL v4 manifests

lazybean:
## Summary

- add a dedicated routed source-install modal with nested Wax dialogs
- start source creation from a URL and discover OpenAPI JSON or YAML metadata through a server resource route
- prefill source name, description, and detected authentication while keeping source type and credentials editable
- keep every authentication variant mounted in an overlapping grid so radio toggles never resize the dialog
- use the new Wax textarea and Base UI-backed radio group for form-appropriate semantics
- add Storybook interactions and server/browser test coverage

## Provenance

This PR continues and replaces [#1779](https://github.com/withcoral/coral/pull/1779), originally opened by @martinrajdl.

The shared `TextInput` `id` forwarding change was split out and merged separately in [#1799](https://github.com/withcoral/coral/pull/1799). The remaining reusable prerequisites are stacked as [#1866](https://github.com/withcoral/coral/pull/1866), [#1868](https://github.com/withcoral/coral/pull/1868), and [#1869](https://github.com/withcoral/coral/pull/1869).

## Validation

- `npm run check --prefix apps/reef`
- `npm run typecheck --prefix apps/reef`
- `npm test --prefix apps/reef` — 90 files, 388 tests passed
- `npm run build --prefix apps/reef`
- Playwright measured the bearer, none, and custom-header variants at the same 532px dialog height

PR #1866, opened 2026-07-20
fix(reef): preserve native label associations in source fields

lazybean:
Constraint: Source creation needs native label associations without duplicating presentation primitives.

Rejected: Keep htmlFor handling in source-create | Shared fields should own their label semantics.

Confidence: high

Scope-risk: narrow

Tested: npm run check and npm run typecheck in apps/reef

PR #1869, opened 2026-07-20
feat(reef): add Wax radio group

lazybean:
## 

##   
  
Summary

- add reusable Wax `Radio.Group` and `Radio.Item` components backed by Base UI
- preserve native radio semantics, focus management, form behavior, disabled behavior, and arrow-key navigation
- render one aria-hidden selection background that slides and resizes between options and wrapped rows
- support `overflow="wrap"` by default and `overflow="scroll"` through Wax ScrollArea when height must remain fixed
- show horizontal edge fades and automatically scroll the selected option into view
- honor `prefers-reduced-motion`
- cover default, disabled, wrapped-many-option, and scrolling-many-option behavior in Storybook

## Stack

- stacked on the Wax textarea parent in [#1868](https://github.com/withcoral/coral/pull/1868)
- consumed by the source-creation flow in [#1800](https://github.com/withcoral/coral/pull/1800)

## Validation

- `npm run check --prefix apps/reef`
- `npm run typecheck --prefix apps/reef`
- `npm test --prefix apps/reef` (87 files, 367 tests)
- `npm run build --prefix apps/reef`
- Playwright Storybook checks for wrap dimensions, ScrollArea overflow, auto-scroll, pointer access, animated selection, reduced motion, and both edge fades

### DIFF
--- a/apps/reef/app/wax/components/radio/group.tsx
+++ b/apps/reef/app/wax/components/radio/group.tsx
@@ -1,0 +1,32 @@
+import { RadioGroup as BaseRadioGroup } from '@base-ui/react/radio-group'
+import classNames from 'classnames'
+
+import { Container as ScrollArea } from '@/wax/components/scroll-area'
+
+import * as styles from './radio.css'
+
+export type GroupProps<Value> = Omit<BaseRadioGroup.Props<Value>, 'className' | 'onValueChange'> & {
+  className?: string
+  onValueChange?: (value: Value) => void
+}
+
+export function Group<Value>({
+  children,
+  className,
+  onValueChange,
+  ref,
+  ...props
+}: GroupProps<Value>) {
+  return (
+    <ScrollArea fade="horizontal" height="auto" scrollDirection="horizontal">
+      <BaseRadioGroup
+        className={classNames(styles.group, className)}
+        onValueChange={(value) => onValueChange?.(value)}
+        ref={ref}
+        {...props}
+      >
+        {children}
+      </BaseRadioGroup>
+    </ScrollArea>
+  )
+}
--- a/apps/reef/app/wax/components/radio/item.tsx
+++ b/apps/reef/app/wax/components/radio/item.tsx
@@ -1,0 +1,20 @@
+import { Radio as BaseRadio } from '@base-ui/react/radio'
+import classNames from 'classnames'
+
+import * as styles from './radio.css'
+
+export type ItemProps<Value> = Omit<React.ComponentPropsWithoutRef<'label'>, 'value'> & {
+  disabled?: boolean
+  value: Value
+}
+
+export function Item<Value>({ children, className, disabled, value, ...props }: ItemProps<Value>) {
+  return (
+    <label className={classNames(styles.item, className)} {...props}>
+      <BaseRadio.Root className={styles.control} disabled={disabled} value={value}>
+        <BaseRadio.Indicator className={styles.indicator} />
+      </BaseRadio.Root>
+      <span className={styles.label}>{children}</span>
+    </label>
+  )
+}
--- a/apps/reef/app/wax/components/radio/radio.css.ts
+++ b/apps/reef/app/wax/components/radio/radio.css.ts
@@ -1,0 +1,76 @@
+import { style } from '@vanilla-extract/css'
+
+import { animation, theme } from '@/wax/theme/theme.css'
+
+export const group = style({
+  alignItems: 'center',
+  display: 'flex',
+  gap: 20,
+  minWidth: 'max-content',
+  paddingBlock: 4,
+  paddingInline: 4,
+})
+
+export const item = style({
+  alignItems: 'center',
+  color: theme.content.secondary,
+  cursor: 'pointer',
+  display: 'inline-flex',
+  flexShrink: 0,
+  gap: 8,
+  userSelect: 'none',
+  whiteSpace: 'nowrap',
+  ...theme.typography.body,
+  selectors: {
+    '&:hover:not(:has([data-disabled]))': {
+      color: theme.content.primary,
+    },
+    '&:has([data-checked])': {
+      color: theme.content.primary,
+    },
+    '&:has([data-disabled])': {
+      color: theme.content.disabled,
+      cursor: 'not-allowed',
+    },
+    '&:has([role="radio"]:focus-visible)': {
+      borderRadius: 4,
+      outline: `2px solid ${theme.stroke.focused}`,
+      outlineOffset: 2,
+    },
+  },
+})
+
+export const control = style({
+  alignItems: 'center',
+  backgroundColor: 'transparent',
+  border: `2px solid ${theme.content.tertiary}`,
+  borderRadius: '50%',
+  cursor: 'inherit',
+  display: 'inline-flex',
+  flexShrink: 0,
+  height: 18,
+  justifyContent: 'center',
+  outline: 'none',
+  padding: 0,
+  transition: animation.colorTransition,
+  width: 18,
+  selectors: {
+    '&[data-checked]': {
+      borderColor: theme.stroke.focused,
+    },
+    '&[data-disabled]': {
+      borderColor: theme.input.stroke.disabled,
+    },
+  },
+})
+
+export const indicator = style({
+  backgroundColor: theme.stroke.focused,
+  borderRadius: '50%',
+  height: 8,
+  width: 8,
+})
+
+export const label = style({
+  pointerEvents: 'none',
+})
--- a/apps/reef/app/wax/components/radio/radio.stories.tsx
+++ b/apps/reef/app/wax/components/radio/radio.stories.tsx
@@ -1,0 +1,116 @@
+import type { Meta, StoryObj } from '@storybook/react-vite'
+import { useState } from 'react'
+import { expect, waitFor } from 'storybook/test'
+
+import { Radio } from '@/wax/components'
+
+const meta = {
+  parameters: {
+    layout: 'centered',
+  },
+  title: 'Wax/Radio',
+} satisfies Meta
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  render: () => {
+    const [value, setValue] = useState('openapi')
+    return (
+      <Radio.Group aria-label="Source type" value={value} onValueChange={setValue}>
+        <Radio.Item value="openapi">REST API (OpenAPI)</Radio.Item>
+        <Radio.Item value="mcp">MCP server</Radio.Item>
+      </Radio.Group>
+    )
+  },
+  play: async ({ canvas, userEvent }) => {
+    await expect(canvas.getByRole('radiogroup', { name: 'Source type' })).toBeInTheDocument()
+
+    const openapi = canvas.getByRole('radio', { name: 'REST API (OpenAPI)' })
+    const mcp = canvas.getByRole('radio', { name: 'MCP server' })
+    await expect(openapi).toBeChecked()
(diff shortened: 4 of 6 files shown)


lazybean:
* **#1800** <a href="https://app.graphite.com/github/pr/withcoral/coral/1800?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1869** <a href="https://app.graphite.com/github/pr/withcoral/coral/1869?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1869?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1868** <a href="https://app.graphite.com/github/pr/withcoral/coral/1868?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1866** <a href="https://app.graphite.com/github/pr/withcoral/coral/1866?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/reef/app/wax/components/radio/radio.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+    )
+  },
+  play: async ({ canvas, userEvent }) => {
+    const mcp = canvas.getByRole('radio', { name: 'MCP server' })
+    await userEvent.click(mcp)
+    await expect(mcp).toBeChecked()
+  },
+}
+
+export const DisabledItem: Story = {
```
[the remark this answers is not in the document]
lazybean: Fixed in 59f30e67d. The DisabledItem story now verifies that clicking the disabled option preserves the current selection, then confirms an enabled option remains selectable.

apps/reef/vitest.config.ts
```diff
@@ -16,7 +16,11 @@ const dirname =
 // More info at: https://storybook.js.org/docs/next/writing-tests/integrations/vitest-addon
 export default defineConfig({
   optimizeDeps: {
-    include: ['@vanilla-extract/recipes/createRuntimeFn'],
+    include: [
+      '@base-ui/react/radio',
+      '@base-ui/react/radio-group',
```
AlbertQM: Don't think we need these?

apps/reef/app/wax/components/radio/group.tsx
```diff
@@ -0,0 +1,139 @@
...
+import { useComposedRefs } from '@/utils/compose-refs'
+import { Container as ScrollArea } from '@/wax/components/scroll-area'
+
+import * as styles from './radio.css'
+
+export type GroupProps<Value> = Omit<BaseRadioGroup.Props<Value>, 'className' | 'onValueChange'> & {
+  className?: string
+  onValueChange?: (value: Value) => void
+  /** How options behave when they exceed the available width. Defaults to `wrap`. */
+  overflow?: 'scroll' | 'wrap'
```
AlbertQM: IMO we could get rid of the `wrap` maybe, since we support scrolling horizontally

If we want to keep it, I think items should fill the space of the containers, rather than have different distances from its edges

Image

apps/reef/app/wax/components/radio/group.tsx
```diff
@@ -0,0 +1,139 @@
+import { RadioGroup as BaseRadioGroup } from '@base-ui/react/radio-group'
```
AlbertQM, line 1: The way we're styling the radio buttons make them look like a [segmented control](https://developer.apple.com/design/human-interface-guidelines/segmented-controls) or [segmented buttons](https://m3.material.io/foundations/interaction/selection#selection-controls-checkbox), not sure how I fee about it 🤔 
Base UI has a component for that too, that we were looking at yesterday: https://base-ui.com/react/components/toggle-group. IMO that would be a better primitive to build on top of

BTW, what do you think of using actual radio buttons in the stacked PR?
Image

Image

apps/reef/app/wax/components/radio/group.tsx
```diff
@@ -0,0 +1,139 @@
...
+  overflow = 'wrap',
+  ref,
+  ...props
+}: GroupProps<Value>) {
+  const groupRef = useRef<HTMLDivElement>(null)
+  const composedRef = useComposedRefs(ref, groupRef)
+  const [selectionChange, setSelectionChange] = useState(0)
+  const [indicatorPosition, setIndicatorPosition] = useState(initialIndicatorPosition)
+
+  const updateIndicatorPosition = useCallback(() => {
```
AlbertQM: Could we use `:has` and get rid of all this logic?

ie something like:

```
 '&:has([data-checked])': {
    background: theme.surface.card,
    color: theme.content.primary,
  }
```

AlbertQM (commented):
Thank you!

Some questions in order of importance

1. This is coming to life to use in the "Create spec from URL" dialog. Could we use/build a simpler radio check box there? (See comments about segmented controls vs radio buttons).
2. If we want this type of UI, is radio the right primitive here? Or would e.g. toggle group be a more semantic accurate one? Given we might want actual radio buttons in the future
3. If yes, then we can make this UI prettier, and I think simplify the code too by moving the active check to CSS!

🙏

apps/reef/app/wax/components/radio/group.tsx
lazybean: I think normal radio button is so much better, at least most people will understand

apps/reef/app/wax/components/radio/group.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const [selectionChange, setSelectionChange] = useState(0)
+
+  useLayoutEffect(() => {
+    const selectedItem = findSelectedItem(groupRef.current)
+    selectedItem?.scrollIntoView({
+      behavior: selectionChange ? 'smooth' : 'instant',
+      block: 'nearest',
+      inline: 'nearest',
+    })
+  }, [children, props.value, selectionChange])
```
AlbertQM:
```suggestion
export function Group<Value>({
  children,
  className,
  onValueChange,
  ref,
  ...props
}: GroupProps<Value>) {
  const groupRef = useRef<HTMLDivElement>(null)
  const composedRef = useComposedRefs(ref, groupRef)
  const [selectionChange, setSelectionChange] = useState(0)
```

The browser should handle the scrolling automatically

apps/reef/app/wax/components/radio/group.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  return (
+    <ScrollArea data-radio-scroll-area fade="none" height="auto" horizontal>
```
AlbertQM: This doesn't seem to be used?

```suggestion
    <ScrollArea fade="none" height="auto" horizontal>
```

apps/reef/app/wax/components/radio/item.tsx
```diff
@@ -0,0 +1,20 @@
...
+export type ItemProps<Value> = Omit<React.ComponentPropsWithoutRef<'label'>, 'value'> & {
+  disabled?: boolean
+  value: Value
+}
+
+export function Item<Value>({ children, className, disabled, value, ...props }: ItemProps<Value>) {
+  return (
+    <label className={classNames(styles.item, className)} {...props}>
+      <BaseRadio.Root className={styles.control} disabled={disabled} value={value}>
+        <BaseRadio.Indicator className={styles.indicator} data-radio-indicator />
```
AlbertQM: This doesn't seem to be used?

```suggestion
        <BaseRadio.Indicator className={styles.indicator} />
```

apps/reef/app/wax/components/radio/radio.css.ts
```diff
@@ -0,0 +1,74 @@
...
+    '&:hover:not([data-disabled])': {
+      borderColor: theme.content.secondary,
+    },
+    '&[data-checked]': {
+      borderColor: theme.stroke.focused,
+    },
+    '&[data-disabled]': {
+      borderColor: theme.input.stroke.disabled,
+    },
+    '&:focus-visible': {
```
AlbertQM: Can we move the focus outline to outside the whole radio item?

Currently looks like this:
Image

Proposed solution:

Image

```ts
export const group = style({
  // ...
  paddingInline: 4,
})

export const item = style({
  // ...
  selectors: {
    // ...
    '&:has([role="radio"]:focus-visible)': {
      borderRadius: 4,
      outline: `2px solid ${theme.stroke.focused}`,
      outlineOffset: 2,
    },
  },
})
```

apps/reef/app/wax/components/radio/group.tsx
AlbertQM: Image

We shouldn't show the horizontal scroll bar, as it covers most of the content -- quickest solution is to remove `horizontal` (just like in Tabs)

```suggestion
    <ScrollArea data-radio-scroll-area fade="none" height="auto">
```

AlbertQM: We should implement horizontal fading into the primitive too; feel free to land as-is and add this in a follow-up PR! 😄 

```suggestion
    <ScrollArea data-radio-scroll-area fade="horizontal" height="auto" horizontal>
```

apps/reef/app/wax/components/radio/radio.css.ts
```diff
@@ -0,0 +1,74 @@
...
+  display: 'inline-flex',
+  flexShrink: 0,
+  height: 18,
+  justifyContent: 'center',
+  outline: 'none',
+  padding: 0,
+  transition: animation.colorTransition,
+  width: 18,
+  selectors: {
+    '&:hover:not([data-disabled])': {
```
AlbertQM: Could we keep the radio circle stable on hover and instead change the label to the same colour as its selected state?

```ts
'&:hover:not(:has([data-disabled]))': {
  color: theme.content.primary,
},
```

ie get rid of this state:

Image

In favour of this
Normal:
Image
On hover:
Image

And no changes when hovering on a selected radio button

AlbertQM (approved):
(*) Accept () Request change -- Thank you!

Few tiny UI adjustments, and one comment which is more for a follow-up 🙏

apps/reef/app/wax/components/radio/radio.css.ts
lazybean: Updated hover styling so only the label moves to the primary color; the radio circle border remains stable, including for selected items.

apps/reef/app/wax/components/radio/group.tsx
lazybean: Removed the custom refs/state/layout effect. Keyboard navigation now relies on the browser to scroll the focused radio into view, and the Many Options story verifies the viewport scrolls.

apps/reef/app/wax/components/radio/radio.css.ts
lazybean: Moved the focus-visible outline to the full label item and added inline group padding so the ring is not clipped. The Default story now verifies the label owns the outline.

lazybean: Thank you!

apps/reef/app/wax/components/radio/radio.css.ts
lazybean: Thank you!

lazybean:
### Merge activity

* **Jul 22, 10:35 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1869).
* **Jul 22, 10:40 AM UTC**: [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1869) rebased this pull request as part of a merge.
* **Jul 22, 10:41 AM UTC**: @lazybean merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1869).

PR #1871, opened 2026-07-20
refactor(reef): unify desktop build marker

AlbertQM:
## Summary

Tiny cleanup.


lazybean (approved):
I looked at it this morning and it looked good.  
Beware of the conflict, it seems in a weird state now

PR #1873, opened 2026-07-20
feat(mcp): add reusable function tool

Bradley-Butcher:
## Intent

Let MCP-connected agents turn a reusable read-only query into a workspace function after discovering a useful SQL pattern, while recording the Coral surface that wrote the current definition.

## What changed

- Add `add_function` with structured schema, name, description, SQL, and replacement inputs.
- Render validated SQL frontmatter in `coral-mcp`; `coral-app` owns atomic create-only and replacement installation.
- Extend the function API response with `replaced`; existing CLI behavior continues to replace by default.
- Persist the current function definition's `write_surface`: `mcp`, `cli`, or `unknown` for functions created before this metadata existed.
- Document persistent workspace functions, `$placeholder` inputs, and the mutating MCP surface.

## Ownership / Boundaries

- The API carries the declared write surface; `coral-app` persists and returns it with function metadata.
- CLI and MCP are thin writers that set `cli` and `mcp` respectively.
- Replacement updates the stored surface because it describes the current definition, not an audit history.

- Existing function-service callers retain replacement semantics because `fail_if_exists` defaults to `false`.
- Existing persisted functions remain readable and are reported as `unknown`.
- MCP creation is create-only by default; `replace_existing: true` permits replacement.
- The `sql` MCP tool remains read-only. `add_function` changes only local workspace state.

## Not in This PR

- Authenticated creator identity, agent identity, timestamps, or function-history/audit records.
- A separate agent skill: the MCP tool description and `coral://guide` provide focused authoring guidance.

## Validation

- `make rust-checks` — format, clippy, 2,033 tests passed (2 skipped; one existing leaky-test note), and rustdoc.
- `make docs-check`
- `cargo test -p coral-mcp add_function_is_create_only_by_default_and_replaces_explicitly`
- `cargo test -p coral-cli --features cli-test-server --test cli_e2e functions_add_sends_file_to_selected_workspace`

The MCP end-to-end test persists `write_surface = "mcp"`; the CLI integration test asserts `cli`; app lifecycle coverage verifies both persistence and replacement.


AGENTS.md
```diff
@@ -191,6 +191,10 @@
   becoming a separate source of truth. Use
   `cargo run --locked -p xtask -- export-skills --dest <path>` for local
   export checks and distribution syncs.
+- `coral-optimise` layers on `$coral` to turn supplied or current task evidence
```
mystic123: is this supposed to be in this PR?

Bradley-Butcher: dang no - llm stacked it wrong. Thanks for catching

plugins/coral/skills/coral-optimise/SKILL.md
```diff
@@ -0,0 +1,47 @@
+---
+name: coral-optimise
```
mystic123: is this supposed to be here?

mystic123 (approved):

PR #1893, opened 2026-07-21
refactor(engine): distinguish SQL catalogs from schemas

iaptsiauri:
## Summary

- represent installed source identity with an explicit schema or catalog qualifier
- preserve schema-backed behavior while enabling catalog-backed database sources
- normalize DataFusion default-catalog handling in provenance and query matching

## Idiot's guide (for Simon)

A fully-qualified table name is, for example, `datafusion.github.pull_requests`

* `datafusion` is the catalog (or database) name
* `github` is the schema
* `pull_requests` is the table name

This won't work for database sources, which already have database name, schema name and table name; if we always use `datafusion` as the catalog name, we'd need four levels of hierarchy for SQL sources. So instead, SQL sources, the name of the source becomes the **catalog** name, rather than the schema name.

> github REST source -> `datafusion.github.users`
> postgres source -> `postgres.my_application.my_users`

## Stack

- Base: main
- Next: #1894

This is PR 1 of 7 in the relational database source MVP stack.

## Notes

- **Breaking:** `datafusion` is now a reserved source SQL name (case-insensitive, as are `coral`/`coral_admin`/`public`). An existing source named `datafusion` no longer passes manifest validation.
- The catalog registration path (`claim_registration_catalogs` and the fail-closed decorator gate) is unreachable here: every backend still returns no catalogs. It first gets test coverage at #2028, which replaces both with `catalog_publications` / `supports_lazy_schemas`.

### DIFF
--- a/crates/coral-engine/src/backends/file/mod.rs
+++ b/crates/coral-engine/src/backends/file/mod.rs
@@ -71,8 +71,8 @@ pub(crate) fn compile_manifest(
 
 #[async_trait]
 impl CompiledBackendSource for FileCompiledSource {
-    fn qualified_name(&self) -> &str {
-        &self.manifest.common.name
+    fn qualified_name(&self) -> SourceQualifiedName {
+        SourceQualifiedName::Schema(self.manifest.common.name.clone())
     }
 
     fn source_name(&self) -> &str {
--- a/crates/coral-engine/src/backends/file/mod.rs
+++ b/crates/coral-engine/src/backends/file/mod.rs
@@ -25,8 +25,8 @@ use datafusion::prelude::SessionContext;
 use crate::backends::{
     BackendCompileRequest, BackendRegistration, BackendRegistrationContext,
     BackendSchemaRegistration, CompiledBackendSource, RegisteredSource, RegisteredTable,
-    build_registered_inputs, build_registered_table, registered_columns_from_schema,
-    registered_columns_from_specs, required_filter_names,
+    SourceQualifiedName, build_registered_inputs, build_registered_table,
+    registered_columns_from_schema, registered_columns_from_specs, required_filter_names,
     validate_lookup_key_filter_backend_support,
 };
 use coral_spec::SourceBackend;
@@ -71,7 +71,7 @@ pub(crate) fn compile_manifest(
 
 #[async_trait]
 impl CompiledBackendSource for FileCompiledSource {
-    fn schema_name(&self) -> &str {
+    fn qualified_name(&self) -> &str {
         &self.manifest.common.name
     }
 
@@ -145,17 +145,17 @@ impl CompiledBackendSource for FileCompiledSource {
             &secret_keys,
         );
 
-        let schema_name = self.manifest.common.name.clone();
         Ok(BackendRegistration {
             schemas: vec![BackendSchemaRegistration {
                 tables,
                 source: RegisteredSource {
-                    schema_name,
+                    qualified_name: SourceQualifiedName::Schema(self.manifest.common.name.clone()),
                     tables: table_infos,
                     table_functions: vec![],
                     inputs,
                 },
             }],
+            catalogs: Vec::new(),
         })
     }
 }
--- a/crates/coral-engine/src/contracts/catalog.rs
+++ b/crates/coral-engine/src/contracts/catalog.rs
@@ -24,6 +24,8 @@ pub struct ColumnInfo {
 /// Describes one queryable table.
 #[derive(Debug, Clone)]
 pub struct TableInfo {
+    /// `SQL` catalog name. Empty for two-part table references.
+    pub catalog_name: String,
     /// `SQL` schema name.
     pub schema_name: String,
     /// Table name within the schema.
--- a/crates/coral-engine/src/contracts/catalog.rs
+++ b/crates/coral-engine/src/contracts/catalog.rs
@@ -24,8 +24,8 @@ pub struct ColumnInfo {
 /// Describes one queryable table.
 #[derive(Debug, Clone)]
 pub struct TableInfo {
-    /// `SQL` catalog name. Empty for two-part table references.
-    pub catalog_name: String,
+    /// `SQL` catalog name. Absent for two-part table references.
+    pub catalog_name: Option<String>,
     /// `SQL` schema name.
     pub schema_name: String,
     /// Table name within the schema.
--- a/crates/coral-engine/src/contracts/query_error.rs
+++ b/crates/coral-engine/src/contracts/query_error.rs
@@ -601,7 +601,9 @@ fn sql_string_literal(value: &str) -> String {
 /// Either a truly unqualified `FROM X` or a qualified `FROM schema.table`.
 #[derive(Debug, Clone, PartialEq, Eq)]
 enum ParsedTableRef {
-    Unqualified { table: String },
+    Unqualified {
+        table: String,
+    },
     Qualified {
         qualifier: ParsedQualifier,
         table: String,
@@ -1071,8 +1073,7 @@ mod tests {
     #[test]
     fn table_not_found_schema_reference_has_no_catalog_metadata() {
         let tables = vec![table("hockey", "master")];
-        let err =
-            StructuredQueryError::table_not_found(&tr(&["hockey", "missing_table"]), &tables);
+        let err = StructuredQueryError::table_not_found(&tr(&["hockey", "missing_table"]), &tables);
 
         assert_eq!(err.metadata().get("catalog"), None);
         assert_eq!(
--- a/crates/coral-engine/src/contracts/query_error.rs
+++ b/crates/coral-engine/src/contracts/query_error.rs
@@ -177,30 +177,52 @@ impl StructuredQueryError {
     /// `missing_ref` preserves the SQL object-name components for the missing
     /// table. `known_tables` is consulted to (a) distinguish `DataFusion`'s
     /// synthetic `public` schema from a real user source also named `public`,
-    /// and (b) recover a correct `(schema, table)` split when the source
-    /// name itself contains a dot.
+    /// (b) recover a correct `(schema, table)` split when the source name
+    /// itself contains a dot, and (c) split a catalog-backed reference into its
+    /// catalog and schema.
+    ///
+    /// Metadata reports `catalog` and `schema` as separate keys, never a
+    /// compound `catalog.schema` string: clients recover from this error by
+    /// feeding those values back into catalog lookups that take the catalog and
+    /// the schema as distinct arguments, and those match the schema on its own.
     pub(crate) fn table_not_found(missing_ref: &TableRefParts, known_tables: &[TableInfo]) -> Self {
         let parsed = parse_table_ref(missing_ref, known_tables);
 
-        let (schema, table) = match &parsed {
+        let (qualifier, table) = match &parsed {
             ParsedTableRef::Unqualified { table } => (None, table.clone()),
-            ParsedTableRef::Qualified { schema, table } => (Some(schema.clone()), table.clone()),
+            ParsedTableRef::Qualified { qualifier, table } => (Some(qualifier), table.clone()),
         };
 
-        let display_ref = match &schema {
-            Some(schema) => format!("{schema}.{table}"),
+        let display_ref = match qualifier {
+            Some(qualifier) => format!("{}.{table}", qualifier.written),
             None => table.clone(),
         };
-        let hint = table_not_found_hint(schema.as_deref(), &table, known_tables);
+        let hint = table_not_found_hint(
+            qualifier.map(|qualifier| qualifier.written.as_str()),
+            &table,
+            known_tables,
+        );
 
         let mut metadata = HashMap::new();
-        if let Some(schema) = &schema {
-            metadata.insert("schema".to_string(), schema.clone());
+        if let Some(qualifier) = qualifier {
+            if let Some(catalog) = qualifier.catalog() {
+                metadata.insert("catalog".to_string(), catalog.to_string());
+            }
+            metadata.insert("schema".to_string(), qualifier.schema().to_string());
         }
         metadata.insert("table".to_string(), table.clone());
 
-        let detail = match schema.as_deref() {
-            Some(schema) => format!("No table `{table}` exists in schema `{schema}`."),
+        let detail = match qualifier {
+            Some(qualifier) => match qualifier.catalog() {
+                Some(catalog) => format!(
+                    "No table `{table}` exists in catalog `{catalog}` schema `{}`.",
+                    qualifier.schema()
+                ),
+                None => format!(
+                    "No table `{table}` exists in schema `{}`.",
+                    qualifier.written
+                ),
+            },
             None => format!("No table `{table}` exists in any registered schema."),
         };
 
@@ -410,10 +432,9 @@ fn table_not_found_hint(
         ));
     };
 
-    let schema_lower = schema.to_lowercase();
     let tables_in_schema: Vec<&TableInfo> = known_tables
         .iter()
-        .filter(|info| table_schema_matches(info, &schema_lower))
+        .filter(|info| table_schema_matches(info, schema))
         .collect();
 
     if tables_in_schema.is_empty() {
@@ -467,7 +488,7 @@ fn quoted_qualified_table_match<'a>(
 
     let mut matches = known_tables
         .iter()
-        .filter(|info| raw_schema_table_name(info).eq_ignore_ascii_case(table));
+        .filter(|info| raw_schema_table_matches(info, table));
     let first = matches.next()?;
     matches.next().is_none().then_some(first)
 }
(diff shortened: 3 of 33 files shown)


crates/coral-engine/src/contracts/catalog.rs
```diff
@@ -24,6 +24,8 @@ pub struct ColumnInfo {
 /// Describes one queryable table.
 #[derive(Debug, Clone)]
 pub struct TableInfo {
+    /// `SQL` catalog name. Empty for two-part table references.
+    pub catalog_name: String,
```
simonwhitaker: nit: I'd prefer `Option<String>` to treating empty strings as special, I think.

crates/coral-engine/src/backends/file/mod.rs
```diff
@@ -71,7 +71,7 @@ pub(crate) fn compile_manifest(
 
 #[async_trait]
 impl CompiledBackendSource for FileCompiledSource {
-    fn schema_name(&self) -> &str {
+    fn qualified_name(&self) -> &str {
```
simonwhitaker: nit: should this be called `qualified_name_str` or similar? To make it clear that it's returning a string, rather than a SourceQualifiedName enum. (I see `qualified_name` used as the identifier for that latter type down on L152.)

simonwhitaker (approved):
It looks like we'll be jumping through some hoops to support some sources that have a source-scoped catalog name and other sources that use `datafusion` as the catalog name. It's always felt to me like the `datafusion` catalog was a leaking abstraction, and maybe this is a good time to kill it, and standardise all sources on <source name>.<category-or-default>.<relation name>.

crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -424,8 +426,8 @@ fn table_not_found_hint(
         // enrich the hint at their layer.
         return Some(format!(
             "Schema `{schema}` is not currently registered. \
-             Query `SELECT DISTINCT schema_name FROM coral.tables` \
-             to see available schemas."
+             Query `{LIST_AVAILABLE_TABLES_SQL}` \
+             to see available catalogs, schemas, and tables."
```
sauldhernandez, line 451: This branch answers "which schemas exist?", but the shared constant returns one row per table. It used to be `SELECT DISTINCT schema_name FROM coral.tables`. In a workspace with 20 sources and ~30 tables each that's ~600 rows instead of 20 — and MCP agents run these recovery queries verbatim, so it lands straight in their context.

The other use of the constant (line 409, no match anywhere) genuinely wants the table list. A second constant like `SELECT DISTINCT catalog_name, schema_name FROM coral.tables` would keep this hint scoped to the question asked and still surface the catalog dimension. Fine to skip if the full listing here was deliberate.

crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -628,7 +656,16 @@ fn schema_is_registered(candidate: &str, known_tables: &[TableInfo]) -> bool {
     let lowered = candidate.to_lowercase();
     known_tables
         .iter()
-        .any(|info| info.schema_name.to_lowercase() == lowered)
+        .any(|info| table_schema_matches(info, &lowered))
+}
+
+fn table_schema_matches(info: &TableInfo, schema_lower: &str) -> bool {
```
sauldhernandez: `table_schema_matches` treats `catalog.schema` as a single schema string for catalog-backed tables, so a table-not-found error carries `metadata.schema = "coral_db.main"` and the detail reads "exists in schema `coral_db.main`" (asserted by `table_not_found_catalog_reference_uses_compound_schema`). The metadata map reaches MCP clients verbatim as JSON.

#1894 gives `describe_table`/`list_columns` separate `catalog` and `schema` arguments, so a client that recovers by feeding `metadata.schema` back in passes `coral_db.main` as the schema and misses — the runtime matches `schema_name == "main"` with catalog compared separately. The hint text is fine (`coral_db.main.users`); only the structured metadata is compound, and no later PR in the stack splits it.

A separate `catalog` key with `schema` staying bare would match the split the rest of the stack makes. Is the compound form deliberate?

crates/coral-engine/src/contracts/query_error.rs
```diff
@@ -628,7 +656,16 @@ fn schema_is_registered(candidate: &str, known_tables: &[TableInfo]) -> bool {
...
+    match info.catalog_name.as_deref() {
+        None => info.schema_name.eq_ignore_ascii_case(schema_lower),
+        Some(catalog_name) => {
+            format!("{catalog_name}.{}", info.schema_name).to_lowercase() == schema_lower
```
sauldhernandez: The two arms fold case differently: `eq_ignore_ascii_case` in the `None` arm, `to_lowercase() ==` in the `Some` arm. Both call sites hand in a Unicode-lowercased candidate, and the code this replaces used `to_lowercase()` on both sides.

Legacy (v1-v3) source names aren't charset-restricted — only v4 enforces `[a-z][a-z0-9_]*` — so a legacy source named `CAFÉ` used to match `FROM CAFÉ.orders` here and now falls through to the generic "schema not registered" hint, since `"CAFÉ".eq_ignore_ascii_case("café")` is false.

Picking one folding for both arms fixes it. If you keep `to_lowercase`, a two-part compare in the `Some` arm also drops the `format!` — it allocates per table, and `schema_is_registered` calls this once per prefix length.

crates/coral-engine/src/runtime/catalog.rs
```diff
@@ -256,6 +256,12 @@ const TABLES_COLUMNS: &[SystemColumnDefinition] = &[
         nullable: true,
         description: "JSON search-limit metadata when the table declares provider search limits.",
     },
+    SystemColumnDefinition {
+        name: "catalog_name",
+        data_type: "Utf8",
+        nullable: false,
```
sauldhernandez, line 263: `catalog_name` is `nullable: false` with `''` for schema-addressed tables — same in the columns/filters/inputs builders — while the contract is `TableInfo.catalog_name: Option<String>` and the other optional columns in these tables (`search_limits_json`, `filter_mode`, `inputs.value`/`hint`) are nullable. So `WHERE catalog_name IS NULL` matches nothing, and users following the new `SELECT catalog_name, schema_name, table_name FROM coral.tables` hint see a blank column they have to filter with `= ''`.

The descriptions do document the sentinel, and there are reasons to keep it — joins on `catalog_name` stay simple, and #1894's proto3 surface forces empty string anyway. If that's the intent, fine. Otherwise `nullable: true` matches both the Rust type and the rest of the SQL surface, and it's easier to settle now than after #1894 freezes the sentinel into the proto/CLI surfaces.

crates/coral-engine/src/runtime/catalog.rs
```diff
@@ -364,14 +376,20 @@ const FILTERS_COLUMNS: &[SystemColumnDefinition] = &[
...
+    },
 ];
 
 const INPUTS_COLUMNS: &[SystemColumnDefinition] = &[
     SystemColumnDefinition {
         name: "schema_name",
         data_type: "Utf8",
         nullable: false,
-        description: "SQL schema for the source that declares the input.",
+        description: "SQL schema of the source that declares the input. Empty for database sources, which are addressed by catalog_name.",
```
sauldhernandez, line 393: This line redefines the `coral.inputs` contract: `schema_name` stays NOT NULL but becomes empty for database sources, with the source name moving to `catalog_name`.

Two documented recipes still key on `schema_name` alone. Once #1896 lands, running them against a Postgres/MySQL source returns zero rows — no error, just an empty result that reads as "this source declares no inputs":

- `apps/docs/reference/source-spec-reference.mdx` — the `coral.inputs` column table (no `catalog_name`) and the `WHERE schema_name = 'datadog'` example
- `plugins/coral/skills/coral-create-source-spec/SKILL.md:175` — `SELECT ... FROM coral.inputs WHERE schema_name = 'my_source'` (the sibling `coral.tables`/`coral.columns`/`coral.filters` queries at 171-174 drift the same way)

#1894 makes exactly this update for the MCP guide (`guide_template.md` gains the "database source inputs are addressed by catalog instead" note), but no PR in the stack touches these two files, and the docs-freshness CI job is path-gated on docs changes so a code-only PR never trips it.

Mirroring the #1894 note into the reference table and the skill would close it, here or in #1896. The alternative is keeping `schema_name` populated with the source's SQL name and treating `catalog_name` as an extra qualifier.

crates/coral-engine/src/runtime/query.rs
```diff
@@ -858,13 +901,14 @@ impl QueryRuntimeAdapter {
...
-        {
+        let catalog_name = non_default_catalog_name(table_reference.catalog());
+        if self.tables.iter().any(|table| {
+            table.catalog_name.as_deref() == catalog_name
+                && table.schema_name == schema_name
+                && table.table_name == table_name
+        }) {
             tables.insert(QueryTableUsage::new(
-                self.source_name_for_schema(schema_name),
+                self.source_name_for(catalog_name.unwrap_or(schema_name)),
```
sauldhernandez, line 974: `record_table_usage` now matches on the catalog, but `QueryTableUsage` still carries only (source, schema, table), so `warehouse.public.users` is recorded as schema_name="public" and the catalog is dropped. `record_query_provenance` in coral-app emits the same three fields to the query-trace span.

Until now schema_name was the source's workspace-unique top-level name, so (schema, table) identified a table. With catalog sources, two databases both exposing `public.users` collapse to the same pair; only source_name still disambiguates, and any consumer keying on schema+table merges them.

Nothing upstack (#1894 through #1896) adds a catalog field. Deliberate scope call for the MVP, or worth carrying the catalog through while the identity model is being reshaped anyway?

crates/coral-engine/src/runtime/registry.rs
```diff
@@ -8,14 +8,14 @@ use datafusion::prelude::SessionContext;
...
-    CompiledBackendSource, RegisteredSource,
+    BackendCatalogRegistration, BackendRegistration, BackendRegistrationContext,
+    BackendSchemaRegistration, CompiledBackendSource, RegisteredSource,
 };
 use crate::runtime::error::{datafusion_to_core, source_decorator_error_to_core};
 use crate::runtime::schema_provider::StaticSchemaProvider;
 use crate::{CoreError, QuerySource, SourceDecorator, SourceFailurePolicy};
 
-const RESERVED_SCHEMA_NAMES: &[&str] = &["coral", "coral_admin", "public"];
+const RESERVED_SCHEMA_NAMES: &[&str] = &["coral", "coral_admin", "datafusion", "public"];
```
sauldhernandez, line 21: Adding `datafusion` here diverges from the spec-side list: `RESERVED_SOURCE_SCHEMA_NAMES` in `crates/coral-spec/src/common.rs:22` only picks it up in #1896, at the top of the stack. Until then a manifest named `datafusion` passes validation and fails later at runtime registration — with a message saying it "cannot be used by manifests" even though the manifest validator accepted it. The spec change is one line plus its comment; pulling it into this PR keeps the intermediate merge states consistent.

Separately, reserving `datafusion` is a breaking change for anyone with an existing source by that name — worth a line in the PR body.

sauldhernandez, line 21: The reservation is exact-match (`is_reserved_schema` uses `contains`), but `non_default_catalog_name` folds the default catalog case-insensitively. Legacy (non-v4) manifests only go through `validate_source_name` — no `[a-z][a-z0-9_]*` rule — so a source named `DataFusion` slips past the reservation and registers, while catalog filters treat that same spelling as the default catalog: `list_tables(Some("DataFusion"), ..)` folds to `None` and matches every schema-backed table. The two checks disagree on what the reserved name is.

`eq_ignore_ascii_case` in `is_reserved_schema` would align them — note it would also start rejecting `Coral`/`Public` case-variants, probably fine but it widens the existing exact-match convention.

crates/coral-engine/src/runtime/registry.rs
```diff
@@ -221,41 +233,102 @@ async fn register_source(
...
     registration: BackendRegistration,
     result: &mut SourceRegistrationResult,
 ) -> std::result::Result<(), CoreError> {
+    // Source decorators wrap table providers at registration time, but catalog
+    // registrations expose providers lazily through the catalog itself.
+    // Decorators must explicitly allow that their table-decoration hook is
+    // skipped so policy decorators fail closed while lifecycle observers can
+    // still participate.
+    if let Some(decorator) = source_decorators.iter().find(|decorator| {
+        !registration.catalogs.is_empty() && !decorator.supports_catalog_sources()
```
sauldhernandez, line 313: `!registration.catalogs.is_empty()` doesn't depend on `decorator`, so having it inside the `find` closure reads as if emptiness could vary per decorator. Hoisting the guard states the intent directly — `if !registration.catalogs.is_empty() && let Some(decorator) = source_decorators.iter().find(|d| !d.supports_catalog_sources())` (edition 2024 let-chains), or a plain outer `if` around the search.

sauldhernandez, line 313: The new catalog registration path is unreachable in this PR: every backend returns `catalogs: Vec::new()` and `catalog_names()` is hard-coded empty, so `claim_registration_catalogs` and this fail-closed gate ship with no test executing them. The gate is what keeps a policy decorator from being silently skipped for catalog-backed sources — if the condition regressed, nothing in the tree would catch it until #1895.

Both are unit-testable today: `claim_registration_catalogs` takes a plain `BackendRegistration`, and a stub registration over a `MemoryCatalogProvider` would drive the gate (`backends/common.rs` already has a `test_support` module). If you'd rather land the tests with the backend in #1895, a note in the PR body would keep that from getting lost.

sauldhernandez (approved):

crates/coral-engine/src/contracts/query_error.rs
iaptsiauri: Not deliberate — fixed in 896b8c889.

`parse_table_ref` now resolves a qualifier to the registered catalog/schema pair instead of keeping the joined string, so metadata carries `catalog: "coral_db"` and `schema: "main"` separately and the detail reads "exists in catalog `coral_db` schema `main`". The summary still echoes the reference as written (`coral_db.main.usrs`), and the hint was already fine.

Worth noting the split now matches the table-function error, which at the top of the stack does emit a separate `catalog` key while this one stayed compound — so the two were inconsistent inside one file. `table_not_found_catalog_reference_uses_compound_schema` became `..._splits_catalog_from_schema`, and `catalog_internal_schema_matches_compound_catalog_schema` now asserts the pair.

One deliberate asymmetry: metadata reports the *registered* spelling (it is what a client feeds back into a lookup), while the human-facing text echoes what the caller wrote. Documented on `table_not_found`.

crates/coral-engine/src/contracts/query_error.rs
iaptsiauri: Real, fixed in 896b8c889 — your `CAFÉ` case reproduces exactly as described.

Both arms now fold through `char::to_lowercase` via one `eq_folded` helper, and I pushed the folding down so callers stop pre-lowercasing: `table_schema_matches` takes the raw candidate and folds both sides identically. That also removes the sigma-style trap of comparing a `str::to_lowercase` result against a per-character folding.

On the allocation: `eq_folded` takes the segments and compares in place, so the catalog arm no longer builds `catalog.schema` per table. I applied the same helper to `quoted_qualified_table_match`, which had the identical ASCII-only fold (`raw_schema_table_name(info).eq_ignore_ascii_case(table)`) plus a per-table `String` — same bug, two hundred lines up.

Added `table_not_found_matches_non_ascii_schema_case_insensitively` to pin it.

crates/coral-engine/src/runtime/registry.rs
iaptsiauri: Both parts done in 29d469afc.

`datafusion` is now in `RESERVED_SOURCE_SCHEMA_NAMES`, so the validator and the registry agree at every merge state instead of only once #1896 lands. Confirmed by walking the stack: the spec side previously picked it up at `feat/database-source-mvp` and nowhere earlier, so 1893 through #1929 all had the "validation accepts it, registration rejects it with `cannot be used by manifests`" window you describe.

The breaking change is called out as a `BREAKING CHANGE:` trailer on that commit, and I have added a line to the PR body.

iaptsiauri: Fixed in 29d469afc — `is_reserved_schema` and the spec-side check both fold with `eq_ignore_ascii_case` now, so they agree with `non_default_catalog_name` on what the reserved name is.

Taking your caveat as intended: `Coral`/`Public` case-variants are rejected too. That looks right rather than merely tolerable — the engine already refused to *serve* those names, so accepting the manifest only deferred the failure.

Covered by `reserved_schema_rejection_ignores_case` (engine) and `reserved_datafusion_source_name_is_rejected_in_any_case` (spec, which drives a real legacy manifest through the parser since that is the path with no charset rule).

crates/coral-engine/src/runtime/query.rs
iaptsiauri: Not a deliberate scope call — carried through in a9c39ad14, since you are right that reshaping the identity model is the moment to do it.

`QueryTableUsage` gains `catalog: Option<String>` (so it is part of the derived `Ord` that drives the `BTreeSet` dedup), `collect_table_scan_usage` passes the catalog it already computed, and the field travels the rest of the chain: the query-trace span JSON, `TraceQueryTableUsage`, and `McpQueryTableUsage`. Stopping at the engine contract would have dropped it again at the app boundary.

The stored row is `#[serde(default)]`, so spans recorded before this change still parse. `QueryTableUsage::new` keeps its name and gains the catalog as its second argument, mirroring `catalog.schema.table` order.

crates/coral-engine/src/runtime/catalog.rs
iaptsiauri: Confirmed live and fixed here in 603c51829, rather than deferring to #1896.

Checked against the top of the stack: `catalog_input_rows` emits `schema_name: ""` for non-default catalogs, and `source-spec-reference.mdx` contained zero occurrences of "catalog" anywhere — so both recipes you name did return zero rows with no error.

- `source-spec-reference.mdx`: `catalog_name` added to the `coral.inputs` column table, with your #1894 note mirrored in, plus the point that the absent side is `` and not NULL (following the old text, a reader would reach for `IS NULL` and match nothing).
- `SKILL.md`: all five validation-loop queries now filter `schema_name = ... OR catalog_name = ...`, which is exact in both shapes because the `` sentinel never equals a real source name. Note this is safe as a *filter* precisely where the `OR` *join* you flag on #1894 is not.

I also documented the asymmetry between the tables, since it is a second trap: on `coral.tables`/`columns`/`filters`/`table_functions`, `schema_name` stays populated for a database source because it names the schema inside the database, while only `coral.inputs` empties it.

Left alone deliberately: `references/http-source-checklist.md:127-130` has the same four queries, but HTTP sources are always schema-addressed there.

Editing `apps/docs/**` also means the docs-freshness job actually runs on this PR now instead of being path-gated out.

PR #1894, opened 2026-07-21
feat(catalog): expose catalog-qualified table discovery

iaptsiauri:
## Summary

- carry catalog names through discovery, API, CLI, search, and MCP surfaces
- keep describe-table and list-columns behavior consistent for DataFusion-qualified input
- format three-part catalog, schema, and table names where public surfaces expose qualified names

## Stack

- Depends on: #1893
- Next: #1927

This is PR 2 of 7 in the relational database source MVP stack.

### DIFF
--- a/crates/coral-api/proto/coral/v1/catalog.proto
+++ b/crates/coral-api/proto/coral/v1/catalog.proto
@@ -256,8 +256,7 @@ message DescribeTableResponse {
   Table table = 1;
   // Suggested nearby tables when the exact table is missing.
   repeated TableSummary suggestions = 2;
-  // Query-visible table namespaces available in this workspace: a schema
-  // name, or the composite `catalog.schema` for database sources.
+  // Query-visible raw schema names available in this workspace.
   repeated string available_schemas = 3;
   // Tables in the requested schema, capped for user-facing missing-table help.
   repeated TableSummary same_schema_tables = 4;
--- a/crates/coral-api/proto/coral/v1/catalog.proto
+++ b/crates/coral-api/proto/coral/v1/catalog.proto
@@ -61,6 +61,9 @@ message Table {
   repeated string required_filters = 6;
   // User-facing query guidance.
   string guide = 7;
+  // SQL catalog containing the schema. Empty for two-part tables; otherwise
+  // the table is addressed as catalog_name.schema_name.name.
+  string catalog_name = 8;
 }
 
 // Table metadata without full column definitions.
@@ -78,6 +81,9 @@ message TableSummary {
   repeated string required_filters = 5;
   // User-facing query guidance.
   string guide = 6;
+  // SQL catalog containing the schema. Empty for two-part tables; otherwise
+  // the table is addressed as catalog_name.schema_name.name.
+  string catalog_name = 7;
 }
 
 // An argument accepted by a query-visible table function.
@@ -184,6 +190,8 @@ message ListCatalogRequest {
   CatalogItemKind kind = 3;
   // Optional page request. Missing or zero limit keeps the unbounded behavior.
   PaginationRequest pagination = 4;
+  // Optional exact SQL catalog name. Empty matches tables in every catalog.
+  string catalog_name = 5;
 }
 
 // Returns queryable catalog items in one workspace.
@@ -210,6 +218,8 @@ message SearchCatalogRequest {
   CatalogItemKind kind = 5;
   // Optional page request. Missing or zero limit applies the search default.
   PaginationRequest pagination = 6;
+  // Optional exact SQL catalog name. Empty searches tables in every catalog.
+  string catalog_name = 7;
 }
 
 // One catalog search match.
@@ -236,6 +246,8 @@ message DescribeTableRequest {
   string schema_name = 2;
   // Exact table name.
   string table_name = 3;
+  // SQL catalog containing the schema. Empty for two-part tables.
+  string catalog_name = 4;
 }
 
 // Returns one table, or app-owned missing-table discovery context.
@@ -244,7 +256,8 @@ message DescribeTableResponse {
   Table table = 1;
   // Suggested nearby tables when the exact table is missing.
   repeated TableSummary suggestions = 2;
-  // Query-visible schemas available in this workspace.
+  // Query-visible table namespaces available in this workspace: a schema
+  // name, or the composite `catalog.schema` for database sources.
   repeated string available_schemas = 3;
   // Tables in the requested schema, capped for user-facing missing-table help.
   repeated TableSummary same_schema_tables = 4;
@@ -266,6 +279,8 @@ message ListColumnsRequest {
   bool required_only = 6;
   // Optional page request. Missing or zero limit applies the column-list default.
   PaginationRequest pagination = 7;
+  // SQL catalog containing the schema. Empty for two-part tables.
+  string catalog_name = 8;
 }
 
 // One column search match.
--- a/crates/coral-app/src/catalog/discovery.rs
+++ b/crates/coral-app/src/catalog/discovery.rs
@@ -389,7 +389,7 @@ impl CatalogDiscovery {
 fn catalog_item_sort_key(item: &CatalogItem) -> (&str, &str, &str, &'static str) {
     match item {
         CatalogItem::Table(table) => (
-            &table.catalog_name,
+            table.catalog_name.as_deref().unwrap_or_default(),
             &table.schema_name,
             &table.table_name,
             "table",
@@ -465,7 +465,10 @@ fn table_matched_fields(table: &TableInfo, regex: &Regex) -> Vec<CatalogMetadata
     let addressable_schema = table_addressable_schema_name(table);
     let name = table_addressable_name(table);
     let mut matches = Vec::new();
-    let catalog_matched = !table.catalog_name.is_empty() && regex.is_match(&table.catalog_name);
+    let catalog_matched = table
+        .catalog_name
+        .as_deref()
+        .is_some_and(|catalog_name| regex.is_match(catalog_name));
     let schema_matched = regex.is_match(&table.schema_name);
     if catalog_matched {
         matches.push(CatalogMetadataField::CatalogName);
@@ -601,17 +604,18 @@ fn table_metadata_contains_literal(table: &TableInfo, literal: &str) -> bool {
     let literal = literal.to_lowercase();
     let schema_name = table_addressable_schema_name(table);
     let name = table_addressable_name(table);
-    let candidates = [
-        table.catalog_name.as_str(),
-        table.schema_name.as_str(),
-        schema_name.as_str(),
-        table.table_name.as_str(),
-        name.as_str(),
-        table.description.as_str(),
-        table.guide.as_str(),
-    ];
-    candidates
+    table
+        .catalog_name
+        .as_deref()
         .into_iter()
+        .chain([
+            table.schema_name.as_str(),
+            schema_name.as_str(),
+            table.table_name.as_str(),
+            name.as_str(),
+            table.description.as_str(),
+            table.guide.as_str(),
+        ])
         .any(|value| value.to_lowercase().contains(&literal))
         || table
             .required_filters
@@ -620,10 +624,9 @@ fn table_metadata_contains_literal(table: &TableInfo, literal: &str) -> bool {
 }
 
 fn table_addressable_schema_name(table: &TableInfo) -> String {
-    if table.catalog_name.is_empty() {
-        table.schema_name.clone()
-    } else {
-        format!("{}.{}", table.catalog_name, table.schema_name)
+    match table.catalog_name.as_deref() {
+        Some(catalog_name) => format!("{catalog_name}.{}", table.schema_name),
+        None => table.schema_name.clone(),
     }
 }
 
@@ -640,7 +643,7 @@ fn table_matches_ref(table: &TableInfo, table_ref: CatalogTableRef<'_>) -> bool 
 }
 
 fn table_qualifier_matches(table: &TableInfo, table_ref: CatalogTableRef<'_>) -> bool {
-    table.catalog_name == table_ref.catalog_name.unwrap_or_default()
+    table.catalog_name.as_deref() == table_ref.catalog_name
         && table.schema_name == table_ref.schema_name
 }
 
@@ -694,7 +697,7 @@ mod tests {
 
     fn database_table(schema_name: &str, table_name: &str) -> TableInfo {
         let mut table = table(Vec::new());
-        table.catalog_name = "coral_db".to_string();
+        table.catalog_name = Some("coral_db".to_string());
         table.schema_name = schema_name.to_string();
         table.table_name = table_name.to_string();
         table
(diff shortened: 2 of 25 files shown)


simonwhitaker (approved):

crates/coral-api/proto/coral/v1/catalog.proto
```diff
@@ -184,6 +190,8 @@ message ListCatalogRequest {
   CatalogItemKind kind = 3;
   // Optional page request. Missing or zero limit keeps the unbounded behavior.
   PaginationRequest pagination = 4;
+  // Optional exact SQL catalog name. Empty matches tables in every catalog.
+  string catalog_name = 5;
```
sauldhernandez, line 194: This new field breaks the `coral-app` `grpc` test binary: `crates/coral-app/tests/grpc/search_tests.rs:146` still builds `ListCatalogRequest` without `catalog_name`, so `cargo test -p coral-app --test grpc` fails to compile (E0063). It's the only call site that wasn't updated — CI didn't surface it because Rust checks aborted at the clippy failure before building tests. Adding `catalog_name: String::new(),` there fixes it.

crates/coral-app/src/catalog/discovery.rs
```diff
@@ -131,14 +137,24 @@ impl ColumnMetadataField {
...
 
 impl<'a> CatalogTableRef<'a> {
-    pub(crate) fn new(schema_name: &'a str, table_name: &'a str) -> Self {
+    pub(crate) fn new(
+        catalog_name: Option<&'a str>,
+        schema_name: &'a str,
+        table_name: &'a str,
+    ) -> Self {
         Self {
+            catalog_name: normalize_catalog_name(catalog_name),
```
sauldhernandez, line 165: `normalize_catalog_name` only drops the `datafusion` sentinel, so `CatalogTableRef::new(Some(""), ..)` or `Some(" warehouse ")` keeps the catalog and then matches nothing in `table_qualifier_matches` — surfacing as a bare "table not found". Safe today because service.rs pre-trims with `optional_trimmed`, but the trimming and the normalization live in different places, and later PRs in this stack add callers. Folding trim + empty-drop into the normalization here would make `catalog_name.is_none()` reliably mean two-part lookup at the type boundary.

crates/coral-app/src/catalog/discovery.rs
```diff
@@ -195,24 +213,27 @@ impl CatalogDiscovery {
...
     }
 
     pub(crate) async fn catalog_info(
         &self,
         workspace_name: &WorkspaceName,
+        catalog_name: Option<&str>,
         schema_name: Option<&str>,
         attribution: &QueryAttribution,
     ) -> Result<CatalogInfo, QueryManagerError> {
+        let catalog_name = normalize_catalog_name(catalog_name);
```
sauldhernandez: This pre-normalization inverts the filter's meaning. Here `None` means "match every catalog", but `normalize_catalog_name` turns `catalog_name = "datafusion"` into `None` — so `list_catalog`/`search_catalog` with that value return tables from every catalog, while the proto promises an exact match and `describe_table` with the same value resolves only two-part tables.

The engine already normalizes this filter itself (`table_qualifier_matches`, and the `includes_schema_sources` check in `catalog_info`), so dropping this line restores exact-match semantics with nothing lost. The normalization in `CatalogTableRef::new` is fine — that one is a reference, not a filter.

Latent today since every source registers `SourceQualifiedName::Schema`. There's also no test using `catalog = "datafusion"` on any surface, and DataFusion-qualified consistency is one of the PR's stated goals — one case in the discovery tests would pin it down.

crates/coral-app/src/catalog/discovery.rs
```diff
@@ -519,15 +559,32 @@ fn column_matched_fields(column: &ColumnInfo, regex: &Regex) -> Vec<ColumnMetada
...
 }
 
+fn available_table_schemas(tables: &[TableInfo]) -> Vec<String> {
+    tables
+        .iter()
+        .map(|table| table.schema_name.clone())
+        .collect::<BTreeSet<_>>()
+        .into_iter()
+        .collect()
+}
```
sauldhernandez, line 581: `same_schema_tables` on the next line is catalog-filtered, but this list collapses schemas across catalogs: with `warehouse.public` and `analytics.public` both installed, a miss returns a single `public` and the agent can't tell which catalog it belongs to. And since `table_reference_matches` requires the catalog for catalog-backed tables, retrying `describe_table` with a schema from this list and no catalog misses again — the only path left is the unfiltered `list_catalog`. The guide resource already renders these as `warehouse.public`.

I can see why raw names — they have to stay usable as the `schema` argument. Would catalog/schema pairs work instead, so the value stays addressable and the qualifier survives? The proto comment change to "raw schema names" reads deliberate, so: planned for a later PR, or is the extra `list_catalog` suggestion meant to cover it?

crates/coral-app/src/catalog/discovery.rs
```diff
@@ -621,4 +718,76 @@ mod tests {
     fn empty_metadata_pattern_is_invalid() {
         compile_metadata_regex(" ", true).expect_err("empty pattern should fail");
     }
+
+    #[test]
+    fn database_catalog_and_schema_match_catalog_discovery_metadata() {
```
sauldhernandez, line 747: This one test asserts ~17 behaviors across six functions (`table_matches_ref`, `table_matched_fields`, `available_table_schemas`, `same_schema_tables`, `missing_table_suggestions`, `table_metadata_contains_literal`). If an early assertion fails — say the `QualifiedSchema` ordering — everything below it never runs, and the generic name doesn't say what broke. The neighbours in this module are one behavior per test with a name that states it (`required_filters_match_each_filter_independently`). Worth splitting into a few named tests, roughly one per helper.

crates/coral-app/src/search/catalog/snapshot.rs
```diff
@@ -193,6 +193,10 @@ fn normalized_runtime_schema_owners(
 }
 
 fn table_documents(table: &TableInfo, documents: &mut Vec<CatalogDocument>) {
+    debug_assert!(
+        table.catalog_name.is_none(),
+        "catalog search documents do not yet support catalog-qualified tables"
+    );
```
sauldhernandez, line 200: `debug_assert!` compiles out in release, so this guards nothing in the build that matters. `doc_id` and `qualified_name` stay keyed on `schema.table` alone, so once a later PR registers catalog-backed sources, two catalogs exposing the same `schema.table` produce the same `doc_id` — that violates `PRIMARY KEY (workspace, doc_id)` on `catalog_documents`, the projection refresh rolls back (`duplicate_document_refresh_rolls_back_documents_fts_owners_and_fingerprint`), and search silently keeps serving a stale index. `owner_source_name` would also be the schema rather than the source, and `searchable_text` omits the catalog, so catalog tables aren't findable by catalog name.

Unreachable today (`SourceQualifiedName` only has `Schema`), and #1927 doesn't close it. Which PR in the stack makes these documents catalog-aware? If it's not soon, folding `catalog_name` into `doc_id`/`qualified_name` — or a `tracing::warn!` + skip — would degrade instead of panicking in debug and colliding in release.

crates/coral-client/src/search.rs
```diff
@@ -719,14 +735,37 @@ pub fn minimal_table_function_call_example(function: &TableFunction) -> String {
...
+
 /// Formats a schema-qualified SQL table or table-function reference.
 #[must_use]
-pub fn format_schema_table_equivalent(schema_name: &str, table_name: &str) -> String {
-    format!(
-        "{}.{}",
-        format_sql_identifier(schema_name),
-        format_sql_identifier(table_name)
-    )
+pub fn format_schema_table_equivalent(
```
sauldhernandez, line 763: The doc line above still says "schema-qualified", but this now emits `catalog.schema.table`. And seven call sites pass a literal `""` for the catalog — table functions (search.rs 206, 604, 727; coral-mcp catalog.rs 627) plus the column-hint/observed-value surfaces whose protos carry no catalog field (371, 407, 638). All three params are `&str`, so a transposed argument compiles cleanly and produces a wrong SQL reference. Options: take `Option<&str>` for the catalog, or keep the two-arg schema-qualified form and add a separate catalog-qualified one for the table paths. Either way the doc line needs a refresh.

crates/coral-engine/src/runtime/mod.rs
```diff
@@ -10,7 +10,7 @@ pub(crate) const DATAFUSION_DEFAULT_CATALOG: &str = "datafusion";
 /// Removes `DataFusion`'s synthetic default catalog from a table qualifier.
 /// Schema-backed Coral tables store no catalog name in metadata, even
 /// when `DataFusion` expands an explicit reference to `datafusion.schema.table`.
-pub(crate) fn non_default_catalog_name(catalog_name: Option<&str>) -> Option<&str> {
+pub fn normalize_catalog_name(catalog_name: Option<&str>) -> Option<&str> {
```
sauldhernandez, line 21: This is what's failing CI: making the function `pub` trips `clippy::must_use_candidate` (pedantic is warn-level workspace-wide), and Rust checks runs clippy with `-D warnings`, so the job aborts at the clippy step before tests even build. Adding `#[must_use]` clears it.

Separate, while you're here: now that this is public engine API for catalog qualifiers, would it fit better in `contracts/catalog.rs` next to `TableInfo`/`CatalogInfo`? `DATAFUSION_DEFAULT_CATALOG` also stays `pub(crate)`, so external callers can't see which name gets filtered.

sauldhernandez, line 21: This now runs on client-supplied catalog names (the four catalog RPCs and the MCP tools), not just DataFusion-generated qualifiers — and `datafusion` isn't in `RESERVED_SOURCE_SCHEMA_NAMES` (`coral`, `coral_admin`, `public`). Once database sources land, a source named `datafusion` gets `catalog_name = Some("datafusion")` in table metadata, but requests for it are stripped to `None`, so `describe_table`/`list_columns` can never find it.

Reserving `datafusion` as a source name — here or in the PR that registers database sources — closes this cheaply. Is that already planned later in the stack?

crates/coral-mcp/src/guide_template.md
```diff
@@ -38,6 +38,8 @@ SELECT schema_name, key FROM coral.inputs
 WHERE kind = 'secret' AND is_set;
 ```
 
+Database source inputs are addressed by catalog instead: their rows have an empty `schema_name` and set `catalog_name`. When joining `coral.inputs` to `coral.tables`, join on `schema_name` for two-part sources and on `catalog_name` for database sources.
```
sauldhernandez: This join rule breaks for a mixed-catalog query. Both `coral.inputs` and `coral.tables` use `''` as `catalog_name` for two-part sources, so the natural rendering — `ON i.schema_name = t.schema_name OR i.catalog_name = t.catalog_name` — matches every two-part input row against every two-part table row via `'' = ''`. Cross product, wrong attributions.

Consider spelling out the guarded form, e.g. `ON (i.catalog_name <> '' AND i.catalog_name = t.catalog_name) OR (i.catalog_name = '' AND t.catalog_name = '' AND i.schema_name = t.schema_name)`, or showing a worked example query instead of the prose rule.

sauldhernandez: Nothing in this PR (or #1927) can produce these rows yet: `SourceQualifiedName` only has a `Schema` variant, and `catalog_input_rows` always fills `schema_name` and leaves `catalog_name` empty. So this documents a row shape no code in the stack implements — if the PR that adds database sources shapes `coral.inputs` differently, `coral://guide` will silently disagree with the engine. Is the docs half meant to land here, or move to that PR? If it stays, worth a follow-up check that the implementing PR matches this exact shape.

crates/coral-mcp/src/surface/catalog.rs
```diff
@@ -159,6 +166,7 @@ pub(crate) fn describe_table_arguments(
     arguments: Option<&Map<String, Value>>,
 ) -> Result<DescribeTableArguments, ErrorData> {
     Ok(DescribeTableArguments {
+        catalog: optional_string_argument(arguments, "catalog")?,
```
sauldhernandez: `optional_string_argument` keeps `Some("")`, so `"catalog": ""` is catalog-absent to the server (`optional_trimmed` in catalog/service.rs) but catalog-present in this response layer. On a missing table that means `requested.catalog` echoes `""`, the first suggested call is `list_catalog(catalog: "")`, and when `same_schema_tables` is empty the `catalog.is_some()` branch appends a second, otherwise-identical unqualified `list_catalog` — two duplicate recovery hints. `list_columns` likewise serializes `"catalog_name": ""` instead of omitting it. Switching the three `catalog` parsers to `optional_non_empty_string_argument` (as `pattern` already does) keeps presence semantics the same on both sides.

crates/coral-mcp/src/surface/resources.rs
```diff
@@ -80,14 +80,14 @@ pub(crate) fn guide_resource_content(
...
+        .filter(|table| !is_coral_system_schema(table))
+        .map(addressable_schema_name)
         .collect::<BTreeSet<_>>();
     schemas.extend(
         table_function_schema_names
             .iter()
-            .map(String::as_str)
-            .filter(|schema| *schema != "coral"),
+            .filter(|schema| *schema != "coral")
+            .cloned(),
```
sauldhernandez: Table schemas now go through `prompt_safe_text` (via `addressable_schema_name`), but this `extend` still inserts table-function schema names raw — `server.rs` pushes `function.schema_name` unsanitized, and per the comment in `source_names.rs` source names are only validated as path segments, so they can contain newlines. A source that publishes table functions can still break out of its `Visible schemas:` bullet — the injection `prompt_safe_text` exists to stop. Mapping this iterator through `prompt_safe_text` too closes the gap.

crates/coral-mcp/src/surface/resources.rs
```diff
@@ -164,14 +174,41 @@ fn tables_resource_description(visible_table_count: usize) -> String {
     format!("Fully qualified database tables in Coral ({visible_table_count} table(s)).")
 }
 
-fn first_visible_table(tables: &[TableSummary]) -> Option<(&str, &str)> {
+fn sql_string_literal(value: &str) -> String {
+    format!("'{}'", value.replace('\'', "''"))
+}
```
sauldhernandez, line 182: Two hardening changes ride along that the catalog work doesn't imply: the generated `coral.columns` example now escapes single quotes (`sql_string_literal`), and the visible-schemas list runs names through `prompt_safe_text` — both previously interpolated raw. Both are good, and output only differs for names containing quotes or control characters. Worth a line in the PR description so a reader diffing guide output knows the change there isn't purely catalog plumbing.

crates/coral-mcp/src/surface/resources.rs
```diff
@@ -164,14 +174,41 @@ fn tables_resource_description(visible_table_count: usize) -> String {
...
+                table.name.as_str(),
+            )
+        })
+}
+
+fn is_coral_system_schema(table: &TableSummary) -> bool {
+    table.catalog_name.is_empty() && table.schema_name == "coral"
+}
+
+fn addressable_schema_name(table: &TableSummary) -> String {
```
sauldhernandez: The "Visible schemas" list now mixes plain schema names with `catalog.schema` composites, and nothing marks which is which. The trap is worst in a mixed deployment: `first_visible_table` sorts empty-catalog tables first, so with any two-part source visible the columns example shows only the `WHERE schema_name = ...` form — an agent that copies `warehouse.public` from the list into `schema_name = 'warehouse.public'` gets zero rows and concludes the table has no columns.

Either render the parts distinctly (e.g. `- public (catalog: warehouse)`) or add a line to the guide saying dotted entries are `catalog.schema` and must be split across `catalog_name` and `schema_name` when filtering.

sauldhernandez (approved):

PR #43, opened 2026-04-09
feat(cli): removes source import, combines into source add --file

iaptsiauri:
## Summary
Removes `source import` in favor of `source add --file`

## Testing plan
`make validate`
Manual tests

PR #1896, opened 2026-07-21
feat(app): connect DSL v4 database sources

iaptsiauri:
## Summary

- parse PostgreSQL, MySQL, and SQLite database surfaces in DSL v4 source manifests
- assemble database runtime components without materializing remote descriptors
- validate database connection templates and reserve the DataFusion namespace
- cover runtime assembly, materialization bypass, snapshot behavior, and parse failures

## Stack

- Depends on: #1929
- Final PR in the stack

## Validation

- make schema-check: passed
- make rust-checks: 1891 tests passed, 2 skipped; clippy and rustdoc passed
- make perf-check: release build passed; benchmark was not executed because hyperfine is unavailable locally

This is PR 7 of 7 in the relational database source MVP stack.

### DIFF
--- a/crates/coral-app/src/sources/materialization.rs
+++ b/crates/coral-app/src/sources/materialization.rs
@@ -846,16 +846,17 @@ fn validate_semantic_ir(
     {
         return Err("semantic IR identity mismatch".to_string());
     }
-    if semantic_ir.importer_version != expected_importer_version(manifest.surface.surface_type) {
+    if semantic_ir.importer_version != expected_importer_version(manifest.surface.surface_type)? {
         return Err("semantic IR importer version mismatch".to_string());
     }
     Ok(())
 }
 
-fn expected_importer_version(surface_type: SurfaceType) -> &'static str {
+fn expected_importer_version(surface_type: SurfaceType) -> Result<&'static str, String> {
     match surface_type {
-        SurfaceType::OpenApi => OPENAPI_IMPORTER_VERSION,
-        SurfaceType::Mcp => MCP_IMPORTER_VERSION,
+        SurfaceType::OpenApi => Ok(OPENAPI_IMPORTER_VERSION),
+        SurfaceType::Mcp => Ok(MCP_IMPORTER_VERSION),
+        SurfaceType::Database => Err("database surfaces are not materialized".to_string()),
     }
 }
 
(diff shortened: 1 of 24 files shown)


crates/coral-app/src/sources/materialization.rs
```diff
@@ -284,13 +284,21 @@ pub(crate) fn build_v4_materialization_tmp(
     }
 }
 
-pub(crate) fn replace_v4_materialization(
+/// Stages the desired materialization state and returns any previous state for rollback.
+///
+/// A missing replacement retires an existing materialization while the source
+/// update is committed.
+pub(crate) fn replace_or_retire_v4_materialization(
```
simonwhitaker, line 291: What's the use case where "a missing replacement retires an existing materialisation"? Is this a workaround so a DB source can pass None as the replacement? In which case it shouldn't have had anything to replace in the first place. Would a better solution be to not call this function in the first place for database sources?

I'm thinking about the case where some code mishap results in us passing None for a source that should be materialised (i.e. a non-database source), and this function will then just happily nuke the existing materialisation. 

(I might be misunderstanding this code!)

crates/coral-app/src/sources/mod.rs
```diff
@@ -9,3 +14,19 @@ pub(crate) mod runtime_package;
 pub(crate) mod service;
 
 pub(crate) use name::SourceName;
+
+pub(crate) fn ensure_database_source_feature_enabled(
```
simonwhitaker, line 18: SimonLint warning #43: agent sneaking code into mod.rs again. 🤖

crates/coral-spec/src/v4/manifest.rs
```diff
@@ -307,6 +341,126 @@ fn parse_mcp_surface(
     })
 }
 
+fn parse_database_surface(
+    source_name: &str,
+    raw_surface: RawV4Surface,
+    surface_value: &Value,
+) -> Result<V4Surface> {
+    if raw_surface.url.is_some() || raw_surface.file.is_some() || raw_surface.server.is_some() {
```
simonwhitaker, line 349: `RawV4Surface` is becoming a beast. I wonder if we can do this better, e.g. using an enum with explicit arms for each surface type. Don't worry about doing that in this PR, but would be fun/interesting as a follow-up if you fancy it.

simonwhitaker (approved):

PR #1900, opened 2026-07-21
fix(telemetry): redact caller-controlled error details

mystic123:
## Summary

- replace caller-controlled gRPC and MCP error text in exported telemetry with stable messages
- preserve categorical gRPC codes and structured Coral error reasons for diagnosis
- keep detailed errors visible to callers while preventing those details from entering span attributes or status
- stop exporting raw `mcp.tool.intent`; task intent is still validated and persisted for task behavior
- add regression tests covering server gRPC, client gRPC, MCP protocol/tool errors, and a real `start_task` call

## Stack boundary

- Layer 1 of 6
- Base: `main`
- Next: #1901
- This PR only hardens telemetry privacy. It does not add Query Stream operation markers, projection, detail selection, or Reef changes.
- It changes newly emitted telemetry only; retained or previously exported telemetry is not rewritten.

## Operational impact

- Caller-visible gRPC and MCP errors keep their original detail, but exported traces intentionally replace free-form error text with stable messages.
- Later layers improve operation attribution and presentation; they do not restore the removed free-form error detail.
- Diagnose trace failures using `error.type`, the RPC status code, service/method fields, and the operation metadata added by later layers.
- `mcp.tool.intent` is no longer exported. Task intent is still validated and persisted for task behavior, but external dashboards or queries that read this telemetry field must migrate.

## Deferred follow-ups

- centralize categorical gRPC error extraction, including an explicit first-versus-last policy if a status contains multiple Coral `ErrorInfo` details
- audit dynamic span events and add a durable sanitization or allowlist boundary; this PR's regression tests prove the exercised request paths, not every future event call site
- review whether the stable `exception.message` attribute should be removed after checking external telemetry consumers

## Validation

- `cargo test -p coral-mcp --all-features`: 87 passed; the end-to-end intent regression explicitly captures MCP, client gRPC, and server gRPC spans
- `cargo fmt --all --check`
- current top-of-stack `npm test --prefix apps/reef`: 88 files and 387 tests passed
- current top-of-stack `npm run check --prefix apps/reef`
- current top-of-stack `npm run typecheck --prefix apps/reef`
- current top-of-stack `npm run build --prefix apps/reef`
- `git diff --check`

### DIFF
--- a/crates/coral-app/src/transport.rs
+++ b/crates/coral-app/src/transport.rs
@@ -24,7 +24,7 @@ use tonic::metadata::MetadataMap;
 use tonic::{Code, Request, Status};
 use tonic_types::{ErrorDetail, StatusExt as _};
 use tower::Layer;
-use tracing::{Instrument as _, field};
+use tracing::Instrument as _;
 use tracing_opentelemetry::OpenTelemetrySpanExt as _;
 
 use crate::bootstrap::{AppError, app_status, core_status, status_with_bounded_detail};
@@ -43,6 +43,7 @@ use crate::workspaces::WorkspaceName;
 struct MetadataExtractor<'a>(&'a tonic::metadata::MetadataMap);
 
 const USER_PRINCIPAL_PROVIDER_TIMEOUT: Duration = Duration::from_secs(30);
+const GRPC_REQUEST_ERROR_MESSAGE: &str = "gRPC request failed";
 
 #[derive(Clone)]
 struct GrpcServerSpan(tracing::Span);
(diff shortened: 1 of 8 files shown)


crates/coral-app/src/transport.rs
```diff
@@ -324,41 +325,24 @@ fn record_grpc_status(span: &tracing::Span, code: Code, status: Option<&Status>)
...
-            },
-            decode_grpc_error,
-        );
+        let error_type = status.map_or_else(|| response_status_code.to_string(), grpc_error_type);
         span.record("status", "error");
-        span.record("error.type", error.error_type.as_str());
-        span.record("exception.message", field::display(error.message.as_str()));
-        span.set_status(OtelStatus::error(error.message));
+        span.record("error.type", error_type.as_str());
+        span.record("exception.message", GRPC_REQUEST_ERROR_MESSAGE);
```
jsummerfield: This redaction policy is now implemented independently in `coral-app`, `coral-client`, and `coral-mcp`, and the exact `"gRPC request failed"` message is duplicated in the app and client. Would it be worth moving a small `coral_telemetry::record_failure(span, error_type, stable_message: &'static str)` helper plus the shared gRPC message into `coral-telemetry`? Classification could remain adapter-local, while the helper would centralize the four field/status writes and make the privacy boundary explicit by preventing dynamic error text at the type level. I would keep the MCP-specific messages in `coral-mcp`; this does not need a cross-protocol error enum.

jsummerfield (commented):
One non-blocking maintainability suggestion.

jsummerfield (approved):

PR #1901, opened 2026-07-21
feat(telemetry): mark user-visible operations

mystic123:
## TL;DR

PR #1901 labels trace spans so later Query Stream code can distinguish real user-visible operations from their internal work.

For example, an agent's Search call may internally run `LIST CATALOG`. This layer marks Search as the outer operation and the nested SQL as an implementation detail. Search-index rebuilds are likewise marked as their own maintenance operation instead of appearing as `LIST CATALOG`.

## Summary

- define shared `coral.stream.entry`, `coral.stream.kind`, and `coral.stream.name` attributes for Query, Search, MCP `tools/call`, and other explicit user-visible operations
- canonicalize MCP operation names through the same `ToolName` parser used for dispatch; rejected names are recorded only as `unknown_tool`
- preserve the active Tool trace context across spawned SQL work
- retain validated, bounded Search text as `coral.local.search.query` only when Coral has installed local trace history
- record Search text directly as an OpenTelemetry span attribute, so it is not formatted to stderr or bridged into OTLP logs
- strip local-only attributes from exported OTLP spans, span events, and links while retaining them in local JSONL history
- keep host-owned and OTLP-only subscribers from receiving populated local-only attributes
- mark search-index rebuilds as explicit `rebuild_search_index` operations above their nested catalog queries
- mark a rebuild trace as failed when an otherwise successful response contains a failed provider, without exporting the provider's detailed note
- report a trace as failed when any operation in a mixed batch fails, even if the primary SQL query succeeded
- share workspace and failure-recording telemetry helpers across the app and MCP crates
- keep MCP discovery calls unmarked and redact caller-controlled Search error details

## Stack boundary

- Layer 2 of 5; #1900 is merged into `main`
- Depends on: `main`
- Next: #2043
- This PR emits operation markers and establishes the local-only span-attribute boundary. It does not read the markers, change the trace API, project Query Stream rows, or change Reef.

## Validation

- `make rust-checks` passed on the final #1901 head: formatting, workspace-wide all-features Clippy, 2,376 nextest tests passed with 6 configured skips, and docs with warnings denied
- focused Search-maintenance tests cover Partial-as-success, failed-provider error classification, unchanged provider results, and provider-note privacy
- focused Search-operation, mixed-batch summary, server-order, host-subscriber, and OTLP loopback tests passed
- the real OTLP loopback proves exact Search text is retained in local history but absent from outgoing OTLP traces and logs
- descendant `range-diff` checks were patch-identical after restacking #2043, #1876, and #1904
- `git diff --check` passed

### DIFF
--- a/crates/coral-app/src/query/manager.rs
+++ b/crates/coral-app/src/query/manager.rs
@@ -1153,6 +1153,7 @@ where
         query_span.set_status(OtelStatus::error(error_message));
     }
 
+    drop(query_span);
     result
 }
 
@@ -1563,6 +1564,51 @@ mod tests {
             span_attr(query_span, coral_telemetry::QUERY_STREAM_NAME_ATTRIBUTE),
             Some("execute_sql".to_string())
         );
+    }
+
+    #[tokio::test(flavor = "current_thread")]
+    async fn query_span_finishes_before_completed_future_is_dropped() {
+        use std::future::poll_fn;
+
+        use opentelemetry::trace::TracerProvider as _;
+        use opentelemetry_sdk::trace::{InMemorySpanExporter, SdkTracerProvider};
+        use tracing_subscriber::layer::SubscriberExt as _;
+
+        let exporter = InMemorySpanExporter::default();
+        let provider = SdkTracerProvider::builder()
+            .with_simple_exporter(exporter.clone())
+            .build();
+        let tracer = provider.tracer("query-span-lifecycle-test");
+        let subscriber = tracing_subscriber::Registry::default()
+            .with(tracing_opentelemetry::layer().with_tracer(tracer));
+        let _guard = tracing::subscriber::set_default(subscriber);
+
+        let workspace = WorkspaceName::default();
+        let mut operation = Box::pin(run_query_operation(
+            QueryOperation::ListTables,
+            &workspace,
+            "LIST TABLES *.*",
+            None,
+            async { Ok::<(), QueryManagerError>(()) },
+            |()| None,
+            |_, ()| {},
+        ));
+        poll_fn(|context| operation.as_mut().poll(context))
+            .await
+            .expect("query operation should succeed");
+
+        provider.force_flush().expect("flush spans");
+        let spans = exporter.get_finished_spans().expect("finished spans");
+        let query_span = spans
+            .iter()
+            .find(|span| span.name == "coral.query")
+            .expect("completed operation should finish its query span");
+        assert_eq!(
+            span_attr(query_span, "operation"),
+            Some("list_tables".to_string())
+        );
+
+        drop(operation);
     }
 
     #[tokio::test(flavor = "current_thread")]
--- a/crates/coral-app/src/query/manager.rs
+++ b/crates/coral-app/src/query/manager.rs
@@ -41,7 +41,7 @@ use crate::sources::runtime_package::{
 };
 use crate::state::{AppConfig, AppStateLayout, ConfigStore};
 use crate::task::id::TaskId;
-use crate::telemetry::WORKSPACE_SPAN_ATTRIBUTE;
+use crate::telemetry::app_error_type;
 use crate::workspaces::{
     WorkspaceLifecycleLock, WorkspaceLifecycleRevision, WorkspaceManager, WorkspaceName,
 };
@@ -577,7 +577,10 @@ impl QueryManager {
             workspace = tracing::field::Empty,
             source.count = tracing::field::Empty,
         );
-        span.record(WORKSPACE_SPAN_ATTRIBUTE, workspace_name.as_str());
+        span.record(
+            coral_telemetry::WORKSPACE_SPAN_ATTRIBUTE,
+            workspace_name.as_str(),
+        );
         let _guard = span.enter();
         let mut loaded_sources = Vec::new();
         let mut failed_source_names = BTreeSet::new();
@@ -1182,7 +1185,10 @@ fn create_query_span(
     if let Some(task_id) = task_id {
         span.record("task.id", tracing::field::display(task_id));
     }
-    span.record(WORKSPACE_SPAN_ATTRIBUTE, workspace_name.as_str());
+    span.record(
+        coral_telemetry::WORKSPACE_SPAN_ATTRIBUTE,
+        workspace_name.as_str(),
+    );
     span
 }
 
@@ -1254,45 +1260,6 @@ fn query_error_message(error: &QueryManagerError) -> String {
         QueryManagerError::App(error) => error.to_string(),
         QueryManagerError::Core(CoreError::QueryFailure(error)) => error.summary().to_string(),
         QueryManagerError::Core(error) => error.to_string(),
-    }
-}
-
-pub(crate) fn app_error_type(error: &AppError) -> &'static str {
-    match error {
-        AppError::Unauthenticated(_) => "UNAUTHENTICATED",
-        AppError::SourceNotFound(_) => "SOURCE_NOT_FOUND",
-        AppError::FunctionNotFound(_) => "FUNCTION_NOT_FOUND",
-        AppError::FunctionAlreadyExists(_) => "FUNCTION_ALREADY_EXISTS",
-        AppError::WorkspaceNotFound(_) => "WORKSPACE_NOT_FOUND",
-        AppError::WorkspaceAlreadyExists(_) => "WORKSPACE_ALREADY_EXISTS",
-        AppError::InvalidInput(_) => "INVALID_INPUT",
-        AppError::FailedPrecondition(_) => "FAILED_PRECONDITION",
-        AppError::MissingSourceInputs { .. } => "MISSING_SOURCE_INPUTS",
-        AppError::UnsupportedV4IdentityRequirements { .. } => {
-            "UNSUPPORTED_V4_IDENTITY_REQUIREMENTS"
-        }
-        AppError::MissingOrIncompatibleV4Materialization { .. } => {
-            "MISSING_OR_INCOMPATIBLE_V4_MATERIALIZATION"
-        }
-        AppError::IncompatibleInstalledV4Manifest { .. } => "INCOMPATIBLE_INSTALLED_V4_MANIFEST",
-        AppError::InvalidV4ProjectionOverride { .. } => "INVALID_V4_PROJECTION_OVERRIDE",
-        AppError::InvalidV4OperationMetadataOverride { .. } => {
-            "INVALID_V4_OPERATION_METADATA_OVERRIDE"
-        }
-        AppError::CredentialRefresh(_) => "CREDENTIAL_REFRESH",
-        AppError::Unavailable(_) => "UNAVAILABLE",
-        AppError::ResourceExhausted(_) => "RESOURCE_EXHAUSTED",
-        AppError::Internal(_) => "INTERNAL",
-        AppError::Io(_) => "IO",
-        AppError::Yaml(_) => "YAML",
-        AppError::TomlDecode(_) | AppError::TomlEditDecode(_) => "TOML_DECODE",
-        AppError::TomlEncode(_) => "TOML_ENCODE",
-        AppError::Json(_) => "JSON",
-        AppError::Transport(_) => "TRANSPORT",
-        AppError::TaskJoin(_) => "TASK_JOIN",
-        AppError::Credentials(_) => "CREDENTIALS",
-        AppError::Database(_) => "DATABASE",
-        AppError::MissingConfigDir => "MISSING_CONFIG_DIR",
     }
 }
 
--- a/crates/coral-app/src/query/manager.rs
+++ b/crates/coral-app/src/query/manager.rs
@@ -1162,6 +1162,9 @@ fn create_query_span(
     let operation = operation.as_str();
     let span = tracing::info_span!(
         "coral.query",
+        coral.stream.entry = true,
+        coral.stream.kind = coral_telemetry::QUERY_STREAM_KIND_QUERY,
+        coral.stream.name = operation,
         otel.name = "coral.query",
         operation = operation,
         workspace = tracing::field::Empty,
@@ -1254,7 +1257,7 @@ fn query_error_message(error: &QueryManagerError) -> String {
     }
 }
 
-fn app_error_type(error: &AppError) -> &'static str {
+pub(crate) fn app_error_type(error: &AppError) -> &'static str {
     match error {
         AppError::Unauthenticated(_) => "UNAUTHENTICATED",
         AppError::SourceNotFound(_) => "SOURCE_NOT_FOUND",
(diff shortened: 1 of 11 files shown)


crates/coral-app/src/search/manager.rs
```diff
@@ -543,6 +559,56 @@ fn workspace_changed_error(operation: &str) -> SearchManagerError {
...
+    let result = operation.instrument(span.clone()).await;
+    if let Ok(response) = &result {
+        span.record(
+            "result_count",
+            u64::try_from(response.results.len()).unwrap_or(u64::MAX),
+        );
+        span.record("status", "ok");
+        span.set_status(OtelStatus::Ok);
+    } else {
+        span.record("error.type", "SearchManagerError");
```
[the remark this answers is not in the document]
mystic123: Updated failed Search spans to reuse the bounded AppError category taxonomy (INTERNAL, UNAVAILABLE, and so on) while keeping the exported exception and status message generic. The regression asserts INTERNAL without exporting the caller-controlled sentinel.

crates/coral-app/src/query/manager.rs
```diff
@@ -1424,6 +1427,18 @@ mod tests {
             task_attr.value.as_str(),
             "550e8400-e29b-41d4-a716-446655440000"
         );
+        assert_eq!(
```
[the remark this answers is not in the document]
mystic123: Changed the Query Stream marker assertion to compare the raw OpenTelemetry value with Value::Bool(true). The current dynamic task ID assertion is also preserved after the rebase.

crates/coral-mcp/src/tests.rs
```diff
@@ -854,6 +884,165 @@ async fn task_intent_is_persisted_but_not_exported_to_telemetry() {
...
+        .build();
+    let tracer = provider.tracer("mcp-unknown-tool-name-privacy-test");
+    let trace_targets = "coral_mcp=trace"
+        .parse::<Targets>()
+        .expect("MCP trace filter");
+    let subscriber = tracing_subscriber::Registry::default().with(
+        tracing_opentelemetry::layer()
+            .with_tracer(tracer)
+            .with_filter(trace_targets),
+    );
```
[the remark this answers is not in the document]
mystic123: Broadened the regression from coral_mcp to all Coral-owned coral_* targets and kept the sentinel scan over every captured first-party span. I did not make the layer fully unfiltered because third-party rmcp debug events can record raw protocol payloads and are intentionally outside Coral telemetry.

mystic123:
* **#1876** <a href="https://app.graphite.com/github/pr/withcoral/coral/1876?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1902](https://github.com/withcoral/coral/pull/1902) <a href="https://app.graphite.com/github/pr/withcoral/coral/1902?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2053](https://github.com/withcoral/coral/pull/2053) <a href="https://app.graphite.com/github/pr/withcoral/coral/2053?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#2043** <a href="https://app.graphite.com/github/pr/withcoral/coral/2043?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1901** <a href="https://app.graphite.com/github/pr/withcoral/coral/1901?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1901?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/coral-app/src/query/manager.rs
```diff
@@ -1162,6 +1162,9 @@ fn create_query_span(
     let operation = operation.as_str();
     let span = tracing::info_span!(
         "coral.query",
+        coral.stream.entry = true,
```
sauldhernandez, line 1169: `rebuild_index` (search/manager.rs) reaches the same `preload_catalog` -> `resolve_catalog` path as `search`, but nothing wraps it in a marked span. Its `coral.query` / `list_catalog` span is then the highest marked span in the trace, so #1876 projects `coral search-index rebuild` as a user-visible Query row whose text is the synthetic `LIST CATALOG` SQL the user never wrote.

Search is fine now that its preload nests under `coral.search`. Is the rebuild meant to show up this way? If not, a wrapper span like search's would suppress the nested operation and name the row after the actual maintenance action. Note that leaving the span unmarked wouldn't help: #1876's legacy fallback still projects any span named `coral.query`.

crates/coral-app/src/query/manager.rs
```diff
@@ -1254,7 +1257,7 @@ fn query_error_message(error: &QueryManagerError) -> String {
     }
 }
 
-fn app_error_type(error: &AppError) -> &'static str {
+pub(crate) fn app_error_type(error: &AppError) -> &'static str {
```
sauldhernandez: Placement thought: `app_error_type` is an AppError-to-`error.type` mapping, and the only reason it becomes `pub(crate)` here is so the search manager can use it for its own spans. That makes search depend on the query manager for something that isn't about querying, and the next subsystem needing an `error.type` will deepen the same edge. `crate::telemetry` already hosts the shared span-attribute helpers (`WORKSPACE_SPAN_ATTRIBUTE`, `record_local_only_span_attribute`) and search/manager.rs imports from it in this PR — moving the function there (or next to `AppError` in bootstrap) keeps ownership where it belongs. Fine as a follow-up.

crates/coral-app/src/search/manager.rs
```diff
@@ -556,6 +573,65 @@ fn workspace_changed_error(operation: &str) -> SearchManagerError {
...
+            );
+            span.record("status", "ok");
+            span.set_status(OtelStatus::Ok);
+        }
+        Err(SearchManagerError::App(error)) => {
+            span.record("error.type", app_error_type(error));
+            span.record("exception.message", SEARCH_TELEMETRY_ERROR_MESSAGE);
+            span.record("status", "error");
+            span.set_status(OtelStatus::error(SEARCH_TELEMETRY_ERROR_MESSAGE));
+        }
```
sauldhernandez, line 670: This is `coral_telemetry::record_failure(&span, app_error_type(error), SEARCH_TELEMETRY_ERROR_MESSAGE)` open-coded — same four recordings, and both arguments already have the `&'static str` shape the helper wants. `transport.rs`, `coral-client/grpc.rs`, and `coral-mcp/telemetry.rs` all go through it, so if that shared failure contract ever grows a field, `coral.search` would be the one span reporting a different error shape. (query/manager.rs can't use the helper — its message is a caller-derived `String` — but this one fits.)

crates/coral-app/src/search/manager.rs
```diff
@@ -556,6 +573,65 @@ fn workspace_changed_error(operation: &str) -> SearchManagerError {
...
+        workspace = field::Empty,
+        query_len_bytes = request.query.len(),
+        limit = request.limit,
+        task.id = field::Empty,
+        result_count = field::Empty,
+        status = field::Empty,
+        error.type = field::Empty,
+        exception.message = field::Empty,
+    );
+    record_local_only_span_attribute(
```
sauldhernandez, line 692: `coral.local.search.query` is stripped from OTLP export and gated on Coral owning the subscriber, but the stderr `fmt` layer still sees it. The compact formatter appends the fields of every span in scope to each event line, including fields added later via `span.record`. Under `coral mcp-stdio` (stderr logs on, `coral_app=info` filter, trace history on by default), any `coral_app` warn/error emitted inside the search span writes the raw query to stderr, which the MCP host captures into its own log files — logs that get no retention pruning and aren't touched by delete-workspace-traces.

Today the only reachable site is the provider-task-failure `error!` in `search/engine.rs:70` (it runs under `Span::current()`), so it takes a panicking provider to trigger — but the exposure grows with every event added inside this span, and the boundary is meant to be structural rather than incidental.

Two options: wrap the stderr layer's field formatting to skip `coral.local.`-prefixed fields, or keep the query out of the span entirely and pass it to the local trace store by a path the fmt layer never sees.

crates/coral-app/src/search/manager.rs
```diff
@@ -669,3 +745,170 @@ fn search_storage_cleanup_result(
...
+                .iter()
+                .any(|attribute| attribute.key.as_str() == "limit")
+        );
+        assert!(
+            search_span
+                .attributes
+                .iter()
+                .any(|attribute| attribute.key.as_str() == "result_count")
+        );
+        assert!(
```
sauldhernandez, line 1077: These two tests are order-dependent: `cargo test -p coral-app --lib` fails them when `bootstrap::server::tests::configured_standalone_grpc_starts_with_configured_bind` runs first in the same process (reproducible with `-- --test-threads=1 configured_standalone_grpc_starts_with_configured_bind search::manager::tests`; they pass in isolation).

Cause: `record_local_only_span_attribute` gates on the process-global `INIT` OnceLock. The server test doesn't call `disable_internal_tracing`, and `TraceHistoryConfig::default().enabled` is true, so its `.start()` installs a local trace store in `INIT` for the rest of the process. The search spans then do record `coral.local.search.query`, and this assertion plus the `!contains(query)` check at line 911 both fire. CI stays green only because `make rust-checks` runs nextest, which gives each test its own process.

Two possible fixes: pass the "local store installed" decision into `create_search_span` as a parameter so the tests control it, or have the server test call `disable_internal_tracing` like its siblings (which papers over the shared-global coupling but fixes the ordering flake).

crates/coral-app/src/telemetry/mod.rs
```diff
@@ -46,12 +46,40 @@ static METER_PROVIDER: Mutex<Option<SdkMeterProvider>> = Mutex::new(None);
...
+    );
+    if !is_local_only {
+        return;
+    }
+    let local_trace_store_is_installed = INIT
+        .get()
+        .and_then(|state| state.as_ref().ok())
+        .is_some_and(|state| state.local_trace_store.is_some());
+    if local_trace_store_is_installed {
+        span.record(field, value);
```
sauldhernandez: Nothing guards that the span actually declared the field — `span.record` on an undeclared field is a silent no-op. `create_search_span` declares the literal `coral.local.search.query = field::Empty` while this records through `QUERY_STREAM_SEARCH_QUERY_ATTRIBUTE`, and nothing links the two. Rename either side and the search text silently stops reaching local trace history, with CI green — and #1876 reads exactly that attribute for the Search operation text.

No test would catch it: every local-only assertion is in the "attribute is absent" direction, and the one positive assertion (`telemetry_otlp_loopback.rs`) uses a hand-written span with `coral.local.future_tool.input`, so it never touches this helper or the search key.

Either fix works: a `debug_assert!(span.is_disabled() || span.has_field(field), ...)` next to the prefix assert, or drive a real Search in the loopback test (it already boots with `[trace_history]` on) and assert the query lands in local history and stays out of the OTLP export.

crates/coral-app/src/telemetry/mod.rs
```diff
@@ -108,6 +143,14 @@ where
...
                 && !span_matches_excluded_rpc_service(span, self.excluded_rpc_services)
         });
+        for span in &mut batch {
+            span.attributes.retain(|attribute| {
+                !self
+                    .stripped_attribute_prefixes
+                    .iter()
+                    .any(|prefix| attribute.key.as_str().starts_with(prefix))
+            });
+        }
```
sauldhernandez, line 198: The strip only walks `span.attributes`. tracing-opentelemetry puts event fields into `span.events[*].attributes` — a `tracing::info!(coral.local.foo = ...)` inside an instrumented span would go out over OTLP unstripped. Nothing emits that today, but the prefix is documented as a generic fail-closed namespace, so this is a trap for the next local-only key. Retaining over `span.events[*].attributes` (and `links`) here would close it.

crates/coral-app/tests/telemetry_host_subscriber.rs
```diff
@@ -54,6 +65,31 @@ async fn host_subscriber_does_not_block_server_startup() {
...
+    );
+    assert!(spans.iter().all(|span| {
+        span.attributes.iter().all(|attribute| {
+            !attribute
+                .key
+                .as_str()
+                .starts_with(coral_telemetry::LOCAL_ONLY_SPAN_ATTRIBUTE_PREFIX)
+        })
+    }));
+    assert!(!format!("{spans:?}").contains(SEARCH_SENTINEL));
```
sauldhernandez, line 95: These new assertions pin a second contract — no `coral.local.*` attribute or query text reaches a host-owned subscriber — but the test name and module doc still only describe startup. If the stripping regresses, CI reports a failure in "does not block server startup", pointing the reader at bootstrap. Co-locating is right (the OnceLock forces one subscriber per binary); just rename to cover both (e.g. `host_subscriber_starts_server_and_never_receives_local_only_attributes`) or add a line to the module doc. The other telemetry tests name the behavior they pin (`target_filtering_exporter_strips_local_only_span_attributes`, etc.).

crates/coral-mcp/src/server.rs
```diff
@@ -556,13 +557,16 @@ impl CoralMcpServer {
...
-            });
+            tasks.spawn(
+                async move {
+                    with_task_metadata(
+                        task_id_metadata,
+                        server.execute_one_sql_query(index, sql, shown_guide_ids),
+                    )
+                    .await
+                }
+                .in_current_span(),
```
sauldhernandez, line 568: Re-parenting the spawned queries fixes the lost trace context, but it also merges a multi-query `sql` batch into one trace, and the default trace list emits one row per trace. `summary_from_list_aggregate` takes status, SQL, and row_count from the primary span — the earliest `coral.query` — and only falls back to `error_count` when that span has no `status` attribute. Successful query spans record `status = "ok"`, so `sql(["SELECT 1", "SELECT bad"])` now lists a single OK row with the first query's SQL and row count; the failure has no row and the trace reads as success. Before this change each query rooted its own trace and got its own row with its own status.

#1876's Query Stream view resolves this for Reef (one row per operation, per-operation status), but it deliberately keeps `TRACE_VIEW_UNSPECIFIED` as one-summary-per-trace for diagnostic callers — and `TraceSummary` carries no error count, so those callers have no other signal. Cheap guard: in the default view, let `error_count > 0` override an OK primary status so the surviving row reports the error.

crates/coral-mcp/src/telemetry.rs
```diff
@@ -59,6 +70,7 @@ pub(crate) fn call_tool_span(tool_name: &str, trace_parent: Option<&str>) -> tra
         otel.name = "coral.mcp.call_tool",
         task.id = field::Empty,
         status = field::Empty,
+        workspace = workspace,
```
sauldhernandez: `workspace` is now a cross-crate span key: this span writes the literal while coral-app reads it through its crate-private `WORKSPACE_SPAN_ATTRIBUTE`. For an unknown or failed tool call there is no child query span, so this one attribute decides whether the trace appears in a workspace-scoped Query Stream list — a spelling drift would silently drop those traces. Since this PR already promotes the `coral.stream.*` keys into `coral-telemetry`, the workspace key arguably belongs there too, with coral-app's constant pointing at it. The macro still needs the literal either way, same as the `coral.stream.*` fields.

sauldhernandez (approved):

mystic123:
### Merge activity

* **Aug 5, 9:52 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1901).
* **Aug 5, 9:53 AM UTC**: [Graphite](https://app.graphite.com/github/pr/withcoral/coral/1901) couldn't merge this PR because it failed for an unknown reason.

PR #1911, opened 2026-07-22
feat(app): persist GUI onboarding completion in local DB

AlbertQM:
## Summary

Adding a table to the local DB that stores userIDs that have completed onboarding; so that we can show them the onboarding only once.

## Test plan

npm run dev
migration runs
check table exists

image

### DIFF
--- a/crates/coral-app/src/gui_onboarding/manager.rs
+++ b/crates/coral-app/src/gui_onboarding/manager.rs
@@ -1,0 +1,34 @@
+use std::sync::Arc;
+
+use crate::bootstrap::AppError;
+use crate::identity::Principal;
+use crate::state::db::{CoralDb, DbRepos};
+
+#[derive(Clone)]
+pub(crate) struct GuiOnboardingManager {
+    db: Arc<CoralDb>,
+}
+
+impl GuiOnboardingManager {
+    pub(crate) fn new(db: Arc<CoralDb>) -> Self {
+        Self { db }
+    }
+
+    pub(crate) async fn is_completed(&self, principal: &Principal) -> Result<bool, AppError> {
+        let mut session = self.db.as_ref();
+        session
+            .gui_onboarding()
+            .is_completed(principal.id().as_str())
+            .await
+            .map_err(Into::into)
+    }
+
+    pub(crate) async fn complete(&self, principal: &Principal) -> Result<(), AppError> {
+        let mut session = self.db.as_ref();
+        session
+            .gui_onboarding()
+            .complete(principal.id().as_str())
+            .await?;
+        Ok(())
+    }
+}
--- a/crates/coral-app/src/gui_onboarding/manager.rs
+++ b/crates/coral-app/src/gui_onboarding/manager.rs
@@ -1,8 +1,10 @@
+//! App-owned per-user GUI onboarding completion behavior.
+
 use std::sync::Arc;
 
 use crate::bootstrap::AppError;
 use crate::identity::Principal;
-use crate::state::db::{CoralDb, DbRepos};
+use crate::state::db::{CoralDb, DbRepos, now_unix_nanos_i64};
 
 #[derive(Clone)]
 pub(crate) struct GuiOnboardingManager {
@@ -27,7 +29,7 @@ impl GuiOnboardingManager {
         let mut session = self.db.as_ref();
         session
             .gui_onboarding()
-            .complete(principal.id().as_str())
+            .complete(principal.id().as_str(), now_unix_nanos_i64()?)
             .await?;
         Ok(())
     }
--- a/crates/coral-app/src/gui_onboarding/service.rs
+++ b/crates/coral-app/src/gui_onboarding/service.rs
@@ -1,0 +1,68 @@
+use coral_api::v1::gui_onboarding_service_server::GuiOnboardingService as GuiOnboardingServiceApi;
+use coral_api::v1::{
+    CompleteGuiOnboardingRequest, CompleteGuiOnboardingResponse, GetGuiOnboardingStateRequest,
+    GetGuiOnboardingStateResponse,
+};
+use tonic::{Request, Response, Status};
+
+use crate::bootstrap::app_status;
+use crate::gui_onboarding::manager::GuiOnboardingManager;
+use crate::identity::Principal;
+use crate::request_context::RequestContext;
+use crate::transport::{grpc_span, instrument_grpc};
+
+#[derive(Clone)]
+pub(crate) struct GuiOnboardingService {
+    gui_onboarding: GuiOnboardingManager,
+}
+
+impl GuiOnboardingService {
+    pub(crate) fn new(gui_onboarding: GuiOnboardingManager) -> Self {
+        Self { gui_onboarding }
+    }
+}
+
+#[tonic::async_trait]
+impl GuiOnboardingServiceApi for GuiOnboardingService {
+    async fn get_gui_onboarding_state(
+        &self,
+        request: Request<GetGuiOnboardingStateRequest>,
+    ) -> Result<Response<GetGuiOnboardingStateResponse>, Status> {
+        let span = grpc_span(&request);
+        let principal = request_principal(&request)?;
+        let gui_onboarding = self.gui_onboarding.clone();
+        instrument_grpc(span, async move {
+            let completed = gui_onboarding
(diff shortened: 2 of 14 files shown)


.github/workflows/validate.yml
```diff
@@ -237,7 +237,7 @@ jobs:
           CORAL_TEST_POSTGRES_URL: postgres://postgres:postgres@localhost:5432/postgres
         run: |
           cargo test --locked -p coral-app --lib \
-            state::db::repositories::workspaces::tests::workspace_repository_round_trips_against_postgres \
+            repository_round_trips_against_postgres \
```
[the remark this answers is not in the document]
AlbertQM: This is intentional; just setting a naming convention so that other tests following that will run automatically

crates/coral-app/migrations/0003_gui_onboarding_completions.sql
```diff
@@ -0,0 +1,3 @@
+CREATE TABLE IF NOT EXISTS gui_onboarding_completions (
+    principal_id TEXT NOT NULL PRIMARY KEY
```
sauldhernandez: Small schema-consistency note, while the table is still empty everywhere: both existing app-state tables record when their rows were written (`workspaces.created_at_unix_nanos` in `0001`, `app_state_migrations.completed_at_unix_nanos` in `0002`), and the plumbing is already in place — `now_unix_nanos_i64()` in `state/db/clock.rs`, with `WorkspaceManager::create` showing the pattern of the manager supplying the timestamp at write time.

As written, *when* a principal completed onboarding is unrecoverable, which is the question you'd want answered if someone reports the flow showing again or never showing, or if you ever want to re-show onboarding after a major UI revision. Adding `completed_at_unix_nanos BIGINT NOT NULL` now costs one extra value in `GuiOnboardingRepo::complete` (and since it inserts with `ON CONFLICT DO NOTHING`, it would naturally record first-completion time); adding it later means a second migration plus a nullable or sentinel-backfilled column for every pre-existing row.

Was leaving it out deliberate (pure set-membership semantics), or just not needed yet?

crates/coral-app/src/gui_onboarding/service.rs
```diff
@@ -0,0 +1,68 @@
+use coral_api::v1::gui_onboarding_service_server::GuiOnboardingService as GuiOnboardingServiceApi;
```
sauldhernandez, line 3: Very minor: the new `gui_onboarding` files (`mod.rs`, `manager.rs`, `service.rs`) start straight into `use`/`mod` statements. Nearly every sibling module in this crate opens with a one-line `//!` boundary statement — 8 of the 9 existing `*/service.rs` files start with ``//! Implements the gRPC `XService`...`` (only `feedback` doesn't), and most `mod.rs`/`manager.rs` files do the same. A couple of `//!` lines here (e.g. ``//! Implements the gRPC `GuiOnboardingService`.`` on this file) would match the convention and give a reader landing in this brand-new module the same orientation the rest of the crate provides.

crates/coral-app/src/gui_onboarding/service.rs
```diff
@@ -0,0 +1,68 @@
...
+}
+
+#[tonic::async_trait]
+impl GuiOnboardingServiceApi for GuiOnboardingService {
+    async fn get_gui_onboarding_state(
+        &self,
+        request: Request<GetGuiOnboardingStateRequest>,
+    ) -> Result<Response<GetGuiOnboardingStateResponse>, Status> {
+        let span = grpc_span(&request);
+        let principal = request_principal(&request)?;
```
sauldhernandez: Observability nit: `request_principal(&request)?` runs before `instrument_grpc` (here and in `complete_gui_onboarding` below), so if it ever returns `Status::internal("missing authenticated request context")` the handler returns before `instrument_grpc` is awaited — and `instrument_grpc` is the only place `record_grpc_status`/`record_failure` run. The span from `grpc_span` then closes with no `grpc.status_code`, `rpc.response.status_code`, or `error.type`, so the failed RPC shows up in traces and error metrics as neither ok nor error. That's unfortunate precisely because this error fires when the service is mounted without `GrpcRequestContextLayer` — the misconfiguration you'd most want visible in telemetry.

The other services in the crate keep only *infallible* extension reads outside the instrumented block (`QueryAttribution::from_extensions` in `search/service.rs`, `.get::<TaskId>().copied()`) and all `?`-returning work inside it. Would it be worth matching that — e.g. capturing `let context = request.extensions().get::<RequestContext>().cloned();` before the block and doing the `ok_or_else(...)?` inside `instrument_grpc`, in both handlers?

Not blocking: with the current wiring (`GrpcRequestContextLayer` wraps all of `application_routes`) the failure path looks unreachable today, so this is a latent gap rather than an active one.

crates/coral-app/src/state/db/repositories/gui_onboarding.rs
```diff
@@ -0,0 +1,162 @@
...
+                .await
+                .expect("get Alice completion")
+        );
+        assert!(
+            !tx.gui_onboarding()
+                .is_completed(BOB_PRINCIPAL_ID)
+                .await
+                .expect("get Bob completion")
+        );
+        tx.rollback().await.expect("roll back state transition tx");
```
sauldhernandez: Coverage note on this harness: every repo call runs on the `tx` opened at line 112 and the test ends with `tx.rollback()` here, so no completion is ever committed and `complete`/`is_completed` are never exercised through `&CoralDb` — the pooled autocommit session `GuiOnboardingManager` actually uses in production (`self.db.as_ref()`). Compare the workspaces harness next door, which commits its `ensure` transaction and then re-reads via `let mut session = db;` — that commit-then-reread is what gives it round-trip value.

As written, the Postgres job that the `validate.yml`/`Makefile` filter was broadened for only proves these statements execute inside an aborted transaction, and the sole committing test (`tests/grpc/gui_onboarding_tests.rs`) runs against SQLite. So nothing verifies that a completion durably persists through the pooled path on Postgres — a Postgres-only defect in the committed write path would ship green.

Would you consider committing one completion through the pooled session (using a unique principal id, workspaces-style, so the shared Postgres database stays isolation-safe), asserting `is_completed` from a fresh session, then cleaning it up? That would match the workspaces harness and make the new Postgres coverage load-bearing.

crates/coral-app/tests/grpc/gui_onboarding_tests.rs
```diff
@@ -0,0 +1,117 @@
...
+            .ok_or_else(|| PrincipalProviderError::unauthenticated("missing test principal"))?
+            .to_str()
+            .map_err(|error| PrincipalProviderError::unauthenticated(error.to_string()))?;
+        Principal::parse(principal_id, PrincipalKind::User)
+            .map_err(|error| PrincipalProviderError::unauthenticated(error.to_string()))
+    }
+}
+
+#[tokio::test]
+async fn gui_onboarding_is_per_principal_idempotent_concurrent_and_persistent_across_restart() {
```
sauldhernandez: Nice coverage — the concurrent-completion and restart-persistence cases are exactly the right properties to pin down. The one thing I'd reconsider is packing them all into a single test: when this fails, the panic line is the only clue to which of the four properties broke, and a failure in the concurrency block means the restart-persistence assertions never run at all. The neighbours split this way — `workspace_lifecycle_tests.rs` alone has 16 narrow, self-describing tests. Something like `..._is_per_principal`, `..._is_idempotent_under_concurrent_completion`, and `..._persists_across_restart` over a small shared server-setup helper would make failures self-locating and fit the directory convention, at the cost of a couple of extra server startups (which every other file here already pays per test).

crates/coral-app/src/gui_onboarding/manager.rs
```diff
@@ -0,0 +1,34 @@
...
+            .is_completed(principal.id().as_str())
+            .await
+            .map_err(Into::into)
+    }
+
+    pub(crate) async fn complete(&self, principal: &Principal) -> Result<(), AppError> {
+        let mut session = self.db.as_ref();
+        session
+            .gui_onboarding()
+            .complete(principal.id().as_str())
```
sauldhernandez: nit: This will cover both user and agent principals. Not sure if that's intended or not, but if not, you could safeguard here a check to ensure the principal is an user.

sauldhernandez (approved):

PR #1914, opened 2026-07-22
feat(reef): skip onboarding after completion

AlbertQM:
## Summary

Let's check if `onboarding` was completed (by reading the local DB), and NOT redirect to onboarding if that's the case.

## Test plan

Open app
See onboarding
Complete onboarding
Re-open app, I land in home rather than onboarding.

### DIFF
--- a/apps/coral-ui/app/lib/gui-onboarding.server.ts
+++ b/apps/coral-ui/app/lib/gui-onboarding.server.ts
@@ -1,8 +1,22 @@
 import { create } from '@bufbuild/protobuf'
 
-import { CompleteGuiOnboardingRequestSchema } from '@/generated/coral/v1/gui_onboarding_pb'
+import {
+  CompleteGuiOnboardingRequestSchema,
+  GetGuiOnboardingStateRequestSchema,
+} from '@/generated/coral/v1/gui_onboarding_pb'
 
 import { guiOnboardingClientForRequest } from './coral-request.server'
+
+export async function getGuiOnboardingCompleted(
+  request: Request,
+  accessToken: string | null,
+): Promise<boolean> {
+  const client = guiOnboardingClientForRequest(request, accessToken)
+  const response = await client.getGuiOnboardingState(create(GetGuiOnboardingStateRequestSchema), {
+    signal: request.signal,
+  })
+  return response.completed
+}
 
 export async function completeGuiOnboarding(
   request: Request,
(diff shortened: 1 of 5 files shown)


apps/reef/app/lib/gui-onboarding.server.ts
```diff
@@ -1,10 +1,36 @@
...
+    const response = await client.getGuiOnboardingState(
+      create(GetGuiOnboardingStateRequestSchema),
+      { signal: request.signal },
+    )
+    return response.completed
+  } catch (error) {
+    if (!request.signal.aborted) {
+      console.error('Failed to read GUI onboarding state:', error)
+    }
+    throw data(null, {
```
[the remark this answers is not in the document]
lazybean: yes

apps/reef/app/lib/gui-onboarding.server.ts
```diff
@@ -1,10 +1,36 @@
...
+      status: CORAL_UNAVAILABLE_STATUS,
+      statusText: CORAL_UNAVAILABLE_STATUS_TEXT,
```
lazybean: I think we should deal with coral unavailable earlier in the route .   
This is clearly not the only place  where we can have an issue.

lazybean: Could be a separate PR

apps/reef/app/root.tsx
```diff
@@ -135,3 +153,26 @@ export function ErrorBoundary({ error }: Route.ErrorBoundaryProps) {
...
+      <EmptyPage
+        action={
+          <Button.TextButton onClick={onRetry} variant="secondary">
+            Retry
+          </Button.TextButton>
+        }
+        description="Coral couldn't read your local onboarding state. Check that Coral is running and try again."
+        iconName="CircleAlert"
+        title="Coral is unavailable"
+      />
```
lazybean: Hmm. We may show this page in many other cases than onboarding

apps/reef/app/routes/onboarding.tsx
```diff
@@ -15,6 +15,13 @@ import { OnboardingView } from '@/views/onboarding/onboarding'
 import { runSourcesAction } from './sources-action'
 import { loadSourcesRouteData } from './sources-loader'
 
+const guiOnboardingMiddleware: MiddlewareFunction = async ({ request }, next) => {
+  if (await getGuiOnboardingCompleted(request)) throw redirect('/')
+  return next()
+}
+
+export const middleware: MiddlewareFunction[] = [guiOnboardingMiddleware]
+
```
lazybean: I see no reason to do that in a middleware instead of the loader. No?

lazybean (changes requested):
I think this PR has too much.  
  
Could you simplify?  
Only deal with the redirect in the routes/onboarding.tsx loader.  
  
The coral unavailable error (throw and handle) should be something separate and have more thought.

lazybean (approved):
thank you

PR #1928, opened 2026-07-23
feat(database): cache remote connection pools across queries

iaptsiauri:
## Summary

- reuse PostgreSQL and MySQL connection pools across query runtime builds
- key cached pools by provider and resolved connection identity without retaining plaintext credentials in cache keys
- bound each attempt, evict failed or stalled pools, and retry once against a fresh pool

## Stack

- Depends on: #1895
- Next: #1929

This is PR 5 of 7 in the relational database source MVP stack.


crates/coral-engine/src/backends/database/pool_cache.rs
```diff
@@ -0,0 +1,464 @@
+//! Cross-query connection pool caching for remote database sources.
+//!
+//! Query runtimes are rebuilt per query, so without caching every query pays
+//! a fresh connection handshake per database source. This module keeps pools
```
simonwhitaker: Unless I've misunderstood, it appears we're building a pool of connection pools here, one pool-of-pools per database source. That seems like an unnecessary nesting of pools. Like, this comment is exactly describing what I understand a connection pool to be.

Either the thing we're calling a "ConnectionPool" is actually just a connection, or we're adding unnecessary complexity, or I am completely misunderstanding. Any of these are possible. But if it turns out that we've just re-invented connection pooling then there might be an easier way to solve this.

crates/coral-engine/src/backends/database/mod.rs
```diff
@@ -1,10 +1,9 @@
 //! Relational database backend registration through `datafusion-table-providers`.
 
 mod catalog;
-
 use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::future::Future;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
```
simonwhitaker: We typically keep mod.rs files slender and put implementation in named sibling files. (We should put something in AGENTS.md about this.) Agents love shoving everything into mod.rs, so it's worth pushing back on that here.

crates/coral-app/src/workspaces/pool_registries.rs
```diff
@@ -0,0 +1,93 @@
...
+
+use crate::workspaces::WorkspaceName;
+
+/// Server-owned database pool registries keyed by workspace.
+#[derive(Default)]
+pub(crate) struct WorkspacePoolRegistries {
+    registries: Mutex<HashMap<WorkspaceName, Arc<DatabasePoolRegistry>>>,
+}
+
+impl WorkspacePoolRegistries {
```
simonwhitaker: nit: this type name being a plural feels weird and confusing. It could legitimately be called `WorkspacePoolRegistry`, I think. Alternatively, pretend it's Java and call it `WorkspacePoolRegistryManager`?

crates/coral-engine/src/backends/database/mod.rs
```diff
@@ -73,6 +174,37 @@ fn database_strategy(connection: &DatabaseConnectionSpec) -> &dyn DatabaseCatalo
...
+        return strategy
+            .build_catalog(catalog_name, context, pool_registry)
+            .await;
+    };
+    let registration = || strategy.build_catalog(catalog_name, context, pool_registry);
+    match tokio::time::timeout(timeout, registration()).await {
+        Ok(result) => result,
+        Err(_elapsed) => {
+            pool_registry.remove_catalog(catalog_name);
+            match tokio::time::timeout(timeout, registration()).await {
```
simonwhitaker: nit: this would be a bit cleaner if you structured it as a loop with a MAX_ATTEMPTS value. If we ever decide we want three tries before giving up, this code would make that an unnecessary palaver.

crates/coral-engine/src/backends/database/mod.rs
```diff
@@ -33,9 +33,104 @@ use crate::backends::{
...
+}
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+struct PoolId {
+    catalog_name: String,
+    connection_fingerprint: [u8; 32],
+}
+
+impl PoolId {
+    fn new(catalog_name: &str, provider: &str, params: &HashMap<String, String>) -> Self {
```
simonwhitaker: This feels a bit heavyweight, but I might be missing something. `catalog_name` will uniquely define a database source (we can't have two DB sources using the same catalog name). The provider and params can all be inferred from the same source spec. So I don't get why we need this complex `PoolId` type, and why `DatabasePoolRegistry.pools` can't just be indexed on catalog name. I *think* that achieves the exact same behaviour with significantly less complexity. Or is there something I'm missing here?

simonwhitaker (approved):

crates/coral-engine/src/backends/database/mod.rs
iaptsiauri: I did have this implementation and then Cubic flagged one issue. There is a chance when configuration changes that we cache an old catalog. 

```
  - Source originally connects to database A.
  - Someone updates the source to connect to database B.
  - A request using the old configuration is still creating its connection pool.
  - That old pool finishes after the update.
```

The old registry could overwrite or be reused by the updated source. Coral would then connect to database A using outdated credentials

iaptsiauri: actually you know what I think you are right this is very  obscure edge case I will revert the change to your suggestion

PR #1929, opened 2026-07-23
feat(engine): fetch database column metadata lazily

iaptsiauri:
## Summary

- fetch database column metadata with one provider-native query per relevant source when coral.columns is scanned
- prune safe catalog, schema, and table filters while degrading per-source failures and timeouts to warnings
- preserve strict catalog and schema identity across discovery and source-scoped validation

## Stack

- Depends on: #1928
- Next: #1896

This is PR 6 of 7 in the relational database source MVP stack.


crates/coral-engine/src/backends/database/catalog.rs
```diff
@@ -92,7 +121,18 @@ async fn load_database_inventory<T: 'static, P: 'static>(
...
         let table_names = inventory_column(&batch, "table_name")?;
         let relation_types = inventory_column(&batch, "relation_type")?;
+        let skip_reasons = inventory_column(&batch, "skip_reason")?;
         for row in 0..batch.num_rows() {
+            if !skip_reasons.is_null(row) {
+                tracing::warn!(
+                    catalog_name = ?catalog_name,
+                    schema_name = ?schema_names.value(row),
+                    table_name = ?table_names.value(row),
+                    unsupported_columns = ?skip_reasons.value(row),
```
simonwhitaker: We're conflating "skip reasons" with "unsupported columns" here. Are there other reasons a column might be skipped? If not, then `skip_reasons` feels mis-named. If so, then the `unsupported_columns` key here feels mis-named.

crates/coral-engine/src/backends/database/columns.rs
```diff
@@ -0,0 +1,390 @@
...
+            .finish_non_exhaustive()
+    }
+}
+
+#[async_trait]
+impl<T: 'static, P: 'static> DatabaseColumnFetcher for DatabaseColumnInventoryFetcher<T, P> {
+    async fn fetch_columns(
+        &self,
+        filter: &ColumnInventoryFilter,
+    ) -> DataFusionResult<Vec<DatabaseColumnRow>> {
```
simonwhitaker, line 95: This response could be arbitrarily large. Is that a problem? Could we return an iterator instead?

crates/coral-api/proto/coral/v1/catalog.proto
```diff
@@ -31,7 +31,8 @@ message PaginationResponse {
 message Column {
   // The column name as exposed by the query engine.
   string name = 1;
-  // The DataFusion/Arrow data type rendered as text.
+  // The column data type rendered as text: DataFusion/Arrow form for static
+  // sources, the provider-native type name for database catalogs.
```
simonwhitaker, line 35: This is a meaningful contract change and I want to make sure it's a deliberate one rather than a side effect of dropping the `information_schema` UNION.

Previously `coral.columns.data_type` reported DataFusion/Arrow type names for database sources; now it reports provider-native names (`character varying`, `TEXT`, `bigint`). But the engine still *executes* against Arrow types, so the two can disagree:

- Postgres registers with `UnsupportedTypeAction::String`, so an `inet` column now reports `inet` while behaving as `Utf8` at query time.
- SQLite `is_nullable` comes from `pragma_table_xinfo` and needn't match the Arrow schema's nullability.

The primary consumer of this field is an LLM writing DataFusion SQL, so a type name it can't cast to or compare against is a real failure mode. If this is intended, could the `coral.columns` guide text in `runtime/catalog.rs` say whose vocabulary the types are in? Right now the caveat only lives in the proto comment, which the agent never sees.

crates/coral-engine/src/backends/database/source.rs
```diff
@@ -292,22 +310,51 @@ impl DatabaseCatalogStrategy for SqliteConnectionSpec {
...
-       TABLE_NAME AS table_name,
-       TABLE_TYPE AS relation_type
-FROM INFORMATION_SCHEMA.TABLES
-WHERE TABLE_SCHEMA NOT IN ('information_schema', 'mysql', 'performance_schema', 'sys')";
+SELECT tables.TABLE_SCHEMA AS schema_name,
+       tables.TABLE_NAME AS table_name,
+       tables.TABLE_TYPE AS relation_type,
+       GROUP_CONCAT(
+           CASE
+               WHEN LOWER(columns.DATA_TYPE) NOT IN (
```
simonwhitaker: This allowlist is a hand-copy of `map_str_type_to_column_type` in `datafusion-table-providers-mysql`, and the consequence of a mismatch is that the **entire table** disappears from the catalog — not just the offending column. Three things compound here:

1. **Nothing ties the two lists together.** A provider upgrade that adds a type leaves tables silently invisible.
2. **It already disagrees with upstream:** `conn.rs:351` maps `geometry` to `MYSQL_TYPE_GEOMETRY`, but `geometry` isn't in this list, so MySQL tables with a `GEOMETRY` column get dropped. (Arguably right, since `arrow_sql_gen.rs` has no `MYSQL_TYPE_GEOMETRY` arm either — but that reasoning isn't written down anywhere.)
3. **This SQL is untested.** It's the most intricate query in the PR (correlated `GROUP_CONCAT` over a `LEFT JOIN`, `NULL`-means-keep semantics), and a typo in one type name silently hides tables. Even asserting the query text against a fixture would catch that.

The user-visible symptom is a table vanishing with only a `tracing::warn!` to explain it. Could this surface as a source-registration warning instead, so it reaches whoever configured the source? Same question for `MYSQL_INVENTORY_SESSION_SQL` just above — nothing covers the `as_sync`/`as_async`/neither branches in `load_database_inventory`.

crates/coral-engine/src/runtime/catalog.rs
```diff
@@ -1273,9 +1238,246 @@ fn build_columns_table(active_sources: &[RegisteredSource]) -> Result<MemTable>
...
+                        &fetcher.relation_names,
+                        rows,
+                    );
+                    batches.push(catalog_columns_batch(Arc::clone(&self.schema), &rows)?);
+                }
+                Err(error) => {
+                    tracing::warn!(
+                        catalog = fetcher.catalog_name.as_str(),
+                        detail = %error,
+                        "failed to fetch database column metadata; omitting catalog from coral.columns"
```
simonwhitaker, line 1423: Degrading to a warning keeps `coral.columns` queryable, which is the right call — but as written a failed fetch is indistinguishable from a table that genuinely has no columns. `describe_table` returns the table with an empty column list and no error, so an agent reads "this table has no columns" rather than "metadata is unavailable, retry or fall back." That's the more damaging of the two interpretations.

`CoralQuery` already carries a `failures` / `registration_failure` channel for exactly this kind of partial degradation. Could the per-catalog fetch error be plumbed through it so callers can tell the two cases apart?

crates/coral-engine/src/backends/database/columns.rs
```diff
@@ -0,0 +1,390 @@
...
+/// `column_name`, `data_type` (provider-native name), `is_nullable`
+/// ('true'/'false').
+pub(super) const POSTGRES_COLUMNS_SQL: &str = "
+SELECT table_schema AS schema_name,
+       table_name,
+       CAST(ordinal_position - 1 AS TEXT) AS ordinal_position,
+       column_name,
+       data_type,
+       CASE WHEN is_nullable = 'YES' THEN 'true' ELSE 'false' END AS is_nullable
+FROM information_schema.columns
```
simonwhitaker, line 32: The SQLite path has real integration tests against `pragma_table_xinfo` (nice — they caught the view and `NOT NULL` cases), but Postgres and MySQL have none, and Postgres is the one with a non-obvious failure mode.

`query_arrow` is handed a forced all-`Utf8` schema, while `information_schema.columns` returns domain types — `table_schema`/`column_name` are `sql_identifier`, `data_type` is `character_data`. If that coercion doesn't land the way it does for the pre-existing relations query, `inventory_column`'s `downcast_ref::<StringArray>()` fails at runtime, and it fails on the main `coral.columns` path rather than at registration. The explicit `CAST(... AS TEXT)` on `ordinal_position` suggests this was already a known concern for at least one column; `data_type` and `is_nullable` are new and uncast.

Worth a test against a real Postgres to confirm, rather than inferring it from the relations query working.

crates/coral-engine/src/runtime/catalog.rs
```diff
@@ -22,7 +31,11 @@ use crate::{
 
 /// Schema name for source metadata tables such as `coral.tables`.
 pub(crate) const SYSTEM_SCHEMA: &str = "coral";
-const STATIC_COLUMNS_TABLE: &str = "_columns_static";
+
+/// Per-source budget for query-time database column inventory. Fetches run
+/// concurrently, so total scan latency is bounded by this duration rather than
+/// multiplying by the number of database sources.
+const DATABASE_COLUMN_FETCH_TIMEOUT: Duration = Duration::from_secs(10);
```
simonwhitaker: Two questions about where the cost ceiling lands now that this is on the scan path:

**No memoization.** Every `coral.columns` scan hits every relevant database. `list_tables` is an MCP tool agents call repeatedly, and each call runs `collect_tables` → one `coral.columns` query → one `information_schema.columns` round trip per database. The PR's goal is trading O(tables) for O(sources), which it achieves per-scan, but the per-*call* cost is now unbounded in the number of calls. Is a short TTL cache keyed on the inventory filter intended follow-up?

**`scan()` runs during physical planning**, so `EXPLAIN SELECT * FROM coral.columns` issues the remote queries too — worth knowing if anything in the planner path speculatively plans.

Relatedly: 10s per source is a long stall on an interactive `list_tables` even though the fetches run concurrently. Is that inside whatever request deadline the MCP/gRPC layer enforces?

simonwhitaker (approved):

PR #1941, opened 2026-07-23
feat(mcp): require task lifecycle

jsummerfield:
## Summary

- enable MCP Task tools for every client without a feature flag or compatibility shim
- remove the obsolete `tasks` feature advertisement and conditional routing
- require data-tool calls to use a Task opened in the current MCP session and close it with a final status
- attribute Tasks to the authenticated principal while sharing them within the Workspace confidentiality boundary
- retain only `CONTEXT.md` as the concise domain model

## Breaking change

The `tasks` feature flag and its CLI/process overrides are removed. The MCP server always exposes `start_task` and `end_task`, and data tools require a `task_id` created in the current MCP session.

## Validation

- `make rust-checks` — 2,066 passed, 5 skipped
- `make postgres-tests` — four repository contracts and Postgres server startup passed
- GitHub performance regression check — passed
- independent Standards and Spec reviews — no unresolved findings

Stacked on #1940.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -1,0 +1,28 @@
+# Coral
+
+Coral exposes connected data to agents and attributes their work without
+conflating current product concepts with retired experimental terminology.
+
+## Language
+
+**Workspace**:
+The confidentiality boundary for Coral data and agent work. Tasks are shared by
+Principals within a Workspace and are never shared across Workspaces.
+
+**Principal**:
+An authenticated actor, stably classified as a User or Agent, whose participation
+in a Task is attributed. Kind informs authorization but does not itself grant
+permission; a Principal does not privately own a Task.
+
+**Task**:
+A Workspace-shared unit of agent work with a declared intent and a terminal
+success or failure outcome. It records the Principal that started it, and Coral
+interactions within it are attributed to its task identifier.
+
+_Avoid_: Using Episode or session as aliases for Task
+
+**Episode**:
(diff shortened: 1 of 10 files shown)


Bradley-Butcher (approved):
Approved.

One follow-up worth considering as Tasks become mandatory at the MCP seam: the model must retain and repeat the task ID on every data call and again at end_task. A UUID is a sound durable database identity, but it is token-heavy and easy for an LLM to copy incorrectly. If low effort, could the MCP surface expose a short, human/LLM-friendly session alias or call-name-style handle while retaining the UUID as the canonical stored identity? If that is not a small change, I would capture it as follow-up work rather than block this PR.

jsummerfield:
> One follow-up worth considering...

Yes - thought about sneaking that one in, but a bigger change than I'd hoped, so will FU

jsummerfield:
* **#1941** <a href="https://app.graphite.com/github/pr/withcoral/coral/1941?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/1941?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1940** <a href="https://app.graphite.com/github/pr/withcoral/coral/1940?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #1947, opened 2026-07-23
ci: centralize postgres test entry point

sauldhernandez:
## Intent

Make `make postgres-tests` the single entry point for Postgres-backed test
coverage in local development and the Validate workflow.

The workflow previously duplicated the target's underlying Cargo invocations.
That created two independently maintained test lists, while the Make target
could only provision its own Docker database and could not reuse CI's Postgres
service.

## What it enables

- CI and local development now execute the same maintained Postgres test suite.
- CI can supply `CORAL_TEST_POSTGRES_URL` and reuse its existing Postgres
  service without starting another container.
- Local runs still provision a fresh database in the reusable Docker container
  and remove that database afterward.
- Changes to the `Makefile` trigger the Postgres validation job.
- Contributor and agent guidance now requires `make postgres-tests` and
  prohibits duplicating its Cargo test invocations in workflows or
  documentation.

## Usage examples

Run the complete suite with a locally managed Postgres:

```bash
make postgres-tests
```

Run the same suite against an existing dedicated database:

```bash
CORAL_TEST_POSTGRES_URL=postgres://user:password@localhost:5432/coral_test \
  make postgres-tests
```

## Validation

- `make postgres-tests` — both the shared repository harness and server
  lifecycle test passed.
- Verified the supplied-URL path skips Docker provisioning.
- Verified the local path provisions and removes a fresh database.
- Parsed `.github/workflows/validate.yml` as YAML.
- `git diff --check`


.github/workflows/validate.yml
```diff
@@ -235,13 +236,7 @@ jobs:
...
         env:
           CORAL_TEST_POSTGRES_URL: postgres://postgres:postgres@localhost:5432/postgres
-        run: |
-          cargo test --locked -p coral-app --lib \
-            state::db::repositories::workspaces::tests::workspace_repository_round_trips_against_postgres \
-            -- --ignored
-          cargo test --locked -p coral-app \
-            --test postgres_database_tests \
-            -- --ignored
+        run: make postgres-tests
```
simonwhitaker, line 239: ❤️

simonwhitaker (approved):

PR #1983, opened 2026-07-28
fix(reef): fix DS button hover background

AlbertQM:
## Summary

Before this PR, hovering on an active element would "downgrade" it visually to hover. We shouldn't.
We also had a bit of unnecessary code handling the hover, which isn't needed because our beautiful DS button already handles that.

## Test plan

Go to schema, hover on the active table, it says `active`


AlbertQM:
@cubic-dev-ai this fix is so simple, you should be able to review and accept, right?

AlbertQM:
@cubic-dev-ai re-review pls

PR #1984, opened 2026-07-28
feat(reef): show saved function bodies

AlbertQM:
## Summary

Initially lifterd the implementation from #19191, then opted for a simpler one.
The goal is retrieving saved function bodies to be shown in the UI

## Test plan

image


AlbertQM:
@cubic-dev-ai if there are no issues, can you accept this?

PR #1997, opened 2026-07-28
feat(spec)!: follow next-page URLs embedded in response bodies

### LINKED ISSUE #1988: microsoft_graph: collection endpoints silently return only the first page
Reported by an external tester evaluating `sources/v4/microsoft_graph` against a production Microsoft 365 tenant with Teams and SharePoint data.

## What happens

Collection endpoints return only the first page, and nothing in the result signals that more data exists. Observed across chat messages, mail, calendar, contacts and files — you get 10–50 items and the query looks complete.

The practical consequence is that an agent answers confidently from a sliver of the data. In testing, no agent pointed at the connector could reliably answer "how many messages are in this chat".

## Why

Microsoft Graph pages collections via an `@odata.nextLink` URL embedded in the JSON response body — not a header, and not an opaque cursor parameter. A client is expected to follow that URL until it stops appearing. Nothing follows it today.

## Constraint worth designing around

This is not simply "loop until done". Some Graph collections are enormous — a busy Teams chat can hold hundreds of thousands of messages — so unbounded following is its own failure mode.

The tester's framing, which seems right: silently stopping at page one is the worst of the available options, because incomplete data is indistinguishable from complete data. An explicit error, or a truncation signal the caller can see, would both be improvements over the current behaviour even before full pagination lands.

## Expected

Either follow `@odata.nextLink` to return complete results, or surface truncation explicitly (error, warning, or a visible marker) rather than returning a partial result that reads as whole.

## Related

Body-embedded next-link pagination overlaps with existing inference work:

- #1755 — response-driven pagination termination
- #1581 — improve OpenAPI next_page response cursor inference
- #1578 — resolve nested OpenAPI response refs for cursor pagination inference
- #576 — DSL: add `response_has_more_path` stop signal for cursor pagination

simonwhitaker:
OData APIs paginate by returning the complete URL of the next page in the response body -- Microsoft Graph's `@odata.nextLink` -- and no pagination mode could express that. `next_url_header` reads response headers only, and `cursor_query` needs a request parameter to place a token into, but there is no token here: there is a whole URL, and the server expects it requested as it stands.

## The new mode

`next_url_body` reads that URL from a declared response path and requests it verbatim, reusing the same-origin check the Link-header path already applies. That check matters more here than it does for a header, because the URL now comes out of the response body, so a compromised response must not be able to redirect a credentialed request off-origin.

Both absolute URLs and relative links are accepted. A relative link is resolved against the request URL that returned it -- `?page=2` resolves against the current path -- and only then checked for origin.

## How it ranks against the other detectors

Inference ranks it above cursor-query and offset detection. A whole URL is unambiguous -- there is no guessing which parameter a token belongs in, and no risk of driving an endpoint with a parameter it rejects. Graph declares `$skip` on collections that reject it at runtime, so that ordering is load-bearing rather than a preference. It stays below `Link` header detection, which is declared and cheaper.

Outranking the input-corroborated modes is also why this detector alone requires the descriptor to declare the property a `string`. Every other detector has a second signal -- a matching cursor, offset or page parameter -- so a name-only false positive costs them nothing. This one commits on the name alone, and the shared schema walk otherwise accepts a property that declares no type at all. If the property turns out not to hold a string, `Value::as_str` reads `None` at runtime, pagination stops, and the query returns page one with no error and no diagnostic, where offset pagination would have fetched everything.

The body name lexicon is deliberately narrower than the header one: a header called `Next` is nearly always a URL, while a body field called `next` is very often a token, and tokens must keep falling through to cursor-query.

## What counts as a page envelope

A next-page URL at the response root is now evidence that the response is an envelope, independently of whether the operation binds a page input. Graph needs that: `{"@odata.nextLink": ..., "value": [...]}` on an operation with no `$top` binds nothing at all, so without it there is nothing to unwrap `value` on -- and a contract only survives once the response reads as a list, so the pagination would have been discarded along with the row path.

Root-only, deliberately. The walk descends into nested objects up to depth 8, but what this unlocks applies at the response root, so deep evidence would answer a question it was not asked. A singleton `GET /tracks/{id}` returning `{id, tags: [...], links: {next_href}}` would otherwise have `tags` promoted to the whole relation and the declared resource discarded.

## Making the fetch loop exhaustive

The fetch loop resolves the next URL once and derives whether it is following one from that single binding, rather than testing the mode at two call sites. Neither site was an exhaustive match, so a mode added later would have built cleanly, reported `next_url_body` in its artifact, and silently stopped after page one. `pagination_next_url_body` is the test that catches exactly that, and it fails without the change.

## Bounding the walk

Unlike `link_header`, where the server must explicitly emit `rel="next"`, this mode is inferred from a property *name*, so the field may not be a next-page link at all -- a related-resource URL the heuristic picked up, or an API that echoes the collection URL instead of omitting the link on its last page. Two stops bound it: a link resolving to the request URL that carried it is not progress, and an empty page ends the walk whatever the body claims. Without them either shape meant `max_pages` credentialed requests and `max_pages` pages of duplicate rows before the query finally errored.

## Breaking change

`OPERATION_METADATA_GENERATOR_VERSION` moves `operation-metadata-v2` -> `operation-metadata-v3` and `PROJECTION_GENERATOR_VERSION` moves `derive-read-v11` -> `derive-read-v12`.

Both bumps invalidate existing materializations, which are regenerated on next import. Anyone holding an `operation-metadata.yaml` or a projection override copied from the previous generator versions will get the "override was copied from generator version ... but this Coral build expects ..." warning until they re-copy it.

## Impact

Measured with `cargo run -p xtask -- v4-metadata-report` (#1996) across all 20,673 operations in the six importable v4 sources, diffed against this stack's base.

**5 Stripe v2 operations** change, all in the same direction -- they return `next_page_url` and were previously reporting `mode: none` with no row path, so each returned a single envelope row:

| Operation | Row path | Mode | `limit` |
| --- | --- | --- | --- |
| `getv2commerceproductcatalogimports` | `""` -> `data` | `none` -> `next_url_body` | lookup key -> `page_size_param` |
| `getv2coreaccounts` | `""` -> `data` | `none` -> `next_url_body` | lookup key -> `page_size_param` |
| `getv2coreaccountsaccountidpersons` | `""` -> `data` | `none` -> `next_url_body` | lookup key -> `page_size_param` |
| `getv2coreeventdestinations` | `""` -> `data` | `none` -> `next_url_body` | lookup key -> `page_size_param` |
| `getv2coreevents` | `""` -> `data` | `none` -> `next_url_body` | lookup key -> `page_size_param` |

Nothing else in the catalog moves. The envelope-evidence widening does not reach Microsoft Graph here: its collection responses are `allOf`-composed, and nothing folds that until #1998, so all 16,702 Graph operations still report `mode: none` at this head.

The `limit` column above is the reason #1996 grew a `lookup_keys` column in this stack. The allowlist is whatever query parameters pagination did not claim, so a pagination edit that changes which detector wins hands its displaced parameters to the dependent-join planner -- and that half of the diff was previously invisible.

Claude-Session: https://claude.ai/code/session_01Rr2wBfz6dZUWkQF58nVBrR

---

### DIFF
--- a/crates/coral-engine/src/backends/http/fetch.rs
+++ b/crates/coral-engine/src/backends/http/fetch.rs
@@ -19,7 +19,7 @@ use crate::backends::http::url::{join_url, normalize_base_url};
 use crate::backends::shared::json_path::get_path_value;
 use crate::backends::shared::response_rows::extract_rows;
 use crate::backends::shared::template::{RenderContext, render_template};
-use coral_spec::{HttpMethod, ValidatedPaginationMode};
+use coral_spec::HttpMethod;
 
 const DEFAULT_MAX_PAGES: usize = 10_000;
 
@@ -94,23 +94,21 @@ pub(super) async fn fetch_rows(
         );
         let base_url = render_template(&client.base_url, &render_context)?;
         let base_url = normalize_base_url(&base_url);
-        let following_link_header = matches!(
-            pagination.mode,
-            ValidatedPaginationMode::LinkHeader | ValidatedPaginationMode::Auto
-        ) && state.next_url.is_some();
+        // Asked once, of the mode itself: a mode that advances by URL must
+        // both request that URL and skip rebuilding the request Coral would
+        // otherwise have made. Two independent `matches!` tests here is how a
+        // new mode silently stops after page one.
+        let follows_next_url = pagination.mode.follows_response_next_url();
+        let following_next_url = follows_next_url && state.next_url.is_some();
 
-        let url = if matches!(
-            pagination.mode,
-            ValidatedPaginationMode::LinkHeader | ValidatedPaginationMode::Auto
-        ) && let Some(next) = state.next_url.clone()
-        {
+        let url = if let Some(next) = state.next_url.clone().filter(|_| follows_next_url) {
             next
         } else {
             let rendered_path = render_template(&active_request.path, &render_context)?;
             join_url(&base_url, &rendered_path)?
         };
 
-        let (query_pairs, body) = if following_link_header {
+        let (query_pairs, body) = if following_next_url {
             (Vec::new(), None)
         } else {
             let mut query_pairs = build_query_pairs(active_request, &render_context)?;
--- a/crates/coral-engine/src/backends/http/fetch.rs
+++ b/crates/coral-engine/src/backends/http/fetch.rs
@@ -94,14 +94,20 @@ pub(super) async fn fetch_rows(
         );
         let base_url = render_template(&client.base_url, &render_context)?;
         let base_url = normalize_base_url(&base_url);
-        // Asked once, of the mode itself: a mode that advances by URL must
-        // both request that URL and skip rebuilding the request Coral would
-        // otherwise have made. Two independent `matches!` tests here is how a
-        // new mode silently stops after page one.
-        let follows_next_url = pagination.mode.follows_response_next_url();
-        let following_next_url = follows_next_url && state.next_url.is_some();
+        // Resolved once, then both decisions read that one binding: a mode
+        // that advances by URL must both request that URL and skip rebuilding
+        // the request Coral would otherwise have made. Deriving "are we
+        // following?" from the URL itself rather than re-testing the mode
+        // means the two can never disagree — and a new mode that forgets to
+        // answer `follows_response_next_url` stops after page one at both
+        // sites together, not at one of them.
+        let next_url = state
+            .next_url
+            .clone()
+            .filter(|_| pagination.mode.follows_response_next_url());
+        let following_next_url = next_url.is_some();
 
-        let url = if let Some(next) = state.next_url.clone().filter(|_| follows_next_url) {
+        let url = if let Some(next) = next_url {
             next
         } else {
             let rendered_path = render_template(&active_request.path, &render_context)?;
--- a/crates/coral-engine/src/backends/http/pagination.rs
+++ b/crates/coral-engine/src/backends/http/pagination.rs
@@ -242,18 +242,35 @@ pub(super) fn advance_pagination_state(
                 .and_then(Value::as_str)
                 .map(str::trim)
                 .filter(|next| !next.is_empty());
-            match next_raw {
-                Some(next_raw) => {
-                    let base = pagination_request_url(context.request_url)?;
-                    state.next_url = Some(resolve_pagination_next_url(
-                        &base,
-                        next_raw,
-                        "next URL body value",
-                    )?);
-                    Ok(PageAdvance::Continue)
-                }
-                None => Ok(PageAdvance::Stop),
+            let Some(next_raw) = next_raw else {
+                return Ok(PageAdvance::Stop);
+            };
+            // A page that carried no rows ends the walk whatever the body
+            // claims. Unlike `page`/`offset` this cannot use
+            // `page_is_exhausted`: the server chooses the page size for a
+            // URL-following mode, so a short page is normal and only an empty
+            // one is evidence.
+            if context.rows_on_page == 0 {
+                return Ok(PageAdvance::Stop);
             }
+            let base = pagination_request_url(context.request_url)?;
+            let next = resolve_pagination_next_url(&base, next_raw, "next URL body value")?;
+            // So does a link back to the page that carried it. This mode is
+            // inferred from a property *name*, so unlike `link_header` — where
+            // the server must explicitly say `rel="next"` — the field may not
+            // be a next-page link at all: a related-resource URL the heuristic
+            // picked up, or an API that echoes the collection URL instead of
+            // omitting the link on its last page. Nothing else bounds the
+            // loop, so that would be `max_pages` credentialed requests and
+            // `max_pages` pages of duplicate rows before the query errors.
+            //
+            // From page two on `request_url` is the previous next link, so a
+            // constant link stops on its second appearance.
+            if next == context.request_url {
+                return Ok(PageAdvance::Stop);
+            }
+            state.next_url = Some(next);
+            Ok(PageAdvance::Continue)
         }
     }
 }
@@ -791,6 +808,14 @@ mod tests {
         path: &[&str],
         payload: &Value,
     ) -> datafusion::error::Result<(PageAdvance, PageState)> {
+        advance_next_url_body_with_rows(path, payload, 10)
+    }
+
+    fn advance_next_url_body_with_rows(
+        path: &[&str],
+        payload: &Value,
+        rows_on_page: usize,
+    ) -> datafusion::error::Result<(PageAdvance, PageState)> {
         let pagination = next_url_body_pagination(path);
         let mut state = PageState::default();
         let advance = advance_pagination_state(
@@ -800,7 +825,7 @@ mod tests {
                 payload,
                 response_headers: &HeaderMap::new(),
                 request_url: "https://api.example.com/v1.0/me/chats",
-                rows_on_page: 10,
+                rows_on_page,
                 page_size: Some(10),
                 source_schema: "demo",
                 table_name: "items",
@@ -851,6 +876,68 @@ mod tests {
             );
             assert!(state.next_url.is_none());
         }
+    }
+
+    #[test]
+    fn advance_next_url_body_stops_on_a_link_back_to_the_page_that_carried_it() {
+        // This mode is inferred from a property name, so the field may be a
+        // related-resource URL or a collection link an API echoes on its last
+        // page rather than omitting. Without this stop nothing bounds the
+        // loop: `max_pages` requests, all returning the same rows.
+        for next in [
+            // Absolute, exactly the request URL.
+            "https://api.example.com/v1.0/me/chats",
+            // Relative, resolving to it.
+            "chats",
+        ] {
+            let (advance, state) = advance_next_url_body_with_rows(
+                &["@odata.nextLink"],
+                &json!({ "@odata.nextLink": next }),
+                10,
+            )
+            .unwrap();
+
+            assert_eq!(
+                advance,
+                PageAdvance::Stop,
+                "a next link pointing at the current page is not progress: {next}"
+            );
+            assert!(state.next_url.is_none());
+        }
+    }
+
+    #[test]
+    fn advance_next_url_body_stops_on_an_empty_page() {
+        // A short page is normal here — the server picks the page size — but
+        // an empty one is the end of the collection whatever the body claims.
+        let (advance, state) = advance_next_url_body_with_rows(
+            &["@odata.nextLink"],
+            &json!({"@odata.nextLink": "https://api.example.com/v1.0/me/chats?$skiptoken=abc"}),
+            0,
+        )
+        .unwrap();
+
+        assert_eq!(advance, PageAdvance::Stop);
+        assert!(state.next_url.is_none());
+    }
+
+    #[test]
+    fn advance_next_url_body_follows_a_short_page_that_still_offers_a_link() {
+        // The mirror of the empty-page stop: `page_is_exhausted` would end the
+        // walk here because 3 < 10, but the server chose that page size and
+        // has said there is more.
+        let (advance, state) = advance_next_url_body_with_rows(
+            &["@odata.nextLink"],
+            &json!({"@odata.nextLink": "https://api.example.com/v1.0/me/chats?$skiptoken=abc"}),
+            3,
+        )
+        .unwrap();
+
+        assert_eq!(advance, PageAdvance::Continue);
+        assert_eq!(
+            state.next_url.as_deref(),
+            Some("https://api.example.com/v1.0/me/chats?$skiptoken=abc")
+        );
     }
 
     #[test]
--- a/crates/coral-engine/src/backends/http/pagination.rs
+++ b/crates/coral-engine/src/backends/http/pagination.rs
@@ -67,7 +67,11 @@ pub(super) fn apply_pagination_query_pairs(
     match &pagination.mode {
         ValidatedPaginationMode::None
         | ValidatedPaginationMode::Auto
-        | ValidatedPaginationMode::CursorBody => {}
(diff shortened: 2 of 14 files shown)


crates/coral-engine/src/backends/http/pagination.rs
```diff
@@ -231,6 +235,26 @@ pub(super) fn advance_pagination_state(
...
+            // Read here rather than in `extract_response_pagination_hints`,
+            // which is a header-only helper and is not given the body.
+            let next_raw = get_path_value(context.payload, next_url_body.path())
+                .and_then(Value::as_str)
+                .map(str::trim)
+                .filter(|next| !next.is_empty());
+            match next_raw {
+                Some(next_raw) => {
+                    let base = pagination_request_url(context.request_url)?;
+                    state.next_url = Some(resolve_pagination_next_url(
```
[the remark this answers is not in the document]
simonwhitaker: `resolve_pagination_next_url` is unchanged from `main` and is shared with `link_header` and `next_url_header`, so this gap isn't introduced here — and the same applies to the initial request, before any pagination.

Sources with a request-identity authenticator already get `credential_safe_redirect_policy` (`crates/coral-engine/src/backends/http/client.rs:122`), which pins redirects to the same origin *and* requires a credential-safe transport on each hop. Extending that to every HTTP source is the real fix, but reqwest's redirect policy is per-`Client`, so it would change behaviour for sources that legitimately redirect cross-origin (CDNs, pre-signed URLs).

Tracking separately rather than widening this PR.

crates/coral-engine/src/backends/http/fetch.rs
```diff
@@ -94,23 +94,21 @@ pub(super) async fn fetch_rows(
...
-        {
+        let url = if let Some(next) = state.next_url.clone().filter(|_| follows_next_url) {
             next
         } else {
             let rendered_path = render_template(&active_request.path, &render_context)?;
             join_url(&base_url, &rendered_path)?
         };
 
-        let (query_pairs, body) = if following_link_header {
+        let (query_pairs, body) = if following_next_url {
```
sauldhernandez: `follows_next_url` and `following_next_url` are one word apart and either compiles at the other's site. Swapped at line 111, page one of a URL-following mode would go out with an empty query and no body; swapped at line 104 it's a silent no-op that reads as meaningful. Binding the next URL once ties both sites to the same value and drops the second name:

```rust
let next_url = state
    .next_url
    .clone()
    .filter(|_| pagination.mode.follows_response_next_url());
let following_next_url = next_url.is_some();
let url = match next_url {
    Some(next) => next,
    None => {
        let rendered_path = render_template(&active_request.path, &render_context)?;
        join_url(&base_url, &rendered_path)?
    }
};
```

crates/coral-engine/src/backends/http/pagination.rs
```diff
@@ -231,6 +235,26 @@ pub(super) fn advance_pagination_state(
...
+                        &base,
+                        next_raw,
+                        "next URL body value",
+                    )?);
+                    Ok(PageAdvance::Continue)
```
sauldhernandez: This `Continue` is unconditional on any non-empty string: nothing compares the resolved URL to `context.request_url`, and `rows_on_page` is ignored. If the detected field holds a constant same-origin link — a related-resource URL the name heuristic picked up, or an API that echoes the collection URL instead of omitting the link on the last page — the fetch loop has no other exit: 10 000 credentialed requests and 10 000 pages of duplicate rows before `max_pages` errors out.

A one-line stop when the resolved next URL equals `context.request_url` bounds it: from page 2 on, `request_url` is the previous next link, so a constant link stops on its second appearance. Stopping on `rows_on_page == 0` would also help.

`link_header` has the same exposure, but there the server must explicitly emit `rel="next"`; this mode is inferred from a property name, so a field that is not actually a next-page link is materially more likely.

crates/coral-spec/src/v4/mod.rs
```diff
@@ -7,8 +7,8 @@ pub const V4_ARTIFACT_SCHEMA_VERSION: u32 = 5;
 pub const SURFACE_IMPORTER_VERSION: &str = "surface-import-v3";
 pub const OPENAPI_IMPORTER_VERSION: &str = "openapi-v7";
 pub const MCP_IMPORTER_VERSION: &str = "mcp-tools-v3";
-pub const OPERATION_METADATA_GENERATOR_VERSION: &str = "operation-metadata-v2";
-pub const PROJECTION_GENERATOR_VERSION: &str = "derive-read-v11";
+pub const OPERATION_METADATA_GENERATOR_VERSION: &str = "operation-metadata-v3";
+pub const PROJECTION_GENERATOR_VERSION: &str = "derive-read-v12";
```
sauldhernandez, line 11: The title carries `!` but the body never says what breaks. These two bumps invalidate existing materializations, and anyone with an `operation-metadata.yaml` or projection override copied from `operation-metadata-v2`/`derive-read-v11` starts getting the "override was copied from generator version ... but this Coral build expects ..." warning until they re-copy. Worth a line in the body saying so — #1791 and #1884 both spelled their breaks out there, and that's what someone hitting the warning after upgrading will search for.

crates/coral-spec/src/v4/surfaces/openapi/operations.rs
```diff
@@ -361,23 +361,34 @@ fn detect_pagination_contract(
     context: &OpenApiResponsePaginationContext,
 ) -> Option<PaginationSpec> {
     detect_link_header_pagination(inputs, context)
+        .or_else(|| detect_next_url_body_pagination(document, inputs, context))
```
sauldhernandez, line 364: Ranking `next_url_body` above offset/cursor/page detection leaves the displaced paging parameter unowned, and unowned query params become lookup keys.

Your own `/next-url` fixture shows it. Before, `detect_offset_pagination` set `offset_param: skip`, so `pagination_query_param_names` covered `skip`: `infer_rest_lookup_keys` excluded it and its exposure was `Internal`. Now `offset_param` is None (the new test asserts that), and `skip` is in neither the presentation nor the search lexicon, so it lands in the lookup-key allowlist and `rest_filter_is_lookup_key` marks it joinable. A dependent join `ON i.skip = o.n` then issues `?skip=n`, receives every row from offset n onward, and attributes all of them to `skip = n` — the wrong-rows failure lookup_keys.rs's module doc warns about. Same for a `cursor`/`starting_after` param on an operation whose body also has a next-URL field.

The same ownership fact also drives relation shape: `has_required_public_rest_input` (projections/derive.rs:207) sees a *required* numeric offset/page param flip from `Internal` to a public `Filter`, so `projection_kind` returns `TableFunction` instead of `Table` and `FROM src.things` becomes `FROM src.things(...)`.

Two ways out: add the offset/page/cursor name lexicons to `joinable_param_name` in lookup_keys.rs so a displaced paging parameter can never become a lookup key regardless of which detector won, or have `detect_next_url_body_pagination` still record the paging inputs it displaced so `pagination_query_param_names` keeps covering them.

Note the #1996 report has no lookup-keys column, so none of this shows up in the before/after inference diff.

sauldhernandez, line 364: Envelope evidence is only read from the detector that wins the mode. An operation that declares both a `Link` response header and a body next-URL, with no page/per-page/cursor input, still imports as a singleton: link-header detection wins, its spec has no bound inputs and an empty `next_url_path`, so `signals_page_envelope` is false and the row path is never inferred — even though the body next-URL is right there. The new `importer_prefers_a_declared_link_header_over_a_body_next_url` test doesn't catch this because its `per_page` input supplies `page_size`.

Narrow, and it was already the behavior before this PR, so feel free to punt. If you want it closed: the next-URL path is a fact about this response schema regardless of which mode wins, so the envelope question could ask the body detector independently (or link-header detection could carry the body path when one exists).

crates/coral-spec/src/v4/surfaces/openapi/operations.rs
```diff
@@ -361,23 +361,34 @@ fn detect_pagination_contract(
...
+/// that header on singleton resources too — GitHub does, on
+/// `GET /orgs/{org}/actions/hosted-runners/{id}` among others. Treating it as
+/// evidence would promote a resource's incidental array, such as that runner's
+/// `public_ips`, to the whole relation.
+fn signals_page_envelope(contract: &PaginationSpec) -> bool {
     contract.cursor_param.is_some()
         || contract.offset_param.is_some()
         || contract.page_param.is_some()
         || contract.page_size.is_some()
+        || !contract.next_url_path.is_empty()
```
sauldhernandez: `signals_page_envelope` counts any non-empty `next_url_path` as envelope evidence, but `find_response_cursor_path` descends into nested `type: object` properties up to depth 8. So the evidence can come from well below the response root, while what it unlocks — `accept_fallback` in `infer_wrapped_list_row_path` — applies at the root.

Concrete misfire: a singleton `GET /tracks/{id}` returning `{id, tags: [...], links: {next_href: string}}`. `nexthref` is in the token list, so `next_url_path = ["links", "next_href"]` and `paginated_operation` flips true. Root evidence is still false (`links` is deliberately absent from `METADATA_OBJECT_NAMES`), but the sole-array fallback now fires anyway: `tags` becomes the whole relation, the declared resource is discarded, and the runtime follows `links.next_href` as page two. That's the promotion the old `binds_pagination_input` comment guarded against, reached by a new route — every pre-existing detector was immune because each required a bound request input.

Both motivating shapes (`@odata.nextLink`, Stripe's `next_page_url`) sit at the root as single segments, so gating this arm on `next_url_path.len() == 1` closes the hole without losing either; a legitimate nested `{data, meta: {next_url}}` still qualifies through `meta` in the metadata lexicon. Alternatively, answer the evidence question with a root-only lookup and keep the deep walk just for reading the URL at runtime. (#1998 folds allOf but doesn't narrow this walk.)

The catalog report only covers today's sources, so the 6-Stripe-operations delta doesn't rule this out for the next source someone imports.

sauldhernandez: The PR body is the first commit's message, so it describes only the new pagination mode. The second commit goes further: a body next-URL path now counts as envelope evidence by itself, so operations that bind no page input at all (the Graph no-`$top` shape) can gain a row path they didn't have before. The "6 Stripe v2 operations" Impact line was written before that commit — Stripe qualifies via `limit`/`page_size` either way, but any catalog spec with a nextlink-named body field and no page inputs may have moved too. Since #1996 exists to make exactly this diff visible, could you re-run the report at this head and fold the widening plus any new deltas into the description?

crates/coral-spec/src/v4/surfaces/openapi/operations.rs
```diff
@@ -400,6 +411,48 @@ fn detect_link_header_pagination(
...
+        "nexturi",
+        "nextpageurl",
+        "nextpageuri",
+        "nextpagelink",
+        "nextpagehref",
+        "nexthref",
+    ];
+
+    let next_url_path =
+        find_response_cursor_path(document, &context.schema, RESPONSE_NEXT_URL_BODY_TOKENS)?;
```
sauldhernandez: `find_response_cursor_path` accepts a name-matching property that declares no `type` (`property_allows_string`, `response_cursors.rs:100`), and the walk resolves `$ref` but not `allOf`. So an untyped or composed `nextLink`/`nextUrl` property matches here. That permissiveness was safe for cursor detection, which still needs a matching cursor query param before it returns a contract. This detector has no second signal, and it now outranks cursor-query and offset.

Concretely: an operation with `skip`+`limit` and a `nextLink` property that isn't actually a string gets `mode: next_url_body`. At runtime `Value::as_str` on a non-string value is `None`, `advance_pagination_state` returns `Stop`, and the query silently returns page one only — no error, no diagnostic. Offset pagination would have fetched everything.

Two options: require a declared `string` type in this detector (a stricter predicate than the cursor helper's), or rank body next-URL below the input-corroborated modes so a name-only guess never displaces one. The second keeps untyped-but-real `@odata.nextLink` detection working on paramless Graph-style operations.

sauldhernandez: This call makes `find_response_cursor_path` a next-page-URL finder, but `response_cursors.rs` still describes itself as cursor-only: the module doc and fn doc both say "continuation token", and the param is `cursor_tokens`. Someone later tightening the walk for cursors — say, making `property_allows_string` reject URL-shaped values — has no signal that `next_url_body` detection rides on it. Nothing in that file's tests would fail; the regression would only surface as a metadata-report diff. Worth widening the module and fn docs (and maybe renaming `cursor_tokens` to `name_tokens`). #1998 reworks the same walk and doesn't close this gap either.

crates/coral-spec/src/v4/surfaces/openapi/operations.rs
sauldhernandez: Human comment: guess this only really matters when you have one API that seems to support both ways, which I find unlikely. Up to you if this is relevant.

sauldhernandez: Human comment: same here, not confidence on relevancy of the corner case. Your call.

sauldhernandez (approved):

crates/coral-engine/src/backends/http/fetch.rs
simonwhitaker: Nice! Will fix

crates/coral-spec/src/v4/surfaces/openapi/operations.rs
simonwhitaker: Agreed. More generally we have to make a call on whether to treat pagination-looking params as always excluded from lookup keys, or only excluded if we infer them to be used for pagination. It's not straightforward. We might decide that a parameter called e.g. `next_page` is always going to be a pagination param, but e.g. `page` could either be a pagination control or an identifier (e.g. it might identify a specific page in a CMS API). I think we should leave this as-is for now, but we should consider adding lookup-keys to the report generated in #1996

PR #1998, opened 2026-07-28
feat(spec)!: read allOf-composed responses as page envelopes

### LINKED ISSUE #1988: microsoft_graph: collection endpoints silently return only the first page

simonwhitaker:
Providers routinely assemble a collection response from a shared pagination base and a branch declaring the rows, rather than declaring it whole. Every OData collection response is written that way, so all 16,702 Microsoft Graph operations were refused by envelope detection: the properties live on the allOf branches, and the walk only looked at the schema itself. Each collection surfaced as a single row holding `odata_count`, `odata_nextlink`, and a JSON `value` blob.

## Folding allOf for inference

`merged_all_of_object_view` folds allOf branches into one property map, shared by wrapped-list inference and cursor discovery so both halves of the decision see the same shape -- if only one could, an operation would be presented as a paginated table that stops after page one.

Only allOf is folded. anyOf, oneOf, and not describe a choice of shapes, not one shape, so there is no single property map to ask questions of and they are still refused. Duplicate names take the first branch and report no conflict: this answers a heuristic question, and an annotation-only disagreement between branches must not discard a whole envelope.

What still has to be declared is the object itself. Folding replaces the response walk's explicit `type: object` check with the merged view's own object test, and that test asks the same question -- some branch must say `type: object`. Declaring `properties` alone is not enough: in JSON Schema `properties` constrains an instance only if it happens to be an object and asserts nothing about whether it is one, so `{properties: {...}}` validates a string quite happily. Reading that as an assertion would let a singleton `{properties: {warnings: [...]}}` reach the sole-array fallback and have `warnings` promoted to the whole relation, which `resolve_output_row_type_ref` cannot catch because type import defaults typeless schemas to object as well.

So the widening here is exactly `allOf` visibility, not a looser notion of what an object is.

## The @odata annotation lexicon

Folding alone is not enough. `normalized_name` keeps the annotation prefix, so `@odata.count` and `@odata.nextLink` normalize to `odatacount` and `odatanextlink` and matched nothing in the metadata lexicon, leaving a three-property envelope with no evidence. They are added explicitly rather than by stripping `@odata.` in `normalized_name`, because that function also decides which arrays are *excluded* as metadata -- stripping there would silently reclassify every `@odata.*` property in every source.

## Folding allOf for type import

Type import had to follow, and did not: it merged only the immediate branches and ignored properties declared alongside allOf. The two halves then disagreed, and a path inference found in a nested branch looked absent from the imported type, so `infer_row_path` discarded it and the collection collapsed to a single JSON row again. Nested branches also lost their fields outright -- `microsoft.graph.drive` reaches `id`, `name`, and `webUrl` through `baseItem` and `entity`, and shipped without those columns.

Folding the full chain forces the property comparison to be annotation-insensitive. Every Graph subtype re-declares the `@odata.type` discriminator with its own `default`, which byte-for-byte equality reads as two branches disagreeing: 1,618 Graph operations lost their response type entirely before the comparison moved to validation semantics, which is a worse outcome than the missing columns it set out to fix. Branches that genuinely disagree about type or constraints are still conflicts.

## Impact

2,592 Graph operations become paginated row tables, and none lands on offset pagination -- which matters, because Graph declares `$skip` on collections that reject it at runtime. Verified against the live v1.0 description: the other six v4 sources are byte-identical, and `v4-metadata-report` does not move by a line across the type-import change either.

Keeping the declared-object requirement costs nothing measurable: every typeless success-response root across the six ingested specs is `allOf` (3,242), `anyOf` (705), or `oneOf` (8), and not one is a bare property bag, so refusing the typeless case moves 0 of 20,673 operations. Worth revisiting if a real spec turns up where accepting it would help.

## Follow-ups

Pre-existing and out of scope: `find_numeric_query_input` selects the first candidate-named query input and only then filters by type, so Graph's boolean `$count` masks its integer `$top` and page-size detection finds nothing. Pagination is unaffected -- Coral just accepts Graph's server-side default page size, and `top` stays a user-facing filter rather than becoming pagination-owned.

The `required` set merged across allOf branches is still dropped, exactly as it was before type import folded the whole tree. Honouring it would flip the nullability of every field of every composed type, which deserves its own commit.

Claude-Session: https://claude.ai/code/session_01Rr2wBfz6dZUWkQF58nVBrR

---

### DIFF
--- a/crates/coral-spec/src/v4/mod.rs
+++ b/crates/coral-spec/src/v4/mod.rs
@@ -5,7 +5,7 @@
 
 pub const V4_ARTIFACT_SCHEMA_VERSION: u32 = 5;
 pub const SURFACE_IMPORTER_VERSION: &str = "surface-import-v3";
-pub const OPENAPI_IMPORTER_VERSION: &str = "openapi-v7";
+pub const OPENAPI_IMPORTER_VERSION: &str = "openapi-v8";
 pub const MCP_IMPORTER_VERSION: &str = "mcp-tools-v3";
 pub const OPERATION_METADATA_GENERATOR_VERSION: &str = "operation-metadata-v4";
 pub const PROJECTION_GENERATOR_VERSION: &str = "derive-read-v13";
--- a/crates/coral-spec/src/v4/mod.rs
+++ b/crates/coral-spec/src/v4/mod.rs
@@ -7,8 +7,8 @@ pub const V4_ARTIFACT_SCHEMA_VERSION: u32 = 5;
 pub const SURFACE_IMPORTER_VERSION: &str = "surface-import-v3";
 pub const OPENAPI_IMPORTER_VERSION: &str = "openapi-v7";
 pub const MCP_IMPORTER_VERSION: &str = "mcp-tools-v3";
-pub const OPERATION_METADATA_GENERATOR_VERSION: &str = "operation-metadata-v3";
-pub const PROJECTION_GENERATOR_VERSION: &str = "derive-read-v12";
+pub const OPERATION_METADATA_GENERATOR_VERSION: &str = "operation-metadata-v4";
+pub const PROJECTION_GENERATOR_VERSION: &str = "derive-read-v13";
 
 mod artifacts;
 mod diagnostics;
--- a/crates/coral-spec/src/v4/openapi_tests.rs
+++ b/crates/coral-spec/src/v4/openapi_tests.rs
@@ -1,4 +1,5 @@
 use std::collections::BTreeMap;
+use std::fmt::Write as _;
 
 use super::*;
 use crate::{
@@ -1231,6 +1232,198 @@ components:
             .diagnostics
             .iter()
             .any(|diagnostic| diagnostic.message.contains("allOf property")),
+        "{:?}",
+        operation.diagnostics
+    );
+    assert_eq!(operation.output.type_ref, "json");
+}
+
+/// A branch may be an alias — `{$ref: Alias}` where `Alias` is itself
+/// `{$ref: Base}` — which one hop of resolution leaves still holding a `$ref`,
+/// with nothing to contribute.
+///
+/// Inference resolves chains, so a branch dropped here would make a row path it
+/// found look absent from the imported type, and `infer_row_path` would discard
+/// the path.
+#[test]
+fn importer_folds_all_of_branches_through_alias_refs() {
+    let manifest = parse_source_manifest_yaml(
+        r"
+name: aliased_all_of
+dsl_version: 4
+surface:
+    type: openapi
+    file: /tmp/openapi.yaml
+    base_url: https://api.example.com
+",
+    )
+    .expect("manifest");
+    let v4 = manifest.as_v4().expect("v4");
+    let ir = import_openapi_surface(
+        v4,
+        &v4.surface,
+        r"
+openapi: 3.0.3
+paths:
+  /items:
+    get:
+      operationId: items/list
+      responses:
+        '200':
+          content:
+            application/json:
+              schema: {$ref: '#/components/schemas/Page'}
+components:
+  schemas:
+    Base:
+      type: object
+      properties:
+        next_cursor: {type: string}
+    Alias: {$ref: '#/components/schemas/Base'}
+    Page:
+      type: object
+      allOf:
+        - {$ref: '#/components/schemas/Alias'}
+        - type: object
+          properties:
+            items:
+              type: array
+              items: {type: object, properties: {id: {type: string}}}
+"
+        .as_bytes(),
+    )
+    .expect("import");
+
+    let operation = ir.operations.first().expect("operation");
+    let row_type = ir
+        .types
+        .iter()
+        .find(|ty| ty.id == operation.output.type_ref)
+        .expect("row type");
+    let IrTypeShape::Object { fields } = &row_type.shape else {
+        panic!("the composed page should import as an object: {row_type:?}");
+    };
+    let names = fields
+        .iter()
+        .map(|field| field.name.as_str())
+        .collect::<Vec<_>>();
+
+    assert!(
+        names.contains(&"next_cursor"),
+        "the aliased base contributes its properties: {names:?}"
+    );
+    assert!(names.contains(&"items"), "{names:?}");
+}
+
+/// The fold tracks the refs it is resolving, so a self-referential branch is
+/// named as a cycle rather than recursing until the depth cap stops it.
+#[test]
+fn importer_reports_a_cyclic_all_of_branch() {
+    let manifest = parse_source_manifest_yaml(
+        r"
+name: cyclic_all_of
+dsl_version: 4
+surface:
+    type: openapi
+    file: /tmp/openapi.yaml
+    base_url: https://api.example.com
+",
+    )
+    .expect("manifest");
+    let v4 = manifest.as_v4().expect("v4");
+    let ir = import_openapi_surface(
+        v4,
+        &v4.surface,
+        r"
+openapi: 3.0.3
+paths:
+  /items:
+    get:
+      operationId: items/list
+      responses:
+        '200':
+          content:
+            application/json:
+              schema: {$ref: '#/components/schemas/Loop'}
+components:
+  schemas:
+    Loop:
+      type: object
+      allOf:
+        - {$ref: '#/components/schemas/Loop'}
+"
+        .as_bytes(),
+    )
+    .expect("a cyclic allOf imports with diagnostics rather than recursing");
+
+    let operation = ir.operations.first().expect("operation");
+    assert!(
+        operation
+            .diagnostics
+            .iter()
+            .any(|diagnostic| diagnostic.message.contains("cyclic")),
+        "the diagnostic must name the cycle, not a depth ceiling: {:?}",
+        operation.diagnostics
+    );
+    assert_eq!(operation.output.type_ref, "json");
+}
+
+/// Composition past the cap reports its own ceiling. Reusing the property
+/// comparison error sent whoever read it to a constant eight times larger.
+#[test]
+fn importer_reports_all_of_composition_past_the_depth_cap() {
+    let mut schemas = String::from(
+        r"
+components:
+  schemas:
+    Deep0:
+      type: object
+      properties:
+        id: {type: string}
+",
+    );
+    for level in 1..=12 {
+        writeln!(
+            schemas,
+            "    Deep{level}:\n      type: object\n      allOf:\n        - {{$ref: '#/components/schemas/Deep{}'}}",
+            level - 1
+        )
+        .expect("writing to a String cannot fail");
+    }
+    let manifest = parse_source_manifest_yaml(
+        r"
+name: deep_all_of
+dsl_version: 4
+surface:
+    type: openapi
+    file: /tmp/openapi.yaml
+    base_url: https://api.example.com
+",
+    )
+    .expect("manifest");
+    let v4 = manifest.as_v4().expect("v4");
+    let document = format!(
+        r"
+openapi: 3.0.3
+paths:
+  /items:
+    get:
+      operationId: items/list
+      responses:
+        '200':
+          content:
+            application/json:
+              schema: {{$ref: '#/components/schemas/Deep12'}}
+{schemas}"
+    );
+    let ir = import_openapi_surface(v4, &v4.surface, document.as_bytes()).expect("import");
+
+    let operation = ir.operations.first().expect("operation");
+    assert!(
+        operation
+            .diagnostics
+            .iter()
+            .any(|diagnostic| diagnostic.message.contains("nests past")),
         "{:?}",
         operation.diagnostics
     );
--- a/crates/coral-spec/src/v4/openapi_tests.rs
+++ b/crates/coral-spec/src/v4/openapi_tests.rs
@@ -2710,3 +2710,248 @@ components:
         "Graph rejects $skip on several collections; the next link is the only contract that works"
     );
 }
+
+/// Graph nests its envelope bases: a delta collection response composes
+/// `BaseDeltaFunctionResponse`, which is itself an `allOf` over
+/// `BaseCollectionPaginationCountResponse`. Row-path inference folds that whole
+/// tree, so type import has to fold it too — otherwise the imported type is
+/// missing the properties the inferred path names and the path is discarded.
+#[test]
+fn importer_folds_nested_all_of_envelope_bases() {
+    let manifest = parse_source_manifest_yaml(
+        r"
+name: graph
+dsl_version: 4
+surface:
+    type: openapi
+    file: /tmp/openapi.yaml
+    base_url: https://graph.microsoft.com/v1.0
+",
+    )
+    .expect("manifest");
+    let v4 = manifest.as_v4().expect("v4");
+    let ir = import_openapi_surface(
+        v4,
+        &v4.surface,
+        r"
+openapi: 3.0.3
+paths:
+  /me/chats/delta:
+    get:
+      operationId: me.chats.delta
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/microsoft.graph.chatDeltaCollectionResponse'
+components:
+  schemas:
+    BaseCollectionPaginationCountResponse:
+      type: object
+      properties:
+        '@odata.count': {type: integer, format: int64, nullable: true}
+        '@odata.nextLink': {type: string, nullable: true}
+    BaseDeltaFunctionResponse:
+      type: object
+      allOf:
+        - $ref: '#/components/schemas/BaseCollectionPaginationCountResponse'
+        - type: object
+          properties:
+            '@odata.deltaLink': {type: string, nullable: true}
+    microsoft.graph.chat:
+      type: object
+      properties:
+        id: {type: string}
+    microsoft.graph.chatDeltaCollectionResponse:
+      type: object
+      allOf:
+        - $ref: '#/components/schemas/BaseDeltaFunctionResponse'
+        - type: object
+          properties:
+            value:
+              type: array
+              items:
+                $ref: '#/components/schemas/microsoft.graph.chat'
+"
+        .as_bytes(),
+    )
+    .expect("import");
+
+    assert_eq!(imported_row_path(&ir, "me_chats_delta"), ["value"]);
+
+    let pagination = imported_rest_pagination(&ir, "me_chats_delta");
+    assert_eq!(pagination.mode, PaginationMode::NextUrlBody);
+    assert_eq!(pagination.next_url_path, ["@odata.nextLink"]);
+
+    // The nested base contributes `@odata.count`/`@odata.nextLink` and the
+    // intermediate one `@odata.deltaLink`. Folding only the immediate branches
+    // kept `value` alone.
+    let IrTypeShape::Object { fields } = &ir
+        .semantic_ir
+        .types
+        .iter()
+        .find(|ty| ty.id == "me_chats_delta_row")
+        .expect("response type")
+        .shape
+    else {
+        panic!("expected an object response type");
+    };
+    let names: Vec<&str> = fields.iter().map(|field| field.name.as_str()).collect();
+    assert_eq!(
+        names,
+        [
+            "@odata.count",
+            "@odata.deltaLink",
+            "@odata.nextLink",
+            "value"
+        ]
+    );
+}
+
+/// `properties` declared alongside `allOf` are as much part of the schema as
+/// the branches are.
+///
+/// Row-path inference reads both, so before type import did too, the inferred
+/// `value` path was rejected as absent from the imported type and the whole
+/// collection collapsed to a single JSON row.
+#[test]
+fn importer_keeps_properties_declared_beside_all_of() {
+    let manifest = parse_source_manifest_yaml(
+        r"
+name: graph
+dsl_version: 4
+surface:
+    type: openapi
+    file: /tmp/openapi.yaml
+    base_url: https://graph.microsoft.com/v1.0
+",
+    )
+    .expect("manifest");
+    let v4 = manifest.as_v4().expect("v4");
+    let ir = import_openapi_surface(
+        v4,
+        &v4.surface,
+        r"
+openapi: 3.0.3
+paths:
+  /me/chats:
+    get:
+      operationId: me.listChats
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/microsoft.graph.chatCollectionResponse'
+components:
+  schemas:
+    BaseCollectionPaginationCountResponse:
+      type: object
+      properties:
+        '@odata.count': {type: integer, format: int64, nullable: true}
+        '@odata.nextLink': {type: string, nullable: true}
+    microsoft.graph.chat:
+      type: object
+      properties:
+        id: {type: string}
+    microsoft.graph.chatCollectionResponse:
+      type: object
+      properties:
+        value:
+          type: array
+          items:
+            $ref: '#/components/schemas/microsoft.graph.chat'
+      allOf:
+        - $ref: '#/components/schemas/BaseCollectionPaginationCountResponse'
+"
+        .as_bytes(),
+    )
+    .expect("import");
+
+    assert_eq!(imported_row_path(&ir, "me_listchats"), ["value"]);
+}
+
+/// A subtype re-declaring an inherited property to pin an annotation is not a
+/// conflict.
+///
+/// Every Graph type re-declares the `@odata.type` discriminator with its own
+/// `default`. Comparing declarations byte-for-byte reads that as two branches
+/// disagreeing and discards the whole type, which costs every column rather
+/// than the one property — so the comparison is on validation semantics, and
+/// [`importer_warns_for_openapi_all_of_property_conflicts`] still holds for branches
+/// that genuinely disagree.
+#[test]
+fn importer_accepts_annotation_only_redeclaration_across_all_of_levels() {
+    let manifest = parse_source_manifest_yaml(
+        r"
+name: graph
+dsl_version: 4
+surface:
+    type: openapi
+    file: /tmp/openapi.yaml
+    base_url: https://graph.microsoft.com/v1.0
+",
+    )
+    .expect("manifest");
+    let v4 = manifest.as_v4().expect("v4");
+    let ir = import_openapi_surface(
+        v4,
+        &v4.surface,
+        r"
+openapi: 3.0.3
+paths:
+  /me/drive:
+    get:
+      operationId: me.getDrive
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/microsoft.graph.drive'
+components:
+  schemas:
+    microsoft.graph.entity:
+      type: object
+      properties:
+        id: {type: string}
+        '@odata.type': {type: string}
+    microsoft.graph.baseItem:
+      type: object
+      allOf:
+        - $ref: '#/components/schemas/microsoft.graph.entity'
+        - type: object
+          properties:
+            name: {type: string}
+            webUrl: {type: string}
+    microsoft.graph.drive:
+      type: object
+      allOf:
+        - $ref: '#/components/schemas/microsoft.graph.baseItem'
+        - type: object
+          properties:
+            driveType: {type: string}
+            '@odata.type': {type: string, default: '#microsoft.graph.drive'}
+"
+        .as_bytes(),
+    )
+    .expect("import");
+
+    let IrTypeShape::Object { fields } = &ir
+        .semantic_ir
+        .types
+        .iter()
+        .find(|ty| ty.id == "me_getdrive_row")
+        .expect("drive response type")
+        .shape
+    else {
+        panic!("a conflict would have left the type as opaque JSON");
+    };
+    let names: Vec<&str> = fields.iter().map(|field| field.name.as_str()).collect();
+    assert_eq!(
+        names,
+        ["@odata.type", "driveType", "id", "name", "webUrl"],
+        "the inherited columns must survive two levels of composition"
+    );
+}
--- a/crates/coral-spec/src/v4/openapi_tests.rs
+++ b/crates/coral-spec/src/v4/openapi_tests.rs
@@ -2904,6 +2904,90 @@ components:
     );
 }
 
+/// The fixture above declares only `$top` and `$skip`, which is not what a real
+/// Graph collection looks like: they also declare a boolean `$count`.
+///
+/// `find_numeric_query_input` picks the first candidate-named input and only
+/// then filters by type, so `$count` is chosen and rejected and `$top` is never
+/// reached — page-size detection finds nothing. Pinning it here so the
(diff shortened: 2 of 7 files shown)


crates/coral-spec/src/v4/mod.rs
```diff
@@ -7,8 +7,8 @@ pub const V4_ARTIFACT_SCHEMA_VERSION: u32 = 5;
 pub const SURFACE_IMPORTER_VERSION: &str = "surface-import-v3";
 pub const OPENAPI_IMPORTER_VERSION: &str = "openapi-v7";
 pub const MCP_IMPORTER_VERSION: &str = "mcp-tools-v3";
-pub const OPERATION_METADATA_GENERATOR_VERSION: &str = "operation-metadata-v3";
-pub const PROJECTION_GENERATOR_VERSION: &str = "derive-read-v12";
+pub const OPERATION_METADATA_GENERATOR_VERSION: &str = "operation-metadata-v4";
+pub const PROJECTION_GENERATOR_VERSION: &str = "derive-read-v13";
```
sauldhernandez, line 11: The importer's output changes in this PR — composed types pick up fields from nested `allOf` bases and from sibling `properties`, and annotation-only redeclarations no longer discard a whole type — but `OPENAPI_IMPORTER_VERSION` stays at `openapi-v7` while both generator versions move.

Since #1772 these stamps are advisory provenance, so a semantic IR persisted by the previous build still passes its `importer_version` check in materialization.rs and reads as current, even though its type shapes predate the fold; the drift warnings only point at the generator versions. #1791 and #1884 both bumped the importer stamp when importer output changed. Should this be `openapi-v8`?

crates/coral-spec/src/v4/openapi_tests.rs
```diff
@@ -2501,3 +2501,334 @@ paths:
...
+
+    let pagination = imported_rest_pagination(&ir, "me_listchats");
+    assert_eq!(pagination.mode, PaginationMode::NextUrlBody);
+    assert_eq!(pagination.next_url_path, ["@odata.nextLink"]);
+    assert_eq!(
+        pagination
+            .page_size
+            .as_ref()
+            .and_then(|size| size.query_param.as_deref()),
+        Some("$top")
```
sauldhernandez, line 2899: The doc comment calls this the Microsoft Graph shape, but the fixture omits `$count`, so this asserts a page size real Graph collections never get: per your follow-ups note, boolean `$count` sorts before `$top` and masks it, so detection returns `None` on the real description.

Keeping this assertion for the wiring is fine — but a companion variant with `- {name: $count, in: query, schema: {type: boolean}}` asserting `page_size.is_none()` would pin the known gap and give the follow-up that fixes `find_numeric_query_input` an assertion to flip.

crates/coral-spec/src/v4/response_cursors.rs
```diff
@@ -55,22 +56,36 @@ fn cursor_path<'a>(
...
+            // half of the decision could see through composition and the other
+            // could not, an operation would be presented as a paginated table
+            // that silently stops after its first page.
+            let Some(view) =
+                merged_all_of_object_view(root, resolved, resolving_refs, next_depth, MAX_DEPTH)
+            else {
                 return Ok(None);
             };
-            for (name, property) in properties {
+            for (name, property) in &view.properties {
```
sauldhernandez: `view.properties` is a `BTreeMap`, so both first-match loops now visit properties lexicographically. The old `resolved["properties"]` was a `serde_json::Map` and the workspace enables `preserve_order`, so it visited them in document order.

That flips ties. A response declaring both `next` and `cursor` (or `next_page_token` and `end_cursor`) goes from whichever the spec declared first to the alphabetically first one, and a `{pageInfo, meta}` envelope now descends `meta` first. If the new winner echoes the request cursor instead of carrying the next one, pagination repeats page one until `max_pages` errors.

No current source looks affected, but the tiebreak is now property spelling. Iterating the token list in the outer loop and looking each name up would make the precedence explicit.

crates/coral-spec/src/v4/response_cursors.rs
```diff
@@ -117,7 +132,11 @@ fn property_is_object(
         resolving_refs,
         depth,
         MAX_DEPTH,
-        |resolved, _, _| Ok(json_schema_type_contains(resolved, "object")),
+        |resolved, resolving_refs, next_depth| {
+            Ok(json_schema_type_contains(resolved, "object")
+                || merged_all_of_object_view(root, resolved, resolving_refs, next_depth, MAX_DEPTH)
```
sauldhernandez: `schema_has_type` in wrapped_lists.rs guards this same fallback with `if resolved.get("allOf").is_none() { return Ok(false) }`; this one doesn't. Without `allOf` the merged view can only re-collect the type `json_schema_type_contains` already checked, so the call is wasted work for every non-object property on every walked level (a small allocation, plus real property clones for untyped schemas that carry `properties`). The same gate here keeps the two checks in sync and skips the no-op case.

crates/coral-spec/src/v4/surfaces/json_schema.rs
```diff
@@ -316,6 +300,149 @@ pub(crate) fn json_schema_scalar_type_or_string(schema: &Value) -> Option<IrScal
...
+/// answer questions against, and this returns `None` for them — which is what
+/// the callers did for all composition before.
+///
+/// Returns `None` when the schema is not object-like, so callers can keep
+/// treating "not an object" and "cannot be read as an envelope" alike.
+///
+/// On duplicate property names the first branch wins, and no conflict is
+/// reported. This answers a heuristic question, so an annotation-only
+/// disagreement between branches must not discard the whole envelope. Type
+/// import has the opposite need and keeps its own exact merge.
```
sauldhernandez: "keeps its own exact merge" points at `merge_json_schema_properties_exact`, which this same PR deletes — type import now goes through `merge_json_object_shape_annotation_insensitive`, which reports conflicts but is deliberately not exact. A reader chasing this sentence finds nothing. The contrast you want is heuristic first-branch-wins vs conflict-reporting: "Type import has the opposite need and uses a merge that reports conflicts", or mirror the `merge_all_of_properties` doc ("a genuine disagreement between branches must still be reported").

crates/coral-spec/src/v4/surfaces/json_schema.rs
```diff
@@ -316,6 +300,149 @@ pub(crate) fn json_schema_scalar_type_or_string(schema: &Value) -> Option<IrScal
...
+    );
+    if alternation {
+        return None;
+    }
+
+    // An explicit `type: object` on any branch settles it. Failing that, a
+    // schema that declares no type at all but does declare properties is an
+    // object by convention, which is how JSON Schema is usually written.
+    let object_like = view.declares_type("object")
+        || (view.declared_types.is_empty() && !view.properties.is_empty());
```
sauldhernandez: This clause widens the pass beyond `allOf`. `candidate_row_path` used to require `json_schema_type_contains(resolved, "object")`, which is false when `type` is absent, so a response declaring `properties` but no `type` was never an envelope candidate. Now every such schema gets a view.

Concretely: `{properties: {warnings: {type: array, items: {type: string}}}}` — no `type`, no `allOf`. Sole property, so `has_envelope_evidence` returns true on the `properties.len() == 1` shortcut; the sole-array fallback makes `warnings` the row path. The relation becomes a list of warning strings where before it was one JSON row — the silent reshaping the module header calls the costly failure. The `resolve_output_row_type_ref` cross-check doesn't catch it, because the importer also defaults typeless schemas to object, so the imported type really does have a `warnings` list.

It also leaves two notions of "object" in one walk: `schema_has_type` and `property_is_object` still require a declared type, so a typeless root is object-like while a typeless child is not.

Intended? Graph declares `type: object` on the envelope and on every base, so the OData fix doesn't need it. If unintended, the fix belongs at the `candidate_row_path` call site (`if !view.declares_type("object") { return Ok(None); }`) rather than here — `cursor_path` never required `type: object` and relies on this fallback. Every new test stays green either way. If it is intended, it deserves a test and a line in the description, since it moves inference for every source, not just Graph.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -50,32 +70,42 @@ impl OpenApiImporter<'_> {
...
+                    property,
+                ))) => {
+                    diagnostics.push(Diagnostic::new(
+                        format!(
+                            "allOf property '{property}' conflicts in operation '{operation_id}'"
+                        ),
+                        Some(operation_id.to_string()),
+                    ));
+                    return None;
+                }
```
sauldhernandez, line 127: Folding the whole ancestor tree also widens what can discard a type: a property conflict (or depth overflow) anywhere in the chain returns `None` here, where before only the immediate branches were ever compared.

A legal narrowing chain triggers it: `Base.status: {type: string}`, `Mid: allOf [Base, {properties: {status: {type: string, enum: [...]}}}]`, `Derived: allOf [Mid, ...]`. Old code never compared the two `status` declarations (Mid has no direct properties), so Derived imported. Now they fingerprint differently — `enum` is a constraint, not an annotation key — so Derived keeps the placeholder `IrTypeShape::Json`, loses every column, and `infer_row_path` drops its row path with it. Under allOf intersection semantics the re-declaration is valid, so well-formed specs can hit this.

Did the v4-metadata-report diff show any new `allOf property '...' conflicts` on Graph? If real specs reach this, dropping just the disputed property (or letting the first declaration win) plus a diagnostic naming the two disagreeing branches would cost one property instead of the whole type.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -150,6 +180,69 @@ impl OpenApiImporter<'_> {
...
+    /// is still a conflict.
+    fn merge_all_of_properties(
+        &self,
+        schema: &Value,
+        merged: &mut JsonObjectShape,
+        operation_id: &str,
+        depth: usize,
+        diagnostics: &mut Vec<Diagnostic>,
+    ) -> Result<(), AllOfMergeError> {
+        if depth > MAX_ALL_OF_DEPTH {
```
sauldhernandez: Nothing tests this cap on the importer path. `A: {allOf: [{$ref: '#/components/schemas/A'}]}` — the exact case the constant's doc names — recurses here, and `resolve_ref` does no cycle tracking, so this check is the only thing between a self-referential spec and unbounded recursion at import time. A refactor that drops or mis-threads `depth` would only fail at runtime on a hostile spec. The inference side got both tests (`ignores_an_all_of_branch_that_cycles`, `ignores_all_of_branches_nested_past_the_depth_cap`); the same pair here, asserting the diagnostic and `output.type_ref == "json"` like `importer_warns_for_openapi_all_of_property_conflicts` does, would pin it.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -150,6 +180,69 @@ impl OpenApiImporter<'_> {
...
+            return Err(AllOfMergeError::Comparison(
+                JsonSchemaComparisonError::DepthExceeded,
+            ));
+        }
```
sauldhernandez: When this trips, the limit hit is `MAX_ALL_OF_DEPTH` (8), but reusing `JsonSchemaComparisonError::DepthExceeded` makes the diagnostic read "exceeds maximum comparison depth" — that's `MAX_PROPERTY_COMPARISON_DEPTH` (64), the ceiling the constant docs above deliberately separate. Whoever debugs a deeply composed spec tunes the wrong knob, and for the cyclic `A: {allOf: [$ref A]}` case this cap exists for, the message never says the schema is cyclic. A dedicated `AllOfMergeError::CompositionTooDeep` with its own wording would name the ceiling actually hit.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -150,6 +180,69 @@ impl OpenApiImporter<'_> {
...
+        .map_err(AllOfMergeError::Comparison)?;
+        for branch in schema
+            .get("allOf")
+            .and_then(Value::as_array)
+            .into_iter()
+            .flatten()
+        {
+            let branch = self
+                .resolve_ref(branch, operation_id, diagnostics)
+                .ok_or(AllOfMergeError::UnresolvedRef)?;
```
sauldhernandez: `resolve_ref` resolves one hop, so an alias branch — `{$ref: Alias}` where `Alias` is itself `{$ref: Base}` — comes back as `{$ref: Base}`. The recursion then sees no `properties` and no `allOf`, and the branch contributes nothing, silently, with no diagnostic. Inference's fold follows `$ref` chains (`with_resolved_json_schema` recurses while the target still has a `$ref`), so it can find a row path behind the alias that the imported type lacks; `infer_row_path` then discards the path and the collection falls back to a single JSON row — the shape this PR is fixing. That breaks the both-halves-fold-the-same-tree invariant this function's doc states.

Resolving branches through `with_resolved_json_schema` closes the gap and gives real cycle detection instead of leaning on the depth cap; the MCP importer's `input_object_shape` already folds that way. Minor aside: `MAX_ALL_OF_DEPTH`'s doc says it matches the inference walk, but inference's 8 is a shared budget also spent on ref hops and the entry resolve, so its effective composition depth is lower than this fold's.

crates/coral-spec/src/v4/wrapped_lists.rs
```diff
@@ -147,17 +162,19 @@ fn candidate_row_path<'a>(
...
+            // `oneOf`, and `not` still bail: they describe a choice of shapes,
+            // not one shape.
+            let Some(view) =
+                merged_all_of_object_view(root, resolved, resolving_refs, next_depth, MAX_DEPTH)
+            else {
                 return Ok(None);
-            }
-            if !json_schema_type_contains(resolved, "object") {
+            };
+            let properties = &view.properties;
```
sauldhernandez, line 174: `has_envelope_evidence` treats exactly-one-property as proof of an envelope, and that count now runs on the folded view. A single-branch `allOf` — the common 3.0 idiom for hanging a `description` on a `$ref` — is enough to trip it:

```yaml
RunnerResponse:
  description: a runner
  allOf: [{$ref: '#/components/schemas/Runner'}]
Runner:
  type: object
  properties:
    public_ips: {type: array, items: {type: string}}
```

On this branch `GET /runners/{id}` infers `row_path: ["public_ips"]` with no pagination contract and no metadata sibling (reproduced on fc03dc656; before this PR any `allOf` bailed out and the response stayed one row). The `infer_row_path` cross-check doesn't catch it, because `public_ips` really is a list on the imported type. So the singleton silently becomes a table of IP strings — the wrong-envelope outcome the module doc calls the costly failure.

To be fair on scope: a bare `$ref: Runner` response already hit the sole-property rule pre-PR, so the rule itself is pre-existing; this change extends it to the wrapped spelling. Two options: make a view that folded any branch earn its evidence from the metadata lexicon only (record on `MergedObjectView` whether folding happened, and skip the shortcut then), or accept the widening deliberately and pin the singleton case with a test.

crates/coral-spec/src/v4/wrapped_lists.rs
```diff
@@ -182,7 +203,9 @@ fn candidate_row_path<'a>(
                         inherited_evidence || evidence,
                         resolving_refs,
                         next_depth,
-                    )?
+                    )
+                    .ok()
+                    .flatten()
```
sauldhernandez, line 211: This swallow does more than the comment above it says. Before, a walk failure under a preferred branch (`items`/`data`/`results`/`rows`) propagated out and aborted the whole inference — empty row path, no envelope. Now the loop continues and the sole-array fallback below becomes reachable.

Concretely: `{total_count: integer, data: {$ref: 'other.yaml#/Thing'}, attachments: [...]}`. `data` is an external ref, so its walk errors; `total_count` supplies evidence; `attachments` is the only walkable array. The relation becomes the attachments list instead of the resource, and `resolve_output_row_type_ref` doesn't reject it because `attachments` really is a list field. That's the trade the module header warns against — a wrong envelope silently discards the declared resource, where the old conservative miss just left an unnestable column.

Either remember that a preferred branch existed but failed to walk and skip the sole-array fallback in that case, or narrow the swallow to `RefCycle`/`DepthExceeded` so unresolvable refs still abort.

crates/coral-spec/src/v4/wrapped_lists.rs
```diff
@@ -260,8 +281,20 @@ fn schema_has_type(
...
+            }
+            if json_schema_type_contains(resolved, expected) {
+                return Ok(true);
+            }
+            if resolved.get("allOf").is_none() {
+                return Ok(false);
+            }
+            Ok(
+                merged_all_of_object_view(root, resolved, resolving_refs, next_depth, MAX_DEPTH)
+                    .is_some_and(|view| view.declares_type(expected)),
```
sauldhernandez, line 301: `merged_all_of_object_view` returns `Some` only when the folded view is object-like, so `declares_type(expected)` here can only succeed for `expected == "object"` — the `pagination`/`meta` lexicon rows. The `"array"` call sites and the string/integer/boolean lexicon checks never benefit: `value: {allOf: [{$ref: ArrayOfChats}]}` (the wrap-a-ref-to-attach-a-description idiom) and `nextLink: {allOf: [{type: string}], description: ...}` still read as untyped. The doc above ("a property assembled from a base and an extension still declares its type") promises more than that. Either narrow the comment to object composition, or have the fold report declared types without the object-likeness gate so non-object composition works too.

sauldhernandez (approved):

crates/coral-spec/src/v4/wrapped_lists.rs
simonwhitaker: Deliberate, and I'd go further: the immunity was a side effect of `schema_uses_composition` bailing on any composed schema, never a decision. `allOf: [X]` means exactly `X` — a single-branch `allOf` is the 3.0 idiom for hanging a `description` on a `$ref`, and it should infer identically to a response that references the branch's contents directly. Pre-PR those two spellings disagreed; folding makes them agree, so this is the inconsistency going away rather than a widening to accept.

Not touching the sole-property rule here. It's pre-existing and applies equally to the bare `$ref` spelling, so whether one property is enough evidence for an envelope is a separate question from composition.

It doesn't reach the catalog either way: across all 20,673 operations the stack changes 2,610 rows (2,605 Graph, 5 Stripe), and none is a singleton gaining a row path. `actions_get_hosted_runner_for_org` has an empty row path before and after — GitHub's `actions-hosted-runner` is a plain `type: object` with ~12 properties, so it never meets the one-property shortcut.

crates/coral-spec/src/v4/surfaces/json_schema.rs
simonwhitaker: Good catch — unintended, and removed in `168d8b7`.

Folding replaced the walk's explicit `type: object` check with the merged view's own object test, and that test had been written to accept a schema declaring no type but some properties. That reads a constraint as an assertion: in JSON Schema `properties` applies only if an instance happens to be an object and asserts nothing about whether it is one, so `{properties: {...}}` validates a string quite happily. Your `{properties: {warnings: [...]}}` really would have had `warnings` promoted to the whole relation, and as you say `resolve_output_row_type_ref` could not have caught it, since type import defaults typeless schemas to object as well.

The declared-object requirement is back, so the widening is now exactly `allOf` visibility — `type: object` on any branch settles it, which is the OData shape — and nothing looser.

It costs nothing to refuse the typeless case: across the six ingested v4 sources every typeless success-response root is `allOf` (3,242), `anyOf` (705), or `oneOf` (8), and not one is a bare property bag, so the report does not move by a line either way. Added a test pinning the refusal, and a note to revisit if a real spec turns up where accepting it would help.

That also disposes of the two notions of "object" you flagged: `schema_has_type` and `property_is_object` both require a declared type, and so does the merged view now, so the walk is consistent again without loosening anything.

crates/coral-spec/src/v4/wrapped_lists.rs
simonwhitaker: You are right that behaviour changed, but the discard is not a choice — it is forced, and I have made the comment say so rather than defending it on the merits.

`JsonSchemaWalkError<'a>` borrows from the schema it walked. Before folding, `properties` came straight out of the document (`resolved.get("properties")`), so an error could carry `'a` and `?` compiled. The merged view owns its `BTreeMap<String, Value>`, so `property` now points into a local and the error cannot outlive the closure. Restoring `?` fails with E0515, `cannot return value referencing local data view.properties` — and narrowing to `RefCycle`/`DepthExceeded` hits the same lifetime, so that option is not available either.

On whether it matters today: I instrumented the arm to print on every `Err` and ran the full catalog. It fires **zero times across all 20,673 operations in the six v4 sources**, and the report is byte-identical either way. Your `{total_count, data: <external $ref>, attachments: [...]}` needs an external `$ref` under a preferred name, and none of the ingested specs has one.

So I have left the behaviour and fixed the comment, which was the misleading part: it now states that propagation is unavailable and why, what the discard actually changes (the sole-array fallback stays reachable where inference used to abort), and that if a spec ever reaches it the fix is your first suggestion — remember that a preferred branch failed and withhold the fallback — since narrowing the swallow cannot work.

`2a8f808`.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
simonwhitaker: Done in `410c2f6`, folded through `with_resolved_json_schema` as you suggested.

The alias case is closed by construction: the walk loops while the resolved value still holds a `$ref`, so `{$ref: Alias}` reaching `{$ref: Base}` now lands on `Base` instead of contributing nothing. Real cycle detection comes with it — `resolving_refs` reports `A: {allOf: [$ref A]}` as a cycle rather than leaning on the depth cap — and it drops the per-branch `resolve_ref` clone.

Your aside about `MAX_ALL_OF_DEPTH`'s doc was right, and it is now accurate rather than aspirational: the cap is a shared budget spent on composition nesting and `$ref` hops alike, which is how inference already spent it. The doc says so.

Two wrinkles worth recording. The merge's own failures have to ride inside the walk's `T`, because `with_resolved_json_schema` fixes the error type and a property conflict must stay distinguishable from a ref that would not resolve — that is why the per-level work moved into its own method. And the walk error is converted to a lifetime-free `AllOfMergeError` before returning, since it borrows from the schema it walked and `import_schema` passes a local.

No catalog movement: 0 of 20,673 operations change. So no ingested spec has an alias-chained branch today, and the shared depth budget drops nothing that imports now — Graph's `entity` → `baseItem` → `drive` chains stay well inside 8.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
simonwhitaker: Fixed in `410c2f6`. `AllOfMergeError::CompositionTooDeep` with its own wording — "allOf composition nests past 8 levels in operation '...'" — so the message names the ceiling actually hit instead of `MAX_PROPERTY_COMPARISON_DEPTH`'s 64.

The cyclic case now reads correctly too, rather than being reported as a depth problem at all. The fold resolves branches through `with_resolved_json_schema`, which tracks the refs it is resolving, so `A: {allOf: [$ref A]}` produces "allOf branch '#/components/schemas/A' is cyclic" and never reaches the cap.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
simonwhitaker: Added in `410c2f6`: `importer_reports_a_cyclic_all_of_branch` and `importer_reports_all_of_composition_past_the_depth_cap`, both asserting the diagnostic and `output.type_ref == "json"` the way `importer_warns_for_openapi_all_of_property_conflicts` does, plus `importer_folds_all_of_branches_through_alias_refs` for the chain case.

I checked they pin behaviour rather than passing incidentally: swapping the previous one-hop `schemas.rs` back in fails all three.

One change to what they cover. The exact case the constant's doc names — `A: {allOf: [{$ref: A}]}` — is no longer a depth-cap case at all. The fold now resolves through `with_resolved_json_schema`, so ref tracking catches it as a cycle and the cap never runs. The cap test therefore uses a 12-deep composition chain to reach it independently, and the cycle test asserts the cycle wording. That is a better split than the original single guard: a refactor that mis-threads `depth` fails the second test, and one that drops cycle tracking fails the first.

crates/coral-spec/src/v4/response_cursors.rs
simonwhitaker: Restructured in `45db037` — tokens in the outer loop, as you suggested.

Worth flagging what that turned up, because it is more than a tiebreak: the loop asked each property whether it was in `name_tokens`, so the order of those lists was never consulted at all. Two of the three callers rank theirs deliberately — the body next-URL lexicon puts `odatanextlink` ahead of `nexthref`, the MCP cursor lexicon puts `nextcursor` ahead of `endcursor` — and neither ranking had ever had any effect. Those lists now mean what they say.

You are right that both old and new orderings were arbitrary in themselves. The reason alphabetical was the worse accident is that it made the answer depend on property spelling, where before folding it depended on the vendor's key order — and these descriptors are generated, so a regeneration could have silently moved a cursor with no semantic change. Neither was a decision; the token list is.

Nothing moves today: 0 of 20,673 operations change, and no schema level in the six sources declares two cursor-named or two next-URL-named properties, so nothing is ranked in this catalog yet. Test asserts the rank beats the property name and flips when the list is reversed; it fails against the previous loop.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
simonwhitaker: > Did the v4-metadata-report diff show any new `allOf property '...' conflicts` on Graph?

Measured it directly — instrumented the conflict arm and imported all six sources on each side. **38 conflicts before the fold, 1 after.** None of them on Graph after; before, 37 of the 38 were Graph.

They were the `@odata.type` restatements. Every subtype re-declares the discriminator with its own `default`, which byte equality read as a disagreement, so those types were discarded. Folding the full chain is what forced the comparison to become annotation-insensitive, and that is what fixed them. So this widened what gets *compared* and narrowed what gets *discarded*, by 37 types.

The one survivor, `apps_create_from_manifest` / `client_id`, is present on both sides — pre-existing, not introduced here. And no operation loses a row path across the stack.

On the narrowing chain itself: your reading of the mechanism is right, and worth recording. Pre-fold, `Derived` did not merely skip the comparison — with `Mid` holding no direct properties it never saw `status` at all, nor anything else inherited through the chain. It imported, but missing its inherited columns, which is the same silent field loss that cost `microsoft.graph.drive` its `id`, `name` and `webUrl`. So the old outcome was not the correct one reached by luck; it was the bug this pass exists to remove.

What remains legitimate is that equality is the wrong test for `allOf`. Composition intersects, so `{type: string}` and `{type: string, enum: [...]}` both apply and the narrowing is valid — treating non-identical as conflicting is a conservative stand-in for "can I reason about these together?", and it can still cost a whole type where it should cost nothing. Your suggestion — drop just the disputed property, or let the first declaration win, plus a diagnostic naming the two branches — is the right shape for that, and I would rather it landed as its own change than be folded into this one, since nothing in the catalog reaches it and the dimension is already 37 types better off.

crates/coral-spec/src/v4/mod.rs
simonwhitaker: Bumped to `openapi-v8` in `bc15776`. You were right on the principle and it has only got stronger since: importer output has moved twice more in this branch — branches now resolve through `$ref` chains, and composed types keep fields the old fold dropped. #1791 and #1884 set the precedent and this matches it.

crates/coral-spec/src/v4/openapi_tests.rs
simonwhitaker: Added in `bc15776`: `importer_misses_the_page_size_a_boolean_count_parameter_masks`, the same fixture plus `- {name: $count, in: query, schema: {type: boolean}}`, asserting `page_size.is_none()` alongside the row path and mode that do survive.

It confirms your reading — `$count` is chosen and rejected and `$top` is never reached, so page-size detection finds nothing while pagination is otherwise intact. The assertion is written to be flipped by whoever fixes `find_numeric_query_input` to filter by type before choosing.

crates/coral-spec/src/v4/surfaces/json_schema.rs
simonwhitaker: Fixed in `bc15776`. It now points at `merge_json_object_shape_annotation_insensitive` and draws the contrast you suggested: heuristic first-branch-wins here, conflict-reporting there.

crates/coral-spec/src/v4/wrapped_lists.rs
simonwhitaker: Corrected in `bc15776` — narrowed to object composition, as you suggested. The doc now says `allOf` branches count for `object`, and that other types must declare themselves outright because the merged view is object-only, so a composed `array` or `string` is not covered.

Also took your `response_cursors.rs:221` point in the same commit: `property_is_object` now skips the merge when there is no `allOf`, matching `schema_has_type`. Worth noting that gate was briefly incompatible with a looser object rule I had tried, which is why it did not land earlier — that rule is gone, so the gate is simply correct now.

crates/coral-spec/src/v4/response_cursors.rs
simonwhitaker: Added in `bc15776` — `property_is_object` now returns early when there is no `allOf`, matching `schema_has_type`. Your reasoning holds exactly: without branches the merge has only this schema to collect, so it could only re-derive what `json_schema_type_contains` already answered, and it runs for every property on every walked level, cloning properties as it goes.

One note in case it looks like it should have been obvious earlier: the gate was briefly incompatible with a looser object rule I had tried, where a schema declaring `properties` and no type counted as an object. Under that rule the no-`allOf` case was not a no-op and the early return would have been wrong. That rule is gone — the merged view requires a declared `type: object` again — so the gate is simply correct.

No catalog movement: 0 of 20,673 operations change.

PR #2005, opened 2026-07-29
feat: expose sources used by functions

AlbertQM:
## Summary

Says on the tin. It'd be cool to exposes a list of all the sources a saved function touches, so we can show them in the UI and e.g. enable filtering too

## Test plan

UI now shows sources.

image


Bradley-Butcher (approved):
lgtm

PR #1855, opened 2026-07-20
feat(spec): add Universal Search authorisation contracts

mystic123:
TL;DR  
  
PR #1855 adds the source-spec contract for deciding which DSL v4 endpoints Coral may use for Universal Search.

In simple terms, a source author can declare:

- Which imported HTTP endpoint or MCP tool may receive search queries.
- Which parameter receives the user’s query.
- Which routes are explicitly denied.
- How returned fields map to result identity, title, description, URL, and attributes.

It validates that Coral can safely call the endpoint using one search string: the query parameter must be a required text parameter, while every other required input must already have an imported default.

It is DSL v4-only. DSL v3 sources cannot opt into fanout.

Crucially, this PR only defines and validates the permission/configuration. It does not yet call endpoints or enable fanout—that happens in later PRs in the stack.

It also hardens manifest parsing against duplicate YAML keys, preventing one authorization rule from silently overwriting another.  
  
Summary

- Adds passive Universal Search authorization only for DSL v4 top-level `universal_search.routes`, with explicit allow/deny decisions, exact imported `operation_id` and query-input identity, and RFC 6901 result mappings.
- Rejects both `functions[].universal_search` and table-function `args[].default` authoring on every DSL v3 HTTP and MCP parse path. DSL v3 functions remain available through ordinary SQL, catalog discovery, and observed-value search, but cannot authorize provider fanout.
- Adds a type-preserving runtime carrier for defaults imported by downstream DSL v4 surface lowering, including the distinction between an omitted default and explicit JSON `null`; this is not a DSL v3 authoring field.
- Rejects duplicate YAML mapping keys in both single-manifest and multi-document bundle parsing, and rejects duplicate DSL v4 operation targets so conflicting policies cannot collapse or alias.
- Validates MCP detail-hint targets manifest-wide and updates schemas, source-spec documentation, and authoring/review skills to describe the DSL v4-only contract.

## Stack boundary

Stack 1/11, based on `main`. #2009 builds on this PR.

This PR accepts and validates source-authored DSL v4 intent and provides only the shared runtime value carrier needed by later v4 lowering. It does not resolve routes into Search, enable fanout, or execute provider calls.

## Test plan

- `cargo test -p coral-spec`
- `make schema-check`
- `make docs-check`
- `make rust-checks` on this standalone PR layer: 2,057 passed, 2 skipped
- `make rust-checks` on the cumulative top branch: 2,275 passed, 2 skipped
[...]

PR #1865, opened 2026-07-20
docs(search): document native provider fanout

mystic123:
## Summary

- Expands the Search guide with the default-off enablement flow, the independent DSL v4 source-policy gate, a native preview example, execution budgets, response caps, partial-result behavior, and rollback.
- States the compatibility boundary explicitly: provider fanout is DSL v4-only. DSL v3 sources remain queryable through SQL and discoverable through local catalog and observed-value search, but Search never executes their functions as fanout routes.
- Keeps the complete DSL v3 provider-ranked function example for direct SQL use while documenting migration to a DSL v4 imported surface plus top-level `universal_search` route for fanout.
- Documents exact DSL v4 imported `operation_id` targets and authored route IDs; generated schema and function names are query-visible projections, not authorization identities.
- Explains that provider fanout is request-triggered read-only upstream work: it adds no background crawl, provider-response cache, or raw-row Search response.
- Documents how successfully decoded provider batches can feed the existing best-effort observed-values pipeline when enabled, while remaining unavailable to the Search response currently being assembled.
- Adds the MCP capability/annotation matrix and expands CLI, configuration, and security references with the 750/600/100/25 ms timing contract, route/row limits, bounded diagnostics, and data-flow boundaries.

## Stack boundary

Stack 11/11. Depends on #1864.

This PR changes documentation only. It describes the default-off, source-authorized DSL v4 capability implemented below it without implying any DSL v3 fanout compatibility.

## Test plan

- `make docs-check`
- `make schema-check`
- `make lint-proto`
- `make rust-checks`: 2,275 passed, 2 skipped

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

PR #1864, opened 2026-07-20
feat(search): expose default-off provider fanout

mystic123:
## Summary

- Adds the default-off `search_provider_fanout` runtime feature and matching process overrides. The disable override wins over stored enablement.
- Wires `NativeFanoutProvider` into bootstrap only when the runtime feature is enabled and at least one source-authorized DSL v4 route resolves. Enabling the feature never authorizes a source by itself.
- Adds a process-level regression that installs both an eligible DSL v4 route and an ordinary DSL v3 `kind: search` function, then proves only the v4 route appears in capabilities and receives an upstream request.
- Adds a workspace-scoped Search capabilities RPC backed by the same runtime catalog resolution and lifecycle snapshot as Search. It exposes at most sixteen sorted safe DSL v4 route identities; the disabled path does not prepare the fanout capability catalog.
- Makes CLI and MCP behavior reflect app-owned capabilities. MCP refreshes capabilities for initialization, tool listing, and dynamic guide reads, and explicitly describes eligible DSL v4 connected-source routes.
- Exposes native results and diagnostics through clients, records bounded low-cardinality fanout metrics, and adds release-safe configuration, security, and rollback documentation.

## Rollout and stack boundary

Stack 10/11. Depends on #1863; #1865 builds on this PR.

This is the first PR in the stack that can make request-triggered provider calls. The built-in default remains off, and both gates are required: the runtime feature and an eligible source-authored DSL v4 route. DSL v3 sources remain available through SQL and local search discovery but are never executed as fanout routes.

## Test plan

- `cargo test -p coral-app fanout --lib`
- `cargo test -p coral-app search::capabilities --lib`
- `cargo test -p coral-mcp --lib`
- `make schema-check`
- `make lint-proto`
- `make rust-checks` on the cumulative top branch: 2,275 passed, 2 skipped

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

PR #1863, opened 2026-07-20
feat(search): orchestrate bounded native fanout

mystic123:
## Summary

- Adds `NativeFanoutProvider`, consuming the request-scoped runtime `CatalogResolution` held under the workspace lifecycle snapshot. Native routing performs no separate manifest scan and maintains no route cache.
- Selects at most four source-authorized DSL v4 routes deterministically, with one route per source in the first pass, and executes them concurrently in one wave.
- Uses exact DSL v4 operation identity for stable ordering and row identity; no DSL v3 route or identity branch exists.
- Enforces the request-relative fanout contract: 750 ms global cutoff, 600 ms per call, 100 ms minimum start budget, 25 ms cancellation cleanup, five rows per route, no retries, and no pagination.
- Normalizes, deduplicates, and fuses safe native previews while emitting bounded stable diagnostics and coverage. Provider failures, timeouts, cancellation, and panics do not discard local catalog or observed-value results.
- Tags observations produced by native calls with a Search origin and cutoff so successfully decoded batches can enter the existing observed-values pipeline only for later Search requests.

## Stack boundary

Stack 9/11. Depends on #1861; #1864 builds on this PR.

Production bootstrap still supplies no native provider. This PR cannot introduce upstream calls from public Search; activation is isolated to #1864.

## Test plan

- `cargo test -p coral-app search::native::provider::tests --lib`
- `cargo test -p coral-app search::engine::tests --lib`
- `cargo test -p coral-app search::observed::sqlite_store::tests --lib`
- `cargo test -p coral-app search_paused_in_catalog_resolution_does_not_recreate_deleted_workspace_storage`
- `cargo check -p coral-app --all-targets`
- `make rust-checks` on the cumulative top branch: 2,275 passed, 2 skipped

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

PR #2009, opened 2026-07-29
feat(spec): preserve DSL v4 operation contracts for fanout

mystic123:
## TL;DR

When Coral imports a DSL v4 OpenAPI endpoint or MCP tool, it converts that external definition into Coral’s internal format. Previously, some important details could be lost during that conversion. #2009 makes those details survive all the way to runtime:

- Exact parameters, types, locations, and defaults.
- The difference between “no default” and “default is null.”
- Original operation IDs.
- MCP read-only and idempotency hints, including explicit `false`.
- Nested response paths and result shapes.

It also prevents source authors from changing the imported contract accidentally—for example, replacing an endpoint’s default `limit: 25` with `limit: 100`, or selecting a synthetic field that does not actually exist in the provider response.

Later fanout PRs can therefore safely answer: “What arguments does this endpoint need, how should we call it, and how should we interpret its response?”

It does not authorize endpoints, make them searchable, enable fanout, or make network calls. It only preserves and validates the information needed by those later layers. DSL v3 is unaffected.

## Summary

- Preserves DSL v4 imported operation contracts through semantic IR, projections, runtime registration, and materialization.
- Carries typed argument defaults without collapsing an omitted default and an explicit JSON `null`; projection overrides must preserve the imported default exactly.
- Gives OpenAPI parameter ref-site defaults precedence over referenced-schema defaults.
- Preserves the original operation ID, read-only and idempotency hints, nested source paths, importer-synthetic field metadata, and MCP structured-output and row-path metadata.
- Rejects authored projection paths that select importer-synthetic compatibility fields while keeping provider-authored fields such as `raw` addressable.
- Derives unique top-level and nested projection column names without repeatedly rebuilding source-path indexes.
- Skips nested numeric object-key paths that the runtime would interpret as array indexes, so generated catalogs remain compatible with their imported schemas.
- Omits absent MCP behavior hints from normalized catalogs while preserving explicit `true` and `false`.
- Rotates the affected schema, surface, importer, operation-metadata, and projection artifact versions so stale generated artifacts fail closed.
- Keeps this layer passive: it does not attach Universal Search authorization, resolve routes, enable fanout, or execute provider calls.

## Stack boundary

Stack 2/11. Depends on #1855; #2010 builds on this PR.

This PR establishes faithful DSL v4 runtime metadata for later route resolution. DSL v3 sources do not gain fanout metadata or compatibility behavior.

## Test plan

- `cargo test -p coral-spec` (481 passed)
- `cargo test -p coral-engine build_query_pairs_omits_missing_args_but_preserves_explicit_values --lib`
- `cargo clippy -p coral-spec -p coral-engine --all-targets --all-features -- -D warnings`
- `make schema-check`
- `make rust-checks` (2,320 passed, 5 skipped)

### DIFF
--- a/crates/coral-spec/src/v4/projections/derive.rs
+++ b/crates/coral-spec/src/v4/projections/derive.rs
@@ -239,7 +239,12 @@ fn generated_projection_name(operation: &IrOperation, is_search: bool) -> String
 }
 
 fn projection_input_required(input: &IrOperationInput) -> bool {
-    input.required && (input.default_value.is_none() || input.location == IrInputLocation::ToolArg)
+    input.required
+        && (input.location == IrInputLocation::ToolArg
+            || input
+                .default_value
+                .as_ref()
+                .is_none_or(|default| default.value().is_null()))
 }
 
 fn rest_input_exposure(
@@ -450,6 +455,15 @@ impl ScalarLeafCollector<'_, '_> {
         description: &str,
         depth: usize,
     ) {
+        // Collector paths traverse object fields only. Runtime reserves a
+        // numeric path segment for list indexing, so a numeric object key and
+        // everything below it could never be read from the projected row.
+        if path
+            .last()
+            .is_some_and(|segment| segment.parse::<usize>().is_ok())
+        {
+            return;
+        }
         let Some(ty) = self.type_by_id.get(type_ref) else {
             return;
         };
--- a/crates/coral-spec/src/v4/projections/derive.rs
+++ b/crates/coral-spec/src/v4/projections/derive.rs
@@ -20,6 +20,8 @@ use super::names::{
 };
 type TypeIndex<'a> = HashMap<&'a str, &'a IrType>;
 
+const MAX_SCALAR_LEAF_PROJECTION_DEPTH: usize = 64;
+
 pub fn generate_projection_catalog(
     manifest: &V4SourceManifest,
     plan: &ValidatedSurfacePlan,
@@ -305,10 +307,11 @@ fn projection_columns(
     operation: &IrOperation,
 ) -> Vec<ProjectionColumn> {
     if matches!(&operation.execution, IrExecutionAttachment::Mcp(_)) {
-        // MCP output schemas drive row cardinality and response extraction, but
-        // SQL columns stay opaque until Coral has stable per-tool payload
-        // normalization semantics.
-        return vec![
+        // Keep the two opaque compatibility columns while also projecting
+        // deterministic scalar leaves from an authored output schema. The
+        // original property segments remain in `source_path`, so route result
+        // pointers can resolve without relying on normalized SQL names.
+        let mut columns = vec![
             ProjectionColumn {
                 name: "result".to_string(),
                 data_type: ManifestDataType::Utf8,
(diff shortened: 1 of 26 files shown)


mystic123:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/withcoral/coral/2009?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1865** <a href="https://app.graphite.com/github/pr/withcoral/coral/1865?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1864** <a href="https://app.graphite.com/github/pr/withcoral/coral/1864?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1863** <a href="https://app.graphite.com/github/pr/withcoral/coral/1863?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1861** <a href="https://app.graphite.com/github/pr/withcoral/coral/1861?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1859** <a href="https://app.graphite.com/github/pr/withcoral/coral/1859?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1858** <a href="https://app.graphite.com/github/pr/withcoral/coral/1858?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1857** <a href="https://app.graphite.com/github/pr/withcoral/coral/1857?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1856** <a href="https://app.graphite.com/github/pr/withcoral/coral/1856?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2010** <a href="https://app.graphite.com/github/pr/withcoral/coral/2010?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2009** <a href="https://app.graphite.com/github/pr/withcoral/coral/2009?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/2009?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1855** <a href="https://app.graphite.com/github/pr/withcoral/coral/1855?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1854** <a href="https://app.graphite.com/github/pr/withcoral/coral/1854?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

crates/coral-spec/src/v4/projections/derive.rs
```diff
@@ -305,10 +312,11 @@ fn projection_columns(
...
     if matches!(&operation.execution, IrExecutionAttachment::Mcp(_)) {
-        // MCP output schemas drive row cardinality and response extraction, but
-        // SQL columns stay opaque until Coral has stable per-tool payload
-        // normalization semantics.
-        return vec![
+        // Keep the two opaque compatibility columns while also projecting
+        // deterministic scalar leaves from an authored output schema. The
+        // original property segments remain in `source_path`, so route result
+        // pointers can resolve without relying on normalized SQL names.
+        let mut columns = vec![
```
simonwhitaker, line 319: The reason I didn't do this previously is that, unlike an OpenAPI spec, an MCP output schema gives no guarantee of stability. It's intended to help the calling agent more efficiently parse the responses it gets _this session_, but beyond that it makes no guaranteees. By baking it into the projections you're making it more likely that a source silently breaks in future.

We deliberately don't ever regenerate projections implicitly; it's always an explicit step. While that remains the case, I think we should not generate projections from MCP output schema.

simonwhitaker (changes requested):
Requesting changes because of the MCP projection generation from output schemas. See inline comment.

PR #2036, opened 2026-07-31
chore(spec): add missing auth block to v4 Gitlab source

simonwhitaker:
Copied from the v3 source. This should have been added in the initial commit of the spec but I missed it.


simonwhitaker:
@cubic you're welcome to approve this. It brings across the auth block from `sources/core/gitlab/manifest.yaml` and transplants it into the Gitlab v4 manifest. It should have been there all along, but I forgot to add it. The manifest is incomplete and unusable without it.

PR #2038, opened 2026-07-31
fix(search): use singular observed source identity

mystic123:
## What

Collapses the observed-values pipeline onto one canonical `source_name` and
migrates the sidecar to schema **v5**, preserving observed data.

Since #1791 one DSL v4 source publishes exactly one surface and one SQL
namespace, so `owner_source_name` (lifecycle owner) and `source_name`
(query-visible schema) are the same thing. The equality this rests on is
checked at the capture seam by #2078, downstack.

## Reviewing this

~55% of the diff is tests, and most of the rest is mechanical. Suggested order:

| Where the thinking is | |
|---|---|
| `migrations/0005_observed_source_identity.sql` | the target schema (104 lines) |
| `search/sqlite_store.rs` **above `mod tests`** | migration hook, validation probes, singular-era repair (~430 lines) |

| Mechanical — forced renames, safe to skim | |
|---|---|
| `sqlite_projection.rs`, `governance.rs`, `observed/sqlite_store.rs` | drop the owner column from structs and SQL |
| `sqlite_store/tests.rs`, inline test modules | fixtures follow the struct change |

## Migration design (v4 → v5, preserve equal rows)

All inside `apply_migration`'s single transaction, so an interrupted upgrade
rolls back to intact v4:

1. Detect the legacy shape, **per table** — `observed_values`,
   `observed_queue_jobs`, `observed_values_fts` independently. A sidecar that is
   singular in one and legacy in another then heals through the *preserving*
   path instead of falling through to the discarding reset.
2. Rename legacy tables aside and drop the five legacy indexes **by name** —
   SQLite keeps index names attached to renamed tables, so 0005's
   `CREATE INDEX` would otherwise silently no-op and leave the new table
   unindexed. This ordering is load-bearing: 0005 uses `IF NOT EXISTS`, so it
   would no-op against legacy tables still holding the canonical names.
3. `execute_batch(0005)` — the complete singular schema.
4. Copy back every row whose two identities agreed, then repopulate FTS from the
   canonical table. Preserved values are searchable the moment the transaction
   commits — no reconcile gap.

Rows with divergent identities are discarded. Unreachable on production paths,
so this only affects synthetic state; preserved and discarded counts are logged
so a nonzero discard in the wild is visible and falsifiable.

## Two failure modes worth your attention

**Repairing a singular schema must not replay owner-era DDL.** Migration 0002
names `owner_source_name` in three of the five index statements. If one of those
indexes goes missing on a v5 database, the full-history replay cannot recreate
it, and the resulting "recoverable" error routed **all observed values and queue
jobs** into `discard_observed_values_schema`. So a singular schema is now
repaired from version 5 onward first. That also makes the later replay inert,
because `CREATE INDEX IF NOT EXISTS` short-circuits on the index name before it
resolves the column list.

**FTS rows carry their canonical `rowid` explicitly.** The projection keys every
FTS row to its canonical rowid and refreshes by inserting at that exact rowid.
Letting fts5 assign fresh sequential rowids misaligns the two whenever canonical
rowids are gapped (after eviction), and the next refresh of a gapped row
collides with whatever now occupies its rowid, failing that queue job until it
dead-letters.

Both are covered by tests verified to fail without their fix.

## Validation gains absence probes

On all three owner-bearing tables. Presence probes cannot see a *surplus*
column, so a surviving legacy queue table (`owner_source_name NOT NULL`, no
default) would pass validation and then reject every rewritten enqueue with no
self-heal.

## Scope ids are byte-stable

The Rust field is renamed with `#[serde(rename = "component_source_name")]` and
`SOURCE_SCOPE_FORMAT_VERSION` stays 1. The hash pinned in #2077 still passes
untouched — that is what proves no bytes moved. Rotating scope ids here would
make every preserved row fail-closed invisible.

## Compatibility

The sidecar becomes **forward-only** at v5: older binaries refuse a newer
`user_version`. Do not run old and new binaries against the same config
directory.

No wire, CLI-flag, or JSON changes: hit `schema_name` was always populated from
`source_name`, and hits never read the owner column. No `github_v4_mcp` alias.
No change to `sources/manager.rs`.

### DIFF
--- a/crates/coral-app/src/bootstrap/server.rs
+++ b/crates/coral-app/src/bootstrap/server.rs
@@ -1178,7 +1178,6 @@ enabled = false
             .enqueue_if_current(
                 &workspace,
                 &ObservedValuesQueueJob {
-                    owner_source_name: "github".to_string(),
                     source_name: "github".to_string(),
                     source_scope_id: "scope".to_string(),
                     surface_kind: ObservedValuesSurfaceKind::Table,
@@ -1638,7 +1637,6 @@ backend = "unsupported"
             .enqueue_if_current(
                 &workspace,
                 &ObservedValuesQueueJob {
-                    owner_source_name: "github".to_string(),
                     source_name: "github".to_string(),
                     source_scope_id: "scope".to_string(),
                     surface_kind: ObservedValuesSurfaceKind::Table,
@@ -1687,7 +1685,6 @@ backend = "unsupported"
             .enqueue_if_current(
                 &workspace,
                 &ObservedValuesQueueJob {
-                    owner_source_name: "github".to_string(),
                     source_name: "github".to_string(),
                     source_scope_id: "scope".to_string(),
                     surface_kind: ObservedValuesSurfaceKind::Table,
--- a/crates/coral-app/src/search/manager.rs
+++ b/crates/coral-app/src/search/manager.rs
@@ -356,8 +356,8 @@ impl SearchManager {
         if request.scope == SearchDataScope::All {
             return match &request.target {
                 SearchClearTarget::Workspace => self.clear_workspace_all(&request.workspace_name),
-                SearchClearTarget::Source(owner_source_name) => {
-                    self.clear_source_all(&request.workspace_name, owner_source_name)
+                SearchClearTarget::Source(source_name) => {
+                    self.clear_source_all(&request.workspace_name, source_name)
                 }
             };
         }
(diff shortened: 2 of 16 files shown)


mystic123:
* **#2038** <a href="https://app.graphite.com/github/pr/withcoral/coral/2038?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/2038?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mystic123:
Both findings were valid and are fixed in the latest push.

**`live_scope.rs:180`** — correct, and it was my own artifact: `main` asserted `first_scope.owner_source_name` then `first_scope.source_name`, and the blanket rename collapsed those into two identical lines. Restored as `second_scope.source_name`, which is the better of the two options here — the test's subject is that reinstalling rotates the scope id, so asserting the *name* is stable while the *id* changes states the claim more sharply than simply deleting the duplicate.

**`provider.rs:744`** — agreed. That string predates this PR, but the line directly below it now reads `failed_source_names()`, so the function contradicted itself in a user-visible diagnostic. Now " source(s)".

Also swept a third instance of the same residue that the review didn't flag: a leftover `owner_while_failed` local in `sqlite_store/tests.rs`.

Remaining `owner_source_name` hits in this PR are deliberate — historical migrations 0002/0004, the version-5 hook's legacy detection and copy, the absence probes, and `#[cfg(test)]` fixtures that construct the pre-#1791 shape on purpose.

Bradley-Butcher (approved):

PR #2039, opened 2026-07-31
refactor(search): remove catalog source ownership projection

mystic123:
## What

Deletes the persisted second copy of a topology that no longer exists —
`catalog_source_owners` plus the runtime `schema → owner` map — and migrates the
sidecar to schema **v6**.

Since #1791 a source publishes exactly one schema, so
`catalog_documents.source_name` already *is* the installed source name. The
ownership projection records nothing the document does not.

No catalog data loss; one forced refresh per workspace.

## Migration 0006 is pure DDL

It drops the index and the table, and deliberately leaves catalog documents and
FTS rows in place. The refresh is forced by bumping
`CATALOG_SEARCH_SNAPSHOT_VERSION`, which guarantees every stored fingerprint
mismatches its recomputed value — so each workspace refreshes exactly once
before its documents are trusted again, and unlike a `DELETE` in the migration
file it does nothing extra on later repair replays.

> **Note the version is `-v5`, not `-v4`.** `-v4` already shipped on `main` in
> #2041 for grouped catalog surfaces. Reusing it would leave workspaces that
> already refreshed under #2041 matching, and the refresh this migration depends
> on would silently never run. Git does not flag this: both sides changed that
> line to the same value, so it auto-merges with no conflict marker.

## The version-4 hook had to be neutered

Its guard is `migration.version == 4 && !tables_exist(["catalog_source_owners"])`.
After 0006 that table is *permanently absent*, so **every** post-v6 repair
replay — including the observed-only reset path whose log line promises to
preserve the catalog — would have fired the hook and deleted all catalog
documents, FTS rows, and fingerprints. Forever.

It cannot be re-gated on the stamped version, because a replay re-stamps
`user_version` to 1, 2, 3 before 0004 runs. So its three DELETEs are removed;
the snapshot-version bump subsumes their invalidation job. Pinned by a
repair-replay test.

## Validation inverts

`catalog_schema_is_current` now requires the ownership table to be **absent**.

Known bounded edge, accepted and not special-cased: a v5 database whose repair
replay fails recoverably *before* 0006 executes still has the table, fails
catalog validation, and routes to the full-schema rebuild — a
refresh-from-provider, not corruption.

## Source clear becomes a direct delete

The FTS delete keeps its own `workspace = ?1` guard because `doc_id` is not
workspace-qualified. Production runs one sidecar per workspace, so that guards
test layouts and defence in depth rather than a live corruption path — it stays
because the schema is workspace-keyed and dropping it buys nothing.

Catalog **lifecycle** invalidation remains workspace-wide by design
(`clear_catalog_projection_for_source_lifecycle_best_effort`) — a refresh cost,
not data loss. Changing it is sources-domain work and out of scope.

## Public wording

`--source` is the installed source name, which is also its SQL schema name. CLI
help and both docs pages drop the multiple-runtime-components language; #2041's
rewrite had reintroduced it in new wording. No wire, CLI-flag, or JSON changes.

### DIFF
--- a/apps/docs/guides/search-with-coral.mdx
+++ b/apps/docs/guides/search-with-coral.mdx
@@ -45,9 +45,8 @@ example.
 
 Here, `github` is the query-visible SQL schema. JSON and MCP results include it
 in `sql_reference`. A catalog-backed table uses a three-part
-`catalog.schema.table` reference. For a source with multiple runtime components,
-the query-visible schema can differ from the installed source name because each
-component can publish its own SQL schema.
+`catalog.schema.table` reference. The schema is also the installed source name:
+one source publishes one SQL namespace.
 
 When more fields matched than were shown, the result reports how many were left
 out so you can widen the query rather than assume the column does not exist.
@@ -102,7 +101,7 @@ To remove observed values for a workspace:
 coral search-index clear --scope observed-values --workspace default --yes
 ```
 
-To remove them for one installed source and every runtime schema and surface it owns, pass the canonical installed source name:
+To remove them for one installed source, pass its name:
 
 ```shellscript
 coral search-index clear --scope observed-values --source github --workspace default --yes
--- a/apps/docs/reference/cli-reference.mdx
+++ b/apps/docs/reference/cli-reference.mdx
@@ -165,9 +165,8 @@ Clear Coral's local search data for one workspace.
 Clear deletes local search data, so Coral requires both `--yes` and an explicit
 `--workspace NAME`.
 
-`--source` accepts the canonical installed source name, not a runtime SQL
-schema. It clears the selected scope for every runtime schema and surface owned
-by that source.
+`--source` accepts the installed source name, which is also its SQL schema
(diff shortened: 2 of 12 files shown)


mystic123:
* **#2039** <a href="https://app.graphite.com/github/pr/withcoral/coral/2039?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/2039?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Bradley-Butcher (approved):

mystic123:
### Merge activity

* **Aug 18, 1:09 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/2039).
* **Aug 18, 1:09 PM UTC**: @mystic123 merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/2039).

PR #2040, opened 2026-07-31
fix(spec): allow injecting OpenAPI parameters via manifest to support…

### LINKED ISSUE #1990: microsoft_graph: createdDateTime filters on chat messages are silently ignored
Reported by an external tester evaluating `sources/v4/microsoft_graph` against a production Microsoft 365 tenant.

## What happens

Filtering chat messages by `createdDateTime` returns the full unfiltered set, with no error. The agent believes it filtered when it did not, which makes this a silent-wrong-answer bug rather than a missing feature.

## Why

`chatMessage` only honours a filter on `createdDateTime` or `lastModifiedDateTime` if the request also passes an `orderby` on the same field. Without the matching `orderby`, Graph quietly ignores the filter and returns everything. This is documented Graph behaviour, but it is an easy trap.

The connector exposes no `orderby` control, so a `createdDateTime` filter cannot be made to work at all.

Confusingly, a `lastModifiedDateTime` filter *does* work as expected — so the failure is inconsistent between two adjacent fields, which makes it harder to spot.

## Expected

Either expose `orderby` so the filter can be paired correctly, or have the connector pair it automatically when a date filter is applied to `chatMessage`. Failing either, reject the filter rather than silently dropping it.

OP-Prajwal:
**What changed**: 
Added `injected_parameters` to the V4 OpenAPI surface configuration so source authors can declare parameters missing from upstream specs. Updated the `microsoft_graph` manifest to inject `$orderby`.

**Why it changed**: 
Upstream OpenAPI specs (like Microsoft Graph) often omit OData query options, which prevented users from combining `$orderby` with date filters on endpoints like `chatMessages`.

Fixes #1990

### DIFF
--- a/crates/coral-spec/src/v4/manifest.rs
+++ b/crates/coral-spec/src/v4/manifest.rs
@@ -83,6 +83,19 @@ pub struct OpenApiRuntimeConfig {
     pub auth: AuthSpec,
     pub request_headers: Vec<HeaderSpec>,
     pub rate_limit: RateLimitSpec,
+    pub injected_parameters: Vec<InjectedParameterSpec>,
+}
+
+/// Parameter to inject into the `OpenAPI` surface when parsed.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
+#[serde(deny_unknown_fields)]
+pub struct InjectedParameterSpec {
+    /// Optional list of operation IDs this parameter applies to. If empty or missing, it applies to all operations.
+    #[serde(default)]
+    pub operations: Option<Vec<String>>,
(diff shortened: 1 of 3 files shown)


simonwhitaker:
I don't think this is the way to solve this, for a few reasons:

1. We want to keep v4 manifests as small as possible. There needs to be a strong argument for adding new parameters to it
2. In order to use this `injected_parameters` field, the author needs to know the internal operation ID of the operation, which is only calculated once the source has been added to Coral. This creates a chicken-and-egg scenario that we should try to avoid.

PR #2049, opened 2026-08-03
refactor(reef): extract a presentational MCP clients list

AlbertQM:
## Summary

Says on the tin. We'll reuse this both in settings and in the onboarding

## Test plan

image


AlbertQM:
@cubic-dev-ai if you have no more comments, can you accept?

AlbertQM:
@cubic-dev-ai then re-review

PR #2056, opened 2026-08-04
docs: document Docker self-hosting

sauldhernandez:
## Intent

Document the public contract for running Coral with the official container image introduced by #2055. The documentation must make the unauthenticated network boundary, once-only seed behavior, writable-volume requirements, secrets posture, and operational limitations explicit.

## What it enables

- Start Coral with the exact one-port, one-volume quickstart.
- Scope the published port to a trusted network or host loopback.
- Supply a complete first-boot config through `CORAL_SEED_CONFIG`.
- Change the server port without breaking the image health check.
- Administer the shared state with `docker exec`.
- Upgrade by recreating the container while preserving its volume.
- Configure compatible Docker UID/GID and Kubernetes `fsGroup` settings.

## Usage examples

```bash
docker run -d -p 14555:14555 -v coral-data:/var/lib/coral ghcr.io/withcoral/coral
```

For host-local access only:

```bash
docker run -d -p 127.0.0.1:14555:14555 -v coral-data:/var/lib/coral ghcr.io/withcoral/coral
```

## Validation

- `make docs-check` passed after rebasing onto current `main`.
- `apps/docs/docs.json` passed JSON parsing and navigation checks.
- Both Compose examples passed `docker compose config --quiet`.
- All YAML examples parsed successfully.
- Diff-scope and whitespace checks passed.
- Fable review was explicitly skipped by the user.

Stack: 2 of 2. Depends on #2055.

### DIFF
--- a/apps/docs/guides/self-host-with-docker.mdx
+++ b/apps/docs/guides/self-host-with-docker.mdx
@@ -1,0 +1,198 @@
+---
+title: "Self-host Coral with Docker"
+description: "Run Coral's long-lived server with the official container image."
+---
+
+<Warning>
+  Coral's native gRPC server does not authenticate clients. Publish its port
+  only inside a trusted network boundary, or put Coral behind an authenticating
+  proxy. Do not expose the default container directly to the public internet.
+</Warning>
+
+The official image packages the same Coral binary as the corresponding GitHub
+release. It currently supports `linux/amd64` and stores all persistent state in
+one writable volume.
+
+## Run the server
+
+You need Docker and a network where every client that can reach Coral is
+trusted. Start the server with one port and one named volume:
+
+```shellscript
+docker run -d -p 14555:14555 -v coral-data:/var/lib/coral ghcr.io/withcoral/coral
+```
+
+The container listens on port `14555` and includes a gRPC health check. The
+`coral-data` volume contains configuration, installed sources, credentials,
+query state, and other durable state. Reuse that volume when you replace the
+container with a newer image.
+
+To limit access to clients on the Docker host, bind the published port to
+loopback instead:
+
+```shellscript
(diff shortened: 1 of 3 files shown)


apps/docs/guides/self-host-with-docker.mdx
```diff
@@ -0,0 +1,198 @@
...
+for the listener and authentication boundary.
+
+## Known limitations
+
+- The image is `linux/amd64` only at launch.
+- Steady-state server logs are quiet; most runtime errors are returned to the
+  requesting gRPC client. Expanded server-side logging is planned separately.
+- One volume supports one running Coral server. Multi-replica and
+  high-availability deployments are not supported.
+- The container does not serve the Coral desktop web UI.
```
AlbertQM, line 198:
```suggestion
- The image is currently `linux/amd64` only.
- One volume supports one running Coral server. Multi-replica and
  high-availability deployments are not supported.
- The container does not serve the Coral desktop web UI.
```

If you're planning on adding/changing that soon, I wouldn't add it to the docs altogether, wdyt?

>- The container does not serve the Coral desktop web UI.

In my head "desktop" and "web" are two separate things. Are you saying we can't reach that coral instance via web? or only that the Desktop app currently can't reach remote servers? (we should fix this). If it's the latter only, I'd remove that point since you should still be able to see the UI for that Coral instance

AlbertQM (approved):
Cool, thanks! Few people mentioned if we could host Coral with Docker, and now it's even easier 😄

sauldhernandez:
Superseded by #2204, which merged a fuller version of this guide at the same path (apps/docs/guides/self-host-with-docker.mdx). The README pointer this PR added may return as a small standalone PR written against the merged guide.

PR #2064, opened 2026-08-05
feat(reef): use native gRPC for hosted Coral requests

sauldhernandez:
> Reef hosted-auth stack: layer 7 of 11. Base: PR #2063.

## Intent

Use Node's native gRPC transport for authenticated BFF-to-Coral calls and attach the server-held bearer token. Preserve gRPC-Web only for unauthenticated local and Desktop traffic.

## What it enables

Hosted Reef can call Coral's authenticated HTTP/2 gRPC server directly, without an MCP URL or browser-facing transport proxy. Explicit loopback HTTP remains supported when the complete Reef and Coral topology is local.

## Usage examples

Hosted deployment:

```bash
CORAL_ENDPOINT=https://coral.internal.example.com
```

All-local authenticated deployment:

```bash
CORAL_ENDPOINT=http://127.0.0.1:14555
```

Non-loopback authenticated endpoints must use HTTPS.


apps/reef/app/lib/coral-request.server.ts
```diff
@@ -54,6 +50,29 @@ export function coralEndpointForRequest(request: Request): string {
   return url.origin
 }
 
-function coralTransportForRequest(request: Request) {
-  return createGrpcWebTransport({ baseUrl: coralEndpointForRequest(request) })
+function coralTransportForRequest(request: Request, accessToken: string | null) {
+  const baseUrl = coralEndpointForRequest(request)
+  if (!accessToken) return createGrpcWebTransport({ baseUrl })
```
lazybean: I am a bit confused and actually wonder if that is necessary.
Couldn't unauthentified local or desktop still use grpc? 
What does grpc web have that make reqest without access token possible, that native grpc cannot do?

lazybean (approved):
Looks good.
I do wonder if the fallback to grpc web is necessary at all.

apps/reef/app/lib/coral-request.server.ts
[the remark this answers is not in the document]
sauldhernandez: Fixed in `733e20b2`. The transport now follows the deployment topology rather than whether a token reached the call, and hosted mode fails closed.

It throws a plain `Error` deliberately, not the expired-session redirect. That redirect carries a header `_protected` answers by clearing the session cookie — so reusing it here would log out a user whose session is fine, walk them through a successful login, and land them back on the same broken caller: a loop reporting nothing. In `required` mode `_protected` has already proven a session before any loader runs, so a null token is a caller that dropped it, and a 500 naming that beats a redirect hiding it. It also matches the sibling endpoint check one branch below.

The same commit stops each transport building its own `Http2SessionManager` — they are now shared per Coral origin. One consequence worth knowing at review time: concurrent Reef→Coral calls share a session's stream limit and flow-control window instead of each holding their own.

sauldhernandez: Good question, and it turned out to be the same fix as cubic's comment above.

**The fallback is load-bearing, but not for the reason the code implied.** `coral ui` — what the Desktop sidecar spawns, and the local dev default — serves gRPC-Web *only*. Its `GrpcWebOnlyService` answers any `application/grpc` content type with HTTP 415: `crates/coral-app/src/bootstrap/server.rs:888-897`, and the round trip is asserted at `:1911-1953` (gRPC-Web POST → 200, native gRPC POST → 415). So an unauthenticated local or Desktop client genuinely cannot use native gRPC — the server on the other end refuses it.

So the branch stays. What was wrong is what it keyed on: **token presence** rather than the deployment topology that actually decides it. Those coincide today only because local mode has no token, and treating them as the same thing meant a hosted request that lost its token got an unauthenticated transport and an opaque transport error, instead of the auth error it is.

`733e20b2` makes the condition say what it means — the branch belongs to `mode === 'disabled'`, and a null token under `required` now fails closed with a message naming the defect.

PR #2071, opened 2026-08-05
feat(app): allow loopback OAuth client metadata

sauldhernandez:
> Coral server prerequisite stack: layer 2 of 2. Base: PR #2070.

## Intent

Permit Coral's authorization server to resolve CIMD over plain HTTP only for an explicit all-loopback topology. Derive the allowed loopback client IDs from configured loopback audiences and retain strict outbound URL checks everywhere else.

## What it enables

Developers can exercise the complete Reef-to-Coral OAuth and native-gRPC flow locally without a public HTTPS tunnel. Hosted, mixed-locality, hostname-alias, redirecting, oversized, and malformed metadata requests remain rejected.

## Usage examples

```toml
[server]
bind_addr = "127.0.0.1:14555"

[auth]
http_bind_addr = "127.0.0.1:9080"
allowed_audiences = ["http://localhost:5173"]
```

Reef then publishes CIMD at `http://localhost:5173/.well-known/oauth-client`, and Coral accepts that exact derived client ID only when its own authorization issuer is also explicit-loopback HTTP.

### DIFF
--- a/crates/coral-app/src/auth/authorization_server.rs
+++ b/crates/coral-app/src/auth/authorization_server.rs
@@ -269,6 +269,13 @@ impl AuthorizationServerHttpState {
         state_store: Arc<InMemoryStateStore>,
         authorization_resources: Arc<BTreeSet<String>>,
     ) -> Result<Self, AuthServerError> {
+        let client_metadata_resolver = Arc::new(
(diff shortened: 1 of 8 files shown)


lazybean (approved):
thank you

PR #2074, opened 2026-08-05
refactor(catalog)!: unify surface description

Bradley-Butcher:
## Intent

Replace the table-only `describe_table` MCP tool with one exact `describe` tool for tables and table functions. The request uses required `schema` and `surface` fields and an optional `catalog` field. `describe` does not perform fuzzy matching or catalog search.

## Changes

- Rename the public MCP tool from `describe_table` to `describe`. There is no compatibility alias.
- Replace the `DescribeTable` gRPC request, response, and RPC with `DescribeCatalogSurface`.
- Add three typed describe results: exact table, exact table function, and missing.
- Return these MCP shapes:
  - Table: `kind`, description, guide, required filters, column count, and a `list_columns` hint.
  - Table function: `kind`, description, guide, arguments, and result columns.
  - Missing: `{ "kind": "missing" }`.
- Resolve a table or table function from one prepared query runtime. A miss does not copy catalog tables or table functions into the response path.
- Reject table and table-function name collisions when the runtime is prepared and when a user function is added. The check uses the normalized SQL namespace, so case-only collisions are also rejected.
- Require nonblank catalog, schema, and surface values, but pass their exact values without trimming.
- Add `catalog`, `schema`, and `surface` to structured search results for tables and table functions. Keep `sql_reference` for direct SQL use.
- Remove the `list_columns` fallback that called the old table-description path. A missing table now returns a `list_columns` tool error.
- Update MCP documentation, the maintained Coral skill, mocks, and contract tests for the new names and result shapes.

## Ownership

- `coral-api` defines the gRPC request and typed result contract.
- `coral-engine` prepares the runtime, enforces the shared table/function namespace, and resolves exact matches.
- `coral-app` loads workspace sources, records query telemetry, and converts engine results to protobuf.
- `coral-mcp` validates the flat JSON input, calls the gRPC operation, and renders the MCP output schema.

## Compatibility

This is a breaking MCP and protobuf change:

- `describe_table` is removed. Callers must use `describe` with `schema` and `surface`.
- `DescribeTable` is removed. gRPC callers must use `DescribeCatalogSurface`.
- The describe response schema changes to the three result shapes above.
- An omitted `catalog` identifies a two-part `schema.surface` target. It does not search all catalogs as `describe_table` did. Callers must provide `catalog` for a three-part database table.
- A missing `list_columns` target now returns a tool error instead of the previous structured recovery response.

Structured search results now include the exact `catalog`, `schema`, and `surface` values that callers can pass to `describe`. `sql_reference` remains available for SQL.

`list_catalog` and exact `list_columns` behavior remain available.

## Not in This PR

- No fuzzy matching, FTS query, recovery candidates, or next-action suggestions in `describe`.
- No full table column list in `describe`; callers still use `list_columns`.
- No removal of `list_catalog` or `list_columns`.

## Telemetry

The existing query-operation instrumentation records an exact table or function as a successful result with row count 1. It records missing as a successful result with row count 0. Errors returned by runtime preparation or exact lookup remain errors. Isolated source registration failures can still leave the requested surface missing.

## Validation

- `make rust-checks`: formatting, Clippy, 2,537 tests, and rustdoc passed; 7 tests skipped.
- Focused client search rendering and MCP output-schema tests passed.
- Focused engine, app gRPC, MCP schema, and MCP integration tests passed.
- `make docs-check` passed.
- `make perf-check` passed: `coral.tables` mean 212 ms against the 750 ms limit.
- `git diff --check` passed.


mystic123 (changes requested):
Read through the whole change. Three points — one small bug, two design questions.

## 1. `describe` throws away the source-failure signal

`manager.rs:443` creates a `CatalogFailureRecorder`, passes it in, and never reads it. It also ignores `source_load.failed_source_names`. `list_catalog` does the identical setup at `:378` and reads both back.

The recorder's `source_failed` returns `Ignore`, so a source that fails to register just vanishes from the catalog. That means a broken source and a typo produce the same output: `{"reason": "missing"}`. Nothing tells them apart. The old response at least listed `available_schemas`, so you could see the schema wasn't there.

Fix: clone the recorder in, read `failed_source_names()` after, and put them on the missing result. `MissingCatalogSurface` is empty right now, so it needs a field.

## 2. An unqualified `describe` can't find database tables, and doesn't say why

`Some(catalog_name.unwrap_or_default())` becomes the predicate `catalog_name = ''` in `append_catalog_filter`, so a `describe` without `catalog` only matches two-part tables. The old code searched every catalog and, when more than one matched, said:

```
table reference `public.events` is ambiguous; qualify the catalog:
`analytics.public.events`, `warehouse.public.events`
```

That told the agent exactly what to do next. The same call now returns `{"reason": "missing"}`.

Right now this only affects users with a database source connected, so it looks small. It won't stay small. As we onboard customers on Postgres, every table becomes three-part and every `describe` needs `catalog`. Then this is the normal failure, not an edge case.

Two things make it worse in that world:

- `search` results only carry `sql_reference` (`coral-client/src/search.rs:63-79`) — one joined string like `warehouse.public.events`. `list_catalog` carries split `catalog_name` / `schema_name` / `name` (`surface/catalog.rs:443`). So an agent that found the table through `search` has to split that string itself to call `describe`, and nothing in the response says which segment is the catalog.
- `format_sql_identifier` quotes when needed. Postgres lowercases unquoted names, so mixed-case tables are common, and the agent sees `warehouse.public."OrderEvents"`. It has to strip the quotes. Leaving them in returns `missing`, with no hint why.

Suggestion: when the two-part lookup misses, do one more exact lookup with the catalog filter dropped.

```rust
let qualified = self.list_tables(None, Some(schema_name), Some(surface_name)).await?;
```

Return `{"reason": "missing", "qualified_matches": ["warehouse.public.events"]}`. Same exact name, one fewer filter. This isn't fuzzy matching, FTS, or the suggestion heuristics you removed — it just restores the old ambiguity message as a field on the miss.

Separately, adding `catalog_name` / `schema_name` / `name` to `search`'s `TableResultValue` would remove the string-splitting step entirely.

## 3. The collision check fails the whole runtime for a `describe`-only problem

`validate_catalog_surface_namespace` runs in `build_registered_runtime` and `install_udfs`. Both are on the path for every operation, so one name conflict stops `sql` as well.

SQL doesn't need this rule. `ScopedTableFunctionCall::parse` only intercepts a relation with `args: Some(_)`. `FROM demo.events` goes to the table planner; `FROM demo.events(...)` goes to the function planner. They can't collide. The only thing that can't cope is `describe`, because `surface` is one flat field.

The case that actually bites: someone adds function `github.foo`, then a later `github` source upgrade ships a table `foo`. The source validates on its own and passes. Then the combined runtime refuses to build and every call in the workspace fails. It's recoverable — `delete_function` doesn't need a runtime — but it's a full outage over a name SQL is indifferent to.

Suggestion: return both.

```proto
message AmbiguousCatalogSurface {
  Table table = 1;
  TableFunction table_function = 2;
}
```

Add it as a fourth arm on the `oneof`. `describe_catalog_surface` already has the dead `(Some(_), Some(_))` branch — point it here instead of `FailedPrecondition`. It renders as `{"kind": "ambiguous", "table": {...}, "table_function": {...}}`.

Then a collision breaks nothing: `sql` works, `list_catalog` works, `list_columns` works, and `describe` answers the question it was asked. Which means the check can come out of `build_registered_runtime` and `install_udfs` altogether. Keep it in `add_function` if you like — refusing to create a function that shadows a table is a sensible nudge — but it stops being something the runtime depends on.

**Small thing while you're in here:** the output union has no single discriminator. Table and function carry `kind`; missing carries `reason` and no `kind`. A client can't branch on one field. If you're adding a fourth arm anyway, worth making it `kind: "table" | "table_function" | "missing" | "ambiguous"` and demoting `reason` to detail.

Bradley-Butcher:
The intended design keeps `describe` as an exact lookup of the current query-visible catalog.

## 1. Source failure and missing results

`describe` returns metadata only for surfaces that are visible in the current query runtime.

A source can fail during load or registration. The catalog runtime excludes that source so that healthy sources remain available.

A `missing` result means:

> No query-visible surface matched the exact target.

It does not state why the surface is not visible. The surface can be absent, or its source can be unavailable.

`failed_source_names` describes the complete catalog preparation operation. It does not prove that a failed source owns the requested surface. The list can also contain failures that are not related to the target. Adding it to the result can therefore suggest an incorrect cause.

Catalog search uses this information to report degraded coverage. Telemetry records the source failure. `describe` does not need this information.

For these reasons, I do not plan to add `failed_source_names` to `MissingCatalogSurface`.

## 2. Exact lookup and catalog qualification

`describe` is not a search operation.

The intended flow is:

```text
search or list_catalog
-> discover a surface and its full identity
-> describe the exact surface
```

An omitted `catalog` identifies an exact two-part name:

```text
schema.surface
```

A database table has a three-part name:

```text
catalog.schema.surface
```

The caller must provide `catalog` for a three-part table. An omitted `catalog` does not mean that Coral must search all catalogs.

Therefore, `describe` must not perform another catalog-wide lookup. It must not return `qualified_matches`.

The current search JSON output makes this transfer more difficult. The app search model already has `catalog_name`, `schema_name`, and `name`, but the client output combines them into `sql_reference`.

That is a separate search-output issue. Search should return an exact structured target that a caller can pass to `describe`. `describe` should not guess a missing qualifier.

## 3. Table and function name collisions

A table and table function cannot have the same case-insensitive name in one Coral SQL namespace. This is an invalid catalog state.

The supported paths handle this state as follows:

- `coral-spec` rejects a source that declares both names.
- Adding a conflicting user function returns `FailedPrecondition`. Coral does not install the function.
- If a later source version adds a conflicting table, Coral marks the installed function as invalid. Coral excludes the function before UDF installation. The query runtime remains available with the table.
- The engine keeps a final check for invalid runtime packages and callers that bypass application validation.

The final engine check does not panic. It returns `CoreError::FailedPrecondition`.

The app converts this to gRPC `FAILED_PRECONDITION`. MCP returns a structured tool error. The process and MCP session remain active.

A runtime operation that uses the invalid package fails until the package is corrected. This is a defensive failure path. It is not a supported `describe` result.

For these reasons, I do not plan to add an `ambiguous` result or remove the runtime check.

The supported `describe` results remain:

- `table`
- `table_function`
- `missing`

## Result discriminator

I agree that all MCP results should use the same discriminator.

I will change the missing output from:

```json
{"reason": "missing"}
```

to:

```json
{"kind": "missing"}
```

The `reason` field is not necessary because it repeats the same information.

mystic123:
Dropping 1 and 3.

On 3 you're right and I missed the app-side guard. `evaluate_function_listings` seeds `sql_publish_targets` from the source's own tables (`functions/manager.rs:308-315`), so a function that collides with a table shipped by a later source version is marked invalid and dropped by `load_runtime_udfs` before `install_udfs` runs. The runtime stays up and the table wins. No outage, no `ambiguous` variant needed.

On 1 I accept the attribution argument. `failed_source_names` describes the preparation, not the target.

2 I want to keep open, narrowed to one thing.

## What changed, precisely

main:

```rust
self.list_tables(catalog_name, Some(schema_name), Some(table_name))
```

this branch, `query.rs:712-714`:

```rust
let table_catalog_filter = Some(catalog_name.unwrap_or_default());
self.list_tables(table_catalog_filter, Some(schema_name), Some(surface_name))
```

`append_catalog_filter` normalizes `""` back to empty and emits `catalog_name = ''`, so an omitted `catalog` matches two-part tables only. Your test names it directly — `describe_catalog_surface_without_catalog_excludes_three_part_tables`, over `analytics` and `warehouse`, asking for `public.events`. That is the exact input the old code answered with:

```
table reference `public.events` is ambiguous; qualify the catalog:
`analytics.public.events`, `warehouse.public.events`
```

I am not arguing the exact lookup is wrong. I am arguing this is a capability removal and Compatibility doesn't list it. "An unqualified `describe` of a three-part table returns `missing`" belongs next to the other breaking bullets regardless of what else we decide.

## The part I want changed

You wrote: "Search should return an exact structured target that a caller can pass to `describe`." Agreed — that is the fix. My objection is ordering.

`coral-client/src/search.rs` is untouched in this PR. On merge, the only tool that emits a describe-ready identity is `list_catalog`. An agent that found the table through `search` holds one joined `sql_reference` string, and `format_sql_identifier` will have quoted the mixed-case segment — `warehouse.public."OrderEvents"`. It has to split on the right dot and strip the quotes, and if it gets either wrong the response is `missing` with nothing indicating that's the problem. Today that costs us little because few workspaces have a database source. It stops being little the moment Postgres onboarding is the normal case, which is the same release train this is on.

Either of these closes it for me:

**(a)** Add `catalog_name` / `schema_name` / `name` to `TableResultValue` in this PR. `describe` keeps its exact contract and the flow you described works end to end on merge.

**(b)** On a two-part miss, one more exact lookup with the catalog filter dropped:

```rust
let qualified = self.list_tables(None, Some(schema_name), Some(surface_name)).await?;
```

returned as `{"kind": "missing", "qualified_matches": ["warehouse.public.events"]}`. Same exact name, one fewer predicate. Not FTS, not the suggestion heuristics you removed.

(a) is the better fix. I'd take either.

Bradley-Butcher:
Implemented option (a) in `f500d1fb0`.

Structured search results now include:

- `catalog` when the surface has a catalog
- `schema`
- `surface`
- the existing `sql_reference`

The field names match the `describe` input. A caller can pass the raw identity to `describe` without parsing the SQL reference. The SQL reference remains available because it is correctly quoted and ready for SQL.

This applies to table and table-function results. I did not add a fallback lookup to `describe`.

I also updated the Compatibility section. It now states that an omitted `catalog` identifies a two-part target and no longer searches all catalogs as `describe_table` did.

Validation: `make rust-checks` passed with 2,537 tests, plus Clippy and rustdoc. The focused client and MCP output-schema tests also passed.

mystic123 (approved):

PR #2075, opened 2026-08-05
fix(reef): show database catalogs in schema explorer

iaptsiauri:
## Summary

- preserve database catalog and provider-schema hierarchy in the Reef schema explorer
- route database table details with the full catalog, schema, and table identity
- include catalog_name in ListColumns requests while retaining existing two-part source and table-function routes
- cover catalog grouping, encoded routes, loader identity, tree navigation, and qualified table headings

## Root cause

Reef discarded catalog_name while grouping ListCatalog results and omitted it from ListColumns requests. Database tables therefore collapsed into ordinary schema groups and could not be addressed reliably when catalogs shared schema and table names.

## Validation

- npm run check --prefix apps/reef
- npm run typecheck --prefix apps/reef
- npm test --prefix apps/reef (108 files, 451 tests)
- npm run build --prefix apps/reef
- git diff --check

### DIFF
--- a/apps/reef/app/views/schema-explorer/schema-table.test.tsx
+++ b/apps/reef/app/views/schema-explorer/schema-table.test.tsx
@@ -1,0 +1,95 @@
+import { createMemoryRouter, Outlet, RouterProvider } from 'react-router'
+import { expect, it } from 'vitest'
+import { render } from 'vitest-browser-react'
+
+import type { SchemaResponse } from '@/lib/schema-explorer'
+import { routePath, routePattern } from '@/routing/routemap'
+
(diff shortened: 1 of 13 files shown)


apps/reef/app/views/schema-explorer/schema-table.test.tsx
lazybean, line 1: Would you mind making a storybook .stories for that instead please?

lazybean (approved):
Thank you!  
  
Could you add a screenshot?  
Also I could find a source that shows catalog (I add to do some mock for verify). Any example?

PR #2077, opened 2026-08-05
test(search): pin the observed scope-id hash

mystic123:
## What

Pins the exact bytes hashed into `source_scope_id`, and their sha256.

## Why

`source_scope_id` is `sha256(serde_json::to_vec(ScopeFingerprint { .. }))`, and
every stored observed row is keyed by it. Retrieval is **fail-closed**: a row
whose scope id no longer matches a live scope is invisible. So a change to this
serialization does not fail a test, return an error, or log anything — it
silently hides the observed corpus.

`ScopeFingerprint` is a private struct with ordinary-looking Rust field names.
Nothing marked those names as an on-disk format, so a routine rename would
rotate every scope id.

## Notes for review

- **No behaviour change.** The asserted constant is what this build already
  produces; the test passes on `main` untouched.
- The value of landing this on its own is ordering: anything downstack that
  touches `ScopeFingerprint` now has to keep a pre-existing assertion green,
  which is stronger evidence than a pin added in the same commit as the change
  it is supposed to guard.

### DIFF
--- a/crates/coral-app/src/search/observed/source_scope.rs
+++ b/crates/coral-app/src/search/observed/source_scope.rs
@@ -177,3 +177,47 @@ struct ScopeFingerprint<'a> {
     surface_kind: &'static str,
     surface_name: &'a str,
 }
+
+#[cfg(test)]
+mod tests {
+    use uuid::Uuid;
+
+    use super::{SOURCE_SCOPE_FORMAT_VERSION, ScopeFingerprint};
+    use crate::hash::sha256_hex;
+
+    /// Pins the exact bytes hashed into `source_scope_id`.
+    ///
+    /// Every stored observed row is keyed by that id and retrieval is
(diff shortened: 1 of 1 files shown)


mystic123:
* **#2077** <a href="https://app.graphite.com/github/pr/withcoral/coral/2077?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/2077?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mystic123:
Folded into #2078. A test-only PR this small costs a full review cycle and a restack on every main move, for a change nobody meaningfully reviews line-by-line. The hash pin still lands ahead of the rename in #2038, which was the only reason to separate it.

PR #2078, opened 2026-08-05
fix(search): fail sources whose runtime component identity diverges

mystic123:
## What

Establishes the two identity invariants the rest of the stack relies on, while
the owner/component pair still exists:

1. **Tripwire** — a runtime component whose name differs from its package's
   `source_name()` fails that source closed, source-locally.
2. **Scope-id pin** — the exact bytes hashed into `source_scope_id`, and their
   sha256.

#2038 downstack collapses the two stored identities into one and renames the
hashed field. Landing these first means that change rests on verified
invariants rather than assumed ones.

## The tripwire

Universal Search stores two identities per observed row: the installed source
that owns lifecycle clears and epochs, and the runtime component name used as
the SQL schema. #1791 made those the same thing — one source, one surface, one
SQL namespace — but nothing in search checked it.

It fails through channels that already exist; no new machinery:

- **capture**: the per-source loop skips the offending source and logs.
  Deliberately *not* the store-error arm above it, which aborts capture for
  every selected source — one source's incoherent identity must not silence its
  neighbours.
- **retrieval**: `load_source_scopes` propagates, and the existing
  `ObservedValuesLiveScopeLoadFailure` channel fails that source closed and
  surfaces a partial-result note.

### This is a tripwire, not an enforcement boundary

Producing singular identity is the sources domain's job. On current paths
divergence is not merely unlikely but impossible: the package `source_name` and
every component name are copied from the same `manifest.common.name` field in
the same function, and installed name == manifest name is enforced at load
(`sources/catalog.rs` `resolve_installed_manifest`).

Search defends at its own seam anyway, because its storage can still
*represent* divergence — nothing in SQLite prevents it, and a synthetic engine
package can construct it. Deliberately no validator in `runtime_package.rs`, no
change to `sources/manager.rs`.

Database components are exempt: they declare no HTTP/MCP observation surfaces,
so they contribute no stored identity for the tripwire to protect.

## The scope-id pin

`source_scope_id` is `sha256(serde_json::to_vec(ScopeFingerprint { .. }))`, and
every stored observed row is keyed by it. Retrieval is **fail-closed**: a row
whose scope id no longer matches a live scope is invisible. So a change to this
serialization does not fail a test, return an error, or log anything — it
silently hides the observed corpus.

`ScopeFingerprint` is a private struct with ordinary-looking Rust field names,
and nothing marked them as an on-disk format. The test asserts both the exact
serialized JSON and its digest.

It passes against today's code unchanged — no behaviour change here. #2038 then
renames that field with `#[serde(rename)]`, and this pre-existing assertion
staying green is what proves the rename moved no bytes.

## Notes for review

`publisher_preserves_owner_and_query_schema_for_multi_surface_v4_package`
asserted the exact topology the tripwire rejects — one package, two
differently-named components. It is replaced by tests that a divergent package
derives no scopes at all, and that skipping it leaves a healthy source
alongside it fully captured.

### DIFF
--- a/crates/coral-app/src/search/observed/live_scope.rs
+++ b/crates/coral-app/src/search/observed/live_scope.rs
@@ -101,10 +101,11 @@ impl ObservedValuesLiveScopeLoader {
             loaded_runtime.runtime_contract_fingerprint.as_str(),
             source.credential_revision,
         );
-        Ok(source_surface_scopes(&loaded_runtime.query_source, seed)
-            .into_iter()
-            .map(|scope| scope.live_scope())
-            .collect())
+        // A component whose name diverges from its package fails this source
+        // closed through the existing partial-result channel.
+        let scopes = source_surface_scopes(&loaded_runtime.query_source, seed)
+            .map_err(|error| AppError::FailedPrecondition(error.to_string()))?;
+        Ok(scopes.into_iter().map(|scope| scope.live_scope()).collect())
     }
 }
 
@@ -262,6 +263,7 @@ mod tests {
                 credential_revision,
             ),
         )
+        .expect("coherent source identity")
         .into_iter()
         .next()
         .expect("publisher scope")
--- a/crates/coral-app/src/search/observed/publisher.rs
+++ b/crates/coral-app/src/search/observed/publisher.rs
@@ -98,7 +98,22 @@ impl SearchObservationHandle {
                 source.runtime_contract_fingerprint,
                 source.credential_revision,
             );
-            for scope in source_surface_scopes(source.query_source, seed) {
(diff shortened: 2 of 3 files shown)


mystic123:
* **#2078** <a href="https://app.graphite.com/github/pr/withcoral/coral/2078?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/2078?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

iaptsiauri (approved):

PR #2085, opened 2026-08-06
feat(app): add workspace membership schema and repositories

sauldhernandez:
## Intent

Workspace access has to be a fact the server can read, not something derived from a token. Before this, the only durable trace of a signed-in person was a one-way SHA-256 digest stored on `tasks.created_by_principal_id`. A digest cannot be enumerated, cannot be shown to an operator, and cannot be granted a role, so no authorization can be built on it. This slice adds the two tables that replace it, and the repositories that read and write them.

## What it enables

Migration `0005_workspace_access_control` creates:

- `users` — keyed on a random internal `user_id`, holding the verified `issuer` and `subject`, an optional `display_name`, and creation and last-login timestamps. `subject` is unique. This is the identity everything later in the stack points at.
- `workspace_members` — `(workspace_id, user_id)` primary key with a `role` column constrained by `CHECK (role IN ('owner', 'member'))`, cascading from `workspaces`, referencing `users`, and indexed in both directions of lookup: a user's workspaces, and a workspace's owners.

Repositories over both tables, on SQLite and Postgres:

- `UsersRepo` — `upsert_login`, `get_by_user_id`, `get_by_subject`, `list`.
- `WorkspaceMembersRepo` — `role_for_user_id` and `workspaces_for_user_id` on any session; `owner_count`, `insert`, and `delete` on a transaction, because they only make sense inside one.
- `MemberRole` (`owner` / `member`) is the app-domain role type, parsed from and rendered to the column's constrained strings. A value the constraint should have excluded is `DbError::CorruptData`, not a default.

No digest participates in authorization anywhere in this stack, and nothing reads these tables yet. The migration is inert until later slices wire it up, and the new modules carry `dead_code` expectations until then.

## Usage examples

The schema and its cascade behavior, and each repository's round trip:

```
cargo nextest run -p coral-app workspace_access_control_migration_enforces_schema_and_cascade
cargo nextest run -p coral-app user_repository_round_trips_against_sqlite
cargo nextest run -p coral-app workspace_member_repository_round_trips_against_sqlite
make postgres-tests
```

Each repository test has a `*_against_postgres` twin that runs the same assertions against the other backend. Those are `#[ignore]`d unless `CORAL_TEST_POSTGRES_URL` is set, which is what `make postgres-tests` does.

---

Slice 1 of 16 in stack #2100. Every slice of this stack now passes `cargo fmt --check` and `cargo clippy --workspace --all-targets --all-features -- -D warnings` independently; the tip additionally passes `make rust-checks` (2,591 tests) and `make postgres-tests`.


sauldhernandez:

is worth carrying over.

PR #2086, opened 2026-08-06
feat(app): add transactional workspace membership state

sauldhernandez:
## Intent

Membership changes have to be all-or-nothing, and they have to hold one invariant no single statement expresses: a workspace must never end up with zero owners. This slice puts the mutations behind transactions on `CoralDb`, above the repositories added in slice 1, so the invariant lives in one place instead of at every future call site.

## What it enables

- `CoralDb::create_workspace_with_owner` inserts the workspace and its creator's `owner` row in one transaction. It holds the creator's `users` row first and answers `WorkspaceCreationOutcome::UserNotFound` rather than leaving an ownerless workspace behind when that user does not exist.
- `CoralDb::provision_login` upserts the login and creates the caller's personal `default-<user_id>` workspace in the same transaction, so a user and their own workspace come into existence together or not at all. `ensure_user_default_workspace` does the same for a user who already exists.
- `CoralDb::add_workspace_member` and `remove_workspace_member` serialize concurrent mutations by holding the parent `workspaces` row first. Removal refuses to take the last owner (`RemoveMemberOutcome::LastOwnerProtected`). Adding someone who already holds a *different* role is refused here as `AddMemberOutcome::RoleConflict`; slice 16 replaces that refusal with a role move.
- `CoralDb::reattribute_pre_v1_task_principal_digest` rewrites historical `tasks` attribution from the old digest to an internal `user_id`. Slice 3 calls it during login.
- `ValidatedOidcIdentity` now carries the verified `issuer` and `subject` beside the principal claim, so provisioning has the values it needs to key a user row. The provider's display-name claim is stored verbatim rather than trimmed, and claim validation is tightened: a missing or non-string `iss`, `aud`, `sub`, `iat`, `exp`, or `nonce` is rejected.

Nothing above the state layer reaches any of this yet; the new modules carry `dead_code` expectations that later slices remove.

## Usage examples

```
cargo nextest run -p coral-app login_creates_only_a_free_personal_workspace
cargo nextest run -p coral-app workspace_member_state_round_trips_against_sqlite
cargo nextest run -p coral-app reattributes_only_the_pre_v1_task_digest
make postgres-tests
```

`workspace_member_state_round_trips_against_sqlite` runs racing adds and racing removals: two identical adds converge on one membership, two adds asking for different roles agree on which one won, and exactly one of two competing last-owner removals is refused. `workspace_member_state_repository_round_trips_against_postgres` asserts the same contract on the other backend under `make postgres-tests`.

---

Slice 2 of 16 in stack #2100. Every slice of this stack now passes `cargo fmt --check` and `cargo clippy --workspace --all-targets --all-features -- -D warnings` independently; the tip additionally passes `make rust-checks` (2,591 tests) and `make postgres-tests`.


sauldhernandez:

is worth carrying over.

PR #2087, opened 2026-08-06
feat(app): provision users at login and key sessions on user IDs

sauldhernandez:
## Intent

A session token used to carry the upstream OIDC `sub`, and task attribution carried a digest of it. Neither can be listed, granted a role, or revoked, so neither can anchor authorization. This slice moves the OIDC callback to provision a real user row first, and mints the session on the internal `user_id` instead.

## What it enables

- The OIDC callback provisions the login before it stores an authorization code, under a two-second timeout. A provisioning failure or a timeout answers `server_error` / `authorization failed` and stores no code, so a login that could not be recorded can never become a session. Failures log `OIDC callback could not provision the authenticated user`, without the issuer or subject value.
- `SessionTokenIssuer::issue_access_token` takes the internal `user_id` and places it in `sub`; `SessionPrincipalProvider` parses that claim into a `PrincipalKind::User` principal. Nothing downstream of the callback ever sees the upstream subject.
- Provisioning goes through `CoralDb::provision_login_and_reattribute_pre_v1_tasks`, so the user row, their personal `default-<user_id>` workspace, and the rekeying of their historical `tasks` rows all commit together. Signing in again returns the same `user_id`.
- A subject presented under an issuer different from the stored one is refused (`LoginProvisionError::IssuerMismatch`) rather than silently rebound to the new provider. That refusal is what stops a provider reusing a subject from inheriting someone's workspaces; recovering from a genuine issuer change is slice 15's admin tool.
- `Principal::for_federated` and its digest derivation are removed. What is left of the digest is `pre_v1_task_attribution_id_for_principal_claim`, used at login for one purpose: rekeying historical `tasks` rows from the old digest to the new `user_id`. It authorizes nothing.

## Usage examples

```
cargo nextest run -p coral-app first_login_owns_default_workspace_and_relogin_preserves_user_id
cargo nextest run -p coral-app issuer_mismatch_and_provisioning_failures_store_no_code
cargo nextest run -p coral-app login_rekeys_legacy_task_attribution_before_code_storage
cargo nextest run -p coral-app valid_session_selects_the_internal_user_principal
make postgres-tests
```

`issuer_mismatch_and_provisioning_failures_store_no_code` drives a database failure, an issuer mismatch, and a hung provisioning call through the real callback, and asserts each one returns `server_error` with no authorization code left behind.

---

Slice 3 of 16 in stack #2100. Every slice of this stack now passes `cargo fmt --check` and `cargo clippy --workspace --all-targets --all-features -- -D warnings` independently; the tip additionally passes `make rust-checks` (2,591 tests) and `make postgres-tests`.


sauldhernandez:

is worth carrying over.

PR #2089, opened 2026-08-06
feat(app): add explicit local-principal policy

sauldhernandez:
## Intent

Every workspace handler needs one policy object that decides whether a principal may read or manage a workspace without leaking the existence of workspaces the caller cannot access. The built-in local principal also needs an explicit deployment-mode policy: shared and embedded server configurations must reject it, while single-user mode may deliberately treat it as the owner.

## What it enables

- `LocalPrincipalPolicy::NoLocalPrincipal` is the safe default. It rejects an injected `coral:local` request with `PermissionDenied` before consulting user or membership state.
- `LocalPrincipalPolicy::ImplicitOwner` is an explicit single-user opt-in. It treats `coral:local` as owner without a membership lookup and lets that principal list users without owning a workspace.
- `WorkspaceAuthorizer::authorize(principal, workspace, action)` resolves `WorkspaceAction::{Read, Manage}` from the workspace-membership row for non-local callers:
  - under the default `NoLocalPrincipal` policy, a workspace with no non-local Owner remains concealed, even from an existing Member;
  - no membership returns `WorkspaceNotFound`, concealing the workspace from outsiders;
  - a Member may read but only an Owner may manage;
  - an agent principal is denied Manage before membership lookup, while an agent acting for a Member may still read.
- `UserManager` uses the same local-principal policy boundary. Its default rejects `coral:local` before user or membership lookup; only the explicit single-user opt-in grants the directory-listing exception.
- Tests create ordinary workspaces and memberships directly through the repositories. No personal or derived default workspace is involved.

This slice defines the domain policy and its enforcement checks; no request handler consults them yet. Later composition slices wire the policy into the serving path, choosing `ImplicitOwner` only for single-user CLI serve and retaining `NoLocalPrincipal` for shared, authenticated, and explicit-provider configurations.

## Usage examples

```console
cargo nextest run -p coral-app strict_policy_rejects_local_principal_before_database_lookup
cargo nextest run -p coral-app implicit_owner_policy_bypasses_database_lookup
cargo nextest run -p coral-app strict_policy_conceals_ownerless_workspace_from_member
cargo nextest run -p coral-app strict_policy_conceals_local_only_workspace_from_member
cargo nextest run -p coral-app owner_manages_while_member_only_reads
cargo nextest run -p coral-app missing_membership_is_concealed_as_workspace_not_found
```


sauldhernandez:

is worth carrying over.

PR #2090, opened 2026-08-06
feat(app)!: scope workspace listing to memberships and add member RPCs

sauldhernandez:
## Intent

Workspace discovery must be membership-scoped, explicit creation must record an Owner, and the membership lifecycle must be available through the internal gRPC contract. This slice adds those behaviors without inventing personal workspaces or reserving ordinary workspace names.

`ListWorkspacesResponse` replaces the legacy `workspaces` field with `memberships`. Each membership contains one authorized workspace and the caller's role. This is a breaking generated-client contract: clients must regenerate their bindings and consume `memberships`; no compatibility `workspaces` field ships.

## What it enables

- `ListWorkspaces` returns only the caller's role-aware `memberships`, scoped to memberships whose workspace has a non-local Owner under the default `NoLocalPrincipal` policy. Ownerless and `coral:local`-only-owned legacy workspaces remain concealed in shared mode.
- `CreateWorkspace` accepts every valid explicit name, including `default` and `default-*`, and atomically creates the workspace with the human creator as Owner.
- Explicit single-user creation under `ImplicitOwner` atomically ensures the reserved local user and creates its Owner membership. Shared/authenticated and explicit-provider configurations retain `NoLocalPrincipal` and never create that local row.
- `WorkspaceService` serves `ListWorkspaceMembers`, `AddWorkspaceMember`, and `RemoveWorkspaceMember`. Each scoped operation requires Manage authorization before accessing membership state.
- Membership changes remain repository operations. Repeated identical adds are idempotent, adding an existing member under a different role fails with `AlreadyExists`, and removing the last Owner is rejected.
- Role conversion rejects unspecified and unknown values at the transport edge. The CLI also rejects malformed membership responses that omit the nested workspace.

The serving composition selects `ImplicitOwner` only when neither session authentication nor an explicit principal provider is present. Later slices extend the same resolved policy to the rest of the service graph.

## Usage examples

```console
cargo nextest run -p coral-app workspace_manager_applies_local_principal_policy
cargo nextest run -p coral-app workspace_service_enforces_membership_lifecycle
cargo nextest run -p coral-app default_workspace_follows_the_regular_delete_lifecycle
cargo nextest run -p coral-app create_workspace_persists_database_row_without_config_scaffolding
cargo nextest run -p coral-cli --features cli-test-server workspace_list_rejects_membership_without_workspace
```


sauldhernandez:

is worth carrying over.

PR #2091, opened 2026-08-06
feat(app): serve users without a default workspace

sauldhernandez:
## Intent

A fresh deployment must begin without an invented workspace, and the internal user API must remain useful when a human has no memberships. This slice removes the implicit config-level `default` workspace and mounts the identity-only user service in the running gRPC graph. The affected CLI, MCP, and performance fixtures now create and select their workspace explicitly instead of relying on an implicit default.

## What it enables

- `WorkspaceCatalog::default()` is empty. Only explicitly created workspaces are rendered or returned, and a fresh local server lists zero workspaces.
- CLI and MCP fixtures create or select the workspace they exercise explicitly. Authenticated fresh-state probes verify transport access and an empty workspace-membership list without depending on a seeded default.
- The performance gate creates and selects an isolated `perf` workspace before installing GitHub and measuring `coral.tables`.
- The internal gRPC `UserService` is registered with the running server. `GetCurrentUser` returns only the authenticated human's stable ID and display name, including when that user has zero workspace memberships.
- `ListUsers` remains owner-scoped. The running-graph test first observes an empty membership list, explicitly creates an owned workspace, and then lists the directory.
- Bootstrap resolves `LocalPrincipalPolicy` once and applies the same value to both workspace and user behavior: session-authenticated and explicit-provider servers use `NoLocalPrincipal`, while the single-user local topology uses `ImplicitOwner`.
- `coral-client` exposes an internal user-directory client for later consumers. This does not add a CLI surface or Reef implementation.

## Usage examples

```console
cargo nextest run -p coral-app running_graph_serves_identity_and_owner_scoped_user_directory
cargo nextest run -p coral-app fresh_server_lists_no_workspaces
cargo nextest run -p coral-app renders_only_explicitly_created_empty_workspaces
cargo nextest run -p coral-app current_user_returns_identity_only_with_zero_workspace_memberships
make perf-check
```


sauldhernandez:

is worth carrying over.

PR #2092, opened 2026-08-06
feat(app): enforce authorization at service boundaries

sauldhernandez:
## Intent

Workspace-scoped services and deployment-global runtime-feature controls must enforce deployment policy before they validate request details or access state. The previous service shape made the workspace authorizer optional, while runtime feature control relied only on transport authentication; either gap could admit work before the required authorization decision.

## What it enables

- The server resolves one `WorkspaceAuthorizer` from `LocalPrincipalPolicy` and requires it when constructing catalog, query, search, function, source, task, and feedback services.
- Catalog, query, search, function listing, task, and feedback RPCs require `Read`.
- Function and search mutations require `Manage`.
- Every source RPC requires `Manage`, including discovery and configuration reads because their responses expose setup and credential metadata.
- Authorization runs immediately after workspace parsing and before attribution, validation, state access, streaming work, or mutation.
- Nonmembers receive the same `WorkspaceNotFound` response as callers naming a nonexistent workspace. Members can read but cannot manage; an Owner role cannot override the restriction on agent credentials.
- Runtime feature listing and mutation require `LocalHostControl`: exactly the built-in local principal under `ImplicitOwner`. Federated humans, agents, and a local principal under `NoLocalPrincipal` are rejected before configuration read, validation, or mutation, and a denied mutation leaves the configuration unchanged.

## Usage examples

```bash
cargo nextest run --locked -p coral-app authorization_service_tests
cargo nextest run --locked -p coral-app source_configuration_requires_manage
cargo nextest run --locked -p coral-app feature_control_requires_an_implicit_local_principal_before_state_access
```

The first matrix exercises every non-source workspace handler with Member, Owner, or nonmember principals and verifies both its assigned action and authorization-before-work ordering. The source matrix exercises all nine source RPCs. The feature-control test covers allowed local access and denial of federated humans, agents, and a local principal under `NoLocalPrincipal` before configuration access.


sauldhernandez:

is worth carrying over.

PR #2093, opened 2026-08-06
feat(app): restrict trace access to workspace owners

sauldhernandez:
## Intent

Traces contain query text, table names, and source names from the workspaces that produced them. An unrestricted trace read would bypass the workspace boundaries enforced by the preceding service slices, especially when a caller does not name a workspace.

## What it enables

- `TraceService` now requires the same bootstrap-resolved `WorkspaceAuthorizer` as the other workspace-scoped services.
- A named-workspace `ListTraces` or `GetTrace` call requires `Manage`, and authorization runs before page-token, view, or trace-ID validation.
- On shared servers, a human caller without a workspace sees traces only when they own every explicitly attributed workspace in that trace. Member-only workspaces, unowned workspaces, and host-only traces are concealed.
- In single-user mode, the built-in local principal retains unrestricted trace access, including host-only traces.
- Repository-backed keyset pagination enumerates only the caller's Owner memberships.
- List and detail reads apply the same trace-wide rule. Unattributed child spans may remain part of an otherwise authorized trace, but one explicitly attributed unowned span conceals the entire trace. This rule also protects the named-workspace Query Stream view before projection and pagination.
- Global owner-scoped Query Stream reads still require an explicit workspace; the later Query Stream isolation slice owns that remaining capability.

## Usage examples

```bash
cargo nextest run -p coral-app named_trace_calls_require_owner_manage_access
cargo nextest run -p coral-app global_trace_calls_page_owned_scope_and_keep_host_rows_local
cargo nextest run -p coral-app query_stream_view_projects_entries_for_an_authorized_workspace
cargo nextest run -p coral-app owned_workspace_filtered_traces_are_complete_scoped_and_globally_paged
make postgres-tests
```

The service tests cover Owner, Member, nonmember, agent, strict-local, and trusted-local callers. The store tests cover pagination, host-only rows, mixed-workspace traces, and the distinction between unattributed child spans and explicit conflicting attribution.


sauldhernandez:

is worth carrying over.

PR #2094, opened 2026-08-06
test(app): classify every served RPC exactly once

sauldhernandez:
## Intent

Authorization tests for individual handlers cannot catch a newly added RPC that nobody classifies. This slice adds a structural completeness guard for the service graph established by the preceding authorization slices.

## What it enables

- The integration test discovers every RPC declared under the canonical `coral/v1` protobuf directory and includes the two mounted gRPC health methods.
- One unified matrix classifies each served method as `Read`, `Manage`, `LocalHostControl`, `SelfScoped`, `AnyAuthenticatedHuman`, `OwnerOfAnyWorkspace`, or `Open`.
- Both runtime FeatureService methods — `coral.v1.FeatureService/ListFeatures` and `coral.v1.FeatureService/SetFeature` — are classified as `LocalHostControl`, matching their handler policy: only the built-in local principal under `ImplicitOwner` may inspect or change deployment-wide feature configuration.
- Duplicate RPC definitions, duplicate classifications, omitted methods, and stale entries fail independently.
- `Open` is restricted to the two mounted gRPC health methods, `Check` and `Watch`; no Coral RPC may use it.
- The parser test covers line and block comments, multiline declarations, nested RPC option bodies, and quoted strings.

The matrix is a completeness guard. The preceding service-authorization slice retains the direct behavioral tests that prove each handler applies its classification before validation or work.

## Usage examples

```bash
cargo nextest run -p coral-app --test authorization_matrix
cargo clippy -p coral-app --all-targets --all-features --locked -- -D warnings
```

Adding or removing a Coral v1 RPC requires updating the matrix. Otherwise the integration test reports the omitted or stale method.


sauldhernandez:

is worth carrying over.

PR #2095, opened 2026-08-06
feat(app)!: migrate local ownership and resolve local workspaces

sauldhernandez:
## Intent

Single-user upgrades need to assign the built-in local identity to legacy ownerless workspaces without creating a workspace or leaking that authority into shared mode. Local CLI and MCP commands also need deterministic workspace selection now that Coral no longer provisions or assumes a default workspace. MCP readiness must follow the app's health contract rather than recreate that removed assumption with a workspace-scoped catalog probe.

## What it enables

- Under `ImplicitOwner`, the one-time `local_workspace_ownership_v1` migration atomically records its marker, ensures the `coral:local` user, and assigns it as Owner only where the owner count is zero.
- A rollback releases the marker and removes every write the migration made, so startup can retry. Workspaces that already have a human owner remain unchanged.
- Under `NoLocalPrincipal`, startup never claims the migration or creates the local user. It reports ownerless and local-only-owned workspaces in separate warnings after telemetry is initialized, then continues serving.
- Local CLI, stdio MCP, and authentication-disabled loopback MCP use an explicit `--workspace` or non-empty `CORAL_WORKSPACE` value when supplied. Otherwise they require exactly one local workspace.
- Empty or ambiguous local state returns guidance to create or specify a workspace. There is no literal-`default` fallback.
- Workspace-free commands, including workspace management and source linting, bypass workspace resolution.
- Authentication-disabled MCP HTTP readiness now delegates to the app's engine-readiness contract instead of issuing `ListCatalog` against a fabricated `default` workspace. Failed or timed-out health checks report the endpoint unavailable.
- The affected CLI and MCP regression fixtures create and bind named workspaces explicitly, so the proof no longer depends on seeded default state.

Authenticated MCP workspace enforcement is intentionally left to the following stack slice; this PR does not add a shared-session fallback.

## Usage examples

```bash
# The sole local workspace is selected automatically.
coral workspace create analytics
coral sql "select * from coral.tables"

# Once multiple workspaces exist, select one explicitly.
coral workspace create finance
coral --workspace analytics sql "select * from coral.tables"
CORAL_WORKSPACE=analytics coral mcp-stdio

# The server passes the selected workspace to an enabled local MCP HTTP companion.
coral --workspace analytics server
```

BREAKING CHANGE: local commands that require a workspace no longer select or assume `default`. With zero workspaces, create one first. With multiple workspaces, pass `--workspace` or set `CORAL_WORKSPACE`.


sauldhernandez:

is worth carrying over.

PR #2096, opened 2026-08-06
feat(app)!: enforce shared MCP and query-stream workspace isolation

sauldhernandez:
## Intent

Shared MCP sessions must bind to one explicitly configured workspace the authenticated caller can actually access, without selecting a default, first membership, or substitute workspace. Global QueryStream reads must follow the same owner-scoped trace isolation as the rest of trace inspection, including mixed-workspace traces and pagination.

## What it enables

- Authenticated MCP initialization requires one configured workspace and lists memberships once through the bearer-bound client.
- Initialization succeeds only when the configured workspace exactly matches a valid returned membership. Missing configuration is actionable; inaccessible or nonexistent configuration is concealed as workspace not found.
- Shared MCP never calls `GetCurrentUser`, selects the first membership, synthesizes `default`, uses the local principal, or falls back to another accessible workspace.
- The established session remains pinned to the configured workspace, while each tool call still observes live authorization; removing membership makes the next call fail with structured `WORKSPACE_NOT_FOUND` instead of switching workspaces.
- QueryStream records the complete retained workspace set for each trace after duplicate precedence. Named reads require an exact singleton match; shared global reads allow only traces whose nonempty workspace set is a subset of the caller's owned workspaces.
- After duplicate-span precedence is resolved, owner-scoped filtering runs before pagination, so pages backfill past newer concealed traces. List and Get use the same workspace-set predicate.
- Callers with no owned workspaces receive an empty global List and NotFound from Get without scanning storage. Host-only traces remain concealed in shared mode.
- `ImplicitOwner` retains unrestricted local QueryStream visibility, including host-only traces. Shared mode remains `NoLocalPrincipal`.

This slice changes the shared MCP surface and its internal gRPC-backed enforcement. It adds no Reef implementation.

## Usage examples

```bash
# With [auth] and authenticated MCP HTTP configured, pin the shared MCP connection.
coral --workspace analytics server

# Equivalent selection via a non-empty CORAL_WORKSPACE value.
CORAL_WORKSPACE=analytics coral server

# Focused security evidence.
cargo test -p coral-app telemetry::local_store::query_stream --lib --all-features --locked
cargo test -p coral-mcp authenticated_sessions_require_exact_configured_workspace_membership --all-features --locked
```

BREAKING CHANGE: authenticated MCP sessions no longer initialize unless the configured workspace exactly matches one of the caller's memberships, and they never fall back to another workspace. Shared global QueryStream results are now restricted to traces attributed only to workspaces the caller owns.


sauldhernandez:

is worth carrying over.

PR #2097, opened 2026-08-06
feat(reef): carry workspace memberships through the app shell

sauldhernandez:
## Intent

Slice 6 changed `ListWorkspaces` to return memberships, and Reef's loader was adapted just far enough to keep compiling: it threw the role away immediately. This slice carries the membership through instead, from the loader to the sidebar, and tolerates a membership the server sent without a workspace resource rather than rendering an empty entry.

## What it enables

- `listWorkspacesForRequest` returns `WorkspaceMembership[]`. `firstWorkspaceForRequest` walks memberships and skips any whose `workspace` is unset, still answering 404 `No Coral workspace is configured.` when none carries one.
- The app-shell loader passes `memberships` to the sidebar rather than flattening them; the sidebar re-derives its workspace list from them, dropping memberships with no workspace. The settings route reads memberships and takes the workspaces out of them, because MCP clients are pointed at a workspace and have no use for the role.
- The role reaches the DOM as a stable name on each workspace-switcher entry (`data-workspace-role="owner" | "member" | "unspecified"`), so policy-aware styling and assertions can read it. Nothing is gated on it and no badge is rendered — this slice is plumbing, not visible behavior.
- No member-management UI ships, and a test asserts the absence rather than leaving it to be inferred: the workspace menu offers only "Create workspace" and "Settings", and the shell renders no invite, member, or manage-access affordance.
- `crates/coral-cli/tests/ownerless_workspace_live.rs` is an `--ignored` live proof over real `coral` processes: a shared deployment with two ownerless workspaces binds and announces its listener, conceals both from a signed-in person and from the host's own commands, and refuses to let either name be squatted by a create. Repair itself is deliberately not exercised — the admin tool lands in slice 15, and writing a membership row by hand would prove only that the test can write a row.

## Usage examples

```
npm test --prefix apps/reef
npm run typecheck --prefix apps/reef
cargo nextest run -p coral-cli

CORAL_BREAK_GLASS_EVIDENCE_PATH=/tmp/break-glass-live.md \
  cargo test -p coral-cli --test ownerless_workspace_live -- --ignored --nocapture
```

The live test points every process at a `TempDir` through `CORAL_CONFIG_DIR` with `HOME` redirected beside it, so no run can reach the operator's own state, and writes a redacted evidence document to the path above.

---

Slice 13 of 16 in stack #2100. Every slice of this stack now passes `cargo fmt --check` and `cargo clippy --workspace --all-targets --all-features -- -D warnings` independently; the tip additionally passes `make rust-checks` (2,591 tests) and `make postgres-tests`.


sauldhernandez:
Closing alongside the superseded implementation stack #2085–#2096 that this builds on.

Note for whoever picks this up: this Reef work is **not** replaced by the new stack
#2207–#2211 (GitHub stack #2212). That stack touches exactly one Reef file
(`apps/reef/app/lib/workspaces.server.ts`) and its docs state plainly that no Reef
member-management UI ships in it. The membership UI here (+764/-53 across 9 files) remains
the only implementation.

The branch is left intact. To revive it, rebase onto `main` or onto the new stack — its base
branch `pr-12-mcp-isolation-and-query-stream-scope` belongs to a now-closed PR.

PR #2098, opened 2026-08-06
docs: document workspace access control and owner recovery

sauldhernandez:
## Intent

Everything this stack introduces is invisible from the CLI, because there is no CLI for it. An operator needs the single-user versus shared distinction stated plainly, the exact error strings they will see, and the out-of-band route back from a workspace nobody owns. The tool that route uses lands in slice 15; this page is what points at it.

## What it enables

A `## Workspace access control` section in `apps/docs/reference/configuration.mdx` covering:

- **Single-user versus shared.** With no `[auth]`, everyone is `coral:local` and has full control. With `[auth]` present, there is no superuser at all: host commands such as `coral sql`, `coral workspace`, and `coral functions` never authenticate and run as `coral:local`, which reaches only what `coral:local` is a member of. A warning states what upgrading changes — "anyone logged in" becomes "members only".
- **What each role may do**, and the two errors: `workspace '<name>' not found` for a non-member, `permission denied: owner access is required for workspace '<name>'` for a member attempting an owner action.
- **Workspaces nobody owns.** Why the server starts anyway, how they arise both on upgrade and on a fresh install (the seeded `default` workspace belongs to nobody), and that their names cannot be squatted — a create answers `workspace '<name>' already exists`, and `coral workspace create` on a shared deployment answers `user 'coral:local' not found`.
- **Personal workspaces.** `default-<user_id>` created at first sign-in, and the reserved `default-` prefix with its rejection message. The shared workspace named exactly `default` does not carry the prefix and is unaffected.
- **Appointing an owner**, including `--state-dir` and the `CORAL_CONFIG_DIR` fallback, that the tool runs no migrations, that Postgres gives no locality guarantee, and the upgrade order: start the server, have people sign in once, then appoint owners. No restart is needed.
- **Managing members**, stated without hedging: this release ships no member-management UI and no membership subcommands. Members are managed over `WorkspaceService` gRPC — `AddWorkspaceMember`, `RemoveWorkspaceMember`, `ListWorkspaceMembers` — with JSON payload examples, pointing at `crates/coral-api/proto` because Coral does not serve gRPC reflection. Slice 16 revises this paragraph once adding an existing member moves their role instead of conflicting.
- **Rebinding the identity-provider issuer**, with the failure symptoms an operator would actually see first (`server_error`, `authorization failed`, and the log line `OIDC callback could not provision the authenticated user`) and a pre-rebind checklist.

## Usage examples

The page documents these four commands. All are repo-side and behind an optional feature, so the shipped binary gains nothing from them; the implementation lands in slice 15:

```
cargo run -p xtask --features admin -- workspace-admin list-workspaces
cargo run -p xtask --features admin -- workspace-admin list-users --show-subjects
cargo run -p xtask --features admin -- workspace-admin set-owner --workspace analytics --user <user-id>
cargo run -p xtask --features admin -- workspace-admin rebind-issuer --from <old-issuer> --to <new-issuer>
```

---

Slice 14 of 16 in stack #2100. Every slice of this stack now passes `cargo fmt --check` and `cargo clippy --workspace --all-targets --all-features -- -D warnings` independently; the tip additionally passes `make rust-checks` (2,591 tests) and `make postgres-tests`.

### DIFF
--- a/apps/docs/reference/configuration.mdx
+++ b/apps/docs/reference/configuration.mdx
@@ -96,6 +96,260 @@ bind = "127.0.0.1:14556"
 
 The unauthenticated MCP listener is restricted to loopback addresses.
 
+## Workspace access control
+
+What a caller may do in a workspace follows from their membership in it. One
+state directory runs in one of two modes:
+
+- **Single-user.** `config.toml` has no `[auth]` section. Every request runs as
+  Coral's built-in local principal, `coral:local`, which reaches every workspace
+  with full control.
+- **Multi-user.** `config.toml` has an `[auth]` section, so `coral server`
+  authenticates callers against your identity provider and each person reaches
+  only the workspaces they belong to.
+
+A workspace member and a workspace owner see different surfaces:
+
+| Action | Member | Owner |
+| ------ | ------ | ----- |
+| Query, search, browse the catalog, and call functions | Yes | Yes |
+| Manage sources and functions, and read traces | No | Yes |
+| Add and remove members, and delete the workspace | No | Yes |
+
+Someone with no membership at all gets the same `workspace '<name>' not found`
+answer as for a workspace that does not exist, so non-members cannot tell that
+it exists. A member who attempts an owner-only action gets
+`permission denied: owner access is required for workspace '<name>'`.
+
+Commands you run on the host — `coral sql`, `coral workspace`, `coral functions`
+— do not authenticate and always run as `coral:local`, in either mode. What that
+principal can reach is not the same in both. In single-user mode it owns every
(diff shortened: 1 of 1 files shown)


sauldhernandez:
Superseded by #2211, whose docs pass rewrote `apps/docs/reference/configuration.mdx` against the
shipped behaviour — including the `server.mcp_http.workspace` key, the auth-disabled
serve-startup resolution (and why creating a second workspace makes the next restart ambiguous),
and the recovery runbook. Every claim there was traced to code on the branch rather than to the
program design.

The branch is left intact if anything here is worth carrying over.

PR #2099, opened 2026-08-06
feat(xtask): add the workspace ownership recovery tool

sauldhernandez:
## Intent

On a shared deployment there is no principal that could appoint an owner over an RPC — that is exactly what having no superuser means. So recovery has to live outside the product, and it has to stay outside: shipping it inside the `coral` binary would hand every operator of every deployment a standing bypass of the policy the rest of this stack builds. Slice 14 documented these commands; this slice implements them.

## What it enables

`crates/coral-app/src/admin.rs`, behind coral-app's `admin` feature, which the shipped binary leaves off. The only place in the workspace that turns it on is `xtask`, and `xtask`'s own `admin` feature is off by default so the docs tooling and the release workflow keep building `-p xtask` without pulling `coral-app` in.

- `AdminDb::open` reads the same state database the server reads, resolving `[database]` the way the server does, and never runs migrations. It authenticates nobody and authorizes nothing — possessing it is equivalent to possessing the state database itself, which is why it is not a product surface.
- `list_workspaces` returns each workspace with its owner and member counts, which is the only way an ownerless workspace becomes visible; concealment hides it from every RPC by design.
- `list_users` returns internal user id, display name, issuer, and subject, withholding subjects unless asked, because subjects identify people and are frequently email addresses.
- `set_owner` makes a user an owner in one transaction, reporting whether they were added, promoted from member, or already an owner. It only ever adds, never removes, so it cannot demote anyone by accident, and repeating it changes nothing.
- `rebind_issuer` moves every identity from one issuer to another and returns the row count.

`xtask/src/admin.rs` exposes all four, renders the output an operator reads — ownerless workspaces are flagged inline — and turns a workspace or user that does not exist into a non-zero exit with a message naming what to run next, rather than an error.

## Usage examples

```
cargo run -p xtask --features admin -- workspace-admin list-workspaces
cargo run -p xtask --features admin -- workspace-admin list-users --show-subjects
cargo run -p xtask --features admin -- workspace-admin set-owner --workspace analytics --user <user-id>
cargo run -p xtask --features admin -- workspace-admin rebind-issuer --from <old-issuer> --to <new-issuer>
```

`--state-dir <path>` is global and may appear before or after the subcommand; without it the tool falls back to `CORAL_CONFIG_DIR`.

```
cargo nextest run -p coral-app --features admin
cargo nextest run -p xtask --features admin
```

---

Slice 15 of 16 in stack #2100. Every slice of this stack now passes `cargo fmt --check` and `cargo clippy --workspace --all-targets --all-features -- -D warnings` independently; the tip additionally passes `make rust-checks` (2,591 tests) and `make postgres-tests`.


sauldhernandez:
Superseded by #2211, which ships the same recovery commands behind an off-by-default `admin`
feature, with mutation-verified non-migration (pinned by a behind-schema fixture, because a
snapshot against an already-current database cannot detect a wired-in migrator), five Postgres
contracts wired into `make postgres-tests`, and a default-surface test wired into
`make rust-checks`.

Like #2177, this PR put the recovery logic in `crates/coral-app/src/admin.rs` behind
`#[cfg(feature = "admin")]` so it could reuse coral-app's repositories, rather than keeping the
shipped crate free of admin code as #2211 does. See #2177's closing comment for that trade-off
and the follow-up it implies; the branch is left intact.

PR #2106, opened 2026-08-07
refactor(search): split sqlite_store into store, migration, and tests

mystic123:
## What

Pure movement. No behaviour change, no logic edited.

`search/sqlite_store.rs` had grown to 2,601 lines carrying three unrelated
things:

- the store's permanent API, capability detection, and schema validation
- a one-time v4→v5 transform that becomes dead weight once every workspace has
  upgraded
- a 1,380-line inline test module

This splits them along the seam that already existed:

| File | |
|---|---|
| `sqlite_store.rs` | 2,601 → **948** |
| `sqlite_store/observed_identity_migration.rs` | **309** (new) |
| `sqlite_store/tests.rs` | **1,380** (new) |

The `tests.rs` shape matches the sibling modules that already do this —
`observed/sqlite_store/tests.rs` and `catalog/sqlite_index/tests.rs`.

## Verified as a pure move, not asserted

Comparing the old single file against the concatenation of the three new ones:

- **declared items:** none lost, none gained
- **`#[test]` functions:** 31 before, 31 after
- **SQL statements:** 120 before, 120 after

The remaining textual differences are rustfmt reflowing (dedenting the test
module by one level lets lines fit that previously wrapped) plus the module
declaration and `pub(super)` markers.

## Visibility barely moves

A child module can read its parent's private items through `super::`, so the
transform reaches `schema_query_is_valid`, `tables_exist`, `apply_migration`,
and `SEARCH_SQLITE_MIGRATIONS` with no widening at all. Only the handful of
items the parent calls back into are marked `pub(super)`.

## One trap worth knowing

`include_str!` paths are file-relative, so the test module's `migrations/...`
references become `../migrations/...` when it moves down a directory — and
rustfmt wraps those lines, which hides them from a naive search. The compiler
catches it immediately, but it means this cannot be done blind.

## Why it is its own PR

It was briefly folded into #2038 and made that PR read +2422/−1583 instead of
+1468/−655 — ~950 lines of relocation on top of a migration that needed real
review. Separating it keeps #2038 at the shape it was reviewed at, and makes
this reviewable as a single question: *did anything change?*

It is also the right place to argue about how these test modules should be
organised, rather than having that discussion attached to a catalog change.


mystic123:
* **#2106** <a href="https://app.graphite.com/github/pr/withcoral/coral/2106?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/2106?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Bradley-Butcher (approved):

mystic123:
Valid observation, but fixed in #2038 rather than here.

`copy_preserved_observed_values` and `copy_preserved_observed_queue_jobs` are now a single `copy_preserved_rows(transaction, legacy_table, target_table, columns)`, with a doc note that every argument is a constant defined in that file — none reaches the interpolation from a request, a manifest, or a stored row.

Doing it **here** would have been wrong: this PR's only claim is that it moves code without changing it, verified by identical declared items, `#[test]` count, and SQL-statement count before and after. Merging two functions breaks all three of those checks and turns a thirty-second review into a real one. The duplication was introduced in #2038, so that is where it is fixed — and #2038 had not been reviewed yet, so the change cost nothing there.

#2106 has been rebuilt on top of the deduplicated base and re-verified: items lost/gained none, 31 tests before and after, 120 SQL statements before and after.

mystic123:
### Merge activity

* **Aug 18, 1:09 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/withcoral/coral/2106).
* **Aug 18, 1:11 PM UTC**: [Graphite](https://app.graphite.com/github/pr/withcoral/coral/2106) rebased this pull request as part of a merge.
* **Aug 18, 1:17 PM UTC**: @mystic123 merged this pull request with [Graphite](https://app.graphite.com/github/pr/withcoral/coral/2106).

PR #2107, opened 2026-08-07
feat(app)!: move an existing member's role on add

sauldhernandez:
## Intent

`AddWorkspaceMember` used to refuse when the user was already a member holding a different role, so changing someone's role meant removing them and adding them back: two calls, with a window in between where they had no access at all, and a sequence that cannot promote anyone in a workspace with a single owner, because the removal is refused first.

**Breaking change.** `AddWorkspaceMember` with a role different from the member's current one now succeeds and returns the moved membership, where it previously failed with `AlreadyExists` and `user '<user-id>' already has a different role in workspace '<workspace>'`. `AppError::WorkspaceMemberRoleConflict` and its telemetry reason are removed. Adding a member with the role they already hold remains a no-op success.

## What it enables

- Promotion and demotion in one authorized transaction. The mutation reads the current role and then `UPDATE`s it, serialized by the same parent-workspace-row hold every other membership mutation takes — not a SQL upsert. Adding an existing member states the role they should hold, and moves them to it.
- The owner floor now covers demotion as well as removal: demoting the last owner is refused with `AppError::LastWorkspaceOwner` — `workspace '<name>' must retain at least one owner` — the same error, and the same gRPC `FailedPrecondition`, that removing them gives. It is the one move that cannot be honored, because it would leave nobody in charge.
- The configuration reference is corrected to describe the move instead of the old remove-and-re-add workaround.
- In-crate service fixtures that model a single-user deployment now say so, constructing `WorkspaceAuthorizer::trusting_local_principal` rather than the strict constructor. Previously they read as if the strict policy admitted `coral:local`, which it does not.

## Usage examples

```
cargo nextest run -p coral-app add_workspace_member_enforces_authorization_and_frozen_add_semantics
cargo nextest run -p coral-app membership_lifecycle_filters_listing_and_preserves_the_last_owner
cargo nextest run -p coral-app workspace_member_state_round_trips_against_sqlite
make postgres-tests
```

`add_workspace_member_enforces_authorization_and_frozen_add_semantics` promotes a member to owner and demotes them again through the real handler, and asserts a non-member and a plain member are stopped before either write.

---

Slice 16 of 16 in stack #2100. Every slice of this stack now passes `cargo fmt --check` and `cargo clippy --workspace --all-targets --all-features -- -D warnings` independently; the tip additionally passes `make rust-checks` (2,591 tests) and `make postgres-tests`.

### DIFF
--- a/apps/docs/reference/configuration.mdx
+++ b/apps/docs/reference/configuration.mdx
@@ -294,11 +294,12 @@ workspace an earlier single-user run left it owning, once a person owns it too:
 `ListWorkspaceMembers` takes `{"workspace": {"name": "analytics"}}` and is how
 an owner verifies either change.
 
-Repeating an identical add succeeds and returns the existing membership. Asking
-for a different role never overwrites the current one; it answers
-`user '<user-id>' already has a different role in workspace 'analytics'`, so
-remove the membership and add it again. Removing the only owner is refused with
-`workspace 'analytics' must retain at least one owner` — which is also why a
+Repeating an identical add succeeds and returns the existing membership. Adding
+someone who already belongs states the role they should hold, so asking for a
+different one moves them to it — that is how you promote a member to owner or
+step an owner back down. The one move Coral refuses is the one that would leave
+the workspace with nobody in charge: demoting or removing the only owner answers
+`workspace 'analytics' must retain at least one owner`, which is also why a
 workspace never becomes ownerless on its own.
 
 ### Rebind the identity-provider issuer
--- a/crates/coral-app/src/admin.rs
+++ b/crates/coral-app/src/admin.rs
@@ -234,7 +234,7 @@ impl AdminDb {
 
         let outcome = match tx
             .workspace_members()
-            .role_for_user_id(workspace, user_id)
+            .role_for_user_id_for_recovery(workspace, user_id)
             .await
             .map_err(describe)?
         {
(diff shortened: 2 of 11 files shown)


sauldhernandez:
Superseded by #2211's stack (#2207-#2211, GitHub stack #2212), which already implements this
same behaviour — verified rather than assumed:

- `AddMemberOutcome::RoleUpdated` in `crates/coral-app/src/state/db/workspace_state.rs`: adding
  an existing member at a different role moves them, the same design this PR takes.
- `AddMemberOutcome::LastOwnerProtected` guards demotion of a workspace's last owner
  (`workspace_state.rs:218`), and `RemoveMemberOutcome::LastOwnerProtected` guards removal.
- Wired through `workspaces/manager.rs:402`, surfacing `AppError::LastWorkspaceOwner` on refusal.
- Covered by tests in `workspace_state.rs` (promotion, demotion, last-owner protection).
- The proto already documents it: "Adds one person to one workspace, or updates the role they
  already hold."

Feature-identical down to the enum variant names. Closing rather than rebasing: both PRs are
written against a `crates/coral-app/src/state/db/workspace_member_state.rs` that the new stack
does not have (membership state lives in `workspace_state.rs`), so this would be a
hand-reimplementation producing code that already exists, not a rebase.

The branch is left intact.

PR #2111, opened 2026-08-10
feat(reef): add Windows MCP setup commands

lazybean:
## Summary
- makes the web MCP setup command detect Windows via client hints with a User-Agent fallback
- emits PowerShell-compatible direct `add-mcp` commands using the resolved `coral` executable
- preserves workspace selection and quotes the selected workspace correctly for PowerShell

## Validation
- targeted Vitest coverage
- `npm run typecheck`
- `npm run build`
- targeted oxfmt and oxlint

PR #2110, opened 2026-08-10
refactor(reef): separate Settings loader and view

lazybean:
## Summary
- Move Settings presentation into `views/settings`.
- Keep `routes/settings.tsx` as a thin React Router framework-mode adapter.
- Load Desktop MCP clients through a hydrated `clientLoader` and apply workspace changes through a `clientAction`.

## Stack
Base refactor for #2112 and #2111. It preserves the existing Desktop-only Settings behavior while moving its data lifecycle into the route.

## Validation
- `npm run check --prefix apps/reef`
- `npm run typecheck --prefix apps/reef`
- `npm test --prefix apps/reef`
- `npm run build --prefix apps/reef`

PR #2112, opened 2026-08-10
feat(reef): configure web MCP clients

lazybean:
## Summary
- adds web Settings commands for the supported MCP-client catalog
- configures remote HTTP MCP through pinned `add-mcp`, with Connector guidance for Claude Desktop
- configures local Coral through the installed `coral` executable and supports choosing a workspace per client
- removes the hosted curl/PowerShell installer routes; configuration is now a transparent copied command
- notes that remote endpoints choose their default workspace

## Validation
- targeted Vitest coverage
- `npm run typecheck`
- `npm run build`
- targeted oxfmt and oxlint

`npm run check` remains blocked by pre-existing unformatted `.omx` runtime files.


lazybean:
* **#2111** <a href="https://app.graphite.com/github/pr/withcoral/coral/2111?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2112** <a href="https://app.graphite.com/github/pr/withcoral/coral/2112?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/2112?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2114** <a href="https://app.graphite.com/github/pr/withcoral/coral/2114?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2110** <a href="https://app.graphite.com/github/pr/withcoral/coral/2110?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lazybean:
This is wrong. We should use the http mcp. It is not necessary local.

apps/reef/app/components/mcp-clients-list/mcp-client-install-list.tsx
```diff
@@ -0,0 +1,59 @@
...
+                  {'installCommand' in client ? (
+                    <div className={styles.installCommand}>
+                      <code>{client.installCommand}</code>
+                      <Button.CopyButton
+                        ariaLabel={`Copy the Coral install command for ${client.name}`}
+                        textToCopy={client.installCommand}
+                      >
+                        Copy
+                      </Button.CopyButton>
+                    </div>
```
AlbertQM: Instead of each row having an always visible secondary button, wdyt of making this bare, icon only and appear only on hover? (3 suggestions in order of how much I think it makes the UI sleeker)

Image

Something like this:

https://github.com/user-attachments/assets/4fea7435-5f24-40ad-bc21-aca530e5ae5f

apps/reef/app/lib/mcp-clients.test.ts
```diff
@@ -0,0 +1,19 @@
...
+    for (const client of webMcpClients) {
+      expect(mcpClientById(client.id)).toEqual(client)
+      expect(mcpClientInstallPath(client.id)).toBe(`/mcp/install/${client.id}`)
+    }
+  })
+
+  it('does not resolve unknown clients into an install route', () => {
+    expect(mcpClientById('not-a-client')).toBeUndefined()
+  })
+})
```
AlbertQM: is this testing anything useful that isn't already covered by TS types?

apps/reef/app/lib/mcp-installer-script.server.ts
```diff
@@ -0,0 +1,130 @@
...
+
+const ADD_MCP_VERSION = '1.11.0'
+
+function installerNodeScript(connection: McpConnection): string {
+  const server =
+    connection.mode === 'remote'
+      ? JSON.stringify({ type: 'http', url: connection.url })
+      : `{ args: ['mcp-stdio'], command: coralBin }`
+
+  return `import { agents, getAgentTypes, listInstalledServers, upsertServer } from 'add-mcp'
```
AlbertQM: 🤯

apps/reef/app/views/settings/settings.tsx
```diff
@@ -31,17 +35,22 @@ export function Settings({
       </header>
 
       <Banner>
-        This page shows only global MCP configurations. Project-specific and other connections will
-        not appear here.
+        {desktop
+          ? 'This page shows only global MCP configurations. Project-specific and other connections will not appear here.'
+          : 'Each command updates one user-global MCP client configuration. It will not replace an existing Coral entry, and project configuration may override it.'}
```
AlbertQM: If the logic that we run is the same, I'd expect the message to be more similar to the already existing desktop one.


```suggestion
          : This settings affect only global MCP configurations. Project-specific and other connections will not be replaced.'}
```

apps/reef/app/views/settings/settings.tsx
```diff
@@ -10,17 +10,21 @@ export function Settings({
...
+
   return (
     <section className={styles.section}>
       <header className={styles.sectionHeader}>
         <Typography.HeadingLarge as="h1">MCP Clients</Typography.HeadingLarge>
         <Typography.Body variant="secondary">
-          Choose the Coral workspace each MCP client can access.{' '}
+          {desktop
+            ? 'Choose the Coral workspace each MCP client can access. '
+            : 'Copy a command to add Coral to a supported MCP client in your user-wide configuration. '}
```
AlbertQM, line 27: Hmmm.. What workspace do we actually connect to? No args, so we connect to the default specified by the user?

AlbertQM (approved):
Thank you, this is cool!

On Desktop, we let users choose which workspace to connect, on web, it doesn't seem to be the case. I think we should also be explicit there (either allow selection, or tell users which workspace it'll be connecting to).

Rest of the comments are just nits 😄

apps/reef/app/views/settings/settings.tsx
lazybean: Default. Soon we will be able to choose

PR #2115, opened 2026-08-10
feat(desktop): download updates only when asked

AlbertQM:
## Summary

Currently, opening the desktop app when there's a new version available initiates the download immediately.
This PR changes that: you'll have to click the download button to initiate the download.

## Test plan

npm run storybook

image

image

image


apps/reef/app/components/desktop-update-indicator/desktop-update-indicator.tsx
```diff
@@ -63,9 +71,36 @@ export function DesktopUpdateIndicator({ isMinimized, state }: DesktopUpdateIndi
...
+        {content}
+      </div>
+    ) : (
+      <button
+        aria-atomic="true"
+        aria-label={accessibleLabel}
+        aria-live="polite"
+        className={className}
+        disabled={isPending}
+        onClick={state.status === 'available' ? onDownload : onInstall}
```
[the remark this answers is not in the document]
AlbertQM: We don't do those anymore for now, after the cleanup

apps/desktop/src/main/index.ts
```diff
@@ -267,9 +268,18 @@ function registerIpcHandlers() {
...
   ipcMain.handle('coral:get-mcp-launch-config', () => getMcpLaunchConfig())
   ipcMain.handle('coral:get-update-state', () => getDesktopUpdateState())
+  ipcMain.handle('coral:download-update', () => downloadDesktopUpdate())
+  ipcMain.handle('coral:install-update', () => installReadyUpdateNow())
   onDesktopUpdateStateChange(publishDesktopUpdateState)
 }
 
+// Quit rather than calling quitAndInstall() directly: the shutdown coordinator
+// stops the sidecar first and only then hands the staged update to Squirrel.
+function installReadyUpdateNow(): void {
```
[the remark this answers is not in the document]
AlbertQM: Nah we're good for now

apps/reef/app/components/desktop-update-indicator/use-desktop-update-state.ts
```diff
@@ -40,5 +52,47 @@ export function useDesktopUpdateState(enabled: boolean): DesktopUpdateState {
...
+    request((desktop) => desktop.downloadUpdate(), 'Couldn’t download the update')
+  }, [request])
+
+  // Installing quits Coral, so the toast only ever reaches the user when the
+  // hand-off refuses to start. A resolved request means the quit was accepted
+  // and teardown is under way, which the window outlives — stay disabled rather
+  // than inviting a second click at a button that is about to disappear.
+  const onInstall = useCallback(() => {
+    request((desktop) => desktop.installUpdate(), 'Couldn’t install the update', {
+      resetOnSuccess: false,
```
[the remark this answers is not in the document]
AlbertQM: Nah we're good

apps/reef/app/components/desktop-update-indicator/desktop-update-indicator.tsx
```diff
@@ -63,9 +71,36 @@ export function DesktopUpdateIndicator({ isMinimized, state }: DesktopUpdateIndi
...
+        aria-atomic="true"
+        aria-label={accessibleLabel}
+        className={className}
+        role="status"
+        tabIndex={isMinimized ? 0 : undefined}
+      >
+        {content}
+      </div>
+    ) : (
+      <button
```
lazybean, line 91: :question:Why not a wax button?  
Especially the green one look the same but not the same.

apps/reef/app/components/desktop-update-indicator/use-desktop-update-state.ts
```diff
@@ -40,5 +52,47 @@ export function useDesktopUpdateState(enabled: boolean): DesktopUpdateState {
     }
```
lazybean: <p class='graphite__hidden'><i>[Re: lines +15 to +47]</i></p>
Added via Giphy
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/withcoral/coral/2115?utm_source=unchanged-line-comment">Graphite</a>.</p>

lazybean: data management + request invinting itself in component + useEffect.   
:no_entry:

lazybean (changes requested):
Sending back to you for visibility.  
  
Few questions/concerns about:  
\- non-wax button  
\- custom data fetching handling in hook VS action  
  
Any reasons for those?

apps/reef/app/components/desktop-update-indicator/desktop-update-indicator.tsx
AlbertQM: Good question! A button should have at most 1 line, here I wanted to show a bit more content than that. Our primary button is also quite attention grabbing (like it should), and I didn't want to have that sit in the sidebar all the time

If we switched to a primary button, how would you design the 3 states?

image

Would you make this more subtle (e.g. bare button that just updates the icon and text), or less subtle (e.g. a primary button for all the states?)

AlbertQM:
@cubic-dev-ai re-review pls

AlbertQM:
> Sending back to you for visibility.
> 
> Few questions/concerns about: - non-wax button - custom data fetching handling in hook VS action
> 
> Any reasons for those?

The UI was intentional (replied in thread), the client fetching wasn't -- added an architecture test to try remove that class of mistakes altogether #2180. Also updated to use an action, thanks!

apps/reef/app/components/desktop-update-indicator/use-desktop-update-state.ts
AlbertQM: 🔥🐞

apps/reef/app/components/desktop-update-indicator/desktop-update-indicator.tsx
lazybean: Thank you for the clarification and confirmation it was on purpose.. Our button has a pending state, but it may not be as nice.  
  
Then I may argue it should not be one button but a `banner` with a button?  
I would need to try out and see how it looks.  
  
If you are satisfied with how that look I am ok as well.

lazybean (approved):

apps/reef/app/components/desktop-update-indicator/desktop-update-indicator.tsx
AlbertQM: >Then I may argue it should not be one button but a banner with a button?
I would need to try out and see how it looks.

Yeah, that can work! My original idea was similar: it would look the same as it does now, but the clickable part just just be the "Coral 0.9.0".
Perhaps you're thinking more of our banner, but with one of our buttons on the right? I think that's even better/more in line with what we have currently. I think we'd need some CSS tweaks too since we have little horizontal space and our banners currently span wide screens in the UI

I can play with it and see how it feels, thank you!

I'll land this so that we stop auto-download+update Coral for everyone, can always adjust UI later 🙏

PR #2117, opened 2026-08-11
feat(reef): suggest importable sources in the catalog

AlbertQM:
## Summary

With DSL v4, we can support OpenAPI specs out of the box; this PR adds a bunch of presets (ie urls to dsl v4 compatible specs), so that users can add those with one click

## Test plan

image
image

https://github.com/user-attachments/assets/c07fc0a8-c57a-4eed-b86e-edd42859e8ea


lazybean (approved):
Good idea! Thank you!  
  
Did you try them all? (at least by checking the schema explorer)

PR #2123, opened 2026-08-11
fix(spec): read OpenAPI type arrays when choosing a schema's shape

simonwhitaker:
Part of #1321. Version-agnostic correctness fix, and the prerequisite that makes 3.1 nullability work.

## Why

Both shape dispatches read `type` with `as_str`, so a schema listing several types came back as `None` and fell through to a default of `"object"`. That default is right often enough to hide the fault: a nullable object is still an object. A nullable collection is not.

`{"type": ["array", "null"]}` was therefore classified as a singleton object rather than a list, so an endpoint returning an optional collection imported as one opaque row and lost every item. The same read in `import_schema` sent a nullable array property down the object branch, where it matched neither `properties` nor `additionalProperties` and collapsed to opaque JSON — losing the element type of every optional list an API returns.

## What changes

Both dispatches now match through `json_schema_type_contains`, which already handled the string and array forms and was already used elsewhere in this module. "Declared no type" stays distinct from "declared a type with no shape here" via a new `json_schema_has_declared_type`, so a typeless schema still reads as an object.

This is version-agnostic — a 3.0 document may carry a type array, and this is how it was always meant to be read — but 3.1 is where it stops being an edge case, since dropping `nullable` makes the type array the only way to spell a nullable collection.

Bumps the importer version: a 3.0 document with a nullable collection now produces different artifacts and needs re-materializing.

## Review follow-up

`classify_response_schema` tested `array` before `object`, so a schema claiming both was read as a list here while `import_schema` built its row type as an object. That also changed behaviour this branch did not mean to change: read as a string, such a schema used to fall through to the typeless default and classify as a singleton. Object is now tested first, in the same order `import_schema` tests it.

The nullable-collection regression test asserted only the cardinality, so it would have passed had the wrong item schema reached `import_schema` and typed every row as opaque JSON — the other half of the same data loss. It now asserts the row type too.

---

### DIFF
--- a/crates/coral-spec/src/v4/surfaces/openapi/responses.rs
+++ b/crates/coral-spec/src/v4/surfaces/openapi/responses.rs
@@ -5,7 +5,9 @@ use serde_json::{Map, Value};
 use crate::ResponseSpec;
 use crate::v4::diagnostics::Diagnostic;
 use crate::v4::ir::{IrOperationOutput, OutputCardinality, RestResponseAttachment};
-use crate::v4::surfaces::json_schema::{json_schema_has_declared_type, json_schema_type_contains};
+use crate::v4::surfaces::json_schema::{
+    json_schema_declares_only_type, json_schema_has_declared_type, json_schema_type_contains,
+};
 
 use super::import::OpenApiImporter;
 
@@ -255,7 +257,10 @@ fn classify_response_schema(
                 .or_else(|| Some(entity_name_from_path(path))),
         );
     }
-    if json_schema_type_contains(schema, "array") {
+    // The sole declared type, matching `import_schema`: a union that only
+    // includes `array` is not a collection, and reading one as a list would
+    // claim a row count for responses that are a single string.
+    if json_schema_declares_only_type(schema, "array") {
         let item = schema.get("items").cloned().unwrap_or(Value::Null);
         return (
             OutputCardinality::List,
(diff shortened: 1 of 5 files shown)


crates/coral-spec/src/v4/surfaces/openapi/responses.rs
```diff
@@ -263,6 +260,16 @@ fn classify_response_schema(
                 .or_else(|| Some(entity_name_from_path(path))),
         );
     }
+    if json_schema_type_contains(schema, "array") {
```
sauldhernandez: `type: [array, string]` is a real union, not a nullable array. This branch always imports `items`, so a valid string response becomes one row of item-derived columns whose values are all NULL. Please select `List` only when `array` is the sole non-null type; otherwise keep the response `Unknown` and import it as JSON. A shared helper with `import_schema` can keep both dispatches aligned.

sauldhernandez (approved):

crates/coral-spec/src/v4/surfaces/openapi/responses.rs
simonwhitaker: Fixed in dd83542.

Both dispatches now go through one helper, `json_schema_declares_only_type`, which asks whether `array` is the sole declared type once `null` is discounted — so `[array, 'null']` stays a `List` and `[array, string]` falls through to `Unknown` / `IrTypeShape::Json`. Covered by `importer_reads_a_union_of_a_collection_and_a_scalar_as_json`, which asserts both halves: the cardinality *and* that no `id` column leaks in from `items`.

One deliberate asymmetry worth your call: I applied the sole-type rule to `array` only, so `[object, string]` is still imported as an object rather than declining to JSON. The reasoning is that `object` is already how a schema declaring *no* type is imported, so a union including it degrades to the same columns — possibly null — whereas choosing `List` makes a claim about how many rows the response has, and getting that wrong restructures the whole relation. That also keeps `[object, array]` answering `Singleton` in both dispatches, which is what the earlier thread on this file settled on. Happy to tighten `object` the same way if you'd rather both be strict.

PR #2124, opened 2026-08-11
feat(spec): import OpenAPI 3.1 documents

### LINKED ISSUE #1321: Add OpenAPI 3.1 support

simonwhitaker:
Part of #1321. This is the PR that actually accepts 3.1.

## Why

DSL v4 rejected 3.1 outright, so a document Coral could otherwise hydrate and discover was turned away at import.

## What changes

`OpenApi31Importer` joins the 3.0 dialect and the version dispatch routes to it.

3.1 realigned its schema object with JSON Schema 2020-12 rather than carrying its own dialect of it, and most of that costs the importer nothing: `$defs` resolves through the same JSON Pointer walk as any other local reference, and the keywords 3.1 gained for binary payloads only matter to file uploads, which Coral does not support. Two spellings do need reading:

- **Nullability.** 3.1 dropped `nullable` for JSON Schema's `null` type, so the dialect reads it out of the type array. The two versions genuinely disagree here rather than one being a superset, and the tests pin both directions: a 3.0 document does not get nullability from a type array, and a 3.1 one does not get it from the keyword.
- **`const`**, which 3.1 documents use to pin a discriminator where a 3.0 one wrote a one-element `enum`. It imports as the enum it is equivalent to.

A 3.1 document still carrying `nullable` now gets a diagnostic. It is usually a 3.0 document whose version string was raised without converting its schemas, and 3.1 gives the keyword no meaning — but dropping it in silence turns every one of those columns non-nullable with nothing said. Only `nullable: true` is reported; `nullable: false` asked for the default under 3.0 as well, and real documents carry leftover `false`s (GitHub's own 3.1 publication has five, and not one `true`).

3.2 stays rejected. It was ratified in April 2026 and is not yet widely adopted.

## Verification

The GitHub fixture is now parameterized by version and imported under both. A document using none of the contested keywords produces identical operations, types, and metadata either way — the claim the dialect split rests on: `$ref` resolution, response selection, wrapped-list row paths, and pagination detection stay one traversal.

## Review follow-up

`const_enum_values` fired for any schema carrying `const`, whatever it held. `const` may hold any JSON value, so `{type: object, properties: {...}, const: {...}}` — legal 3.1 — imported as a one-value enum whose single value was the stringified object, standing where the declared fields should have been and dropping every column. Only string, number and boolean constants are read as enums now; the rest fall through to the shape dispatch. Scalars are still stringified exactly as `enum` members already are, so `const: 4` and `enum: [4]` agree.

The cross-version test compared operation and type ids, which would have held just as well if 3.1 gave every one of those types a different shape. It now compares the whole serialized semantic IR.

---

### DIFF
--- a/crates/coral-spec/src/v4/surfaces/openapi/import.rs
+++ b/crates/coral-spec/src/v4/surfaces/openapi/import.rs
@@ -13,6 +13,7 @@ use crate::v4::{
 use crate::{ManifestError, Result};
 
 use super::dialect::OpenApiDialect;
+use super::json_schema_dialect::validate_json_schema_dialect;
 use super::v3_0::OpenApi30Importer;
 use super::v3_1::OpenApi31Importer;
 use super::version::{OpenApiVersion, parse_openapi_version};
@@ -26,7 +27,13 @@ pub fn import_openapi_surface(
         serde_yaml::from_slice(document_bytes).map_err(ManifestError::parse_yaml)?;
     let dialect: &dyn OpenApiDialect = match parse_openapi_version(&document)? {
         OpenApiVersion::V3_0 => &OpenApi30Importer,
-        OpenApiVersion::V3_1 => &OpenApi31Importer,
+        OpenApiVersion::V3_1 => {
+            // Only 3.1 can choose a dialect. 3.0's schema object is its own,
+            // fixed by that specification, so neither keyword means anything
+            // there and a document carrying one is not selecting anything.
+            validate_json_schema_dialect(&document)?;
+            &OpenApi31Importer
+        }
     };
 
     let mut importer = OpenApiImporter::new(manifest, surface, &document, dialect);
--- a/crates/coral-spec/src/v4/surfaces/openapi/import.rs
+++ b/crates/coral-spec/src/v4/surfaces/openapi/import.rs
@@ -14,6 +14,7 @@ use crate::{ManifestError, Result};
 
 use super::dialect::OpenApiDialect;
 use super::v3_0::OpenApi30Importer;
+use super::v3_1::OpenApi31Importer;
 use super::version::{OpenApiVersion, parse_openapi_version};
 
 pub fn import_openapi_surface(
@@ -25,6 +26,7 @@ pub fn import_openapi_surface(
         serde_yaml::from_slice(document_bytes).map_err(ManifestError::parse_yaml)?;
     let dialect: &dyn OpenApiDialect = match parse_openapi_version(&document)? {
         OpenApiVersion::V3_0 => &OpenApi30Importer,
+        OpenApiVersion::V3_1 => &OpenApi31Importer,
     };
 
     let mut importer = OpenApiImporter::new(manifest, surface, &document, dialect);
--- a/crates/coral-spec/src/v4/surfaces/openapi/import.rs
+++ b/crates/coral-spec/src/v4/surfaces/openapi/import.rs
@@ -149,6 +149,33 @@ impl<'a> OpenApiImporter<'a> {
         }
     }
 
+    /// Reports a schema reaching for a keyword its document's version removed.
+    ///
+    /// Every path that reads a schema has to ask, not just the one that imports
+    /// a type from it. A parameter resolves to a scalar through
+    /// `import_parameter_scalar` and never reaches `import_schema`, and a
+    /// top-level collection response hands `import_schema` its `items` and sets
+    /// the schema carrying the keyword aside — so a `nullable: true` in either
+    /// place was dropped with nothing said, which is the silence this diagnostic
+    /// exists to break.
+    ///
+    /// `subject` names what carries the keyword — a type, a parameter, a
+    /// response — because the operation alone does not locate it.
+    pub(super) fn warn_removed_keywords(
+        &self,
+        schema: &Value,
+        subject: &str,
+        operation_id: &str,
+        diagnostics: &mut Vec<Diagnostic>,
+    ) {
+        if let Some(warning) = self.dialect.removed_keyword_warning(schema) {
+            diagnostics.push(Diagnostic::new(
+                format!("{subject} in operation '{operation_id}': {warning}"),
+                Some(operation_id.to_string()),
+            ));
+        }
+    }
+
     pub(super) fn ref_error_diagnostic(
         error: RefError<'_>,
         context: &RefDiagnosticContext<'_>,
--- a/crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
+++ b/crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
@@ -1,6 +1,6 @@
 use std::collections::BTreeSet;
 
-use serde_json::Value;
+use serde_json::{Map, Value};
 
 use crate::v4::diagnostics::Diagnostic;
 use crate::v4::ir::{IrField, IrType, IrTypeShape};
@@ -30,6 +30,35 @@ const MAX_ALL_OF_DEPTH: usize = 8;
 /// type. Matches the MCP importer's ceiling for the same comparison.
 const MAX_PROPERTY_COMPARISON_DEPTH: usize = 64;
 
+/// Keywords beside a `$ref` that describe it rather than narrow it.
+///
+/// An annotation says nothing about which instances the schema admits, so it
+/// cannot make the reference mean something the reference does not already
+/// mean — a `description` written beside a `$ref` documents the field that
+/// holds it, and the schema is still exactly what it resolves to.
+const REF_SIBLING_ANNOTATIONS: [&str; 11] = [
+    "description",
+    "title",
+    "summary",
+    "example",
+    "examples",
+    "deprecated",
+    "readOnly",
+    "writeOnly",
+    "externalDocs",
+    "xml",
+    "$comment",
+];
+
+/// Assertions beside a `$ref` that the `allOf` fold can compose.
+///
+/// The fold merges properties and answers with an object, so these are the
+/// keywords whose composition it actually implements. Anything else — a
+/// `maxLength` narrowing a referenced string, a `type` contradicting the
+/// reference — is reported instead of being folded into a shape it does not
+/// describe.
+const COMPOSABLE_REF_SIBLINGS: [&str; 3] = ["properties", "required", "additionalProperties"];
+
 /// Why folding a schema's `allOf` tree stopped.
 enum AllOfMergeError {
     /// A branch could not be resolved. A diagnostic naming it has already been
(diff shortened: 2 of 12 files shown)


simonwhitaker:
@cubic re-review

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -150,6 +156,8 @@ impl OpenApiImporter<'_> {
             IrTypeShape::Enum {
                 values: values.iter().map(enum_value).collect(),
             }
+        } else if let Some(values) = self.dialect.const_enum_values(&resolved) {
```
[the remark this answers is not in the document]
simonwhitaker: Fixed in 7821158: `const_enum_values` now reads only scalar constants (`String`/`Number`/`Bool`), so an object- or array-valued `const` falls through to the shape dispatch and keeps its declared fields.

crates/coral-spec/src/v4/surfaces/openapi/v3_1.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// equivalent to, so those branches keep their fixed value.
+    ///
+    /// Only a scalar constant is read this way. `const` may hold any JSON value,
+    /// and an object or array one describes a shape rather than a value: taking
+    /// `{type: object, properties: {...}, const: {...}}` for an enum would stand
+    /// the stringified constant where the declared fields should be and drop
+    /// every column. Those fall through to the shape dispatch instead, which is
+    /// what describes them.
+    fn const_enum_values(&self, schema: &Value) -> Option<Vec<String>> {
+        let value = schema.get("const")?;
```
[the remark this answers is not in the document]
simonwhitaker: Fixed in ba0ab8a: `const_enum_values` is now read before the `enum` arm, so a branch declaring both keeps the value its `const` pins rather than the whole family its inherited `enum` lists. A structured `const` still falls past it to `enum` and then to the shape dispatch, since `const_enum_values` reads only scalars.

Both directions are pinned by tests: `a_const_narrows_the_enum_it_is_declared_beside` fails without the reorder (it reports `["invoice", "receipt", "credit_note"]`), and `a_structured_const_leaves_a_neighbouring_enum_to_be_read` holds the boundary the other way.

crates/coral-spec/src/v4/surfaces/openapi/v3_1.rs
```diff
@@ -0,0 +1,48 @@
...
+/// reference, and the keywords 3.1 gained for describing binary payloads only
+/// matter to file uploads, which Coral does not support. What is left is the
+/// handful of spellings below.
+pub(super) struct OpenApi31Importer;
+
+impl OpenApiDialect for OpenApi31Importer {
+    /// 3.1 dropped `nullable` in favour of JSON Schema's own `null` type, which
+    /// a schema admits by listing it alongside the type it otherwise is.
+    fn schema_nullable(&self, schema: &Value) -> bool {
+        json_schema_type_contains(schema, "null")
```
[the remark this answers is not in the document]
simonwhitaker: Not acting on this one — I think the premise is wrong in both directions, and the suggested change would make the reading less correct rather than more.

**The sibling cannot add nullability.** In 2020-12 a `$ref` and its siblings are a conjunction: `{$ref: Party, type: [object, "null"]}` means *satisfies `Party` **and** is object-or-null*. If `Party` declares `type: object`, the intersection excludes null, so the schema is not nullable and reading it as nullable would be wrong. Merging siblings as the comment suggests would flip a correct answer to an incorrect one. (The 3.1 spelling that genuinely does make a reference nullable is `anyOf: [{type: "null"}, {$ref: ...}]`, which is #2128's subject and is handled there.)

**No field imports as non-nullable regardless.** `IrField.nullable` is hard-coded `true` in `import_object_fields`, and every projected column in `projections/derive.rs` is nullable. The only thing the sibling could affect is `IrType.nullable`, which nothing downstream reads for column nullability. I probed this directly — a field written `{$ref: Party, type: [object, "null"]}`, one written as a bare `$ref: Party`, and one referencing a target that declares `type: [object, "null"]` all import with `nullable: true` on the field.

**And there is nowhere to put a site-specific answer.** Both the sibling form and the bare `$ref` intern to the same shared `party` type — that is the point of the interning, and the same reason the target's description is used rather than one written beside the `$ref` (see the comment above the `description` binding in `import_schema`). Storing per-site nullability on a shared type would let the first referencing site win and every later one lose.

Happy to revisit if `IrType.nullable` ever becomes load-bearing — a non-null column type would be the trigger — but at that point the fix is to stop sharing the type, not to merge siblings into it.

crates/coral-spec/src/v4/surfaces/openapi/import.rs
```diff
@@ -25,6 +26,7 @@ pub fn import_openapi_surface(
         serde_yaml::from_slice(document_bytes).map_err(ManifestError::parse_yaml)?;
     let dialect: &dyn OpenApiDialect = match parse_openapi_version(&document)? {
         OpenApiVersion::V3_0 => &OpenApi30Importer,
+        OpenApiVersion::V3_1 => &OpenApi31Importer,
```
sauldhernandez: A 3.1 response schema with `$ref: Base` plus sibling `properties.extra` and `required: [extra]` resolves to `Base` alone, so `extra` disappears from the IR and projection. Please merge assertion siblings into the resolved schema, or reject this unsupported shape instead of silently narrowing it.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -72,6 +72,12 @@ impl OpenApiImporter<'_> {
             .unwrap_or_default()
             .to_string();
         let nullable = self.dialect.schema_nullable(&resolved);
+        if let Some(warning) = self.dialect.removed_keyword_warning(&resolved) {
```
sauldhernandez: This warning only runs through `import_schema`. Parameter schemas use `import_parameter_scalar`, and a top-level array response imports only its items, so `nullable: true` in either location is silently ignored. Please centralize this diagnostic across schema entry paths and add coverage for both cases.

crates/coral-spec/src/v4/surfaces/openapi/v3_1.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// discriminator on each branch of a union. Read as the enum it is
+    /// equivalent to, so those branches keep their fixed value.
+    ///
+    /// Only a scalar constant is read this way. `const` may hold any JSON value,
+    /// and an object or array one describes a shape rather than a value: taking
+    /// `{type: object, properties: {...}, const: {...}}` for an enum would stand
+    /// the stringified constant where the declared fields should be and drop
+    /// every column. Those fall through to the shape dispatch instead, which is
+    /// what describes them.
+    fn const_enum_values(&self, schema: &Value) -> Option<Vec<String>> {
```
sauldhernandez, line 51: This applies 2020-12 `const` semantics even when `jsonSchemaDialect` or `$schema` selects another dialect. For example, draft-04 treats `const` as unknown, but Coral persists `{type: string, const: 4}` as the closed enum `["4"]`. Please propagate supported dialect declarations or reject unsupported ones before import.

crates/coral-spec/src/v4/surfaces/openapi/v3_1.rs
```diff
@@ -0,0 +1,62 @@
...
+        let value = schema.get("const")?;
+        // Scalars are stringified exactly as `enum` members already are, so
+        // `const: 4` and `enum: [4]` agree rather than the newer keyword
+        // inventing its own reading. `null` is excluded with the structured
+        // values: it constrains the schema to hold nothing, which the type
+        // already says better than a one-value enum of "null" would.
+        matches!(value, Value::String(_) | Value::Number(_) | Value::Bool(_))
```
sauldhernandez, line 58: A typeless `{const: null}` or enum containing only `null` admits null, but this leaves `IrType.nullable` false and the shape opaque. Please derive nullability from `const` and `enum` while preserving the existing shape behavior.

sauldhernandez (approved):

crates/coral-spec/src/v4/surfaces/openapi/import.rs
simonwhitaker: Fixed in 9ef18e8. `{$ref: Base, properties: {extra}, required: [extra]}` now composes rather than narrowing to `Base`.

It is rewritten as the `allOf` it is equivalent to, so the existing fold does the merge with its depth ceiling, cycle guard and conflict reporting rather than a second merge growing beside it. Whether siblings apply is a dialect question — 3.0's Reference Object ignores them — so it goes through `OpenApiDialect::ref_siblings_apply`, and `openapi_30_ignores_members_written_beside_a_ref` pins the 3.0 half against the same document.

Two things worth flagging. The composition had to go in the response path as well as `import_schema`: `import_response` resolves the schema before classifying it, so by the time `import_schema` sees it the siblings are already gone — your example goes through exactly that path. And only the object assertions (`properties`, `required`, `additionalProperties`) compose, because the fold answers with an object: routing `{$ref: Name, maxLength: 5}` through it would put a fieldless object where a referenced string was. Those are reported instead — `openapi_31_reports_a_ref_sibling_it_cannot_compose` covers it. Annotations beside a `$ref` still resolve straight through, since they admit no different instances.

One known gap I did not chase: a composed schema takes its nullability from the wrapper, so `{$ref: NullableBase, properties: {...}}` loses the base's own `null` type. Say the word if you want that too.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
simonwhitaker: Fixed in f20891c. There is now one `warn_removed_keywords` on the importer, and all three entry paths report through it — `import_schema`, `import_parameter_scalar`, and the response path for the schema a collection sets aside.

`openapi_31_warns_about_a_removed_keyword_on_every_schema_path` covers both of the cases you named, asserting the exact count so a warning cannot go missing behind a passing `any`. The response path asks only when the classification set the schema aside: a singleton hands `import_schema` that very schema, and reporting in both places would say it twice — `openapi_31_reports_a_singleton_response_keyword_once` pins that.

crates/coral-spec/src/v4/surfaces/openapi/v3_1.rs
simonwhitaker: Fixed in b7207ae. A 3.1 document declaring a dialect Coral does not read is now refused before import rather than read as 2020-12 anyway — your draft-04 `{type: string, const: 4}` example is the test.

I went with rejecting rather than propagating: the failure mode is a persisted surface whose columns are wrong under rules the author did not choose, and unlike a diagnostic that cannot be noticed after the fact. Both spellings are checked — `jsonSchemaDialect` for the document, `$schema` for a single schema resource — and the OAS 3.1 base dialect is accepted alongside plain 2020-12, since its extra vocabulary describes `discriminator` and `xml` rather than anything read here.

The `$schema` walk skips `example`, `examples`, `default`, `const` and `enum`: their values are instance data, so a document that describes JSON Schema documents carries dialect identifiers in its examples that declare nothing about how it is itself read. A property *named* `$schema` is unaffected — in that position the value is a schema object, not a string.

crates/coral-spec/src/v4/surfaces/openapi/v3_1.rs
simonwhitaker: Fixed in c2ab6f4. `schema_nullable` now derives nullability from `const` and `enum` as well as `type`, so `{const: null}` and `{enum: [null, ...]}` mark the column nullable.

Any null enum member counts rather than only a lone one — `enum: [null, open]` admits null exactly as `enum: [null]` does. Shape behaviour is untouched, and the test asserts that half too: the enum still imports as its values, and a null `const` still falls through to the shape dispatch because `const_enum_values` declines to read it as a one-value enum.

PR #2125, opened 2026-08-11
feat(spec): accept OpenAPI 3.1 documents that declare no paths

simonwhitaker:
Part of #1321.

## Why

3.1 made `paths` optional, so a document may describe only `webhooks` or only reusable `components`. The importer required it unconditionally and refused such a document outright, which meant rejecting a spec its own author's tooling considers valid.

## What changes

`paths` is now required only where the version requires it, and a document with nothing under it imports to an empty surface instead.

An empty surface is a confusing thing to receive in silence, so a document declaring webhooks now says so. Coral models a source as endpoints it calls and has nothing to map provider-initiated callbacks onto, but the author deserves to be told that rather than left to work out why their spec imported nothing.

## `$ref` siblings: pinned, not changed

3.1 gives keywords beside a `$ref` meaning where 3.0 ignored them, so a description written at a reference site looked like a third thing for the dialects to disagree about. It is not.

A type here is interned by the name it is referenced under and shared by every site that references it, so a site's description cannot go on the type without overwriting what every other site says — first reference winning, the rest silently lost. The field is where a site-specific description fits, and `field_description` already reads the site in preference to the target, which is 3.1's rule already applied to both versions. Two tests fix that split in place, one of them covering the two-references case that makes it necessary.

## Review follow-up

Reading `paths` straight through `as_object` collapsed absent and malformed into the same answer. Once 3.1 made an absent `paths` legal, that collapse meant a 3.1 document whose `paths` was a string or a list imported as an empty surface — reported as a source with no tables rather than the broken descriptor it is. A present `paths` must now be a mapping under either version; only an absent one is subject to the version's rule.

The webhook diagnostic assertion searched for the character `2` anywhere in the message, so it would have passed on any count containing one.

---

### DIFF
--- a/crates/coral-spec/src/v4/openapi_tests.rs
+++ b/crates/coral-spec/src/v4/openapi_tests.rs
@@ -4465,6 +4465,7 @@ fn openapi_31_reads_a_scalar_const_the_way_it_reads_a_one_value_enum() {
     );
 }
 
+#[test]
 fn openapi_30_still_requires_paths() {
     let error = import_document(
         r"
@@ -4507,7 +4508,7 @@ webhooks:
         .find(|diagnostic| diagnostic.message.contains("webhook"))
         .expect("a webhook-only document should say why it imported nothing");
     assert!(
-        diagnostic.message.contains('2'),
+        diagnostic.message.contains("2 webhook(s)"),
         "the count makes it clear what was skipped: {}",
         diagnostic.message
     );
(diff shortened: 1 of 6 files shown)


crates/coral-spec/src/v4/surfaces/openapi/import.rs
```diff
@@ -72,15 +72,30 @@ impl<'a> OpenApiImporter<'a> {
...
+        let paths = match self.document.get("paths") {
+            Some(paths) => Some(paths.as_object().ok_or_else(|| {
+                ManifestError::validation("OpenAPI document paths must be a mapping")
+            })?),
+            None if self.dialect.paths_required() => {
+                return Err(ManifestError::validation(
+                    "OpenAPI document is missing paths",
+                ));
+            }
+            None => None,
```
sauldhernandez, line 92: OpenAPI 3.1 requires at least one of `paths`, `components`, or `webhooks`. This arm accepts a document with none of them and installs an empty source. Please accept missing `paths` only when `components` or `webhooks` is a mapping, and test absent and malformed alternatives.

crates/coral-spec/src/v4/openapi_tests.rs
```diff
@@ -4417,6 +4465,234 @@ fn openapi_31_reads_a_scalar_const_the_way_it_reads_a_one_value_enum() {
...
+    // 3.1 gives keywords beside a `$ref` meaning where 3.0 ignored them, but the
+    // distinction lands in the same place under both versions, so neither
+    // dialect needs a say in it.
+    //
+    // A type here is interned by the name it is referenced under and shared by
+    // every site referencing it, so a description written beside one `$ref`
+    // cannot go on the type without overwriting what the other sites say. The
+    // field is where a site-specific description fits, and it already reads the
+    // site in preference to the target — which is 3.1's rule, applied to both.
+    for version in ["3.0.3", "3.1.0"] {
```
sauldhernandez, line 4586: This loop locks in the wrong OpenAPI 3.0 behavior. A `$ref` at a 3.0 `properties` entry is a Reference Object, so its sibling `description` is ignored; only 3.1 gives that sibling meaning. Please expect the target description for 3.0 and the reference-site description for 3.1.

sauldhernandez (approved):

PR #2126, opened 2026-08-11
fix(spec): treat OpenAPI 3.0 'example' as an annotation

simonwhitaker:
Part of #1321. Drive-by — not needed for 3.1, and droppable if it muddies the stack.

## Why

`ANNOTATION_KEYS` listed `examples` but not `example`. They are the same annotation under two spellings — `examples` is JSON Schema's, which 3.1 adopted, and `example` is the singular one 3.0 defined on its own schema object — so reading only the plural form left the singular one counting as a constraint.

Composed types routinely re-declare an inherited property to narrow an annotation. Where that annotation was a 3.0 `example`, the fold read the re-declaration as a genuine disagreement about validation, and a property conflict discards the whole type: the composed object collapsed to opaque JSON and took every nested type it owned with it.

## Verification

Measured against the six bundled v4 descriptors, this recovers one type in GitHub's spec and the 22 nested types it owned — 15,364 imported types to 15,387. The other five are unaffected, and no inferred row path, pagination contract, or lookup key changes on any of them.

## Note on the version bumps

Bumps both importer versions. The fingerprint this feeds is shared with the MCP importer, which is unaffected in practice — 2020-12 has no singular `example` — but its artifacts are keyed on the same comparison and should not be read back under the old version.

---

### DIFF
--- a/crates/coral-spec/src/v4/surfaces/json_schema.rs
+++ b/crates/coral-spec/src/v4/surfaces/json_schema.rs
@@ -4,7 +4,22 @@ use serde_json::Value;
 
 use crate::v4::ir::IrScalarType;
 
-const ANNOTATION_KEYS: &[&str] = &["$comment", "default", "description", "examples", "title"];
+/// Keywords that describe a schema without constraining what validates against
+/// it, and so cannot make two declarations of one property disagree.
+///
+/// `example` and `examples` are the same annotation under two spellings:
+/// `examples` is JSON Schema's, which 3.1 adopted, and `example` is the
+/// singular one 3.0 defined on its own schema object. Both have to be here, or
+/// a property re-declared only to carry a different `example` reads as a
+/// genuine conflict.
+const ANNOTATION_KEYS: &[&str] = &[
+    "$comment",
+    "default",
+    "description",
+    "example",
+    "examples",
+    "title",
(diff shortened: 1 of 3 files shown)


crates/coral-spec/src/v4/surfaces/json_schema.rs
```diff
@@ -4,7 +4,22 @@ use serde_json::Value;
...
+/// `example` and `examples` are the same annotation under two spellings:
+/// `examples` is JSON Schema's, which 3.1 adopted, and `example` is the
+/// singular one 3.0 defined on its own schema object. Both have to be here, or
+/// a property re-declared only to carry a different `example` reads as a
+/// genuine conflict.
+const ANNOTATION_KEYS: &[&str] = &[
+    "$comment",
+    "default",
+    "description",
+    "example",
```
sauldhernandez, line 19: This list also controls MCP input-schema folding, where `$schema` may select another dialect. A custom vocabulary can define singular `example` as an assertion; dropping it here can merge incompatible `allOf` branches and expose a tool whose calls always fail validation. Scope this key to OpenAPI, or reject unsupported MCP dialects before folding.

sauldhernandez (approved):

PR #2128, opened 2026-08-11
fix(spec): read a union of a schema and null as that schema

simonwhitaker:
Part of #1321. Found while investigating a review comment on #2124; it is the largest single fidelity gap on the 3.1 path.

## Why

Once 3.1 removed `nullable`, a nullable reference had nowhere to say so: a `$ref` carries no type of its own to list `null` in. Documents spell it as a union instead — `anyOf: [{type: 'null'}, {$ref: ...}]` — which GitHub's 3.1 publication uses 217 times.

That matched no arm of the shape dispatch, so every one of them imported as opaque JSON and lost every field of the referenced type.

## Measured impact

Against GitHub's own two publications of the same API (`descriptions` is 3.0.3, `descriptions-next` is 3.1.0):

| | 3.0.3 | 3.1.0 before | 3.1.0 after |
|---|---|---|---|
| opaque JSON types | 237 | 621 | 234 |
| object types | 2,000 | 1,965 | 1,976 |

Parity, on a document whose every nullable reference is spelled this way.

## What changes

A union that reduces to exactly one real branch is imported as that branch. The branch is handed back to `import_schema` rather than having its shape inlined, which keeps a `$ref` branch on the type id it already has — expanding it under the referring site's name instead would copy the target, and every inline schema beneath it, once per nullable reference: 25k imported types against 15k for the same API.

Two or more real branches is a genuine choice between shapes, which this IR has no type for and which keeps importing as opaque JSON. Reading the first branch and discarding the rest would describe columns the response does not always carry.

The six bundled 3.0 descriptors are unaffected: no inferred row path, pagination contract, or lookup key changes on any of them.

## Review follow-ups

Four, all from review, in the second commit.

**Unbounded recursion.** Unwrapping discarded the identity of the reference it arrived by, which took the type id with it — and the type id is what ends a `$ref` cycle here, through the interning check. `Node: anyOf: [{type: 'null'}, {$ref: Node}]` unwrapped to a schema resolving back to the same union, and the inline variant grew a fresh `node_row_child_child…` on every pass, so neither ever met a type it had already interned. Both overflowed the stack on documents that imported fine before unions were read at all. The id now comes from the referring site first and the branch's own only as a fallback, and the walk is bounded twice over: a branch leading back to a reference already followed stops it, with a depth cap behind that.

**Response classification.** It was left reading the wrapper. The union declares no type of its own, so a nullable collection looked like a singleton while its row type was built from the branch — publishing one row holding the whole list. It now unwraps through the same walk.

**Lost descriptions.** A nullable reference lost the referenced schema's description, because the wrapper has neither a description nor a `$ref` for `field_description` to follow. It now describes the branch that carries the shape.

**Other null spellings.** `declares_only_null` recognised only the `type` spelling, so `const: null` and `enum: [null]` branches — both valid — made the union look like a choice between two real shapes and sent it back to opaque JSON.

---

### DIFF
--- a/crates/coral-spec/src/v4/surfaces/openapi/responses.rs
+++ b/crates/coral-spec/src/v4/surfaces/openapi/responses.rs
@@ -106,8 +106,7 @@ impl OpenApiImporter<'_> {
         // For a collection that published one row holding the whole list.
         let resolved = self
             .schema_through_null_unions(&resolved, operation_id, diagnostics)
-            .and_then(|site| self.resolve_ref(&site, operation_id, diagnostics))
-            .unwrap_or(resolved);
+            .map_or(resolved, |walk| walk.resolved);
         let (cardinality, row_schema, schema_entity_name) =
             classify_response_schema(path, &resolved);
         // Only when the classification set this schema aside. A singleton hands
--- a/crates/coral-spec/src/v4/surfaces/openapi/responses.rs
+++ b/crates/coral-spec/src/v4/surfaces/openapi/responses.rs
@@ -99,6 +99,15 @@ impl OpenApiImporter<'_> {
                 },
             );
         };
+        // Classified through the same unwrapping `import_schema` does. A
+        // nullable response — `anyOf: [{type: 'null'}, {$ref: ...}]` — otherwise
+        // has its cardinality read off the union, which declares no type and so
+        // looks like a singleton, while the row type is built from the branch.
+        // For a collection that published one row holding the whole list.
+        let resolved = self
+            .schema_through_null_unions(&resolved, operation_id, diagnostics)
+            .and_then(|site| self.resolve_ref(&site, operation_id, diagnostics))
+            .unwrap_or(resolved);
         let (cardinality, row_schema, schema_entity_name) =
             classify_response_schema(path, &resolved);
         // Only when the classification set this schema aside. A singleton hands
--- a/crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
+++ b/crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
@@ -94,6 +94,23 @@ impl OpenApiImporter<'_> {
             return self.import_schema(&composed, suggested_id, operation_id, diagnostics);
         }
         let resolved = self.resolve_ref(schema, operation_id, diagnostics)?;
+        // A union of one real branch and `null` is imported as that branch.
+        //
+        // It is how a nullable reference has to be written once `nullable` is
+        // gone: a `$ref` carries no type of its own to list `null` in, so 3.1
+        // documents spell it `anyOf: [{type: 'null'}, {$ref: ...}]`. Left alone
+        // it matches no arm of the shape dispatch below and imports as opaque
+        // JSON, taking every field of the referenced type with it.
+        //
+        // Handing the branch back to this function rather than inlining its
+        // shape here is what keeps a `$ref` branch on the type id it already
+        // has. Expanding it under this site's name instead would copy the whole
+        // target — and every inline schema beneath it — once per site that
+        // references it nullably, which on GitHub's 3.1 descriptor is the
+        // difference between 15k imported types and 25k.
+        if let Some(branch) = null_union_branch(&resolved).cloned() {
+            return self.import_schema(&branch, suggested_id, operation_id, diagnostics);
+        }
         let type_id = schema.get("$ref").and_then(Value::as_str).map_or_else(
             || normalize_identifier(suggested_id, "type"),
             type_id_from_ref,
@@ -532,6 +549,43 @@ impl OpenApiImporter<'_> {
             .and_then(Value::as_str)
             .unwrap_or_default()
             .to_string()
+    }
+}
+
+/// The single real branch of a union between a schema and `null`.
+///
+/// Only a union that reduces to exactly one real branch is unwrapped. Two or
+/// more is a genuine choice between shapes, which Coral has no type for and
+/// which has to keep importing as opaque JSON — reading the first branch and
+/// discarding the rest would describe a column the response does not have.
+fn null_union_branch(schema: &Value) -> Option<&Value> {
+    for keyword in ["anyOf", "oneOf"] {
+        let Some(branches) = schema.get(keyword).and_then(Value::as_array) else {
+            continue;
+        };
+        // A union has to have lost something to `null` to be one of these; a
+        // single-branch `anyOf` is just a wrapper and keeps its own reading.
+        if branches.len() < 2 {
+            continue;
+        }
+        let mut real = branches.iter().filter(|branch| !declares_only_null(branch));
+        let only = real.next()?;
+        if real.next().is_some() {
+            return None;
+        }
+        return Some(only);
+    }
+    None
+}
+
+/// Whether a branch admits nothing but `null`.
+fn declares_only_null(schema: &Value) -> bool {
+    match schema.get("type") {
+        Some(Value::String(value)) => value == "null",
+        Some(Value::Array(values)) => {
+            !values.is_empty() && values.iter().all(|value| value.as_str() == Some("null"))
+        }
+        _ => false,
     }
 }
 
--- a/crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
+++ b/crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
@@ -65,6 +65,42 @@ const COMPOSABLE_REF_SIBLINGS: [&str; 3] = ["properties", "required", "additiona
 /// for ever.
 const MAX_NULL_UNION_DEPTH: usize = 8;
 
+/// Where a walk through a chain of `[T, null]` unions ended.
+pub(super) struct NullUnionWalk {
+    /// The schema carrying the shape, with the walk's last `$ref` hop already
+    /// followed.
+    pub(super) resolved: Value,
+    /// The reference the result is interned under, when the walk crossed one.
+    ///
+    /// The site the walk ended on supplies this, so a chain of aliases still
+    /// interns under the name it finally arrived at and every route to that
+    /// name shares one type. Only when the walk ends somewhere unnamed — a
+    /// union whose real branch is written inline — does the first reference it
+    /// crossed stand in.
+    ///
+    /// That fallback is what terminates a recursive nullable reference. A
+    /// self-referential type reached through a wrapper — `node: anyOf: [null,
+    /// {type: object, properties: {child: {anyOf: [null, {$ref: node}]}}}]` —
+    /// ends its walk on the inline branch, and without a name to intern under
+    /// every nested `child` took the suggested id instead, growing a fresh
+    /// `node_child_child…` that the interning check could never match until the
+    /// stack ran out.
+    pub(super) reference: Option<String>,
+}
+
+impl NullUnionWalk {
+    fn ending_at(site: &Value, resolved: Value, first_reference: Option<String>) -> Self {
+        Self {
+            reference: site
+                .get("$ref")
+                .and_then(Value::as_str)
+                .map(ToString::to_string)
+                .or(first_reference),
+            resolved,
+        }
+    }
+}
+
 /// Why folding a schema's `allOf` tree stopped.
 enum AllOfMergeError {
     /// A branch could not be resolved. A diagnostic naming it has already been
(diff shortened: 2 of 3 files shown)


crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -59,6 +59,23 @@ impl OpenApiImporter<'_> {
...
+        // it matches no arm of the shape dispatch below and imports as opaque
+        // JSON, taking every field of the referenced type with it.
+        //
+        // Handing the branch back to this function rather than inlining its
+        // shape here is what keeps a `$ref` branch on the type id it already
+        // has. Expanding it under this site's name instead would copy the whole
+        // target — and every inline schema beneath it — once per site that
+        // references it nullably, which on GitHub's 3.1 descriptor is the
+        // difference between 15k imported types and 25k.
+        if let Some(branch) = null_union_branch(&resolved).cloned() {
```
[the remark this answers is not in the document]
simonwhitaker: Not accurate — this path has three regression tests, added in the same commit as the change:

- `a_union_of_a_ref_and_null_imports_as_the_referenced_type` is the case described here: a schema with `anyOf: [{type: 'null'}, {$ref: ...}]` asserted to import with the target's fields rather than as opaque JSON, for both `anyOf` and `oneOf`, plus an assertion that both land on the shared `party` type id rather than a per-site copy.
- `a_union_of_two_real_branches_stays_opaque` pins the boundary in the other direction.
- `both_spellings_of_a_nullable_reference_import_alike` asserts the 3.0 `nullable` spelling and the 3.1 union spelling produce the same shape.

I verified the first and third fail without the change by stashing `schemas.rs` and re-running: both fail, so they discriminate rather than passing vacuously.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -420,6 +437,43 @@ impl OpenApiImporter<'_> {
...
 }
 
+/// The single real branch of a union between a schema and `null`.
+///
+/// Only a union that reduces to exactly one real branch is unwrapped. Two or
+/// more is a genuine choice between shapes, which Coral has no type for and
+/// which has to keep importing as opaque JSON — reading the first branch and
+/// discarding the rest would describe a column the response does not have.
+fn null_union_branch(schema: &Value) -> Option<&Value> {
+    for keyword in ["anyOf", "oneOf"] {
```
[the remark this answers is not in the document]
simonwhitaker: Correct on the JSON Schema semantics, but not acting on it, for two reasons.

The shape is the same either way. Unwrapping decides which branch supplies the type; whether `null` is *also* admitted only affects nullability, and the strict `oneOf` reading differs from the loose one on exactly that. So the overlapping case cannot produce a wrong shape here — only a nullability claim that nothing downstream reads: `IrField.nullable` is set for every field regardless, and every projected column is nullable. There is an assertion recording that boundary in `both_spellings_of_a_nullable_reference_import_alike`.

Proving the real branch non-null is also not cheap in this pass. The branch is usually a `$ref`, so it means resolving the target and deciding whether *it* admits null — itself possibly through a union or a composition — which is the cycle-prone walk that just had to be bounded twice over in this file. That is a real cost to buy a distinction that does not reach output.

If nullability ever becomes load-bearing (a non-null column type, say), this is the right place to tighten and the assertion above is what will flag it.

simonwhitaker:
@cubic re-review

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -59,6 +59,23 @@ impl OpenApiImporter<'_> {
...
+        // JSON, taking every field of the referenced type with it.
+        //
+        // Handing the branch back to this function rather than inlining its
+        // shape here is what keeps a `$ref` branch on the type id it already
+        // has. Expanding it under this site's name instead would copy the whole
+        // target — and every inline schema beneath it — once per site that
+        // references it nullably, which on GitHub's 3.1 descriptor is the
+        // difference between 15k imported types and 25k.
+        if let Some(branch) = null_union_branch(&resolved).cloned() {
+            return self.import_schema(&branch, suggested_id, operation_id, diagnostics);
```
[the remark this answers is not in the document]
simonwhitaker: Fixed in 63ecd9d: response classification now runs the same unwrap, via `schema_through_null_unions` in `responses.rs`, so cardinality and pagination inference see the real branch rather than the outer union.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -58,11 +64,37 @@ impl OpenApiImporter<'_> {
...
+        // that terminates every other `$ref` cycle here — with a name that never
+        // repeats, growing a fresh `root_child_child…` on each pass until the
+        // stack ran out.
+        //
+        // Preferring the branch's id when this site has none is what keeps a
+        // nullable reference on the type it already has, rather than copying the
+        // target and every inline schema beneath it once per referring site —
+        // on GitHub's 3.1 descriptor, the difference between 15k imported types
+        // and 25k.
+        let type_id = schema
```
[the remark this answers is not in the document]
simonwhitaker: Confirmed and fixed in 5329e8b. It reproduces as a stack overflow, with the recursive `$ref` wrapped at the use site and the walk ending on an inline branch:

```yaml
node:
  anyOf:
    - type: 'null'
    - type: object
      properties:
        child:
          anyOf: [{type: 'null'}, {$ref: '#/components/schemas/node'}]
```

The terminal site is that inline object, which carries no `$ref`, so the type id fell back to the suggested one and each nested `child` got a fresh `node_child_child…` the interning check could never match.

`schema_through_null_unions` now reports the first reference it crossed alongside the site it ended on, and the type id falls back to that when the site is unnamed. The site still wins where it has a name, so a chain of aliases keeps interning under the name it arrives at and existing sharing is unchanged. `a_recursive_nullable_reference_reached_through_a_wrapper_terminates` covers it — it overflows the stack without the change.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -58,11 +64,37 @@ impl OpenApiImporter<'_> {
...
-            type_id_from_ref,
-        );
+        // A union of one real branch and `null` is imported as that branch.
+        //
+        // It is how a nullable reference has to be written once `nullable` is
+        // gone: a `$ref` carries no type of its own to list `null` in, so 3.1
+        // documents spell it `anyOf: [{type: 'null'}, {$ref: ...}]`. Left alone
+        // it matches no arm of the shape dispatch below and imports as opaque
+        // JSON, taking every field of the referenced type with it.
+        let site = self.schema_through_null_unions(schema, operation_id, diagnostics)?;
```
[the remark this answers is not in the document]
simonwhitaker: Fixed in 5329e8b. `schema_through_null_unions` now returns what it resolved alongside the site, so neither caller resolves the terminal `$ref` a second time — `import_schema` and the response classification in `responses.rs` both had that duplicate hop.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -401,10 +433,55 @@ impl OpenApiImporter<'_> {
...
+
     fn field_description(&self, schema: &Value) -> String {
         if let Some(description) = schema.get("description").and_then(Value::as_str) {
             return description.to_string();
         }
+        // A nullable spelling wraps the schema that has the description. The
+        // wrapper says nothing itself, so without this a field written
+        // `anyOf: [{type: 'null'}, {$ref: ...}]` lost the column description
+        // that the same field written as a bare `$ref` keeps.
+        if let Some(branch) = null_union_branch(schema) {
```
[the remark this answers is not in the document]
simonwhitaker: Addressed in 5329e8b, though not by borrowing the import walk's budget — that would have been a regression.

The stack exhaustion is not reachable. Document nesting is capped well below anything that could exhaust a stack: a schema nested past roughly a dozen levels is refused by the YAML parser itself with `recursion limit exceeded`, so no document that parses can drive this lookup deep enough to matter. I checked by generating wrappers at increasing depth — depth 12 imports, depth 60 and beyond fail to parse.

Bounding it at `MAX_NULL_UNION_DEPTH` would also have lost real data. That budget is 8, and a chain nested 9–12 deep does parse, so capping there silently dropped descriptions that were previously found. I measured that too: with the cap, depth 8 and 12 both returned `""` where the recursive version returned the description.

So the lookup is a loop now rather than a recursion — no depth of nesting can put it on the stack — with no depth cap, which it does not need. Unlike the import walk it never follows a `$ref`, only steps to a branch written inside the schema it already holds, so every step descends into one finite tree and it terminates because the document does. `a_chain_of_nullable_wrappers_yields_its_description_at_any_depth` pins that at depths either side of the import walk's budget.

crates/coral-spec/src/v4/surfaces/openapi/responses.rs
```diff
@@ -101,6 +101,14 @@ impl OpenApiImporter<'_> {
                 },
             );
         };
+        // Classified through the same unwrapping `import_schema` does. A
+        // nullable response — `anyOf: [{type: 'null'}, {$ref: ...}]` — otherwise
+        // has its cardinality read off the union, which declares no type and so
+        // looks like a singleton, while the row type is built from the branch.
+        // For a collection that published one row holding the whole list.
+        let resolved = self
+            .schema_through_null_unions(&resolved, operation_id, diagnostics)
```
sauldhernandez, line 108: With nine nested `anyOf: [null, …]` wrappers around an array, this eight-step walk returns the ninth wrapper. Response classification marks it `Singleton`, but `import_schema` starts a fresh walk, reaches the array, and builds a list type, silently disabling pagination. Preserve the exhausted-walk state across both consumers, or make both stop at the same opaque result and report the depth limit.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -93,11 +135,38 @@ impl OpenApiImporter<'_> {
...
-            type_id_from_ref,
-        );
+        // A union of one real branch and `null` is imported as that branch.
+        //
+        // It is how a nullable reference has to be written once `nullable` is
+        // gone: a `$ref` carries no type of its own to list `null` in, so 3.1
+        // documents spell it `anyOf: [{type: 'null'}, {$ref: ...}]`. Left alone
+        // it matches no arm of the shape dispatch below and imports as opaque
+        // JSON, taking every field of the referenced type with it.
+        let walk = self.schema_through_null_unions(schema, operation_id, diagnostics)?;
```
sauldhernandez, line 145: This changes persisted Semantic IR for identical OpenAPI input, but the importer version remains `openapi-v10`. A source materialized by #2126 will load without the intended mismatch diagnostic even though regeneration now converts nullable refs from opaque JSON and may change response cardinality. Bump `OPENAPI_IMPORTER_VERSION` to `openapi-v11` so provenance remains truthful.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -516,9 +585,68 @@ impl OpenApiImporter<'_> {
...
+    ) -> Option<NullUnionWalk> {
+        let mut site = schema.clone();
+        let mut followed = BTreeSet::new();
+        let mut first_reference = None;
+        if let Some(reference) = site.get("$ref").and_then(Value::as_str) {
+            followed.insert(reference.to_string());
+            first_reference = Some(reference.to_string());
+        }
+        for _ in 0..MAX_NULL_UNION_DEPTH {
+            let resolved = self.resolve_ref(&site, operation_id, diagnostics)?;
```
sauldhernandez, line 614: OpenAPI 3.1 `$ref` branches can carry assertion siblings. For `anyOf: [{type: null}, {$ref: Base, properties: {extra: ...}, required: [extra]}]`, this resolves directly to `Base`, silently dropping `extra`. Compose the selected branch before resolving it while retaining its reference identity, or leave assertion-bearing branches opaque.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -516,9 +585,68 @@ impl OpenApiImporter<'_> {
...
+            let Some(branch) = null_union_branch(&resolved).cloned() else {
+                return Some(NullUnionWalk::ending_at(&site, resolved, first_reference));
+            };
+            if let Some(reference) = branch.get("$ref").and_then(Value::as_str) {
+                first_reference.get_or_insert_with(|| reference.to_string());
+                if !followed.insert(reference.to_string()) {
+                    return Some(NullUnionWalk::ending_at(&site, resolved, first_reference));
+                }
+            }
+            site = branch;
```
sauldhernandez, line 624: Stepping into the real branch discards removed keywords on the wrapper before either caller can report them. For example, a 3.1 schema with `nullable: true` around a `[null, string]` union no longer emits the existing removed-keyword warning. Preserve or report each resolved wrapper before stepping into its real branch.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -516,9 +585,68 @@ impl OpenApiImporter<'_> {
...
+        let mut schema = schema;
+        loop {
+            if let Some(description) = schema.get("description").and_then(Value::as_str) {
+                return description.to_string();
+            }
+            // A nullable spelling wraps the schema that has the description. The
+            // wrapper says nothing itself, so without this a field written
+            // `anyOf: [{type: 'null'}, {$ref: ...}]` lost the column description
+            // that the same field written as a bare `$ref` keeps.
+            let Some(branch) = null_union_branch(schema) else {
```
sauldhernandez, line 646: `field_description` stops when the unwrapped branch is a `$ref` and reads only that target's direct description. For `field: anyOf: [null, $ref A]` where A is `anyOf: [null, $ref B]` and only B is described, `import_schema` reaches B but the field description stays empty. Continue through local refs with cycle protection and add a ref-separated wrapper-chain test.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -535,6 +663,54 @@ impl OpenApiImporter<'_> {
...
+    for keyword in ["anyOf", "oneOf"] {
+        let Some(branches) = schema.get(keyword).and_then(Value::as_array) else {
+            continue;
+        };
+        // A union has to have lost something to `null` to be one of these; a
+        // single-branch `anyOf` is just a wrapper and keeps its own reading.
+        if branches.len() < 2 {
+            continue;
+        }
+        let mut real = branches.iter().filter(|branch| !declares_only_null(branch));
```
sauldhernandez, line 682: `declares_only_null` never follows `$ref`, so `anyOf: [{$ref: '#/components/schemas/Null'}, {$ref: '#/components/schemas/Party'}]` treats both as real even when `Null` is `{type: null}`. This valid 3.1 nullable union remains opaque JSON and drops `Party`'s fields. Resolve local refs while classifying branches, leaving unresolved or cyclic unions opaque.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -535,6 +663,54 @@ impl OpenApiImporter<'_> {
...
+        let only = real.next()?;
+        if real.next().is_some() {
+            return None;
+        }
+        return Some(only);
```
sauldhernandez, line 687: This returns only the non-null arm even when the union wrapper has assertion siblings. In OpenAPI 3.1, `{anyOf: [null, {$ref: Base}], properties: {extra: ...}, required: [extra]}` requires `extra`, but this path resolves and imports only `Base`, silently dropping that field. Unwrap only annotation-only wrappers, or compose supported assertions with the selected arm and diagnose unsupported ones.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -535,6 +663,54 @@ impl OpenApiImporter<'_> {
...
+/// shapes and send it back to opaque JSON.
+fn declares_only_null(schema: &Value) -> bool {
+    let declares_null_type = match schema.get("type") {
+        Some(Value::String(value)) => value == "null",
+        Some(Value::Array(values)) => {
+            !values.is_empty() && values.iter().all(|value| value.as_str() == Some("null"))
+        }
+        _ => false,
+    };
+    let is_null_constant = schema.get("const").is_some_and(Value::is_null)
```
sauldhernandez, line 706: This treats `const: null` as a null-only branch in OpenAPI 3.0, where `const` is unsupported and deliberately ignored. For `anyOf: [{const: null}, {$ref: Party}]`, the first branch can match arbitrary values, but this drops it and imports the union as `Party`, exposing fields those values need not have. Gate `const` on the dialect and add a 3.0 regression case; `enum: [null]` can remain shared.

sauldhernandez (approved):

PR #2132, opened 2026-08-11
feat(reef): add catalog-qualified schema routes

iaptsiauri:


apps/reef/app/routes/schema-table-route.tsx
lazybean, line 1: The name of this file is confusing.   
It is not a route, but a route component.  
  
The route is actually apps/reef/app/routes/schema-table.tsx

lazybean (approved):

PR #2139, opened 2026-08-11
feat(reef): add a runtime features list component

AlbertQM:
## Summary

Adding a presentation component for Coral "Runtime features" (kinda like our own feature flags)

## Test plan

image

### DIFF
--- a/apps/reef/app/components/runtime-features-list/index.ts
+++ b/apps/reef/app/components/runtime-features-list/index.ts
@@ -1,0 +1,7 @@
+export {
+  RuntimeFeatureRow,
+  RuntimeFeaturesList,
+  type RuntimeFeatureListItem,
+  type RuntimeFeatureRowProps,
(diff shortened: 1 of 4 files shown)


lazybean:
Interestingly the header in light mode is same color as the rest of the surface. Is that accidental, or on purpose? Or a wax issue?  
  
https://6a3cfca9530ba1e4c7574952-ouwukxkocn.chromatic.com/?path=/docs/components-runtimefeatureslist--docs&globals=themeComparison:!true

lazybean:
> Interestingly the header in light mode is same color as the rest of the surface. Is that accidental, or on purpose? Or a wax issue?
>
> https://6a3cfca9530ba1e4c7574952-ouwukxkocn.chromatic.com/?path=/docs/components-runtimefeatureslist--docs&globals=themeComparison:!true  
> 

​It is a wax/table feature/issue

lazybean (approved):

PR #2143, opened 2026-08-11
feat(reef): add a runtime features settings page

AlbertQM:
## Summary

Adding a new page to settings, which allows users to update Coral's runtime features.

## Test plan

Update a value, check it is correctly written in the config.
Restart Coral, check flag is correctly enabled/disabled

image


lazybean (commented):
How does this work with user role? can anyone change any feature?

lazybean (approved):
thank you

AlbertQM:
> How does this work with user role? can anyone change any feature?

Anyone can change a feature, and features are global (not per workspace). The UI is following the Coral CLI logic 1:1 -- I see how we'll want to tweak this once we have user roles (and probably restrict these to Coral admins)!
I can see how we'll also need some user-specific features at some point (usually called user preferences)

PR #2145, opened 2026-08-12
feat(engine): add query memory tracking

sauldhernandez:

The first two PRs build the engine-side measurement path. This PR defines the callback contract and the execution-scoped tracker; the next PR wires them through query execution.

## Intent

Measure the highest total number of bytes reserved through DataFusion's memory pools during one query execution without coupling `coral-engine` to OpenTelemetry.

## What it enables

`QueryRuntimeConfig` can now accept an optional `QueryMemoryObserver`. A shared tracker follows every memory-pool wrapper created for the execution, records the combined reservation peak, and reports a `QueryMemoryObservation` of `Success`, `Error`, or `Cancelled` only after all tracked reservations have been released.

Observer panics are contained. If a reservation leaks, Coral omits the observation instead of reporting an incomplete value.

## Usage examples

Set `QueryRuntimeConfig::query_memory_observer` to a `QueryMemoryObserver` callback when a caller wants the final measurement. Leave it unset for the existing behavior.


sauldhernandez:

PR #2146, opened 2026-08-12
feat(engine): observe top-level query memory

sauldhernandez:

The first two PRs build the engine-side measurement path. The previous PR defines the tracker; this PR applies it to complete top-level query executions.

## Intent

Produce one coherent DataFusion reservation-peak observation for each `execute_prepared` execution, including a dependent-join fallback attempt when one is required.

## What it enables

Primary and fallback task contexts now share the same execution tracker, so overlapping reservations from both attempts contribute to one peak. Successful executions report `Success`, executions that return an error report `Error`, and dropping an in-flight execution reports `Cancelled`. Coral waits for tracked reservations to be released before calling the observer.

## Usage examples

Execute prepared queries as before. When `QueryRuntimeConfig::query_memory_observer` is configured, the callback receives one `QueryMemoryObservation` after execution and reservation cleanup finish.


sauldhernandez:

PR #2147, opened 2026-08-12
feat(app): add query memory metric primitives

sauldhernandez:

This PR starts the app-telemetry part of the stack by defining the instruments that the following PRs populate.

## Intent

Add the app-side metric primitives for query concurrency, DataFusion reservation peaks, and estimated Arrow result memory.

## What it enables

The telemetry layer now provides:

- the attribute-free `coral.query.active` gauge;
- the `coral.query.memory.datafusion_reserved_peak` byte histogram; and
- the `coral.query.result.arrow.estimated_occupied_memory` byte histogram.

The histograms use explicit byte buckets from 4 KiB through 16 GiB and only the bounded `operation` and `status` attributes. The recorders omit measurements that cannot be represented as `u64` values.

## Usage examples

No new telemetry configuration is required. The next layers call these primitives from the query lifecycle, and the existing OTLP metrics pipeline exports their data.

### DIFF
--- a/crates/coral-app/src/telemetry/metrics.rs
+++ b/crates/coral-app/src/telemetry/metrics.rs
@@ -1,19 +1,48 @@
 //! Shared query metric instruments.
 
-use std::sync::RwLock;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::{Arc, RwLock};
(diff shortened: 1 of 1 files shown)


sauldhernandez:

PR #2148, opened 2026-08-12
feat(app): count active query operations

sauldhernandez:

This and the next two PRs connect the app-side instruments to real query work. This PR covers in-flight query concurrency.

## Intent

Count query operations from their first poll until they complete or are dropped.

## What it enables

The shared `run_query_operation` wrapper now counts `execute_sql`, `explain_sql`, `list_tables`, `list_catalog`, and `describe_table` in `coral.query.active`. The count increments on the first poll and is released on success, handled error, cancellation, or any other future drop, so abandoned work cannot leave the gauge stuck above zero.

## Usage examples

Collect `coral.query.active` through the existing OTLP metrics pipeline. It reports the current number of app query operations routed through `run_query_operation`, has no attributes, and reports zero when Coral is idle.

### DIFF
--- a/crates/coral-app/src/query/manager.rs
+++ b/crates/coral-app/src/query/manager.rs
@@ -1186,6 +1186,7 @@ where
     RowCount: FnOnce(&T) -> Option<u64>,
 {
     let started_at = Instant::now();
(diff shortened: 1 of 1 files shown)


sauldhernandez:

PR #2149, opened 2026-08-12
feat(app): export query memory measurements

sauldhernandez:

The surrounding app PRs cover concurrency and encoded output size. This PR connects engine reservation tracking and successful Arrow result estimates to app telemetry.

## Intent

Export two distinct measurements for `execute_sql`: DataFusion's peak reserved bytes during execution and the estimated occupied memory of the returned Arrow batches.

## What it enables

The `execute_sql` runtime now installs the engine memory observer; ordinary runtimes and source-validation runtimes remain outside this metric population. Coral records reservation peaks for success, handled error, and cancellation, and records the Arrow estimate only for successful results.

The measurements are exported as `coral.query.memory.datafusion_reserved_peak` and `coral.query.result.arrow.estimated_occupied_memory`. Matching `coral.query.memory.datafusion` and `coral.query.result.arrow` spans keep the observations in the query trace without copying SQL or identity fields onto the measurement spans.

## Usage examples

Collect both histograms through the existing OTLP pipeline. Interpret them separately: one measures DataFusion reservations during execution, while the other estimates the memory occupied by the returned Arrow batches. They cover different scopes and are not additive.

### DIFF
--- a/crates/coral-app/src/query/manager.rs
+++ b/crates/coral-app/src/query/manager.rs
@@ -7,13 +7,14 @@ use std::time::Instant;
 
 use coral_engine::{
     CatalogInfo, CoralQuery, CoreError, DescribeTableInfo, PreparedQueryRuntime, QueryExecution,
-    QueryExecutionProvenance, QueryPlan, QueryRuntimeConfig, QueryRuntimeContext, QuerySource,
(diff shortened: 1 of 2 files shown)


sauldhernandez:

PR #2150, opened 2026-08-12
feat(app): export Arrow IPC encoded size

sauldhernandez:

The previous PR measures in-memory Arrow results. This PR adds the corresponding serialized-output measurement at the gRPC response boundary.

## Intent

Measure the exact byte length of successfully encoded Arrow IPC query responses.

## What it enables

Successful encoding now records `coral.query.result.ipc.encoded_size` and an `encoded_size_bytes` field on the `coral.query.result.ipc` span. Encoding failures mark the span as failed and emit no size metric; an unrepresentable payload length is also omitted instead of being truncated.

## Usage examples

Collect `coral.query.result.ipc.encoded_size` through the existing OTLP pipeline. Treat it as a separate output-size signal alongside `coral.query.result.arrow.estimated_occupied_memory`: it measures serialized bytes, not live memory, and is not an additive component of the Arrow estimate.

### DIFF
--- a/crates/coral-app/src/query/service.rs
+++ b/crates/coral-app/src/query/service.rs
@@ -1,6 +1,7 @@
 //! Implements the gRPC `QueryService`.
 
 use std::collections::HashSet;
+use std::io::Write;
 
 use arrow::datatypes::SchemaRef;
 use arrow::ipc::writer::StreamWriter;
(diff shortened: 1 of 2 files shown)


sauldhernandez:

PR #2151, opened 2026-08-12
feat(app): export process memory telemetry

sauldhernandez:

This final layer adds process-level context for the query measurements and documents how to interpret the complete set of signals.

## Intent

Export the current Coral process's resident memory and document the scope, limits, and privacy properties of every new query and process memory metric.

## What it enables

The attribute-free `coral.process.memory.resident` gauge samples Linux `statm` resident pages, macOS physical resident memory, or the Windows working set. Unsupported platforms, failed reads, and overlapping collection attempts omit that data point rather than repeating a stale value; a genuine zero is still exported.

The OpenTelemetry guide now lists all query metrics, corrects the operation coverage of the existing count, duration, and row metrics, and explains platform differences, measurement exclusions, failure behavior, and why the memory signals must not be added together.

## Usage examples

Collect `coral.process.memory.resident` through the existing OTLP pipeline and view it beside `coral.query.active` for nearby context. The two gauges are sampled independently, so do not treat them as an atomic snapshot or divide resident memory by the active-query count to estimate per-query memory.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1368,11 +1368,13 @@ dependencies = [
  "fs2",
  "jsonwebtoken",
  "keyring-core",
+ "memory-stats",
  "opentelemetry",
  "opentelemetry-appender-tracing",
  "opentelemetry-otlp",
  "opentelemetry-proto",
  "opentelemetry_sdk",
+ "page_size",
  "prost",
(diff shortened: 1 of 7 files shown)


sauldhernandez:

PR #2154, opened 2026-08-12
feat(spec)!: import OpenAPI 3.1 documents

### LINKED ISSUE #1321: Add OpenAPI 3.1 support

simonwhitaker:
Closes #1321.

The v4 importer rejected any document that did not declare `3.0.x`, so a provider shipping 3.1 could not be imported at all. OpenAI, Discord and Tailscale all publish `3.1.0` today.

## Not versioned importer logic

The issue anticipated one importer per version. Measuring three real 3.1 documents says that is not what the delta costs us: the only 3.1 change this importer can even observe is how nullability is spelled. Everything else it reads — `type`, `properties`, `items`, `allOf`, `$ref`, `enum`, `required` — is unchanged, and the shared JSON Schema helpers already tolerate `$defs` and `prefixItems`.

The two spellings also coexist in the wild, which is the argument against dispatching on the declared version at all: OpenAI publishes a `3.1.0` document that still uses the 3.0 `nullable` keyword in 112 places. A 3.1 branch that stopped reading `nullable` would break a document that imports correctly today.

So the gate accepts `3.0.` and `3.1.`, and a normalization pre-pass rewrites the 3.1 nullability spellings into the 3.0 forms the pipeline already reads. There is no version branch anywhere else. The duplicated gate in `import.rs` and `document.rs` becomes one `validate_supported_openapi_version`.

## What the pre-pass rewrites

Two rules, both requiring a literal `null` type:

- `anyOf`/`oneOf` with exactly one non-`null` variant becomes that variant, merged over the keys the union sat beside.
- `type: [T, "null"]` becomes `type: T`.

A union with more than one declared shape is left alone. It describes a choice between shapes, which is what it did before this PR and what the inference passes are right to refuse.

The merge direction matters for the `$ref` case. A nullable ref is routinely paired with a sibling `description`, and `field_description` prefers that sibling over the referent's own, so replacing the node outright would drop the documentation. Keeping the `$ref` is most of the value: `import_schema` names a type after the ref it arrived through, so unwrapping recovers 291 named types in Discord's document and 209 in OpenAI's that would otherwise import as anonymous JSON.

Nothing synthesizes `nullable: true` in its place. `IrType.nullable` has no non-test reader on this path, and `nullable` is absent from `ANNOTATION_KEYS`, so it reaches `schema_validation_fingerprint` — a subtype re-declaring an inherited property as nullable would then read as a `PropertyConflict` and discard the whole composed type. #2167, stacked on this PR, removes the write-only `IrType.nullable` and `IrField.nullable` outright.

## Why one pre-pass rather than teaching each reader

The parsed document is threaded into every consumer, so normalizing it once reaches schema import, response classification, row-path inference, cursor detection and the validation fingerprints together. Patching read sites instead means six of them, including `schema_uses_alternation`, which the MCP importer shares.

Two of those consumers are wrong today in ways nothing else would catch. `{type: ["array","null"]}` reads as `None` through `Value::as_str`, so it lands on the `object` default and imports as opaque JSON — and at a response root the same array is classified `Singleton` when it is a collection. Behind a union it is worse: row-path inference refuses alternation outright, so an envelope whose rows sit behind `anyOf: [{type: array}, {type: "null"}]` loses its row path and the whole relation collapses to a single JSON row.

## Applied unconditionally

No version branch guards the pass. Both rules need a literal `null` type, which 3.0 does not have — it spells the same thing with `nullable` — so the pass is provably inert on a valid 3.0 document, the existing 3.0 suite exercises it for free, and a 3.0-labelled document that smuggles in a 3.1 spelling is fixed rather than misread.

A single-variant union with no `null` member is deliberately not unwrapped: GitHub's 3.0 document spells nullable refs as `anyOf: [{$ref: T}]`, and unwrapping those would change the output of a source that imports correctly today. `leaves_3_0_alternation_schemas_unchanged` is the regression guard.

## Breaking change

`OPENAPI_IMPORTER_VERSION` moves `openapi-v9` → `openapi-v10`. The bump invalidates existing OpenAPI materializations, which are regenerated on next import.

## Evidence

Counted across the three real 3.1 documents that motivated the change:

| construct | OpenAI | Discord | Tailscale |
| --- | --- | --- | --- |
| `anyOf`/`oneOf` with one non-`null` variant | 844 | 305 | 1 |
| `type: [T, "null"]` | 0 | 911 | 16 |
| `nullable` (the 3.0 keyword) | 112 | 0 | 0 |

Unwrapping recovers what was being lost to `IrTypeShape::Json`: 381 `string`, 209 `$ref`, 67 `integer`, 59 `object` and 58 `array` in OpenAI's document; 291 `$ref` in Discord's.

Importing those documents end-to-end, via a throwaway harness:

| document | before | after |
| --- | --- | --- |
| Discord 3.1.0 | rejected outright | 242 operations, 2398 types, 12 row paths, 0 diagnostics |
| Tailscale 3.1.0 | rejected outright | 93 operations, 871 types, 13 row paths, 0 diagnostics |

## Impact

Measured with `cargo run -p xtask -- v4-metadata-report`, diffed against this branch's base.

The six v4 sources the base also has — 21,525 operations — come out byte-identical. Nothing in the catalog moves, which is what "provably inert on 3.0" is supposed to mean: every one of those sources declares 3.0, so neither rewrite has anything to match.

This branch adds a seventh, `integrations_sh_v4` (4 operations), for 21,529 in total.

## Out of scope

`paths` is optional in 3.1, but all three real documents declare it and a webhooks-only source would import zero operations, so the existing error is kept. Top-level `webhooks` is already ignored. `$ref` siblings are annotations in every observed case. Discord's 659 `const` declarations all sit inside multi-variant `oneOf` branches, which stay opaque either way. File-upload keywords (`contentMediaType`/`contentEncoding`) are excluded by design.

## Known gap, not addressed here

OpenAI's document still does not import, for a reason that predates this PR and is not version-specific: it declares `minimum: -9223372036854776000`, outside `i64` range, and `serde_yaml::from_slice::<serde_json::Value>` fails the whole parse. It reproduces on a minimal `openapi: 3.0.3` document with the same literal. The importer never reads `minimum`, so a keyword it ignores should not be able to fail the parse.

Filed as #2187. A `sources/v4/openai` manifest was prepared on this branch and has been dropped again until that lands: `v4-metadata-report` propagates the first import failure and writes its CSV only after the loop, so one unimportable source leaves no report for any source — including the baseline this PR is measured by.

## Testing

`make rust-checks` passes. 21 new tests, 505 passing in `coral-spec`:

- 10 unit tests in `normalize.rs` covering the sibling merge, every schema position, a variant that is itself a union, per-read normalization leaving children to their own reads, property names that collide with keywords (`type`, `anyOf`, `default`), example payloads, and the non-matching unions that must be left alone.
- 11 integration tests in `openapi_tests.rs`: a 3.1 document carrying all three spellings side by side, the version gate through both entry points (previously untested), a nullable array response root classified as a list, a row path inferred through a nullable-union envelope, a `$ref` pointing into a nullable union, `allOf` branches spelling nullability differently, a null parameter default read as no default, a cursor declared as a nullable union, multi-variant unions staying opaque, and the 3.0 regression guard.


crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -271,7 +275,7 @@ impl OpenApiImporter<'_> {
         diagnostics: &mut Vec<Diagnostic>,
     ) -> Result<(), AllOfMergeError> {
         let walked = with_resolved_json_schema(
-            self.document,
+            self.schema_root(),
```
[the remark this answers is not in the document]
simonwhitaker: Confirmed and fixed in 96fb5be84.

Reproduced end to end: a base declaring `id: {type: [string, "null"]}` and a subtype re-declaring it as `anyOf: [{type: string}, {type: "null"}]` imported as `allOf property 'id' conflicts` with the whole composed type falling back to an opaque JSON column — every field, not just the one in dispute.

It was a regression from moving to read-time normalization. Running the same document against the previous commit imports the object with both fields and no diagnostics: the document-wide pre-pass had rewritten every property in place, so the two spellings arrived identical. `direct_json_object_shape` and `collect_all_of_branches` clone `properties` entries straight out of the branch, one level below where the walk normalizes.

Both now read through the `SchemaRoot`, and `direct_json_object_shape` takes one rather than leaving a raw variant beside it. `v4-metadata-report` across all 21,524 operations is unchanged, and the case is covered by `importer_folds_all_of_branches_that_spell_nullability_differently`.

crates/coral-spec/src/v4/surfaces/openapi/normalize.rs
```diff
@@ -0,0 +1,470 @@
...
+/// reusing the narrower list would refuse 21 of the 844 reducible unions in
+/// `OpenAI`'s document over a sibling `example`.
+fn annotates(key: &str) -> bool {
+    // A vendor extension cannot constrain anything the importer reads.
+    key.starts_with("x-")
+        || matches!(
+            key,
+            "$comment"
+                | "default"
+                | "deprecated"
```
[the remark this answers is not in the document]
simonwhitaker: Fixed in 5aefa5eab — `discriminator` is now in the allowlist.

Agreed on the reasoning: a union that reduces to a single declared variant has nothing left to choose between, so the mapping describes the shape being promoted rather than constraining it further.

For the record it changes no document we have: neither Discord's 305 reducible unions nor OpenAI's 844 sit beside a discriminator, and the metadata report is byte-identical. This is hardening for specs we have not read.

sources/v4/openai/manifest.yaml
```diff
@@ -0,0 +1,14 @@
+name: openai
```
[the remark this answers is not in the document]
simonwhitaker: Fixed in 3dff7a06d — the manifest is now `openai_v4`. The directory stays `openai`, matching `github` holding `github_v4`.

crates/coral-spec/src/v4/surfaces/openapi/normalize.rs
```diff
@@ -0,0 +1,470 @@
...
+fn unwrap_nullable_union(object: &mut Map<String, Value>) -> bool {
+    let mut unwrapped = false;
+    for keyword in ["anyOf", "oneOf"] {
+        let Some(variant) = sole_declared_variant(object.get(keyword)).cloned() else {
+            continue;
+        };
+        if object.keys().any(|key| key != keyword && !annotates(key)) {
+            continue;
+        }
+        object.remove(keyword);
```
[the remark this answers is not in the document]
simonwhitaker: Addressed one layer up rather than here: #2167, stacked on this PR, removes `IrType.nullable` and `IrField.nullable` outright.

Both were write-only. Every `column.nullable` downstream — the Arrow field in `backends/common.rs`, the catalog snapshot hash, the transport proto — reads a *projection* column, and `projections/derive.rs` hardcodes `nullable: true` at every column site without consulting the IR. `IrField.nullable` was redundant on top of that, since `IrField` already carries `required` and the MCP importer computed the flag as `!required`.

Synthesizing `nullable: true` while unwrapping was the other option, and it has a real cost: `nullable` is absent from `ANNOTATION_KEYS`, so it reaches `schema_validation_fingerprint`, where a subtype re-declaring an inherited property as nullable then reads as a `PropertyConflict` and discards the composed type. That is the same failure this PR just fixed from a different direction.

crates/coral-spec/src/v4/surfaces/openapi/normalize.rs
```diff
@@ -0,0 +1,470 @@
...
+    let object = schema.as_object()?;
+    if !carries_a_rewritable_keyword(object) {
+        return None;
+    }
+    let mut object = object.clone();
+    let mut rewritten = false;
+    // A promoted variant can carry a union of its own. The document-wide pass
+    // this replaced normalized children first and never had to look twice;
+    // reading one schema at a time, the second look happens here.
+    for _ in 0..MAX_NORMALIZE_PASSES {
```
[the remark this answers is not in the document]
simonwhitaker: Half agreed, so I have left the cap and will explain rather than change it.

The loop already exits as soon as a pass rewrites nothing — `if !(unwrapped || collapsed) { break; }` — so the cap is not the normal exit, only a ceiling on how deep a chain of nested nullable unions is followed. Eight matches the ceilings the rest of the crate already imposes on the same document: `MAX_ALL_OF_DEPTH` is 8 and the inference walks use `MAX_DEPTH` 8, so a schema nesting deeper than that is truncated by those walks regardless of what this pass does.

Iterating unbounded would make this the one walk in the importer with no ceiling, on input we do not control. If a real document turns up nesting nullable unions more than eight deep, raising all of these together is the right change.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -57,7 +57,11 @@ impl OpenApiImporter<'_> {
         operation_id: &str,
         diagnostics: &mut Vec<Diagnostic>,
     ) -> Option<String> {
-        let resolved = self.resolve_ref(schema, operation_id, diagnostics)?;
+        // The type is named after the ref it arrived through, and a nullable
+        // `$ref` union only carries one once it is unwrapped — so the referring
+        // schema is read the same way the referent is.
+        let schema = self.read_schema(schema);
```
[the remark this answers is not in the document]
simonwhitaker: Real, but pre-existing rather than introduced here, so I would rather not fold it into this PR.

`resolve_ref` follows exactly one hop by design — the crate already documents that for `allOf` branches — so a component that is itself an alias (`A: {$ref: B}`) came back still holding a `$ref` before this PR too. Normalizing a nullable ref alias to `{$ref: B}` produces the same node the alias case already produced; it does not add a hop that was previously resolved.

Worth its own issue with a real document behind it. Neither 3.1 document I measured (Discord, OpenAI) contains a nullable ref alias chain — zero `$ref`s pointing at a component whose own schema is a single `$ref`.

crates/coral-spec/src/v4/surfaces/openapi/import.rs
```diff
@@ -124,14 +120,39 @@ impl<'a> OpenApiImporter<'a> {
...
         value: &Value,
         operation_id: &str,
         diagnostics: &mut Vec<Diagnostic>,
     ) -> Option<Value> {
-        match resolve_local_ref(self.document, value) {
-            Ok(resolved) => Some(resolved.clone()),
+        // Normalized before resolving as well as after: a nullable `$ref` union
+        // only becomes a `$ref` once it is unwrapped, and resolving the union
+        // itself would hand back the union.
+        let value = self.read_schema(value);
```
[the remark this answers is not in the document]
simonwhitaker: The behaviour is as described, but it is the intended reading rather than a contradiction.

The commit says `field_description` prefers the sibling over the referent's own, and that still holds — a *field* whose schema is a nullable `$ref` with a sibling `description` keeps the sibling, because `field_description` reads it off the referring node before resolution.

What the sibling does not do is rename the *type*. `import_schema` interns one `IrType` per component and takes its description from the component, which is right: the same component referenced from ten sites with ten different sibling descriptions has one type, and the last site to import it should not decide what that type is called. The per-site wording lives on the field, where it belongs.

crates/coral-spec/src/v4/surfaces/openapi/operations.rs
```diff
@@ -207,9 +208,9 @@ impl OpenApiImporter<'_> {
                     .get("in")
                     .and_then(Value::as_str)
                     .and_then(parse_parameter_location)?;
-                let schema = parameter_obj.get("schema").unwrap_or(&Value::Null);
+                let schema = self.read_schema(parameter_obj.get("schema").unwrap_or(&Value::Null));
```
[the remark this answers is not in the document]
simonwhitaker: I do not think there is a behaviour shift to document here.

The siblings win a conflict when a union is unwrapped, so a parameter declaring `default` keeps its own; the variant's `default` is only adopted when the parameter declares none. Reading it from a schema the importer understands is the point of normalizing before extraction — the alternative is reading `default` off a node whose `type` we then read differently.

The comparison with `import_request_body` does not quite land either: request bodies never read `default` at all. Their schema goes through `import_schema`, which normalizes via `read_schema` on the way in, so both paths do read normalized schemas — they just read different keywords from them.

crates/coral-spec/src/v4/mod.rs
```diff
@@ -5,7 +5,7 @@
 
 pub const V4_ARTIFACT_SCHEMA_VERSION: u32 = 6;
 pub const SURFACE_IMPORTER_VERSION: &str = "surface-import-v4";
-pub const OPENAPI_IMPORTER_VERSION: &str = "openapi-v9";
+pub const OPENAPI_IMPORTER_VERSION: &str = "openapi-v10";
```
sauldhernandez, line 8: The description has drifted from the branch: the bump here is `openapi-v9` -> `openapi-v10`, but the body's breaking-change section says v8 -> v9. Same for the testing section — the diff adds 20 tests (10 in `normalize.rs`, 10 in `openapi_tests.rs`), not 13 — and "the six importable v4 sources" predates this branch adding `sources/v4/integrations.sh` and `sources/v4/openai`. Worth refreshing before merge: this body becomes the squash commit documenting the materialization invalidation, so the wrong version string lands in history.

crates/coral-spec/src/v4/surfaces/json_schema.rs
```diff
@@ -242,7 +325,7 @@ pub(crate) fn direct_json_object_shape(schema: &Value) -> JsonObjectShape {
         .map(|properties| {
             properties
                 .iter()
-                .map(|(name, property)| (name.clone(), property.clone()))
+                .map(|(name, property)| (name.clone(), root.read(property).into_owned()))
```
sauldhernandez, line 328: Normalization here is one node deep, but the allOf conflict check compares whole subtrees: `json_schema_property_schemas_conflict` fingerprints via `schema_validation_fingerprint`, which recurses through `properties`/`anyOf` without reading through the `SchemaRoot`. So two allOf branches that spell a *nested* property's nullability differently — `meta: {type: object, properties: {id: {type: [string, "null"]}}}` vs the `anyOf` spelling — still fingerprint as a conflict, and `import_schema` drops the whole composed type to opaque JSON, every field included.

Narrow trigger and no worse than before this PR, so not blocking. Threading the `SchemaRoot` through `schema_validation_fingerprint` (calling `root.read` at each node) would close it where the comparison happens; otherwise a line in the module docs noting the rewrite is per-node would save the next reader the trace.

crates/coral-spec/src/v4/surfaces/openapi/document.rs
```diff
@@ -13,18 +13,38 @@ pub struct OpenApiDocumentMetadata {
...
-pub fn openapi_document_metadata(document_bytes: &[u8]) -> Result<OpenApiDocumentMetadata> {
-    let document: Value =
-        serde_yaml::from_slice(document_bytes).map_err(ManifestError::parse_yaml)?;
+/// Accepts the `OpenAPI` versions the importer reads, and rejects the rest.
+///
+/// 3.0 and 3.1 share one code path rather than one per version. The 3.1 changes
+/// this importer can even observe are how nullability is spelled — a `null`
+/// member in a `type` array, or a union whose only other variant is `null` —
+/// and [`super::normalize::normalize_nullable_unions`] rewrites both into the
+/// 3.0 forms before anything reads the document. Everything downstream is
```
sauldhernandez: `normalize_nullable_unions` doesn't exist — `normalize.rs` exports `normalized_schema`, so this intra-doc link resolves to nothing (rustdoc warns under `broken_intra_doc_links`). The same sentence says the rewrites happen "before anything reads the document", but normalization is applied per read through `SchemaRoot::normalized_by` — `normalize.rs`'s module doc says exactly that. Link `normalized_schema` and reword to "as each schema is read".

crates/coral-spec/src/v4/surfaces/openapi/normalize.rs
```diff
@@ -0,0 +1,487 @@
...
+/// document, and the existing 3.0 suite exercises it for free. A 3.0-labelled
+/// document that smuggles in a 3.1 spelling is read rather than misread, which
+/// is the same tolerance the importer already extends in the other direction by
+/// reading `nullable` out of 3.1 documents.
+pub(crate) fn normalized_schema(schema: &Value) -> Option<Value> {
+    let object = schema.as_object()?;
+    if !carries_a_rewritable_keyword(object) {
+        return None;
+    }
+    let mut object = object.clone();
```
sauldhernandez, line 66: This clone runs before either rewrite decides anything, and `carries_a_rewritable_keyword` only checks that a keyword is present. So any union that doesn't reduce — GitHub's `anyOf: [{$ref: T}]` spelling, any genuine alternation — is cloned and discarded on every read, on 3.0 documents as much as 3.1.

That contradicts three doc claims: line 50 here ("nothing is cloned unless a rewrite applies"), the module doc ("costs a clone only where a schema actually needs rewriting"), and `SchemaRoot::normalize` in json_schema.rs ("walked without a single clone").

No correctness impact, and the cloned node is small in the `$ref` case. Either decide reducibility before cloning (dry-run `sole_declared_variant` / the type-array check against the borrowed map), or soften the three comments to match the code.

crates/coral-spec/src/v4/surfaces/openapi/operations.rs
sauldhernandez, line 211: `default: null` becomes the literal string "null" on the wire. `annotates` includes `default`, so `{"anyOf": [{"type": "integer"}, {"type": "null"}], "default": null}` normalizes to `{"type": "integer", "default": null}`, and the `default_value` read below stringifies it — `json_schema_default_to_string` only special-cases `Value::String`, so `Value::Null` falls through to `to_string()` and yields `Some("null")`. That survives into `ValueSourceSpec::Filter { default: Some("null") }`; an unbound filter falls back to its default (`string_runtime_value`) and `build_query_pairs` sends any resolved value, so an unfiltered query on the operation sends `?limit=null` — which an integer parameter will typically 400 on.

This spelling is what Pydantic/FastAPI emit for `param: int | None = None`, i.e. the common case in the 3.1 documents this PR unlocks. Before the PR, 3.1 was rejected outright, so the union form never reached this read. The tests cover a sibling `default: 20` beating a variant's `default: 100`, but not `default: null`.

Dropping null defaults at the read site covers both cases and restores the no-default behavior (param omitted):

```rust
default_value: schema.get("default").filter(|v| !v.is_null()).map(json_schema_default_to_string),
```

crates/coral-spec/src/v4/surfaces/openapi/operations.rs
```diff
@@ -356,11 +357,11 @@ fn parameter_is_required(parameter_obj: &Map<String, Value>, location: IrInputLo
 /// the question wrapped-list inference needs — is this operation paginated? —
 /// without the two inferences having to predict each other.
 fn detect_pagination_contract(
-    document: &Value,
+    document: SchemaRoot<'_>,
```
sauldhernandez: Four functions here (360, 408, 447, 486) name this parameter `document`, while the other 17 `SchemaRoot<'_>` parameters in the PR — including `response_cursor_header`/`response_next_url_header` below, which receive this same value — call it `root`. Since `SchemaRoot::document()` returns the inner `&Value`, `document: SchemaRoot` reads like it's already the unwrapped document. Rename these four to `root`?

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -57,7 +57,11 @@ impl OpenApiImporter<'_> {
...
+        let resolved = self.resolve_ref(&schema, operation_id, diagnostics)?;
```
sauldhernandez, line 64: Two subschema pointers with the same trailing index collide on one type id. `type_id_from_ref` names the type after the last pointer segment, which for `#/components/schemas/Page/anyOf/0` and `#/components/schemas/Other/anyOf/0` is `0` — both become `type_0`. The second import hits the `self.types` cache and silently reuses the first schema's shape, with no diagnostic. I reproduced it: a response with `page` and `other` referencing those two pointers imports both as `type_0` with Page's fields; `Other`'s field disappears.

The naming predates this PR, but `importer_resolves_a_ref_that_points_into_a_nullable_union` now blesses subschema pointers as supported, and `anyOf/0` is the index nearly every 3.1 nullable-union pointer uses, so two such refs in one document collide by default. Deriving the id from more of the pointer when the last segment is numeric (`page_any_of_0`), or falling back to `suggested_id`, keeps them distinct.

crates/coral-spec/src/v4/surfaces/openapi/schemas.rs
```diff
@@ -294,7 +298,7 @@ impl OpenApiImporter<'_> {
             Ok(merge_result) => merge_result,
             // Converted rather than propagated: the walk error borrows from the
             // schema it walked, and the top-level caller passes a local.
-            Err(error) => Err(Self::all_of_walk_error(error, operation_id, diagnostics)),
+            Err(error) => Err(Self::all_of_walk_error(&error, operation_id, diagnostics)),
```
sauldhernandez, line 302: Two comments still justify this code by a borrow this PR removed — `JsonSchemaWalkError` owns its `String`s now (you already updated the doc comment on `all_of_walk_error` itself):

- Here (schemas.rs:299): "the walk error borrows from the schema it walked, and the top-level caller passes a local" — no longer true; the conversion survives only as an error-type translation.
- wrapped_lists.rs:196-199: "`?` would hand back a reference to a local" — also stale; the surviving reason is the second half, that a branch which cannot be walked holds no rows.

Worth rewording both so the next reader doesn't inherit a false rule.

sources/v4/openai/manifest.yaml
```diff
@@ -0,0 +1,14 @@
+name: openai_v4
```
sauldhernandez: This manifest points at a descriptor the PR body says cannot be parsed yet: `minimum: -9223372036854776000` overflows i64 and `serde_yaml::from_slice::<serde_json::Value>` fails the whole document. Reproduced on serde_yaml 0.9.34, and the live descriptor at this URL still carries the literal.

That breaks the default `cargo run -p xtask -- v4-metadata-report`. It walks every source under `sources/v4`, propagates the first import failure with `?` (xtask/src/metadata_report.rs:198), and writes the CSV only after the loop — so it now aborts at `openai_v4` and produces no report for any source, including the ones that do import. AGENTS.md names that command as the before/after baseline for v4 inference changes. `coral source add --file sources/v4/openai/manifest.yaml` fails the same way, since materialization parses the same bytes.

Hold this manifest until the `minimum` fix lands (the body already calls that "worth its own change"), or make the report warn and skip a source it can't import instead of aborting.

sources/v4/openai/manifest.yaml
```diff
@@ -0,0 +1,14 @@
...
+dsl_version: 4
+inputs:
+  OPENAI_API_KEY:
+    kind: secret
```
sauldhernandez: Both new manifests skip two conventions the other seven `sources/v4` manifests follow: a `test_queries` entry (all seven carry the key, five with real smoke queries) and a `hint` on secret inputs. The hint shows next to the field during `coral source add --interactive`, so `OPENAI_API_KEY` currently asks for a secret with no pointer to where to get one (platform.openai.com/api-keys). `x/manifest.yaml` is the closest shape to copy — openapi surface, single bearer secret.

sauldhernandez (approved):
Approving — the importer change itself is well-designed and well-tested, and the normalize pass held up when I traced it against every schema read site. The `SchemaRoot` seam is in the right place, and the reverse direction is unusually well guarded (`annotates()` as an allowlist, `sole_declared_variant` refusing genuine multi-shape unions).

Two things worth handling, both outside the importer:

- `sources/v4/openai/manifest.yaml` points at a descriptor that can't parse yet, and `v4-metadata-report` propagates the first import failure with `?` before writing anything — so the baseline command AGENTS.md mandates for inference changes currently produces no report for any source, including the ones that do import. Holding the manifest until the `minimum` fix lands, or making the report warn-and-skip, closes it.
- `default: null` now reaches `json_schema_default_to_string` and stringifies to the literal `"null"`. That's the spelling Pydantic/FastAPI emit for `param: int | None = None`, so it's the common case in the 3.1 documents this unlocks.

The rest are minor or advisory: a doc link to a function that no longer exists, some description numbers that drifted from the branch (the bump is v9→v10), and a few consistency nits. Details inline.

crates/coral-spec/src/v4/surfaces/json_schema.rs
simonwhitaker: Confirmed, and filed as #2190 rather than fixed here.

Reproduced end to end with your exact shape — a base declaring `meta.id` as `{type: [string, "null"]}` and a subtype re-declaring it as `anyOf: [{type: string}, {type: "null"}]`:

```
diagnostics: ["allOf property 'meta' conflicts in operation 'things_get'"]
output type_ref: json
types: things_get_row -> Json
```

Worth noting how far the damage reaches: I gave the subtype an unrelated `extra: {type: string}` with no competing declaration anywhere, and it disappears from the catalog too. The conflict is reported against one property and discards every field of the composed type.

Your diagnosis is exactly right about the cause. `schema_validation_fingerprint` takes a bare `schema: &Value` and recurses through `properties`/`anyOf` on raw nodes, so the per-node rewrite never reaches them. `schema_type_validation_fingerprint` only sorts a `type` array — it does not collapse `["string", "null"]`, and nothing reconciles it with the `anyOf` spelling.

Taking the follow-up rather than the doc note. A comment saying the rewrite is per-node would describe the gap accurately but still leave a document silently importing as opaque JSON, and threading the root through the fingerprint helpers reaches the MCP importer as well, which is more than this branch should carry. The issue writes up all three options; (1), threading `SchemaRoot` through and calling `root.read` at each node, is where I would start — it finishes what `96fb5be84` began one level up.

Agreed it is not blocking: before this PR a 3.1 document was rejected outright, so the shape was unreachable. It is a gap in new capability, not a break in old behaviour.

PR #2155, opened 2026-08-12
feat(reef): enforce same-origin OAuth metadata

sauldhernandez:
## Intent

Keep Reef's OAuth client and protected-resource metadata on the validated issuer origin, including token exchange redirects.

## What it enables

Reef can participate in Coral OAuth without trusting attacker-controlled metadata URLs or leaking authorization codes and PKCE verifiers across redirects.

## Usage examples

Configure `REEF_AUTH_MODE=required`, `REEF_AUTH_ISSUER`, `REEF_PUBLIC_URL`, and a session secret. Reef serves its client metadata at `/.well-known/oauth-client` and validates discovered OAuth endpoints against the configured issuer.


lazybean (approved):
thank you

PR #2157, opened 2026-08-12
feat(reef): serve Reef from a production Node entry

sauldhernandez:
## Intent

Provide a production Node entrypoint that serves Reef's SSR build and rejects invalid runtime configuration before accepting traffic.

## What it enables

Operators get static asset serving, graceful shutdown, disconnect cancellation, `/healthz`, and a production boot probe through the real React Router handler.

## Usage examples

Build Reef with `npm run build --prefix apps/reef`, set the required runtime variables, and run `node apps/reef/server.js`. Probe `GET /healthz` for process liveness.

### DIFF
--- a/apps/reef/server.js
+++ b/apps/reef/server.js
@@ -1,0 +1,187 @@
+import { createReadStream } from 'node:fs'
+import { stat } from 'node:fs/promises'
+import { createServer } from 'node:http'
+import { extname, resolve, sep } from 'node:path'
+import { pipeline } from 'node:stream/promises'
+import { Readable } from 'node:stream'
+import { fileURLToPath, pathToFileURL } from 'node:url'
+
+import { createRequestHandler, RouterContextProvider } from 'react-router'
+
+const CLIENT_ROOT = resolve(fileURLToPath(new URL('./build/client/', import.meta.url)))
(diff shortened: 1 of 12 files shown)


apps/reef/server.js
lazybean, line 1: react-router provide a template there. It seems simpler?  
  
https://github.com/remix-run/react-router-templates/blob/main/node-custom-server/server.js

lazybean (approved):
Looks good, but consider using one of react-router template instead of making or own server.js perhaps?

apps/reef/server.js
sauldhernandez: Thanks — I ran a spike using @react-router/express 7.17.0 to match Reef. It reduced our server from 208 to 131 lines, but added four direct dependencies and 73 package versions absent from the current lockfile (about 2.5 MiB installed). In a bounded local benchmark, median startup was about 11% slower, health-route throughput about 20% lower, and static throughput about 39% lower. We still needed custom pre-listen validation, HOST/PORT handling, cache policy, the disabled-auth warning, graceful shutdown, and failure diagnostics. Given that tradeoff, I am keeping the current Node adapter, which adds no new dependencies.

PR #2158, opened 2026-08-12
build(reef): add the Reef container image

sauldhernandez:
## Intent

Package the production Reef server as a COPY-only, non-root linux/amd64 container with executable topology evidence.

## What it enables

Reef can run beside Coral through shared loopback, HTTPS termination, or explicitly opted-in h2c, with separate liveness and Coral readiness probes.

## Usage examples

Run `make reef-docker-build` to create `reef:local`, then `make reef-docker-smoke` to exercise all three topologies, invalid configuration, non-root identity, and a read-only filesystem.


lazybean (approved):
thank you!

PR #2159, opened 2026-08-12
ci(reef): publish and coordinate Reef release images

sauldhernandez:
## Intent

Publish immutable Reef release images and coordinate Coral and Reef moving aliases through one release workflow.

## What it enables

Each `X.Y.Z` Reef tag is compare-and-set immutable, signed, and source-verified. Eligible `X.Y` and `latest` aliases advance as a verified Coral/Reef pair before the GitHub release is marked complete.

## Usage examples

Publishing a `v1.2.3` prerelease builds `ghcr.io/withcoral/reef:1.2.3`. If it is the newest stable release, the coordinator advances and re-reads both repositories' `1.2` and `latest` aliases; retries converge partial failures idempotently.

### DIFF
--- a/.github/workflows/docker-aliases.yml
+++ b/.github/workflows/docker-aliases.yml
@@ -1,0 +1,78 @@
+name: Coordinate Docker aliases
+on:
+  workflow_call:
+    inputs:
+      tag:
+        required: true
+        type: string
+      coral-digest:
+        required: true
+        type: string
+      reef-digest:
+        required: true
+        type: string
+
+permissions:
+  contents: read
+concurrency:
+  # Alias publication is one cross-repository state transition at a time.
+  group: coordinated-docker-aliases
+  queue: max
+
+jobs:
+  aliases:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    env:
+      TAG: ${{ inputs.tag }}
+      CORAL_DIGEST: ${{ inputs.coral-digest }}
+      REEF_DIGEST: ${{ inputs.reef-digest }}
+    steps:
+      - name: Log in to GHCR
+        uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up Buildx
+        uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
+        with:
+          cache-binary: false
+
+      - name: Advance and verify coordinated aliases
+        shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          version="${TAG#v}"
+          line="${version%.*}"
+          # The release remains a prerelease until both aliases verify.
+          draft="$(gh release view "$TAG" --repo "$GITHUB_REPOSITORY" --json isDraft --jq .isDraft)"
+          test "$draft" = false
+
+          stable="$(gh release list --repo "$GITHUB_REPOSITORY" \
+            --exclude-pre-releases --exclude-drafts --limit 1000 \
+            --json tagName --jq '.[].tagName')"
+          candidates="$(printf '%s\n%s\n' "$stable" "$TAG" | grep -v '^$' | sort -u)"
+          newest="$(printf '%s\n' "$candidates" | sort -V | tail -1)"
+          newest_line="$(printf '%s\n' "$candidates" | grep -E "^v${line//./\\.}\." | sort -V | tail -1)"
+
+          aliases=()
+          [ "$TAG" != "$newest_line" ] || aliases+=("$line")
+          [ "$TAG" != "$newest" ] || aliases+=(latest)
+          for alias in "${aliases[@]}"; do
+            # A partial failure is deliberately a failed release workflow. A
+            # retry is idempotent and converges both aliases before completion.
+            docker buildx imagetools create --tag "ghcr.io/withcoral/coral:${alias}" \
+              "ghcr.io/withcoral/coral@${CORAL_DIGEST}"
+            docker buildx imagetools create --tag "ghcr.io/withcoral/reef:${alias}" \
+              "ghcr.io/withcoral/reef@${REEF_DIGEST}"
+            coral_actual="$(docker buildx imagetools inspect "ghcr.io/withcoral/coral:${alias}" --format '{{.Manifest.Digest}}')"
+            reef_actual="$(docker buildx imagetools inspect "ghcr.io/withcoral/reef:${alias}" --format '{{.Manifest.Digest}}')"
+            test "$coral_actual" = "$CORAL_DIGEST"
+            test "$reef_actual" = "$REEF_DIGEST"
+          done
--- a/.github/workflows/docker-aliases.yml
+++ b/.github/workflows/docker-aliases.yml
@@ -11,6 +11,10 @@ on:
       reef-digest:
         required: true
         type: string
+    outputs:
+      make-latest:
+        description: Whether this tag is the newest eligible stable release.
+        value: ${{ jobs.aliases.outputs.make-latest }}
 
 permissions:
   contents: read
@@ -29,6 +33,8 @@ jobs:
       TAG: ${{ inputs.tag }}
       CORAL_DIGEST: ${{ inputs.coral-digest }}
       REEF_DIGEST: ${{ inputs.reef-digest }}
+    outputs:
+      make-latest: ${{ steps.advance.outputs.make-latest }}
     steps:
       - name: Log in to GHCR
         uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
@@ -43,6 +49,7 @@ jobs:
           cache-binary: false
 
       - name: Advance and verify coordinated aliases
+        id: advance
         shell: bash
         env:
           GH_TOKEN: ${{ github.token }}
@@ -62,8 +69,13 @@ jobs:
           newest_line="$(printf '%s\n' "$candidates" | grep -E "^v${line//./\\.}\." | sort -V | tail -1)"
 
           aliases=()
+          make_latest=false
           [ "$TAG" != "$newest_line" ] || aliases+=("$line")
-          [ "$TAG" != "$newest" ] || aliases+=(latest)
+          if [ "$TAG" = "$newest" ]; then
+            aliases+=(latest)
+            make_latest=true
+          fi
+          echo "make-latest=${make_latest}" >> "$GITHUB_OUTPUT"
           for alias in "${aliases[@]}"; do
             # A partial failure is deliberately a failed release workflow. A
             # retry is idempotent and converges both aliases before completion.
--- a/.github/workflows/docker-aliases.yml
+++ b/.github/workflows/docker-aliases.yml
@@ -31,7 +31,7 @@ jobs:
       packages: write
     env:
       TAG: ${{ inputs.tag }}
-      CORAL_DIGEST: ${{ inputs.coral-digest }}
+      PUBLISHED_CORAL_DIGEST: ${{ inputs.coral-digest }}
       REEF_DIGEST: ${{ inputs.reef-digest }}
     outputs:
       make-latest: ${{ steps.advance.outputs.make-latest }}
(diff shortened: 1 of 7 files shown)


.github/workflows/release.yml
```diff
@@ -608,3 +545,88 @@ jobs:
...
+    permissions:
+      contents: read
+      packages: write
+    uses: ./.github/workflows/docker-aliases.yml
+    with:
+      tag: ${{ needs.validate-release-ref.outputs.tag-name }}
+      coral-digest: ${{ needs.publish-docker.outputs.digest }}
+      reef-digest: ${{ needs.publish-reef-docker.outputs.digest }}
+
+  promote-release:
```
simonwhitaker, line 577: **Blocker: this job is missing `environment: release`, so it can never run.**

`RELEASE_APP_ID` exists only as a `release`-environment variable, and `RELEASE_APP_PRIVATE_KEY` / `DISCORD_CHANGELOG_WEBHOOK` exist only as `release`-environment secrets — there are no repo-level equivalents (repo secrets are just `CHROMATIC_PROJECT_TOKEN`, and there are no repo variables).

The promotion step used to live in `publish-version`, which declares `environment: release`. This new standalone job doesn't, so `vars.RELEASE_APP_ID` and `secrets.RELEASE_APP_PRIVATE_KEY` resolve to empty strings and `actions/create-github-app-token` fails on the first step. Every release then fails at promotion: since nothing else promotes the release under the new design, the GitHub release stays a prerelease forever and no Discord announcement is posted.

Adding `environment: release` also restores the environment's deployment protection to the step that flips a release to stable — right now that protection is silently dropped.

.github/workflows/reef-docker-publish.yml
```diff
@@ -0,0 +1,157 @@
...
+          VERSION: ${{ steps.tag.outputs.version }}
+        run: |
+          set -euo pipefail
+          version_ref="${IMAGE}:${VERSION}"
+          existing_digest="$(gh api --paginate \
+            "/orgs/withcoral/packages/container/reef/versions?per_page=100" \
+            --jq ".[] | select(any(.metadata.container.tags[]; . == \"${VERSION}\")) | .name")"
+          if [ -n "$existing_digest" ]; then
+            test "$(printf '%s\n' "$existing_digest" | wc -l)" -eq 1
+            if [ "$existing_digest" != "$BUILD_DIGEST" ]; then
```
simonwhitaker: **The compare-and-set wedges the release on any re-run, because the build isn't reproducible.**

`existing_digest` is the digest already published for `reef:X.Y.Z`, but `BUILD_DIGEST` comes from a *fresh rebuild* a few steps up. Those won't match across runs: `--provenance=true` embeds `buildStartedOn` / `buildFinishedOn` in the attestation, and the image config carries a `created` timestamp, so the index digest differs every time.

Concretely: `coordinate-docker-aliases` hits a transient GHCR error, someone clicks "Re-run all jobs" → `publish-reef-docker` rebuilds → digest mismatch → "refusing to replace it" → `exit 1`. `coordinate-docker-aliases` and `promote-release` both transitively need this job, so the release is now permanently stuck as a prerelease with no convergent path. That's the opposite of the PR description's "retries converge partial failures idempotently".

The same wedge hits `workflow_dispatch` re-runs of `release.yml` against an already-released tag, which was previously supported (e.g. to re-point the `docs` branch).

Comparing against the *signed, verified* digest recorded from the original publish — rather than a rebuild — would give you the immutability guarantee without the wedge.

.github/workflows/reef-docker-publish.yml
```diff
@@ -0,0 +1,157 @@
...
+              echo "${version_ref} already points to ${existing_digest}; refusing to replace it with ${BUILD_DIGEST}." >&2
+              exit 1
+            fi
+            echo "${version_ref} already points to the verified digest."
+          else
+            docker buildx imagetools create --tag "$version_ref" "${IMAGE}@${BUILD_DIGEST}"
+          fi
+          published_digest="$(docker buildx imagetools inspect "$version_ref" --format '{{.Manifest.Digest}}')"
+          if [ "$published_digest" != "$BUILD_DIGEST" ]; then
```
simonwhitaker: **This digest-stability assertion contradicts what the sibling Coral workflow states as fact — in the same PR.**

Here we assert `published_digest == BUILD_DIGEST` after `docker buildx imagetools create`. But `docker-publish.yml:232` says the opposite outright: *"Tagging republishes the index, so re-read the digest the tags resolve to"* — and it acts on that by signing the **post-tag** digest (`docker-publish.yml:243`), whereas Reef signs the **pre-tag** `BUILD_DIGEST` at line 127.

Both can't be right. If the Coral comment is accurate, every Reef publish hard-fails at this check (blocking the whole release), and the cosign signature ends up attached to a digest that no tag resolves to.

Worth settling explicitly rather than shipping two contradictory assumptions about the same buildx operation.

.github/workflows/docker-publish.yml
```diff
@@ -251,32 +218,17 @@ jobs:
...
-          # `X.Y` and `latest` on whichever release is newest by semver.
-          tags=(--tag "${IMAGE}:${VERSION}")
-          if [ "$PRERELEASE" = "false" ] && [ "$NEWEST_LINE_TAG" = "$TAG" ]; then
-            tags+=(--tag "${IMAGE}:${MAJOR_MINOR}")
-          fi
-          if [ "$PRERELEASE" = "false" ] && [ "$NEWEST_TAG" = "$TAG" ]; then
-            tags+=(--tag "${IMAGE}:latest")
-          fi
-          docker buildx imagetools create "${tags[@]}" "${IMAGE}@${BUILD_DIGEST}"
+          docker buildx imagetools create --tag "${IMAGE}:${VERSION}" "${IMAGE}@${BUILD_DIGEST}"
```
simonwhitaker: **CVE rebuilds can no longer advance `latest` / `X.Y`.**

The `workflow_dispatch` (CVE rebuild) path now re-tags only `:VERSION`. The alias logic moved to `docker-aliases.yml`, which is `workflow_call`-only and reachable solely from `release.yml` — so there is no path, automatic or manual, to re-point the aliases after a rebuild.

Scenario: a base-image CVE is patched by dispatching `docker-publish` for the newest release. `ghcr.io/withcoral/coral:1.2` and `:latest` keep resolving to the vulnerable digest indefinitely.

The deleted code handled exactly this case — *"a CVE rebuild of a superseded release re-publishes its own version tag, but leaves `X.Y` and `latest` on whichever release is newest by semver."*

.github/workflows/docker-aliases.yml
```diff
@@ -0,0 +1,90 @@
...
+          for alias in "${aliases[@]}"; do
+            # A partial failure is deliberately a failed release workflow. A
+            # retry is idempotent and converges both aliases before completion.
+            docker buildx imagetools create --tag "ghcr.io/withcoral/coral:${alias}" \
+              "ghcr.io/withcoral/coral@${CORAL_DIGEST}"
+            docker buildx imagetools create --tag "ghcr.io/withcoral/reef:${alias}" \
+              "ghcr.io/withcoral/reef@${REEF_DIGEST}"
+            coral_actual="$(docker buildx imagetools inspect "ghcr.io/withcoral/coral:${alias}" --format '{{.Manifest.Digest}}')"
+            reef_actual="$(docker buildx imagetools inspect "ghcr.io/withcoral/reef:${alias}" --format '{{.Manifest.Digest}}')"
+            test "$coral_actual" = "$CORAL_DIGEST"
```
simonwhitaker: Same digest-stability assumption as `reef-docker-publish.yml:153`: `imagetools create --tag …@DIGEST` followed by asserting the alias resolves to a byte-identical index digest.

Lower risk here — the input is already the output of a previous `create`, so it's more plausibly a fixed point — but if `create` re-serializes the index at all, this hard-fails the alias job and therefore the release.

Comparing what the alias resolves to against the *image content* would be sturdier than asserting byte-identical index digests.

.github/workflows/reef-docker-publish.yml
```diff
@@ -0,0 +1,157 @@
...
+      - name: Install cosign
+        uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
+
+      - name: Sign the staged manifest
+        shell: bash
+        env:
+          DIGEST: ${{ steps.build.outputs.digest }}
+        run: cosign sign --yes "${IMAGE}@${DIGEST}"
+
+      - name: Publish immutable version tag
```
simonwhitaker, line 225: **The Reef image is signed, published, and aliased without ever being started.**

Coral's publisher gates its version tag behind a full container smoke battery (`docker-publish.yml:186-219`). This publisher only checks two OCI labels via `docker image inspect`.

So a Reef image that fails to boot — a missing runtime dependency in the `deps` stage, an entrypoint regression — is still signed, published as `reef:X.Y.Z`, promoted to `reef:latest` and `reef:X.Y` by the coordinator, and the release is marked complete.

The PR-time `reef-image` validate job that *does* smoke it is path-filtered, so a release commit whose changes miss the filter never exercised that image either.

simonwhitaker (approved):
Approving so this isn't blocked on me — but there's one hard blocker plus five issues that I'd want resolved before merge rather than closed out.

`promote-release` is missing `environment: release`, which means it cannot run at all: the App-token secrets exist only in that environment. As it stands every release will fail at promotion and stay a prerelease. That one is not optional.

The rest, in rough severity order: the Reef compare-and-set wedges the release on any re-run because the build isn't reproducible; two digest-stability assertions rest on an assumption that `docker-publish.yml`'s own comment contradicts; CVE rebuilds lose their ability to advance `latest` / `X.Y`; and the Reef image ships signed and aliased without ever being booted.

Worth noting `docker/reef-publish.test.mjs` passes locally (8/8), but it asserts on workflow *text*, so it can't catch any of the above.

One operational heads-up: `ghcr.io/withcoral/reef` will be created as a new **private** GHCR package on first publish — visibility has to be flipped manually before `reef:latest` is pullable.

.github/workflows/docker-aliases.yml
```diff
@@ -0,0 +1,90 @@
...
+        required: true
+        type: string
+      coral-digest:
+        required: true
+        type: string
+      reef-digest:
+        required: true
+        type: string
+    outputs:
+      make-latest:
```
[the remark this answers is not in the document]
sauldhernandez: Resolving as invalid. `validate-release-ref` only accepts semantic stable tags of the form `vX.Y.Z`, so semantic prerelease tags cannot reach this workflow. The GitHub prerelease flag is deliberately the release transaction staging state until both immutable images and coordinated aliases succeed. `gh release list --exclude-pre-releases` therefore necessarily omits the current tag; appending `$TAG` lets it participate in the newest-version calculation before final promotion. Checking `isPrerelease` here would suppress aliases and `make_latest` for every normal release.

.github/workflows/release.yml
sauldhernandez: Fixed on the PR head in `70541095d`: both `promote-release` and `notify-release` now declare `environment: release`. That restores the release-scoped app variable and secrets and keeps promotion behind the environment protection.

.github/workflows/reef-docker-publish.yml
sauldhernandez: Spiked this rather than relying on the Coral comment. For a single source index, `docker buildx imagetools create --dry-run docker.io/library/alpine:3.22` produced JSON structurally identical to `docker buildx imagetools inspect docker.io/library/alpine:3.22 --raw`; Docker also defines the one-source operation as a carbon copy. The Coral comment means we re-read the published reference defensively, not that Buildx changes the index bytes. Keeping the exact digest assertion is intentional because Reef signs that index before exposing the tag.

.github/workflows/docker-aliases.yml
sauldhernandez: The same spike covers this path, and this case is stronger because the source is already a published OCI index. A one-source `imagetools create` preserves the index digest. The equality check ensures both aliases resolve to the exact signed index; comparing only child image content would weaken that guarantee. Keeping the assertion.

.github/workflows/reef-docker-publish.yml
sauldhernandez: Fixed in `a5446da8d`. The Reef publisher now resolves the existing immutable version before building, cross-checks the package API digest against the registry, verifies the original cosign signature and image metadata, and smoke-tests that exact digest. When the version exists, the build, sign, and tag-creation steps are skipped, so release and `workflow_dispatch` reruns converge instead of rebuilding into a digest mismatch.

.github/workflows/docker-publish.yml
sauldhernandez: Fixed in `a5446da8d`. Direct `docker-publish` dispatches again compute stable eligibility and move `X.Y` and `latest` for an eligible Coral CVE rebuild. Release calls pass `defer-aliases: true`, leaving paired Coral/Reef alias movement to the coordinator, and both mutation paths share the same concurrency group.

.github/workflows/reef-docker-publish.yml
sauldhernandez: Fixed in `a5446da8d`. Before signing or exposing the version tag, the Reef publisher now starts the exact candidate digest with a read-only root filesystem, waits for the image health check to report healthy, and verifies the runtime UID:GID is `1000:1000`. A boot or entrypoint regression now fails the release gate.

.github/workflows/reef-docker-publish.yml
```diff
@@ -0,0 +1,241 @@
...
+        id: publish
+        shell: bash
+        env:
+          DIGEST: ${{ steps.candidate.outputs.digest }}
+          EXISTING_DIGEST: ${{ steps.existing.outputs.digest }}
+          VERSION: ${{ steps.tag.outputs.version }}
+        run: |
+          set -euo pipefail
+          version_ref="${IMAGE}:${VERSION}"
+          if [ -z "$EXISTING_DIGEST" ]; then
```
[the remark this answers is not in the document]
sauldhernandez: Resolved without a code change. Same-tag executions are serialized for the full reusable workflow by `concurrency.group: reef-docker-publish-${{ inputs.tag }}`, and the only caller, `release.yml`, is itself serialized end to end by its workflow-level `concurrency.group: release`, so another run cannot publish this version tag between the initial lookup and the final tag step. No other repository workflow writes Reef exact-version tags; `docker-aliases.yml` writes only X.Y and `latest`. A second read immediately before `imagetools create` would therefore be redundant under this single-writer contract, and even with it, a TOCTOU window would remain if an out-of-contract writer existed. The post-create digest assertion retains fail-fast verification of the published result.

PR #2166, opened 2026-08-13
test(cli): prove ownerless workspace isolation in shared mode

sauldhernandez:
## Intent

Prove against a real shared-mode server that configured authentication selects `NoLocalPrincipal`: the local ownership migration remains unclaimed, the legacy workspaces remain concealed, and the authenticated test identity gains access only to the workspace it explicitly creates.

## What it enables

- Starts shared mode with two explicitly configured legacy ownerless workspaces and verifies `local_workspace_ownership_v1` remains unclaimed.
- Calls the production login-identity transaction through a test helper that itself runs neither the catalog nor local-ownership migration and creates no workspace or membership.
- Proves the identity starts with zero memberships, then receives exactly Owner access after explicitly creating a third workspace through `WorkspaceService`.
- Proves authenticated reads of both legacy ownerless workspaces return NotFound while attempts to recreate their existing names return AlreadyExists.
- Proves ordinary host workspace listing and SQL against a named workspace fail with an authentication-required error; shared mode does not give host commands a local principal.
- Restarts the server and verifies the explicit Owner membership, ownerless concealment, stable identity, and unclaimed migration marker all persist.
- The requested evidence file contains only deliberately redacted factual results; all other test state remains in temporary directories. It makes no OAuth-provider, UI, Reef, or recovery-tool claim.

## Usage examples

```bash
CORAL_OWNERLESS_EVIDENCE_PATH=/private/tmp/coral-ownerless-live-proof.md \
  cargo test -p coral-cli --test ownerless_workspace_live \
  -- --ignored --exact shared_ownerless_workspaces_remain_isolated_across_restart \
  --nocapture --test-threads=1
```


sauldhernandez:
Closing alongside the rest of the superseded workspace access-control work. The stack that
replaces it is #2207-#2211 (GitHub stack #2212), which covers the individual properties proven
here — ownership bootstrap leaving an authenticated deployment unclaimed, ownerless concealment,
and `NoLocalPrincipal` refusing host commands in shared mode — across milestones m6 and m8.

Recording what this PR had that the new stack does not, so it is not lost with the branch:

1. RESTART PERSISTENCE. This proved the Owner membership, the ownerless concealment, the stable
   identity, and the unclaimed `local_workspace_ownership_v1` marker all survive a server
   restart. The new stack's tests establish those properties within a single process lifetime.
   A reboot is exactly when a bootstrap migration could re-fire and claim something, so this is
   a real gap rather than a duplicate.
2. COMPOSITION. The new stack proves these properties separately; this exercised them together
   against a real shared-mode server, where an interaction between them could surface.

Caveat on its own coverage: the test is `#[ignore]`d and touches no Makefile or workflow, so no
gate ever ran it — it required an explicit invocation with `CORAL_OWNERLESS_EVIDENCE_PATH`. Any
revival should wire it into a gate, or it is evidence that runs once and then silently stops
being true.

The branch is left intact.

PR #2177, opened 2026-08-13
feat(xtask): add workspace ownership and issuer recovery commands

sauldhernandez:
## Intent

Provide an off-by-default, repository-only recovery path for shared deployments whose persisted workspace ownership or identity-provider issuer needs repair, without adding a product administration API.

## What it enables

- Inspect every persisted workspace and authenticated non-local identity, including state concealed from ordinary shared callers.
- Add or promote an existing identity to workspace Owner; an already-running server observes the change on the person's next request.
- Rebind non-local identities to the renamed issuer while the server is stopped; provider subjects, internal IDs, and memberships are preserved.
- Open existing SQLite or Postgres state without migrations, startup repair, or filesystem creation.
- Keep provider subjects opt-in, escape terminal output, and reject `coral:local` as a recovery target.

The commands are absent from shipped `coral` and default `xtask` builds. Database possession is the authorization boundary. Because the tool never runs migrations, operators must use a checkout matching the deployment, and they should back up state first.

## Usage examples

```bash
cargo run -p xtask --features admin -- workspace-admin list-workspaces --state-dir /path/to/coral-state
cargo run -p xtask --features admin -- workspace-admin list-users --state-dir /path/to/coral-state
cargo run -p xtask --features admin -- workspace-admin set-owner --state-dir /path/to/coral-state --workspace work --user 00000000-0000-4000-8000-000000000000
cargo run -p xtask --features admin -- workspace-admin rebind-issuer --state-dir /path/to/coral-state --from https://old.example --to https://new.example
```

Run issuer rebinding only after confirming subjects are stable, stopping Coral, and taking a backup. Compare the reported count, update authentication configuration, restart, and test a login.


sauldhernandez:
Superseded by #2211, which ships the same four recovery commands (`list-workspaces`,
`list-users`, `set-owner`, `rebind-issuer`) behind an off-by-default `admin` feature, with
mutation-verified non-migration, five Postgres contracts wired into `make postgres-tests`, and
a default-surface test wired into `make rust-checks`.

Recording one design point from this PR that #2211 did **not** adopt, because it is the better
call on that axis and should not be lost with the branch:

This PR put the recovery logic in `crates/coral-app/src/admin.rs` behind
`#[cfg(feature = "admin")]`, so it reuses coral-app's real repositories and types — 987 lines
total. #2211 instead keeps the shipped crate entirely free of admin code by implementing
everything in `xtask/src/admin.rs` (2,876 lines). That buys stronger isolation at the cost of
hand-duplicating server behaviour, which #2211's own reviews flagged twice:

- `LOCAL_PRINCIPAL_ID` is hand-copied into `xtask/src/admin.rs` from coral-app's crate-private
  constant. It already changed once ("local" -> "coral:local"); if it changes again, recovery
  silently refuses the wrong string and no gate catches it.
- The `[database]` config parse and the `sslmode=verify-full` transport rule are
  re-implemented in xtask with nothing tying them to the server's originals.

Follow-up worth doing on #2211: a drift tripwire pinning those literals against coral-app's
originals. The branch is left intact if the coral-app-gated approach is ever preferred.

PR #2178, opened 2026-08-13
feat(app)!: move existing workspace members between roles

sauldhernandez:
## Intent

Align `AddWorkspaceMember` with the membership role-management contract. Repeating the current role succeeds without rewriting the membership, while requesting a different role moves the existing member atomically.

## What it enables

Existing members can be promoted or demoted through the existing RPC without a separate role-change method. Membership writes and last-owner checks remain serialized under the workspace parent lock, and committed role changes affect authorization on the next request. The obsolete role-conflict error and telemetry classification are removed without changing the protobuf wire shape.

## Usage examples

- Requesting `Member` for an existing user who is not yet in the workspace creates and returns the membership.
- Requesting `Member` for an existing Member returns the same membership without rewriting its creation time.
- Requesting `Owner` for an existing Member promotes the member atomically and returns the resulting Owner membership.
- Requesting `Member` for an Owner demotes them when another Owner remains.
- Attempting to demote the sole Owner returns `FailedPrecondition` without changing the membership.

BREAKING CHANGE: `WorkspaceService.AddWorkspaceMember` now changes an existing member's role and returns success instead of returning `AlreadyExists` when the requested role differs. Attempting to demote the sole owner now returns `FailedPrecondition`.

### DIFF
--- a/apps/docs/reference/configuration.mdx
+++ b/apps/docs/reference/configuration.mdx
@@ -109,6 +109,8 @@ that needs operator attention, not an alternate access path.
 The gRPC API is Coral's internal contract boundary. It is not a supported
 end-user workspace-administration or recovery surface.
 
+Adding an existing member with the same role is idempotent; requesting a different role atomically moves the membership to that role. Coral refuses to demote or remove the sole owner.
+
 ### Repository recovery tool
 
 The repository contains an off-by-default tool for recovering a legacy workspace with no non-local owner and for rebinding identities after an issuer rename. It is absent from shipped `coral` and default `xtask` builds. Use a checkout matching the deployment and back up first. Database possession is the authority: SQLite requires state-directory access; Postgres uses `[database].url_env` without a host-locality guarantee. The tool opens existing state without migrations or startup repair. Start the upgraded shared server, have the intended owner sign in once, then run:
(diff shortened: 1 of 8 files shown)


sauldhernandez:
Superseded by #2211's stack (#2207-#2211, GitHub stack #2212), which already implements this
same behaviour — verified rather than assumed:

- `AddMemberOutcome::RoleUpdated` in `crates/coral-app/src/state/db/workspace_state.rs`: adding
  an existing member at a different role moves them, the same design this PR takes.
- `AddMemberOutcome::LastOwnerProtected` guards demotion of a workspace's last owner
  (`workspace_state.rs:218`), and `RemoveMemberOutcome::LastOwnerProtected` guards removal.
- Wired through `workspaces/manager.rs:402`, surfacing `AppError::LastWorkspaceOwner` on refusal.
- Covered by tests in `workspace_state.rs` (promotion, demotion, last-owner protection).
- The proto already documents it: "Adds one person to one workspace, or updates the role they
  already hold."

Feature-identical down to the enum variant names. Closing rather than rebasing: both PRs are
written against a `crates/coral-app/src/state/db/workspace_member_state.rs` that the new stack
does not have (membership state lives in `workspace_state.rs`), so this would be a
hand-reimplementation producing code that already exists, not a rebase.

The branch is left intact.

PR #2179, opened 2026-08-14
ci: validate the container images without building Rust

sauldhernandez:
Validates the Coral and Reef container images without compiling Rust, taking
image validation from ~5.9 min to ~39s.

## Intent

The `Reef image` job ran 5.4–9.0 min (median 5.9, n=36) on every PR touching
`crates/**`, `apps/reef/**`, or the Docker files. Step timings showed almost
none of it was Reef:

| phase | time |
| --- | --- |
| `cargo build --locked --release -p coral-cli` | **4m23s** |
| coral image assembly | 10s |
| reef image build (`npm ci` + vite) | 16s |
| reef smoke | 13s |

That compile existed only to produce a Coral peer for `docker/reef-smoke.sh`.
Reading what the peer was actually asked to do: `wait_for_coral` probed
`grpc_health_probe -addr=127.0.0.1:14555`, and `assert_ready` checked that
Reef's `/readyz` reported `"coral":"reachable"`. No Coral business logic, no
proto surface. Two of the four scenarios used no peer at all — and the two that
did were already covered by mocks:

| existing test | smoke scenario it duplicated |
| --- | --- |
| `readyz.server.test.ts` — ready only when Coral reports serving | A, `"coral":"reachable"` |
| `readyz.server.test.ts` — rejects a non-serving status | — |
| `readyz.server.test.ts` — propagates an upstream health failure | B, peer stopped → not ready |
| `coral-app/tests/grpc/health_service_tests.rs` | the peer's health contract itself |

A 4m23s release build was buying a gRPC health responder whose contract was
already under test on both sides.

## What it enables

Image validation is split by what each image owns, and neither half builds Rust.

**`reef-image`** runs a peer-free matrix over the runtime configurations: which
boot healthy (explicit-loopback cleartext, and non-loopback *with*
`REEF_ALLOW_INSECURE_CORAL_ENDPOINT`) and which fail fast with a named cause.
Reef probes its own `/healthz` before listening (`apps/reef/server.js:31`), so a
peer was never required for any of it. Non-root identity and the read-only root
are still asserted.

**`coral-image`** is new and owns `docker/Dockerfile` and
`docker/entrypoint.sh` — which previously had no dedicated job and were built
only as a side effect of the Reef smoke. The entrypoint is 108 lines of pure
shell up to its closing `exec`, so a stub binary reaches every branch: starter
seeding, seed-once semantics, verbatim `CORAL_SEED_CONFIG`, and the
unwritable-volume failure. That is the same matrix `docker-publish.yml` runs at
release, minus the compile. The real binary stays covered by `rust-checks`, the
real image by the release smoke.

Both jobs run on 4vcpu, in parallel. Neither triggers on `crates/**`, so
Rust-only PRs no longer run either.

## Usage examples

```bash
make reef-docker-test REEF_DOCKER_IMAGE=reef:local
```

```bash
make coral-docker-stub-test CORAL_DOCKER_IMAGE=coral:stub
```

`make docker-build` is unchanged for building a real local Coral image from the
current checkout.

## Coverage this gives up, deliberately

- **No live Reef↔Coral handshake anywhere in CI.** Both sides are covered by
  mocks, and clients are generated from shared protos, but nothing now proves
  the two real processes talk to each other. This is the one genuine reduction.
- **`docker/Dockerfile.local` is no longer built in CI.** It was only ever
  exercised as a side effect of the Reef smoke; it remains the local
  `make docker-build` path.
- **The Coral image `HEALTHCHECK` is not exercised on PRs**, since a shell stub
  cannot serve gRPC. The release smoke covers it against the real binary.

## Review notes

Three places in this workflow can swallow a job failure, and all three are
handled here — worth checking if you touch the filters:

- `validate` is the single required status check and enumerates results in an
  explicit bash array. Adding a job to `needs` alone lets its failure pass.
- A new filter key needs a matching entry in the `changes` job's `outputs`
  block, or the job silently never runs.
- `docker/reef-publish.test.mjs` reads the `reef-image` filter by slicing to the
  next key, so the new `coral-image` filter is placed *before* `reef-checks:`
  rather than between `reef-image:` and `desktop-checks:`.

The first two commits are superseded and contribute nothing to the net diff:
`94a6978` tried a persistent BuildKit cache and made the job **slower** (43 min
against a 5.9 min median); `1313521` swapped in a released Coral peer but left
the filter watching paths it no longer built. Both are retained rather than
force-pushed away.

## Validation

Measured on this PR's run, which touches `validate.yml`, `Makefile`, and both
smoke scripts, so both image jobs execute:

| job | result | duration |
| --- | --- | --- |
| Coral image | success | **18s** |
| Reef image | success | **39s** |

Both scripts printed their closing line — reachable only after every assertion,
since both run under `set -eu` with no conditionals:

```
Coral image entrypoint matrix passed
Reef image smoke matrix passed
```

Also: `node --test docker/reef-image.test.mjs docker/reef-publish.test.mjs`
12/12; `sh -n` clean on both scripts; `make -n` resolves every new target; the
workflow YAML parses and the new `changes` output is declared.

Not proven by this run: that a Rust-only PR no longer triggers either job. That
follows from the filters, but the first `crates/**`-only PR after merge is what
confirms it.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -508,9 +508,12 @@ jobs:
 
   reef-image:
     name: Reef image
-    # Building the real Coral peer dominates this job. Give it a larger
-    # runner than the other jobs while Dockerfile.local's locked cache mounts
-    # retain registry, git, and target artifacts within the BuildKit builder.
+    # Building the real Coral peer dominates this job: the release build of
+    # coral-cli is ~85% of its wall time. Dockerfile.local's locked cache mounts
+    # hold the registry, git, and target artifacts, but they live in the
+    # BuildKit state dir, so they only survive across runs on the persistent
+    # builder configured below. Keep the larger runner until that cache is
+    # observed warming; it is sized for the cold-build case.
     runs-on: blacksmith-16vcpu-ubuntu-2404
     needs: changes
     if: ${{ needs.changes.outputs.reef-image == 'true' }}
@@ -519,9 +522,13 @@ jobs:
         with:
           persist-credentials: false
 
-      - uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
+      # Backs BuildKit with a sticky disk mounted at /var/lib/buildkit, so
+      # Dockerfile.local's cargo cache mounts persist across runs instead of
+      # recompiling every dependency from scratch. Falls back to a local
+      # builder if the Blacksmith setup fails.
+      - uses: useblacksmith/setup-docker-builder@a5256a73e30f09e37e3eceb8ca36043d17621d24 # v2.1.0
         with:
-          cache-binary: false
+          cache-key: coral/reef-image
 
       - name: Check Reef image contract
         run: node --test docker/reef-image.test.mjs docker/reef-publish.test.mjs
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -508,35 +508,52 @@ jobs:
 
   reef-image:
     name: Reef image
-    # Building the real Coral peer dominates this job: the release build of
-    # coral-cli is ~85% of its wall time. Dockerfile.local's locked cache mounts
-    # hold the registry, git, and target artifacts, but they live in the
-    # BuildKit state dir, so they only survive across runs on the persistent
-    # builder configured below. Keep the larger runner until that cache is
-    # observed warming; it is sized for the cold-build case.
-    runs-on: blacksmith-16vcpu-ubuntu-2404
+    # Pull requests smoke against a released Coral peer and never compile Rust,
+    # so they fit comfortably on a small runner. Pushes still build the peer
+    # from HEAD and keep the larger one for that release build.
+    runs-on: ${{ github.event_name == 'pull_request' && 'blacksmith-4vcpu-ubuntu-2404' || 'blacksmith-16vcpu-ubuntu-2404' }}
     needs: changes
     if: ${{ needs.changes.outputs.reef-image == 'true' }}
+    env:
+      # The smoke matrix only needs a peer that serves gRPC health on 14555, so
+      # it tracks the newest released Coral rather than pinning a version that
+      # would go stale. Pin this to a version or digest if a release ever breaks
+      # the matrix; pushes cover the HEAD peer either way.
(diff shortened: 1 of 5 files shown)


sauldhernandez:
Thanks — five of the seven are fixed in 1a16ee1. Two are deliberate trade-offs
rather than defects, so I'm answering them here instead of changing code.

**Fixed** (`Makefile:102`, `Makefile:141`, `coral-smoke.sh:22`, `:27`, `:99`)

Both `-j` findings and the subshell-bookkeeping one were real and reproduced
locally: a smoke prerequisite printed before its build finished, and an
accumulator assigned inside `$(...)` came back empty. The smoke now runs from
the recipe in both targets, and the helpers assign in the calling shell.

On `:99` the problem was slightly worse than described. Because command
substitution strips trailing newlines from *both* sides, the assertion could not
observe a trailing-newline difference at all — a test advertising "verbatim" was
not verifying it. It now compares sha256 digests computed by the same binary in
the same image, and the fixture carries a trailing newline so the property is
actually exercised.

`:22` was correct, and `docker/reef-smoke.sh` had the same pattern already, so
both scripts now follow the convention `docker/entrypoint.sh` uses: cleanup on
EXIT, explicit exit per signal.

---

**`validate.yml:532` — no real-peer integration smoke**

The technical claim is right: mocking `healthClientForRequest` cannot catch a
mismatch between Reef's health client and Coral's actual health surface. That
gap is intentional and is listed under "Coverage this gives up, deliberately" in
the description.

The "contradicts the PR description" part is stale — that promise was in an
earlier revision describing a released-peer approach that this PR no longer
takes. The description was rewritten to describe the net diff before this review
ran.

**`validate.yml:104` — `Dockerfile.local` has no CI coverage**

Correct, and disclosed in the description. But the suggested fix — adding it to
the `coral-image` filter — would make that job trigger on a file it does not
build, which is exactly the false-green pattern this PR removes: on `main`,
`reef-image` triggers on `docker/Dockerfile` and `entrypoint.sh` while the new
arrangement is what finally gives those files a job that builds them.

Restoring real coverage means a job that actually builds `Dockerfile.local`, and
that is the ~4m23s release compile this PR exists to remove from the PR path. It
is a local-developer convenience (`make docker-build`), so leaving it uncovered
is the deliberate call. A push-only or nightly job would be the way to change
that, and is worth its own PR rather than reinstating the compile here.

lazybean (approved):
LGTM

PR #2180, opened 2026-08-14
test(reef): keep data fetching in route handlers

AlbertQM:
## Summary

Adding an architecture test to try and prevent LLMs from fetching data from the client, and giving some guidance on what to do instead.


lazybean (approved):
Thank you

PR #2205, opened 2026-08-18
refactor(ui): Rebuild wax  table

AlbertQM:
## Summary

Rewriting our Table, and moving away from `<table>` in favour of divs.

Standardised design language: variant card only for settings tables, rest is normal.

Fixed the header token, so the sticky header isn't transparent and blends with the table content.

## Test plan

Look at all the tables in the product

image
this is now not the card variant
image
image

Now we have sticky headers in the settings tables
image
image

Schema and functions both look the same now (and long content is accessible by horizontal scrolling)
image
image


lazybean (approved):
thank you for lavishing the table

PR #2217, opened 2026-08-18
feat(desktop): package the Linux desktop app as AppImage and deb

AlbertQM:
## Summary

Ships Coral Desktop on x64 Linux as an AppImage and a deb. The Rust CLI already builds for Linux and the main process was already platform-branched, so this is packaging plus one runtime fix.

- **Auto-update stays macOS-only.** `desktopUpdatesSupported()` already returns `false` off darwin, so `linux.publish: null` stops electron-builder from writing a `latest-linux.yml` feed nothing serves, or embedding `app-update.yml` in the package.
- **One runtime fix:** the packaged sidecar ran with `cwd: process.resourcesPath`, a read-only squashfs mount inside an AppImage. Now `app.getPath('userData')`.
- **`CORAL_DESKTOP_RELEASE=1` is rejected off macOS** — it's the Apple signing path. `createConfig()` takes the platform as an argument so the config tests stay host-independent (`desktop-checks` runs on Ubuntu).
- **`verify-dist.mjs` takes a platform argument** (`mac` | `linux`) and asserts the per-platform artifact shape, including that Linux produces no update metadata.
- CI: `desktop-package.yml` becomes a mac+linux matrix; `release.yml` gains `build-desktop-linux`, which reuses the CLI artifact the `build` job already produced — no Rust rebuild, no signing, no xtask.

User-facing docs (the Install and Upgrade tabs in `installation.mdx`) are deferred to a follow-up so this PR.

Four config details are load-bearing and not obvious:

| Setting | Why |
|---|---|
| `executableName` / `deb.packageName` avoid `coral` | The deb symlinks its executable into `/usr/bin`; the default would install `/usr/bin/coral` pointing at the Electron app and **shadow the CLI**. |
| `linux.icon` names the **directory**, not `icon.png` | electron-builder returns a lone `.png` verbatim instead of generating a set, and hicolor's `index.theme` declares nothing above `512x512` — so naming the file installs a single 1024×1024 icon launchers can't find. The directory yields 8 sizes, 16→512. |
| `desktopName` + `syncDesktopName` | Otherwise `WM_CLASS` is `withcoral-desktop` (from the package name) while the desktop entry declares `Coral`, and no desktop environment can associate a running window with the installed launcher. |
| `homepage` + `maintainer` | fpm refuses to build a deb without a project URL or contact address; `author` carries no email. |

## Test plan

Verified against a `coral-cli` built from the main, in an `ubuntu:24.04` container (Apple `container`) matching the CI runner's distro.

- [x] `npm run check` (12 config tests) and `npm test` (49 vitest) pass; `actionlint` clean; `cargo build -p xtask` clean
- [x] `npm run package:linux` produces `coral-desktop-linux-x86_64.AppImage` + `coral-desktop-linux-amd64.deb`; `verify-dist … linux` passes and `… mac` correctly fails
- [x] No `latest-linux.yml` and no embedded `app-update.yml`
- [x] deb: `Package: coral-desktop`, `Architecture: amd64`, `Maintainer: With Coral <eng@withcoral.com>`, `Homepage`/`License` set, `StartupWMClass=coral-desktop`, icons at 8 hicolor sizes, postinst symlinks `/usr/bin/coral-desktop`
- [x] Bundled CLI is `-rwxr-xr-x` in **both** the deb payload and the AppImage squashfs — no `afterPack` re-chmod needed
- [x] App launched (arm64 build, native in the VM): sidecar reached ready on the new `Coral gRPC server listening on` line, Coral UI rendered over `coral-app://`, window `WM_CLASS` measured as `coral-desktop` via `xprop`
- [x] Sidecar `cwd` read off `/proc/<pid>/cwd` → userData, not resourcesPath; Coral state landed in the isolated `CORAL_CONFIG_DIR` under userData
- [x] Closing the window (under a real WM) quits app + sidecar; relaunch reaches ready, so the state lock was released

Not covered: Desktop Settings → MCP clients (needs UI interaction), and `desktop-package.yml` via `workflow_dispatch` plus a release dry-run, which need CI.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -332,9 +332,86 @@ jobs:
             apps/desktop/dist/coral-desktop-*.blockmap
             apps/desktop/dist/latest-mac.yml
 
+  # Far simpler than the macOS leg: no universal binary, no signing, no
+  # notarization, and no updater, so no xtask and no credentials.
+  build-desktop-linux:
+    needs:
+      - validate-release-ref
+      - build
+    runs-on: blacksmith-4vcpu-ubuntu-2404
+    env:
+      TAG_NAME: ${{ needs.validate-release-ref.outputs.tag-name }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          ref: ${{ needs.validate-release-ref.outputs.commit-sha }}
+          persist-credentials: false
+
+      # A manual tag on a commit whose desktop version was never bumped would
+      # otherwise ship an AppImage labelled with the wrong version. The macOS
+      # leg gets this from xtask; see ensure_version_matches_tag in
+      # xtask/src/release.rs.
+      - name: Verify desktop version matches the release tag
+        run: |
+          set -euo pipefail
+          version="$(jq -r .version apps/desktop/package.json)"
+          if [ "v${version}" != "$TAG_NAME" ]; then
+            echo "desktop version v${version} does not match release tag ${TAG_NAME}" >&2
+            exit 1
+          fi
+
+      - name: Download Linux CLI artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
+        with:
+          name: coral-x86_64-unknown-linux-gnu
+          path: ${{ runner.temp }}/cli
+
+      # The `build` job already compiled this target; reuse it rather than
+      # rebuilding Rust here.
+      - name: Unpack the prebuilt Coral CLI
+        run: |
+          set -euo pipefail
+          tar xzf "$RUNNER_TEMP/cli/coral-x86_64-unknown-linux-gnu.tar.gz" -C "$RUNNER_TEMP/cli"
+          "$RUNNER_TEMP/cli/coral" --version
+
+      - name: Set up Node
+        uses: actions/setup-node@249970729cb0ef3589644e2896645e5dc5ba9c38 # v6.5.0
+        with:
+          node-version-file: apps/desktop/.node-version
+          package-manager-cache: false
+
+      - name: Install desktop dependencies
+        run: npm ci --prefix apps/desktop
+
+      # fpm shells out to `ar` (binutils) to assemble the .deb and to `xz` for
+      # its control tarball. Both ship in the runner image; installing them
+      # explicitly keeps the dependency visible instead of implicit.
+      - name: Install deb packaging tools
+        run: sudo apt-get install -y --no-install-recommends binutils xz-utils
+
+      # No CORAL_DESKTOP_RELEASE: release mode is the Apple signing path, and
+      # Linux ships without an updater.
+      - name: Package desktop app
+        env:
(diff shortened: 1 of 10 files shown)


AlbertQM:
Addressed two of the three; pushed in 5a44521.

**Upgrade instructions (P2) — valid, and my miss.** The Upgrade section told every reader that Coral shows a notification and that the menu bar can check manually. On Linux neither is true: `desktopUpdatesSupported()` is false, so nothing polls *and* the menu item is never built (`index.ts:301` composes it conditionally). Split the tab into macOS / Linux, mirroring the Install section.

**Sidecar test (P3) — right conclusion, wrong mechanism.** The `TypeError` described cannot occur: `stopChild` returns early on `child.exitCode !== null`, and on the mock that is `undefined !== null` → `true`, so `kill` is never reached. The real defect is the one underneath — an unsettled promise and a live 30s timer. Rather than capture-and-terminate, the fake child now emits the ready line, so the promise resolves, the timer clears itself, and the test covers the handshake that timer exists for.

**README wording (P3).** Reworded, though not for the reason given — "off macOS" leans on an idiomatic preposition, so it now reads "it fails on any platform other than macOS", which states the condition directly.

.github/workflows/release.yml
```diff
@@ -332,9 +332,86 @@ jobs:
...
+        run: npm ci --prefix apps/desktop
+
+      # fpm shells out to `ar` (binutils) to assemble the .deb and to `xz` for
+      # its control tarball. Both ship in the runner image; installing them
+      # explicitly keeps the dependency visible instead of implicit.
+      - name: Install deb packaging tools
+        run: sudo apt-get install -y --no-install-recommends binutils xz-utils
+
+      # No CORAL_DESKTOP_RELEASE: release mode is the Apple signing path, and
+      # Linux ships without an updater.
```
AlbertQM, line 393: All these comments that say Linux doesn't ship with an updater will go away in the next pr, which implements exactly that

apps/desktop/src/main/sidecar.ts
```diff
@@ -81,7 +81,11 @@ function packagedSidecarCommand(): { command: string; args: string[]; cwd: strin
   return {
     command: bundledCoralPath(),
     args: ['server'],
-    cwd: process.resourcesPath,
+    // The binary is addressed absolutely and its state directory arrives as
+    // CORAL_CONFIG_DIR, so cwd only decides where stray relative work lands.
+    // `resourcesPath` is a read-only squashfs mount inside an AppImage;
+    // userData is writable on every platform.
+    cwd: app.getPath('userData'),
```
simonwhitaker, line 88: Will this change affect people who already have the desktop app installed (on MacOS) and then upgrade to this version?

(I don't think we need to go overboard on migrating data if so; just make Coral folks aware if it's likely to cause them a problem.)

simonwhitaker (approved):

apps/desktop/src/main/sidecar.ts
AlbertQM: Good shout! Shouldn't require any action; the other PR which landed (that split the CONFIG_DIR per binary was the one with consequences)

PR #2219, opened 2026-08-19
feat(engine): observe values from file-backed sources

mystic123:
> **Stack 2 of 2**, on top of #2220, which adds the decorator this wires up.

Wire the file backend to the observing table-provider decorator, so values in
parquet, csv, jsonl, and json tables reach `observed_values` and surface in
`coral search` under the `observed_values` provider. Until now the backend
received `source_observation_publishers` and dropped them.

`compile_manifest` keeps the publishers on `FileCompiledSource` and `register()`
wraps each provider, whether it is a DataFusion `ListingTable` behind
`FileTableProvider` or the custom `JsonFileTableProvider`. Wrapping at the
provider boundary covers both without touching either. This retires the staged
`dead_code` expectation the decorator carried.

Capture stays gated on `observed_values_search`. With the flag off the publisher
list is empty, `SourceObservationConfig::new` returns `None`, and `register()`
hands back the provider itself rather than a passthrough wrapper. Anyone already
running with the flag on will start capturing values from their file sources.

No app-side change was needed, verified rather than assumed:
`source_surface_scopes` already enumerates file table surfaces, `QueryManager`
builds publishers for every selected source with no backend filter, and
`credential_revision` is stably nil for credential-less sources, so scope ids
match between capture and retrieval.

Engine tests cover parquet, csv, and the custom JSON provider publishing under
the right source and surface names, a partition-pruned scan publishing only the
surviving file's rows, a predicate selecting what is observed while the query
result stays identical to an unwrapped scan's, and capture-off registering
`FileTableProvider` itself.

App tests carry the lifecycle requirements: a file source's value reaches
`coral search`, removing the source clears its rows and advances its epoch,
editing its location does the same, and a source removed while the flag was off
cannot resurface, because retrieval filters to live scopes. That last test
asserts the rows really do linger on disk first, so it fails rather than passes
vacuously if the flag-off clear gap is ever closed. Each of the four fails if
this wiring is removed.

`FileCompiledSource` loses an unused `#[derive(Debug)]`; the publisher trait is
not `Debug`, and the sibling `HttpCompiledSource` holding the same field derives
only `Clone`.

## Behaviour change

**Anyone already running with `observed_values_search` on starts capturing values
from their file sources when this lands.** No separate flag, no per-source
opt-out; a per-backend opt-out is not worth the surface area until someone asks.
Off by default, unchanged.

File values also enter the same ranking path (`search/observed/ranking.rs`) that
the observed-ranking rework is changing. A large file corpus may crowd out
API-source hits until that lands. Not a blocker, but worth knowing before this
is enabled anywhere real.

## What gets observed

This table's rows that satisfy the query's predicates on it, roughly its
contribution to the result. Deliberately not a post-pushdown scan head: an HTTP
response is shaped by the user's request, the first rows of a parquet file are
not, so a head would index query-irrelevant noise. Two accepted approximations:
rows that match this table's predicates but are dropped by a later join are
observed anyway, and a predicate spanning tables narrows this table only where
DataFusion can infer an equivalent single-table predicate from an equijoin.

There is no engine-side budget and no new config key. Inflow is contained by
machinery that already exists: the per-publish collector caps, the bounded writer
queue that sheds load by dropping with a warn, and store eviction. Consequence
accepted knowingly: an unfiltered scan of a large file publishes every batch, so
collector CPU rides on the query for its duration, and a very large scan's
coverage is whatever survived the queue.

Observation is strictly query-driven. Installing a file source never triggers a
scan — verified by hand below.

## Verification

Draft PRs report every real check as `skipping`, so CI says nothing. Locally on
this branch:

- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets
  --all-features -D warnings`, `RUSTDOCFLAGS="-D warnings" cargo doc` all clean
- `cargo nextest run --workspace --all-targets --all-features` — 2680 passed
- each of the four app tests fails if this wiring is reverted (checked by
  reverting it, not assumed)

Driven by hand with the real CLI against a real `.jsonl`, in an isolated
`CORAL_CONFIG_DIR`:

| step | result |
| --- | --- |
| install source, no query yet | `observed_values: eligible 0` — no install-time scan |
| `coral sql "SELECT id, species, habitat …"` | 3 rows, normal output |
| `coral search kestrel` | `results_found` → `birdlog.sightings`, `matched species = kestrel`, eligible 9 (3 rows × 3 cols) |
| `… WHERE species = 'osprey'` | eligible 2 (1 row × 2 projected cols); `kestrel` **not** captured though it is in the same file |
| `coral source remove birdlog` | store 9 rows → 0, search returns nothing |
| same query, flag off | 0 observed values, 0 queue jobs, provider reports `not_enabled` |

Not run locally: `make docs-check`, `make schema-check`.

## Follow-ups, not in this PR

- Removing a source while `observed_values_search` is off skips the store-side
  clear, so rows linger until eviction. Pre-existing and not file-specific; the
  live-scope filter keeps them unretrievable and the last test here pins that.
- Merge pushed partition-filter *values* into observations, as
  `http/provider.rs::merge_filter_values` does.
- Wrap the `database` backend's providers with the same decorator.
- Rate-limit the writer queue's "dropped" warn if a large scan turns out to spam
  logs. Nothing pre-emptive.

### DIFF
--- a/crates/coral-app/tests/grpc.rs
+++ b/crates/coral-app/tests/grpc.rs
@@ -9,6 +9,8 @@
 mod catalog_discovery_tests;
 #[path = "grpc/feature_service_tests.rs"]
 mod feature_service_tests;
+#[path = "grpc/file_source_observed_values_tests.rs"]
+mod file_source_observed_values_tests;
 #[path = "grpc/function_lifecycle_tests.rs"]
 mod function_lifecycle_tests;
 #[path = "grpc/gui_onboarding_tests.rs"]
--- a/crates/coral-app/tests/grpc/file_source_observed_values_tests.rs
+++ b/crates/coral-app/tests/grpc/file_source_observed_values_tests.rs
@@ -1,0 +1,266 @@
+//! End-to-end proof that file-backed sources contribute observed values.
+//!
+//! The engine tests prove the file backend publishes scan batches. These prove
+//! the published rows survive the whole app path — collector, writer queue,
+//! store, search retrieval — and that source removal and edit take them back
+//! out again.
+
+use std::path::Path;
+use std::time::Duration;
+
+use coral_api::v1::{DeleteSourceRequest, DrainSearchQueueRequest, SearchProvider, SearchRequest};
+use coral_client::default_workspace;
+use serde_json::json;
+use tonic::Request;
+
+use super::harness::{GrpcHarness, manifest_yaml};
+
+const SOURCE_NAME: &str = "file_observed_fixture";
+const OBSERVED_TERM: &str = "kestrel";
+
+fn write_events_fixture(dir: &Path, kind: &str) {
+    std::fs::create_dir_all(dir).expect("fixture dir");
+    std::fs::write(
+        dir.join("events.jsonl"),
+        format!("{{\"id\":1,\"kind\":\"{kind}\"}}\n"),
+    )
+    .expect("jsonl fixture");
+}
+
+fn file_source_manifest_yaml(dir: &Path) -> String {
+    manifest_yaml(&json!({
+        "name": SOURCE_NAME,
+        "version": "0.1.0",
+        "dsl_version": 3,
+        "backend": "file",
+        "tables": [{
+            "name": "events",
+            "description": "Observed-value file fixture",
+            "format": "jsonl",
+            "source": {
+                "location": format!("file://{}/", dir.display()),
+                "glob": "**/*.jsonl",
+            },
+            "columns": [
+                {"name": "id", "type": "Int64"},
+                {"name": "kind", "type": "Utf8"},
+            ],
+        }],
+    }))
+}
+
+fn search_sqlite_path(harness: &GrpcHarness) -> std::path::PathBuf {
+    harness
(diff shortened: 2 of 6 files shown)


mystic123:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/withcoral/coral/2219?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2219** <a href="https://app.graphite.com/github/pr/withcoral/coral/2219?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/2219?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2220** <a href="https://app.graphite.com/github/pr/withcoral/coral/2220?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #2220, opened 2026-08-19
feat(engine): add an observing table-provider decorator

mystic123:
> **Stack 1 of 2.** Adds the decorator; nothing uses it yet. PR #2219 wires the
> file backend to it and is where the behaviour actually changes. Reviewing them
> in order is easiest: this one is the hard part, and it is deliberately free of
> file-backend and app-level noise.

Observed-value capture fires only in the http and mcp backends. Both wire
publishers through their own provider types, which does not generalise: the
file backend registers plain `Arc<dyn TableProvider>` values, and so will the
database backend.

Add a decorator those backends can wrap a provider in. Nothing wires it yet;
the file backend does that in the next PR of this stack, so the module carries
a staged `dead_code` expectation until then.

`ObservedTableProvider` delegates every `TableProvider` method and upgrades
`supports_filters_pushdown`'s `Unsupported` verdicts to `Inexact`. `Inexact`
and `Unsupported` both leave DataFusion's own `FilterExec` above the scan, so
plan shape and results are unchanged; the upgrade only makes the query's
predicates over this table visible to `scan()`, where they select the rows
worth observing. Observing this table's filter-matching rows approximates its
contribution to the result, which a scan head over a file would not: an HTTP
response is shaped by the request, the first rows of a parquet file are not.

`ObservingExec` publishes those rows and forwards every batch untouched. It is
a transparent proxy rather than a new plan node: name, display, plan
properties, children, and every optimizer hook delegate to the wrapped plan,
and it rewraps whenever a rule rewrites that plan. DataFusion added
`ExecutionPlan::downcast_delegate` for this shape. A visible single-child node
would hide the file scan from every rule that downcasts to `DataSourceExec`,
costing filter pushdown, limit pushdown, and file-group repartitioning.

Delegation was checked against both traits rather than by eye. That turned up
`with_preserve_order`, which `datafusion-tracing`'s `InstrumentedExec` omits:
`LimitPushdown` calls it straight after `with_fetch` and reads `None` as "no
order-preserving variant", which would leave a parquet source free to skip row
groups in an order the plan above already assumed was preserved.

The plan-shape tests were written first, to fail fast if the `Inexact` claim
moved plans at all: wrapping an ordered parquet `ListingTable` scan, and
wrapping a provider that pushes nothing down, each produce a byte-identical
physical plan and identical rows. The rest cover predicate selection,
unfiltered capture, projected columns only, both scan entry points, and that an
empty publisher list returns the provider itself rather than a passthrough
wrapper.

## Reviewing this

The two questions worth your attention:

1. **Is the `Unsupported` → `Inexact` upgrade really invisible?** The two
   plan-shape tests are the argument, and they were written before the
   implementation so a wrong answer would have killed the approach early. Worth
   knowing: `ListingTable::supports_filters_pushdown` already returns `Inexact`
   for every non-partition filter, so the upgrade is a no-op for parquet, csv,
   and jsonl, and only takes effect for `MetadataFileTableProvider` and
   `JsonFileTableProvider`.
2. **Is the transparent-proxy shape right?** The alternative, a visible
   single-child node, is what the design doc originally called for. I changed it:
   a visible node hides the file scan from every rule that downcasts to
   `DataSourceExec`. `downcast_delegate` exists in DataFusion 54 for exactly
   this, and `datafusion-tracing`'s `InstrumentedExec` is the same shape.

## Verification

Draft PRs report every real check as `skipping`, so CI says nothing here. Run
locally on this branch:

- `cargo fmt --all --check` clean, under both the `test` and `not(test)` cfgs
- `cargo clippy --workspace --all-targets --all-features -D warnings` clean
- `cargo nextest run --workspace --all-targets --all-features` — 2680 passed
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace` clean

Not run locally, so genuinely unverified: `make docs-check` and
`make schema-check`. Neither should fire — no public API, manifest schema, or
generated docs change — but I have not proven it.

### DIFF
--- a/crates/coral-engine/src/backends/shared/mod.rs
+++ b/crates/coral-engine/src/backends/shared/mod.rs
@@ -4,6 +4,7 @@ pub(crate) mod filter_expr;
 pub(crate) mod json_exec;
 pub(crate) mod json_path;
 pub(crate) mod mapping;
+pub(crate) mod observing_provider;
 pub(crate) mod response_rows;
 pub(crate) mod scalar;
 pub(crate) mod source_observation;
--- a/crates/coral-engine/src/backends/shared/observing_provider.rs
+++ b/crates/coral-engine/src/backends/shared/observing_provider.rs
@@ -1,0 +1,887 @@
+//! Shared `TableProvider` decorator that publishes scanned rows as source
+//! observations.
+//!
+//! Backends whose tables are plain `Arc<dyn TableProvider>` values (file
+//! today, database next) get observed-value capture by wrapping the provider
+//! instead of threading publishers through every provider implementation.
+//!
+//! Two properties keep the decorator invisible to query results:
+//!
+//! - [`ObservedTableProvider::supports_filters_pushdown`] mirrors the inner
+//!   provider's verdicts and upgrades `Unsupported` to `Inexact`. Both keep
+//!   `DataFusion`'s own `FilterExec` above the scan, so plan shape and results
+//!   are unchanged, but the predicates now reach `scan()` where the tap can
+//!   use them to pick the rows worth observing.
+//! - [`ObservingExec`] is a transparent proxy, not a new plan node: it
+//!   delegates its name, display, plan properties, children, and every
+//!   optimizer hook to the plan it wraps, rewrapping whenever an optimizer
+//!   rewrites that plan. `DataFusion` added
+//!   [`ExecutionPlan::downcast_delegate`] for exactly this shape, and
+//!   `datafusion-tracing`'s `InstrumentedExec` is the reference
+//!   implementation. A visible single-child node would instead hide the file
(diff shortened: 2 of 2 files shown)


mystic123:
* **#2219** <a href="https://app.graphite.com/github/pr/withcoral/coral/2219?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2220** <a href="https://app.graphite.com/github/pr/withcoral/coral/2220?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/2220?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #2221, opened 2026-08-19
feat(desktop): package the Windows desktop app as an NSIS installer

AlbertQM:
## Summary

Building an unsigned desktop app for Windows x64.

## Test plan

Manually run desktop package action ([results](https://github.com/withcoral/coral/actions/runs/32240419196/job/96029528534)).
Download installer from there.
Open it on a Windows machine (not a VM on apple silicon, but an actual windows machine).
Go through installation and then onboarding.

Since it's still unsigned, I got this message:

image

image

image

image

I was also able to uninstall it correctly

<details>
Phase 2 of the desktop port. An unsigned x64 NSIS installer, `coral-desktop-win-x64.exe`, labelled a preview: SmartScreen shows an "unrecognized app" warning on install. Signing is a later decision.

## No updater, and that takes two settings

`desktopUpdatesSupported()` already returns `false` for win32, so nothing should ship update metadata. Two independent gates in `app-builder-lib` have to be closed for that to actually hold:

- `win.publish: null` — `getPublishConfigs` reads the platform block before the top-level `publish`, so this suppresses both the `latest.yml` feed and the `app-update.yml` the package would otherwise embed.
- `nsis.differentialPackage: false` — `NsisTarget` gates the `.exe.blockmap` on `!isPortable && differentialPackage !== false`, **not** on `publish`. Without it a blockmap lands in `dist` even with `publish: null`, and `release.yml`'s existing `coral-*.blockmap` glob would sweep it into the release with no warning.

`verify-dist.mjs` gains a `win` platform that fails on either, so this cannot regress silently.

## Deliberate asymmetries with the Linux block

- **`icon: 'resources/icons/icon.ico'`** — the file, not the directory. The lone-`.png` trap that forces `icon: 'resources/icons'` on Linux is PNG-only; the ICO branch validates a `>=256` entry and uses the file as is.
- **No `executableName`** — Linux renames the executable because the deb symlinks it into `/usr/bin` and would shadow the `coral` CLI. `WinPackager` writes `Coral.exe` into the install directory, so there is nothing to collide with.
- **`perMachine: false`** — per-user under `%LOCALAPPDATA%`, no UAC. A per-machine install puts `resourcesPath` under `C:\Program Files`, which a standard user cannot write to.

## Windows polish

- `app.setAppUserModelId('com.withcoral.desktop')` before the app is ready. Windows silently drops toasts from a process without one. Pre-emptive — only the updater uses `Notification` today and that is macOS-gated — but it is one line.
- A conventional **File / Edit / View / Help** menu off darwin, with About and updates under Help. A top-level "Coral" menu sitting next to Edit and View is not where Windows users look for either. The menu bar is now visible off darwin too, since `autoHideMenuBar` put both behind the Alt key. **This changes Linux as well**, which is intended — the same reasoning applies there.
- The sidecar probe asks for `F_OK` on win32. libuv maps `X_OK` to `F_OK` there, so the check silently degraded to "does it exist" while the error message claimed otherwise.

## CI

A `blacksmith-4vcpu-windows-2025` leg in `desktop-package.yml`, and a `build-desktop-windows` job in `release.yml` consuming the CLI artifact the `build` job already produces for `x86_64-pc-windows-msvc`. No `CORAL_DESKTOP_RELEASE` — `createConfig()` rejects it on a win32 host, since release mode means an updater.

Packaging has to run on Windows: `NsisTarget.computeScriptAndSignUninstaller` builds the uninstaller by *executing* the freshly built installer, and off-Windows it does that through wine. The only bypass is owning a custom `installer.nsi`, which is not a trade worth making.

Note the `desktop-package.yml` leg compiles the Rust CLI from source; the release job reuses the prebuilt artifact instead.

## Verification

Local, on macOS:

- `npm run check` and `npm test` in `apps/desktop` — 14 + 50 pass, including a new test asserting the NSIS/no-updater shape.
- The config passes electron-builder's own `scheme.json` validation for `win32`, `darwin`, and `linux`.
- `verify-dist … win` exercised against synthetic dists: happy path, stray blockmap, stray feed, two installers, unknown platform name.
- `electron-builder --win --dir` really packaged: the executable is `Coral.exe`, no `app-update.yml` appears in resources, and all seven ICO entries (16 → 256) are byte-present in the exe.

**Not yet verified: NSIS itself.** Nothing off Windows can cover it. Still to do on a real Windows machine, before this leaves draft:

- [ ] Installer is `coral-desktop-win-x64.exe`; `verify-dist … win` passes on a real build.
- [ ] Installs per-user under `%LOCALAPPDATA%` with no UAC prompt; the directory picker works.
- [ ] Sidecar becomes ready, UI renders over `coral-app://`, no read-only-filesystem errors.
- [ ] Quit is clean, no orphaned `coral.exe` in Task Manager, second launch succeeds immediately.
- [ ] State lands under `%APPDATA%\@withcoral\desktop\coral` via `CORAL_CONFIG_DIR`, and nothing appears in the legacy `config_dir()` location.
- [ ] Desktop Settings → MCP clients writes under `%APPDATA%` through `add-mcp`'s win32 branch.

Drive the real title-bar close, not a scripted one. Phase 1 produced a convincing false "the app does not quit on window close" under bare Xvfb; treat any Windows equivalent as suspect until a minimal control app reproduces it.

## Out of scope

- **Signing.** Azure Trusted Signing (~$10/mo, no HSM, first-class `win.azureSignOptions` support) is the recommended option when we do this. It is also what would unblock a Windows updater later.
- **`aarch64-pc-windows-msvc`.** The CLI release matrix does not build it, so there is no binary to bundle. Windows ARM runs the x64 build under emulation.
- **Graceful sidecar shutdown.** On Windows both `SIGTERM` and `SIGKILL` become `TerminateProcess`, so the five-second graceful window is dead code there. It leaks nothing: the state lock is an advisory `LockFileEx`, which the kernel releases when the handle closes. If it ever matters, the portable fix is a stdin-close handshake, not `taskkill`.
</details>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -420,11 +420,84 @@ jobs:
             apps/desktop/dist/coral-desktop-*.deb
             apps/desktop/dist/latest-linux.yml
 
+  # An unsigned preview: SmartScreen warns on install. No signing, no
+  # notarization, and no updater, so no xtask and no credentials — and unlike
+  # the Linux leg, no update feed either (see win.publish in
+  # apps/desktop/electron-builder.config.ts).
+  build-desktop-windows:
+    needs:
+      - validate-release-ref
+      - build
+    runs-on: blacksmith-4vcpu-windows-2025
+    env:
+      TAG_NAME: ${{ needs.validate-release-ref.outputs.tag-name }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          ref: ${{ needs.validate-release-ref.outputs.commit-sha }}
+          persist-credentials: false
+
+      # Nothing downstream carries the version — the installer filename holds
+      # only the arch — so a manual tag on a commit whose desktop version was
+      # never bumped would ship an About panel and an Add/Remove Programs entry
+      # that disagree with the release. The macOS leg gets this from xtask; see
+      # ensure_version_matches_tag in xtask/src/release.rs.
+      - name: Verify desktop version matches the release tag
+        shell: pwsh
+        run: |
+          $version = (Get-Content apps/desktop/package.json | ConvertFrom-Json).version
+          if ("v$version" -ne $env:TAG_NAME) {
+            Write-Error "desktop version v$version does not match release tag $env:TAG_NAME"
+            exit 1
+          }
+
+      - name: Download Windows CLI artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
+        with:
+          name: coral-x86_64-pc-windows-msvc
+          path: ${{ runner.temp }}/cli
+
+      # The `build` job already compiled this target; reuse it rather than
+      # rebuilding Rust here.
+      - name: Unpack the prebuilt Coral CLI
+        shell: pwsh
+        run: |
+          $cli = "$env:RUNNER_TEMP\cli"
+          Expand-Archive -Path "$cli\coral-x86_64-pc-windows-msvc.zip" -DestinationPath $cli -Force
+          & "$cli\coral.exe" --version
+
+      - name: Set up Node
+        uses: actions/setup-node@249970729cb0ef3589644e2896645e5dc5ba9c38 # v6.5.0
+        with:
+          node-version-file: apps/desktop/.node-version
+          package-manager-cache: false
+
+      - name: Install desktop dependencies
+        run: npm ci --prefix apps/desktop
+
+      # No CORAL_DESKTOP_RELEASE: it bakes in the updater, and Windows has none.
+      # createConfig() rejects the flag on a win32 host for that reason.
+      - name: Package desktop app
+        env:
+          CORAL_DESKTOP_CLI_PATH: ${{ runner.temp }}/cli/coral.exe
+        run: npm run package:win --prefix apps/desktop
+
+      - name: Verify desktop package artifacts
+        run: node apps/desktop/scripts/verify-dist.mjs apps/desktop/dist win
+
+      - name: Upload desktop artifact
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
+        with:
+          name: coral-desktop-windows
+          if-no-files-found: error
+          path: apps/desktop/dist/coral-desktop-*.exe
+
   checksums:
     needs:
       - build
       - build-desktop-linux
       - build-desktop-macos
+      - build-desktop-windows
       - sign-notarize-macos
     runs-on: blacksmith-4vcpu-ubuntu-2404
     steps:
@@ -448,6 +521,7 @@ jobs:
             coral-*.blockmap
             coral-desktop-*.AppImage
             coral-desktop-*.deb
+            coral-desktop-*.exe
           )
           if [ "${#archives[@]}" -eq 0 ]; then
             echo "No Coral release archives found" >&2
@@ -508,6 +582,7 @@ jobs:
             artifacts/coral-*.blockmap
             artifacts/coral-desktop-*.AppImage
             artifacts/coral-desktop-*.deb
+            artifacts/coral-desktop-*.exe
           )
           if [ "${#archives[@]}" -eq 0 ]; then
             echo "No Coral release archives found" >&2
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -429,6 +429,14 @@ jobs:
       - validate-release-ref
       - build
     runs-on: blacksmith-4vcpu-windows-2025
+    # The Windows installer is an unsigned preview, so it must not gate the whole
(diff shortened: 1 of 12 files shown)


apps/desktop/electron-builder.config.ts
```diff
@@ -142,6 +141,30 @@ export function createConfig(
...
+      allowToChangeInstallationDirectory: true,
+      // Blockmaps are differential-update metadata, gated on this rather than on
+      // `publish`, so without it a .exe.blockmap lands in dist and release.yml's
+      // `coral-*.blockmap` glob sweeps it into the release.
+      differentialPackage: false,
+      oneClick: false,
+      // Per-user, under %LOCALAPPDATA%. A per-machine install puts
+      // resourcesPath under Program Files, which a standard user cannot write
+      // to, and costs a UAC prompt for a developer tool that needs none.
+      perMachine: false,
```
simonwhitaker, line 189: **`perMachine: false` sets the default install scope, it does not force per-user.**

In app-builder-lib 26, `NsisTarget` defines `INSTALL_MODE_PER_ALL_USERS` only when `perMachine === true`. With it undefined, `templates/nsis/assistedInstaller.nsh` inserts `PAGE_INSTALL_MODE` — the "Who should this application be installed for?" page — because `oneClick` is false:

```nsis
!ifndef INSTALL_MODE_PER_ALL_USERS
  !insertmacro PAGE_INSTALL_MODE
!endif
```

So a user who picks "Anyone who uses this computer" runs `setInstallModePerAllUsers`: UAC elevation and `$PROGRAMFILES64\Coral`. On a standard non-admin account the prompt demands admin credentials and the install cannot proceed without them.

That makes two claims inaccurate as written:

- this comment's "Per-user, under %LOCALAPPDATA%" and its rationale;
- `apps/desktop/README.md:96`, which states as fact that the installer "installs per user under `%LOCALAPPDATA%`, needs no UAC prompt".

Neither `verify-dist` nor the new config test can observe the difference, so this only surfaces as a support question. Worth either softening both to "defaults to per-user", or closing the door — `allowElevation: false` removes the elevation path so the all-users choice can't succeed.

apps/desktop/src/main/index.ts
```diff
@@ -356,11 +368,18 @@ function installMenu() {
...
   ]
   Menu.setApplicationMenu(Menu.buildFromTemplate(template))
 }
 
 function startApplication(): void {
+  // Windows keys toasts, taskbar grouping, and jump lists off this id, and
+  // silently drops a notification from a process that never set one. Must match
+  // `appId` in electron-builder.config.ts, which is what the installer stamps on
+  // the shortcut, and must run before the app is ready. Inert elsewhere.
+  app.setAppUserModelId('com.withcoral.desktop')
```
simonwhitaker: The comment says this "must match `appId` in electron-builder.config.ts" — but nothing enforces it. Two string literals in two files, no shared constant, and `scripts/electron-builder-config.test.mjs` (extended in this PR) asserts the whole win/nsis shape without ever comparing `appId` to this one.

If `appId` is changed later, the NSIS shortcut carries the new AUMID while the running process registers the old one, and Windows treats them as two applications: the running window gets its own taskbar button instead of grouping under the pinned shortcut, and toasts are dropped as coming from an unregistered AUMID — the exact failure this line exists to prevent. Silent in every test and every CI leg.

A grep-based assertion in the config test would pin it:

```js
test('the main process sets the AppUserModelID electron-builder stamps on the shortcut', () => {
  const main = readFileSync(new URL('../src/main/index.ts', import.meta.url), 'utf8')
  const { appId } = createConfig({}, 'win32')
  assert.ok(main.includes(`app.setAppUserModelId('${appId}')`))
})
```

.github/workflows/release.yml
```diff
@@ -420,11 +420,84 @@ jobs:
...
+          name: coral-desktop-windows
+          if-no-files-found: error
+          path: apps/desktop/dist/coral-desktop-*.exe
+
   checksums:
     needs:
       - build
       - build-desktop-linux
       - build-desktop-macos
+      - build-desktop-windows
```
simonwhitaker, line 508: This makes the explicitly-preview, unsigned Windows leg a hard gate on the entire release.

If `build-desktop-windows` fails for any reason — Windows runner unavailability, electron-builder's NSIS toolchain download flaking, or the new version-vs-tag check firing on a manually dispatched tag — `checksums` is skipped, and so is `publish-version`. The release then publishes with **no assets at all**: no CLI `.tar.gz`/`.zip`, no macOS DMG/ZIP, no Linux AppImage/deb, and no `latest-mac.yml` / `latest-linux.yml`, so existing macOS and AppImage users' auto-update stalls too.

The downstream jobs already tolerate a missing installer: both archive lists run under `shopt -s nullglob` and stay non-empty from the other platforms, and the two `test -s` literals are the mac and linux feeds. So `continue-on-error: true` on `build-desktop-windows` would degrade this to "no Windows installer this release" rather than "no release" — which reads closer to how the PR describes the platform.

simonwhitaker (approved):

PR #2224, opened 2026-08-20
feat(desktop): ship auto-updates to the Linux AppImage

AlbertQM:
## Summary

Wiring up auto-updates for the new Linux desktop app package

## Test plan

Just like for the MacOS version, this path isn't easily tested unless we cut an actual release.

<details>
The Linux packaging commit left updates macOS-only. The AppImage can in fact replace itself: electron-updater's AppImageUpdater downloads against $APPIMAGE, unlinks it, and moves the new image over that path. This turns that on. The deb stays out, since its files belong to dpkg.

src/shared/release-targets.ts is now the one list of platforms a release build ships to, and whether each signs with Apple. Both electron-builder.config.ts and main/auto-update.ts read it, so packaging and the running app cannot disagree about who gets an update feed. Release mode is no longer the Apple path: the flag now means "bake in an active updater", and only darwin adds signing and notarization on top, so a Linux release build takes no credentials. Windows is absent from the table and still rejects the flag; adding it there is what puts it on par.

Three details in the update path are load-bearing:

- $APPIMAGE is the gate, not any AppImage detection of our own. AppImageUpdater reads the variable itself and refuses to run when it is unset or relative, so desktopUpdatesSupported() matches the updater's own precondition. An extracted AppDir never sets it.
- autoRunAppAfterInstall = false, on Linux only. AppImageUpdater starts the new image from inside its install step, while this process still holds the single-instance lock, so the replacement would find it taken. app.relaunch() from before-quit-for-update starts it after the exit instead. The flag must stay true on macOS: MacUpdater reads it to choose between the Squirrel hand-off and a plain quit, and false there would leave the staged update uninstalled.
- app.relaunch() passes empty args, not Electron's default process.argv.slice(1). Those are the flags AppRun added for this launch, and AppRun adds them again.

release.yml's Linux leg now builds with CORAL_DESKTOP_RELEASE=1, runs the same verify:release-build check as macOS, packages the prepared bundle instead of rebuilding it, and uploads latest-linux.yml with the archives. verify-dist's feed check is now shared by both platforms: every file a feed names must exist in the dist, or it 404s for every updater in the wild. The Linux leg also tightened from "at least one" AppImage and deb to exactly one, since the feed pins a specific file.

tsconfig gains allowImportingTsExtensions, because electron-builder.config.ts is loaded by jiti and by node --test, and neither resolves an extensionless specifier to a .ts file.
</details>


apps/desktop/electron-builder.config.ts
```diff
@@ -115,25 +121,16 @@ export function createConfig(
...
+      // does not declare. A directory routes it through the size generator.
       icon: 'resources/icons',
       // Debian requires a contact address for the Maintainer field, and fpm
       // refuses to build without one — `author` carries no email.
       maintainer: 'Coral Eng Team <eng@withcoral.com>',
-      // Linux has no updater (see desktopUpdatesSupported in src/main/auto-update.ts).
-      // `null` keeps electron-builder from writing a latest-linux.yml feed nobody
-      // serves and from embedding app-update.yml in the package.
-      publish: null,
       synopsis: 'Coral desktop app',
```
simonwhitaker, line 131: **Dropping `linux.publish: null` also turns the updater on for the deb — and that leaks into the AppImage.**

`FpmTarget.build()` (app-builder-lib 26) now resolves a publish config for the deb, and when it does it writes two files into the *shared* unpacked app dir:

```js
await outputFile(path.join(resourceDir, "app-update.yml"), ...)
await outputFile(path.join(resourceDir, "package-type"), target)   // "deb"
```

`resourceDir` is `dist/linux-unpacked/resources` — the very directory `AppImageTarget` squashes (`copyDir(appDir, stageDir)`), and both targets are started concurrently by `PlatformPackager.buildAsyncTargets`. FpmTarget writes those files within its first few awaits, while AppImageTarget still has to stage AppRun, fetch the appimage toolset and copy the runtime libs — so `package-type: deb` ends up *inside the shipped AppImage*.

At runtime electron-updater's `doLoadAutoUpdater()` reads `process.resourcesPath/package-type` and, seeing `deb`, builds a **DebUpdater instead of an AppImageUpdater**. For an AppImage user who clicks "Install update":

1. `DebUpdater.doDownloadUpdate()` picks the `.deb` (which this change also adds to `latest-linux.yml`).
2. `doInstall()` runs `dpkg -i` via `pkexec`/`sudo` — a root-password prompt, and Coral gets installed system-wide on a machine that never used dpkg.
3. The AppImage on disk is never replaced, and since `autoRunAppAfterInstall = false` DebUpdater skips its own relaunch, so `scheduleAppImageRelaunch()` restarts `$APPIMAGE` — the **old** version.
4. That old version fails `shouldExitForPendingDesktopUpdate()` (its version < the recorded intent target) and exits immediately at startup, so the app refuses to launch for the whole 10-minute intent TTL.

Suggested fix: keep the deb out of the updater explicitly — `deb: { publish: null }` in the config below. A target-level `publish: null` short-circuits `getPublishConfigs()`, so no `package-type`, no `app-update.yml`, and the deb also stops being written into `latest-linux.yml`.

apps/desktop/src/main/auto-update.ts
```diff
@@ -23,12 +28,27 @@ declare const __CORAL_DESKTOP_RELEASE__: boolean
...
+// to quit. Electron starts the replacement only after this process exits, so the
+// new instance finds the single-instance lock free.
+function scheduleAppImageRelaunch(): void {
+  if (process.platform !== 'linux') return
+  const imagePath = appImagePath(process.env)
+  if (!imagePath) return
+
+  // Empty args, not Electron's default `process.argv.slice(1)`: that carries
+  // the flags AppRun added for this launch, and AppRun adds them again.
+  app.relaunch({ execPath: imagePath, args: [] })
```
simonwhitaker, line 60: **The relaunch assumes the new image landed on `$APPIMAGE`, which is not always where `AppImageUpdater` puts it.**

`AppImageUpdater.doInstall()` overwrites in place only when the current basename equals the downloaded one *or* contains no `x.y.z`:

```js
if (path.basename(installerPath) === existingBaseName || !/\d+\.\d+\.\d+/.test(existingBaseName)) {
  destination = appImageFile
} else {
  destination = path.join(path.dirname(appImageFile), path.basename(installerPath))
}
...
unlinkSync(appImageFile)   // the old path is gone either way
```

So for a user who keeps versioned copies — `~/Applications/coral-desktop-0.14.0.AppImage`, a very common AppImage habit — the old file is unlinked, the new image is written next to it under the artifact name, and `app.relaunch({ execPath: imagePath })` then points at a path that no longer exists. The app quits and never comes back, and their `.desktop` launcher is dangling too.

electron-updater emits `appimage-filename-updated` with the real destination for exactly this case; capturing it and relaunching that path (falling back to `$APPIMAGE`) would close the gap.

apps/desktop/scripts/verify-dist.mjs
```diff
@@ -72,18 +75,15 @@ function verifyMac() {
...
   }
-  // Linux has no updater, so a feed here would be published and then served to
-  // nobody. Treat one as a packaging mistake.
-  const feeds = entries.filter((f) => /^latest(-\w+)?\.yml$/.test(f))
-  if (feeds.length > 0) {
-    fail(`linux ships no updater, but the build produced update metadata: ${feeds.join(', ')}`)
-  }
-  return [...appImages, ...debs].join(', ')
+
+  const referenced = verifyFeed('latest-linux.yml', appImages[0])
```
simonwhitaker: **This check can't tell that the feed also advertises the deb.**

`FpmTarget` emits update info too (`isWriteUpdateInfo: true` whenever its publish config resolves), and `writeUpdateInfoFiles()` merges both tasks into the one `latest-linux.yml`, pushing the deb into `files:`. Its sort key is only zip-vs-non-zip and then arch — both artifacts tie on both — so the surviving top-level `path:`/`sha512:` come from whichever target's `hashFile()` resolved first. Across builds the published feed can legitimately say `path: coral-desktop-linux-amd64.deb`, i.e. name the one Linux package that must never be installed by the updater.

`verifyFeed(..., appImages[0])` only asserts the AppImage appears *somewhere* in the feed, so this passes either way. Worth asserting `path:` is the AppImage — or better, dropping the deb from the feed entirely (see the `deb: { publish: null }` comment on `electron-builder.config.ts`), which makes this deterministic.

apps/desktop/src/main/auto-update.ts
```diff
@@ -23,12 +28,27 @@ declare const __CORAL_DESKTOP_RELEASE__: boolean
...
-  return isReleaseBuild && app.isPackaged && process.platform === 'darwin'
+  if (!isReleaseBuild || !app.isPackaged) return false
+  if (process.platform === 'darwin') return true
+  return process.platform === 'linux' && appImagePath(process.env) !== null
+}
+
+// Runs once the updater has moved the new image over the old path and is about
+// to quit. Electron starts the replacement only after this process exits, so the
+// new instance finds the single-instance lock free.
+function scheduleAppImageRelaunch(): void {
```
simonwhitaker, line 53: **Worth confirming on a real AppImage that the relaunch actually fires.**

Electron does not exec the new image from this process — `Browser::Relaunch()` forks a *relauncher helper* from `CHILD_PROCESS_EXE`, i.e. the electron binary inside the currently-mounted AppDir (`/tmp/.mount_XXXX/...`). `RelauncherSynchronizeWithParent()` (relauncher_linux.cc) then arms `PR_SET_PDEATHSIG` and blocks until this process exits, and only afterwards calls `LaunchProgram()`.

The AppImage runtime tears down that squashfs mount the moment this process exits, so the helper resumes and forks while executing from a filesystem that is disappearing underneath it — a page fault on non-resident text there is a SIGBUS, and the user is left with no running app and no error. The comment above claims the ordering as a guarantee; since this path can't be exercised without cutting a release, it would be good to validate it on a real AppImage (or fall back to relaunching from the new image if nothing comes back).

simonwhitaker (approved):

PR #2233, opened 2026-08-21
feat(coral-ui): Add CSS @layers to WAX

AlbertQM:
## Summary

I noticed a CSS bug which only appeared in prod. 
Turns out a component (WorkspaceSelector) was using a CSS rule with 0,1,0 specificity, same as our WAX button.
In prod, the WAX stylesheet came after, overriding the style added by WorkspaceSelector.

Rather than fixing that one occurrence, I'm adding layers to enforce the order of styles in a predictable manner:
1. Reset css first (so they are always overridden)
2. Vendor-specific stylesheets (only toast atm)
3. WAX components next
4. Everything else (so they always take precedence)

Also add a lint rule to help with the migration and prevent diverging in the future

## Test plan

Ask Claude to review the outputs of dev and build, before and after this change. Styles are now always emitted in the correct order.

### DIFF
--- a/apps/coral-ui/.oxlintrc.json
+++ b/apps/coral-ui/.oxlintrc.json
@@ -21,6 +21,27 @@
       }
     },
     {
+      "files": ["app/wax/**/*.{ts,tsx}"],
+      "excludeFiles": ["app/wax/css.ts", "app/wax/theme/layers.css.ts"],
+      "rules": {
+        "no-restricted-imports": [
+          "error",
+          {
+            "paths": [
+              {
+                "name": "@vanilla-extract/css",
+                "message": "Import from '@/wax/css' instead, so wax rules land in the wax cascade layer."
+              },
+              {
+                "name": "@vanilla-extract/recipes",
+                "message": "Import from '@/wax/css' instead, so wax rules land in the wax cascade layer."
+              }
+            ]
+          }
+        ]
+      }
+    },
+    {
       "files": ["app/routes/**/*.{ts,tsx}"],
       "plugins": ["eslint", "typescript", "unicorn", "oxc", "react", "react-hooks", "import"],
       "rules": {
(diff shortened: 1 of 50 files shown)


lazybean:
## Automated React Router framework-mode review

Head: `b7a0a9d44df0ee2fd3cdf93a58edde42d121be02`

## React Router framework-mode review

  No actionable findings.

  The PR changes CSS cascade-layer infrastructure and related architecture tests. It does not modify route configuration, loaders, actions, forms, fetchers, URL state, revalidation, error boundaries, or pending/mutation behavior. The component-scoped toast stylesheet remains compatible with SSR and hydration: its hosts render through existing routes,
  while the root stylesheet establishes the global layer order.

  ### Checks performed

  - Read the supplied Coral AGENTS.md instructions.
  - Applied the installed react-router-framework-mode guidance.
  - Confirmed checkout HEAD exactly matches b7a0a9d44df0ee2fd3cdf93a58edde42d121be02.
  - Confirmed PR #2233 metadata, commits, and changed-file list.
  - Reviewed the complete diff from merge base 7e521caac25c0f3356e95d299554f2009c978976.
  - Inspected neighboring framework code:
      - app/root.tsx
      - app/routes.ts
      - react-router.config.ts
      - vite.config.ts
      - routes/app-shell.tsx
      - onboarding’s toast host

  - Checked stylesheet entry points, route-scoped toast loading, SSR document integration, and theme bootstrap/hydration behavior.
  - git diff --check passed.

  ### Not verified

  - npm run check --prefix apps/coral-ui could not run because dependencies are not installed (oxfmt: command not found).
  - Typecheck, tests, production build, server smoke test, and browser rendering were therefore not executed.
  - No files were modified and no external review comments were posted.

lazybean (approved):
Thank you! Looks good!

PR #2256, opened 2026-08-26
refactor(search): add the storage seam with the SQLite adapter

mystic123:
Introduce `SearchStorage` and `SearchStore` with the `CatalogStore` and
`ObservedValuesStore` component traits, following the `CoralDbBackend`
precedent. The SQLite store moves behind the seam unchanged; the catalog
provider, search manager, and source manager open per-workspace stores
through it instead of constructing `SqliteSearchStore` directly, and the
source manager takes its storage as a required dependency so a lifecycle
change can never clear the wrong backend in silence.

The seam is synchronous on purpose: the provider registry keeps the
blocking boundary. The paired contract suite (catalog contract and the
benchmark match strata) runs against SQLite here and against every later
backend unchanged.

No behavior change: existing SQLite search tests run as they were.

https://claude.ai/code/session_01WGJRTvciKJkDniTJVSwSY1

PR #2255, opened 2026-08-26
refactor(search): move catalog index types out of the SQLite module

mystic123:
# WAITING ON LAGOON 

  
The catalog projection's document, snapshot, result, and hit types, the
query-term normalization, the searchable-text composition, and the
closed vocabularies for surface kinds and field roles move from
`catalog::sqlite_index` to a backend-neutral `catalog::index`. The
retrieval class now names document kinds instead of carrying SQL, so a
second storage backend can share the vocabulary and render its own
predicate.

No behavior change; the SQLite index keeps its statements and tests.

https://claude.ai/code/session_01WGJRTvciKJkDniTJVSwSY1

PR #2257, opened 2026-08-26
feat(search): add the Postgres catalog search store, not yet selectable

mystic123:
Catalog Universal Search gains a Postgres backend in the same database
as app state: one schema per Workspace named through the
`search_registry.workspaces` surrogate registry, which also serves as
the per-schema migration ledger. Workspace names never reach SQL as
identifiers. The backend opens its own small pool through the shared
`connect_postgres_pool` helper (same URL and TLS posture as CoralDb),
sets `search_path` per transaction, runs its own migration stream —
non-idempotent by design, since DDL and the ledger bump share one
transaction — and fails open loudly when `pg_trgm` is missing.

Writers take the Workspace's advisory lock without waiting, so a
contended projection is reported as such and the provider serves its
cached projection, as it does on SQLite. Deletion takes the same lock
before dropping a schema.

The backend keeps no observed values; the seam, the provider registry,
and maintenance report that absence explicitly. Nothing selects the
backend yet: configuration lands after lifecycle wiring, so no
deployment changes behavior with this change.

## Retrieval and ranking

The engine is `pg_trgm` + weighted `tsvector` on vanilla Postgres (RDS-installable, per the locked plan). The ranking is BM25 in plain SQL — the SQLite side's `bm25()` rebuilt, with the same constants and field weights — because the obvious alternative measured badly: `ts_rank_cd` has no corpus statistics, and under coral's disjunctive `tsquery` its cover-density machinery never engages, so it degenerates to a weighted term-frequency count. On the benchmark catalog (21k documents) the word `github` appears in 97.7 % of documents; FTS5 clamps it to nothing, `ts_rank_cd` ranks on it. That version placed the table the agent went on to use in the top 3 **14 %** of the time against SQLite's 54 %.

What ships:

- **Corpus statistics in the projection** (`0001_catalog_documents.sql`): `catalog_documents.doc_len`, `catalog_terms(term, ndoc)` filled by `ts_stat`, and a one-row `catalog_stats(total_docs, avgdl)`. `replace_documents` rewrites all three inside the projection's transaction (~150 ms at 21k docs), so statistics and rows can never disagree.
- **Score**: every query word joins the BM25 sum with its true IDF (`ln((N - n + 0.5) / (n + 0.5))`, floored at FTS5's `1e-6`), FTS5's `k1` = 1.2, and `b` = 0.4 — FTS5's 0.75 is a prose constant; on short structured catalog documents it made a column-rich canonical table pay in its name field for its column count. Lexeme occurrences carry the 8/6/2/1 field weights; a lexeme-prefix join lets `pull` score `pulls`.
- **`.`-split identifier fields**: the `tsv` generated column tokenizes `replace(qualified_name, '.', ' ')` (and `title`, `searchable_text`). The default parser fuses `linear.issues` into one host lexeme no query word can match, while `linear.initiative_projects` splits its underscore tail into real A lexemes — the field-weight intent inverted. `description` is prose and keeps the plain expression.
- **Candidates**: prefix `tsquery` (`word:*`) OR trigram `ILIKE '%word%'` per word — substring semantics preserved — inside a `WITH candidates AS MATERIALIZED` wall (planned together with the scoring join, the planner abandons the bitmap index scans for a seq scan and the median doubles). Only candidate selection is restricted to words below a 5 % document-frequency cap; it is a latency lever (uncapped candidates measured 5× the cost for identical relevance), not a relevance rule — every word still scores. If every word is common, the 3 rarest survive as candidate keys.
- **Order**: whole-query phrase boost (8/6/2/1, what keeps `"issue labels"` ahead of co-occurrence matches), BM25 score, `doc_id`. Zero-score candidates (mid-word substring matches like `enchmark` → `benchmark_runs`) keep their rank through a `LATERAL` + `coalesce(sum(…), 0)`; an inner-join `GROUP BY` shape would silently drop them.
- **Compact variants earn their place from the corpus.** Query construction adds a compact identifier variant to every term with a separator (`deploy_url` → `deployurl`), including the whole-query term it appends (`issue labels` → `issuelabels`), so a name typed with spaces meets the identifier. The store now keeps each term's provenance and drops a variant word the statistics show absent (`ndoc = 0`) before candidate selection, scoring, and the rarest-words fallback. Without that, the variant of a sentence — a 60-character word nothing starts with — was always under the cap, so a query made only of common words kept it as its sole candidate key and returned nothing; it also cost an `ILIKE` pattern and a score row on every multi-word query. Typed words are never subject to the rule (`enchmark` still selects `benchmark_runs` by substring). Measured on the 402-query offline replay: identical aggregates, one ranked list changed (used table rank 4 → 3), SQL median 85 → 58 ms, p90 450 → 315 ms.

Measured and rejected along the way: an exact-token qualified-name tier (as a hard sort tier it promotes underscore compounds without the split and wrecks top-k/#1/Slack with it; a soft ×2 multiplier is a wash), `ts_rank_cd` normalization flags (insufficient), and dropping words above the cap from scoring (buried `linear.issues`: `issues` is in 10 % of documents and carries a real IDF ≈ 2.2 that FTS5 keeps).

### Measured

Offline replay of 395 recorded agent queries (387 labeled with the table the agent went on to use), same corpus for both engines:

| | top-3 | #1 | top-k | med rank | GitHub | Linear | Notion | Slack |
|---|---|---|---|---|---|---|---|---|
| SQLite baseline | 52 % | 28 % | 83 % | 2.0 | 38 % | 65 % | 19 % | 71 % |
| `ts_rank_cd` (first cut) | 14 % | 3 % | 51 % | 8.0 | | | | |
| **this PR** | **52 %** | 28 % | 84 % | **2.0** | 40 % | 58 % | 28 % | 70 % |

The paid duel run (`coral-benchmarks/docs/findings/postgres-search-duel.md`, run 3, 2026-09-10) confirms it end to end: used-table top-3 53 % vs SQLite 52 %, at #1 29 % vs 26 %, GitHub 63 vs 50, Linear 54 vs 62, task creation 17/17, cost +5 %, paired outcomes 1 W / 2 L / 98 D.

### Known limit

Search cost scales with query length because every word scores: warm medians of 291 / 480 / 719 ms at 3–8 / 9–12 / 13–25 query words, where SQLite is flat. The 400 ms warm budget passes on one recorded query mix (393 ms) and fails on another (566 ms); in-field agent call time is within noise of the capped variant (0.64 vs 0.62 s). The candidate follow-up is capping the *scored* word count to the k rarest words instead of by document frequency, which should keep this ranking at a flat latency; it validates in minutes on the offline replay before any paid run. Round-trip consolidation (~17 round trips per search) is the other independent latency follow-up.

## Validation

- `make rust-checks` green.
- `make postgres-tests` green against Postgres 17: the full store contract (substring / word / phrase / 3-char floor / negative strata), the common-word-noise ranking contract (twelve documents dense in a common word must lose to the one document the query's rare words name — passes on FTS5 and Postgres, fails on `ts_rank_cd`), the common-word retrieval contract (`firehose events` on a corpus where both words are common and nothing is named `firehoseevents` must return all twelve event tables — 0 of 12 without the variant rule, 12 with it; FTS5 passes either way), registry naming with hostile Workspace names, schema isolation, deletion, the ledger sweep, pruning, contention.
- Offline mirror and real-binary `mcp-stdio` replay agree on every aggregate and per-source figure.

https://claude.ai/code/session_01WeuXrBPjT8Q5nPTruS8xjX

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -249,6 +249,9 @@ postgres-tests: $(POSTGRES_TESTS_PREREQS)
 	CORAL_TEST_POSTGRES_URL="$$url" cargo test --locked -p coral-app --lib \
 	  contract_on_postgres \
 	  -- --ignored; \
+	CORAL_TEST_POSTGRES_URL="$$url" cargo test --locked -p coral-app --lib \
+	  search::store::postgres:: \
+	  -- --ignored; \
 	CORAL_TEST_POSTGRES_URL="$$url" cargo test --locked -p coral-app \
 	  --test postgres_database_tests \
 	  -- --ignored
--- a/crates/coral-app/src/search/catalog/index.rs
+++ b/crates/coral-app/src/search/catalog/index.rs
@@ -128,26 +128,53 @@ impl CatalogDocumentClass {
     }
 }
 
+/// One normalized query term and where it came from.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct NormalizedSearchTerm {
+    pub(crate) text: String,
+    /// Synthesized by removing every non-alphanumeric character from a term
+    /// the caller gave (`deploy_url` → `deployurl`); never typed by the
+    /// caller. Documents index the same variant of each of their names, so
+    /// the only designed match for one is a whole lexeme. A backend that can
+    /// tell whether its corpus holds that lexeme may drop a variant it does
+    /// not; the whole query's variant of a sentence is the common case.
+    pub(crate) compact_variant: bool,
+}
+
 /// Lowercases, trims, and de-duplicates query terms, adding the compact
 /// identifier variant of each term so `deploy_url` and `deployurl` meet.
 pub(crate) fn normalized_search_terms(terms: &[String]) -> Vec<String> {
+    normalized_search_term_variants(terms)
+        .into_iter()
+        .map(|term| term.text)
+        .collect()
+}
+
+/// [`normalized_search_terms`] with each term's provenance kept.
+pub(crate) fn normalized_search_term_variants(terms: &[String]) -> Vec<NormalizedSearchTerm> {
     let mut normalized = Vec::new();
     for term in terms {
         let term = term.trim().to_lowercase();
         if term.is_empty() {
             continue;
         }
-        push_search_term(&mut normalized, term.clone());
+        push_search_term(&mut normalized, term.clone(), false);
         if let Some(compact) = compact_identifier_variant(&term) {
-            push_search_term(&mut normalized, compact);
+            push_search_term(&mut normalized, compact, true);
         }
     }
     normalized
 }
 
-fn push_search_term(terms: &mut Vec<String>, term: String) {
-    if !terms.iter().any(|existing| existing == &term) {
-        terms.push(term);
+fn push_search_term(terms: &mut Vec<NormalizedSearchTerm>, text: String, compact_variant: bool) {
+    match terms.iter_mut().find(|existing| existing.text == text) {
+        // A term the caller typed is never demoted to a variant, whichever
+        // of the two spellings arrived first.
+        Some(existing) => existing.compact_variant &= compact_variant,
+        None => terms.push(NormalizedSearchTerm {
+            text,
+            compact_variant,
+        }),
     }
 }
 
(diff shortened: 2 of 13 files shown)


mystic123:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/withcoral/coral/2257?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2259** <a href="https://app.graphite.com/github/pr/withcoral/coral/2259?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 3  dependent PRs ([#2274](https://github.com/withcoral/coral/pull/2274) <a href="https://app.graphite.com/github/pr/withcoral/coral/2274?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2308](https://github.com/withcoral/coral/pull/2308) <a href="https://app.graphite.com/github/pr/withcoral/coral/2308?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2310](https://github.com/withcoral/coral/pull/2310) <a href="https://app.graphite.com/github/pr/withcoral/coral/2310?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#2258** <a href="https://app.graphite.com/github/pr/withcoral/coral/2258?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2257** <a href="https://app.graphite.com/github/pr/withcoral/coral/2257?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/withcoral/coral/2257?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2256** <a href="https://app.graphite.com/github/pr/withcoral/coral/2256?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2255** <a href="https://app.graphite.com/github/pr/withcoral/coral/2255?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #2263, opened 2026-08-27
feat(coral-ui): add workspace users with a selected role

AlbertQM:
## Summary

- Add a multi-select so I can add multiple users in one go
- Update button to say add members/owners based on what we selected
- Removed the "can't remove last owner" copy since that's not true anymore -- we now have org owners who can control every workspace

## Test plan

image

### DIFF
--- a/apps/coral-ui/app/components/workspace-users/workspace-users.stories.tsx
+++ b/apps/coral-ui/app/components/workspace-users/workspace-users.stories.tsx
@@ -45,16 +45,16 @@ const meta: Meta<StoryArgs> = {
       const submittedIntent = formData.get('intent')
       const intent =
         submittedIntent === 'add' || submittedIntent === 'remove' ? submittedIntent : 'role'
-      const result = {
+      const result: WorkspaceUsersActionData = {
         intent,
         message:
           intent === 'add'
-            ? 'Workspace user added.'
+            ? 'Workspace users added.'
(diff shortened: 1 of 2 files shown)


apps/coral-ui/app/components/workspace-users/workspace-users.css.ts
```diff
@@ -61,6 +61,24 @@ export const memberNameLine = style({
   gap: '4px 8px',
 })
 export const memberTableRow = style({})
+
+// The last owner takes no action, so the row must not read as a target. The table
+// paints its hover colour from a rule inside the wax layer, and this file is
+// unlayered, so this wins whatever the specificity.
```
lazybean: Ok.

apps/coral-ui/app/components/workspace-users/workspace-users.stories.tsx
```diff
@@ -73,6 +76,49 @@ type Story = StoryObj<StoryArgs>
...
+          status: 'success' as const,
+          userIds,
+        }
+      }
+
+      return {
+        failures: userIds.map((userId) => ({
+          message: 'The directory no longer lists this user.',
+          userId,
+        })),
```
lazybean: Does the actual grpc endpoint returns `failures` ?

lazybean (approved):
So much better thank you

